PR #11, opened 2018-03-07
cleanup polyline layer, add example

johnpryan:


haydenflinner:
Thanks so much for this!

PR #15, opened 2018-03-12
Fixed marker redraw on map rotation.

### LINKED ISSUE #1: Rotating map doesn't re-draw markers
steps to reproduce:

1. run example
2. rotate map, observe marker locations
3. pan the map
4. observe marker location jumps to the correct location

Max053:
Should fix #1 .


johnpryan:
Solid. Thanks!

PR #24, opened 2018-04-11
Expose MapState

avioli:
Allow users to pass an instance of MapState, so they can get access to map's properties and be able to programmatically control the map.


johnpryan:
Thanks for the contribution! I've included this in #25

Issue #26, opened 2018-04-21
How to set Marker anchors?


squashmode:
I'm using markers on my flutter_map, using the built-in Icon class as the child, and they are drawn with the anchor at the top-left corner of the icon.  This is different than the FlutterLogo() that is used in the examples that has its anchor at the middle of the logo (and therefore is centered on the location of the marker).  I've played with Container alignment, Align class, etc, but cannot seem to change the anchor location of the drawn marker. 

Following the leaflet.js implementation, it would be great to be able to change the anchor of the drawn marker. See, e.g., http://leafletjs.com/examples/custom-icons/. for a good example.

Can this currently be done? If so, how, and if not, do you plan to add support for that?

Thanks!

squashmode:
Work fine in version 0.0.4.  The example app is excellent, thank you!

PR #27, opened 2018-04-23
Add marker anchor support

milare:
Implements #26

### DIFF
--- a/flutter_map/lib/src/layer/marker_layer.dart
+++ b/flutter_map/lib/src/layer/marker_layer.dart
@@ -13,11 +13,13 @@ class Marker {
   final WidgetBuilder builder;
   final double width;
   final double height;
+  final List<double> anchor;
   Marker({
     this.point,
     this.builder,
     this.width = 30.0,
     this.height = 30.0,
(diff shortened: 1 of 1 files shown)


johnpryan:
Thanks for the contribution!

Just a heads up, we are going to also make the offset calculation automatic:

```
      new Marker(
          width: 80.0,
          height: 80.0,
          point: new LatLng(53.3498, -6.2603),
          builder: (ctx) => new Container(
                child: new FlutterLogo(
                  colors: Colors.green,
                ),
              ),
          anchor: MarkerAnchor.center),
```

so that users won't need to think about dividing 80.0 / 2 and setting the anchor point manually.

Let me know if that doesn't work for your use case or you have any questions!

johnpryan:
New API: 48a5a3e06204c8c480e097da0c6ce7cf69070222

milare:
That's even better, thanks!

Leffe108:
It may help other people to know that this is how you specify anchorPos to Marker constructor.

```dart
Marker(
  anchorPos: AnchorPos.align(AnchorAlign.bottom),
  ...
```

PR #32, opened 2018-05-01
remove quiver dep

avioli:
Removed quiver dependency, since the only use is for the `hash3` method.

Copy/pasted the `JenkinsSmiHash` from flutter's `front_end` lib and added attribution where needed.

BTW: I'm wondering if this hash method is tested/proven useful for doubles, given it is hashing coordinates (which are usually doubles, not integers).

`dart:math` uses it internally, but doesn't expose the `hashX`, `combine` nor `finish` methods, so that is of no use. It's attribution is also missing, thus using the `front_end` codebase.


flutter_map/lib/src/layer/tile_layer.dart
```diff
@@ -395,5 +394,62 @@ class Coords<T extends num> extends Point<T> {
...
+/// are hash codes to be combined.
+///
+/// For ease of use, you may also use this pattern:
+/// `(new JenkinsSmiHash()..add(a)..add(b)....add(z)).hashCode`, where a..z are
+/// the sub-objects whose hashes should be combined.  This pattern performs the
+/// same operations as the performance critical variant, but allocates an extra
+/// object.
+class _JenkinsSmiHash {
+  /// Accumulates the hash code [value] into the running hash [hash].
+  static int combine(int hash, int value) {
```
johnpryan: This makes me think that this will only work with integers, not doubles. Have you tested with doubles in strong mode?

avioli: I've removed the `JenkinsSmiHash` altogether, since `dart:ui` exposes `hashValues()`

PR #36, opened 2018-05-10
Add backgroundColor option to tile layer, fix #35

### LINKED ISSUE #35: Map background color for unloaded tiles
Is there a way to use a more appropriate background color or placeholder tile for the map when tiles are loading?

image

kaetemi:


johnpryan:
thanks for the contribution!

kaetemi:
You're welcome :)

PR #50, opened 2018-06-08
Review/alfanhui/offline maps example + update to .gitignore to root directory

alfanhui:
Added offline map example code, including offline zoom capability. 
Added blurb to README to talk about how to use offline maps, and updated feature list check boxes
Updated .gitignore to root directory, and added temp file removals for /flutter_map_example/ DIR.


avioli:
Just a side note re git history - the offline map example tiles that are added in `f652dd0` and then some of them removed in `25fdfcd` are retained in history - the "removal" does not slim down the .git size, but simply does not create them after the latter commit. No need to comment here - I just wanted to point it out.

alfanhui:
Closing due to git history to cluttered with deleted files.

PR #76, opened 2018-07-26
Add double-tap to zoom

vinicentus:
Added the necessary code to support double-tap to zoom. Also added a dispose function. Note that it zooms in the center, not where the user tapped. See https://github.com/flutter/flutter/issues/10048.

Closes https://github.com/apptreesoftware/flutter_map/issues/73


johnpryan:
Looks good to me, thanks!  added https://github.com/apptreesoftware/flutter_map/issues/80 to track the todo.

PR #85, opened 2018-08-09
Project refactor and changes to offline map

LJaraCastillo:
* Project reorganization to match the current template for flutter plugins.
* Added mechanism to read the tiles from the local storage when in offline map.

Please check if the reorganization is something that workout for you guys.


johnpryan:
looks good to me - thanks for the PR and for your patience!

PR #94, opened 2018-09-05
Added polyline customisation options

RaimundWege:
I've added some polyline customisations like a border and the option to draw it as a dotted line:
custom_polyline


vinicentus:
+1

samatzhusup:
@RaimundWege 
Hello from 2022! 
this app source code in public? 
I'm interesting how realized draw walk point --> bus point logic)

PR #99, opened 2018-09-11
Expose map bounds

### LINKED ISSUE #87: Load only visible markers
How would one go about adding only the visible markers to the map. I basically want to have a list of markers, get the visible area of the map, and display only the markers that are within that box. I found that adding about 8300 markers was a bit too heavy...

xqwzts:
Provides a way for clients to get the current map bounds.

Fixes #87 

- Calculates the map bounds on `move`.
- Caches the value in `MapState` and exposes a `bounds` getter.
- Adds `bounds` property to `MapPosition`, and provides it upon triggering the `onPositionChanged` callback.
- Adds `bounds` getter to `MapController`/`MapControllerImpl`.
- Updates the MapController example with a button that shows the current bounds in a `SnackBar`.

### DIFF
--- a/example/lib/pages/map_controller.dart
+++ b/example/lib/pages/map_controller.dart
@@ -13,6 +13,7 @@ class MapControllerPage extends StatefulWidget {
 }
 
 class MapControllerPageState extends State<MapControllerPage> {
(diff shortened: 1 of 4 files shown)


johnpryan:
Thanks for the PR - I'll take a look as soon as I can

johnpryan:
Thanks for the contribution!

PR #107, opened 2018-09-17
Marker tap callback

tomwyr:
Adds `onTap` callback function to `MarkerLayerOptions` as a field. If it's specified, marker `Widget` is wrapped in `GestureDetector` that will trigger callback with the marker as parameter. Also, if marker widget builder wraps it in its own detector, then the ancestor callback won't be invoked.


vinicentus:
+1

gerry05:
Marker onTap isn't working if there's a polyline.

johnpryan:
Can you help clarify why this is a better or simpler API than users using a GestureDetector themselves?

tomwyr:
It makes markers creating process more straightforward by just specifying callback method which handle tap events. Users can still wrap markers in custom GestureDetector if they need to respond to different touch event. But yeah, this is rather minor improvement.

johnpryan:
I think we should leave this out of the API. Using a GestureDetector is the normal way of handling tap events, it also is more flexible because it can be used to handle more than just a single tap gesture (e.g. double tap gestures). Please let me know if I am misunderstanding this PR. Thanks for your contribution regardless!

tomwyr:
I think you've a got a point there. My intent was to provide simple way of registering a callback that is commonly used in map components, however, the "everything-is-widget" nature of Flutter makes it easy to add your own detector . And as you said, it's more flexible to allow user decide, what gesture event is he interested in listening to.

johnpryan:
Yeah, I appreciate the feedback. There's some ongoing discussions about how we can improve gesture detection here: https://github.com/johnpryan/flutter_map/issues/174

PR #109, opened 2018-09-17
Filter map position changes with distinct and debounce operators

tomwyr:
Map position changes are emitted via stream and transformed with `distinct` and `debounce`. It will prevent two or more identical consecutive positions to invoke callback multiple times and will allow to receive callback once the movement is finished (and ignore all preceding position changes), e.g.:
```
FlutterMap(
  options: MapOptions(
    // ...
    positionListener: PositionListener(
      debounce: Duration(milliseconds: 100),
      onChanged: (position) => print("Position changed: ${position.center}"),
    ),
  ),
)
```

### DIFF
--- a/lib/flutter_map.dart
+++ b/lib/flutter_map.dart
@@ -79,7 +79,7 @@ class MapOptions {
   final bool debug;
   final bool interactive;
   final TapCallback onTap;
-  final PositionCallback onPositionChanged;
+  final PositionListener positionListener;
   final List<MapPlugin> plugins;
   LatLng center;
   LatLng swPanBoundary;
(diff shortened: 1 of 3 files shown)


tomwyr:
It adds RxDart as a dependency to make use of `debounce` operator. I don't know how expensive it is for the project though.

johnpryan:
Thanks for the PR! At first glance, I'm not sure this makes sense to add to the map component itself... this seems like something that is up to the user to decide how to debounce the stream. That said, I'll have to take a closer look at this before i can have an informed opinion :)

tomwyr:
Listetning to an event when the map is fully settled seems to be quite common use case. I mean, current api `onPositionChanged` parameter may be a little bit misleading and suggest the movement is ended. But I also totally understand if you prefer to keep it raw, as it is.

PR #122, opened 2018-09-26
Update the version number in the README

bendiken:
The pubspec example in the README is badly out of date in its version number, which isn't good as people will tend to copy and paste that line as-is into their pubspecs.


johnpryan:
good catch, thanks!

PR #129, opened 2018-10-04
Added isStaticScale boolean option to Marker layer options

alfanhui:
When not set/true, marker pixel size remains constant during zooms. When isStaticScale is set to false, the marker size adjusts to the zoom level. 

Could be a potential fix for [issue 88](https://github.com/apptreesoftware/flutter_map/issues/88)

### DIFF
--- a/lib/src/layer/marker_layer.dart
+++ b/lib/src/layer/marker_layer.dart
@@ -61,15 +61,17 @@ class Marker {
   final double width;
   final double height;
   final Anchor _anchor;
+  final bool isStaticScale;
 
-  Marker({
-    this.point,
-    this.builder,
-    this.width = 30.0,
-    this.height = 30.0,
-    AnchorPos anchor,
-    Anchor anchorOverride,
-  }) : this._anchor = anchorOverride ?? new Anchor._(width, height, anchor);
+  Marker(
+      {this.point,
+      this.builder,
+      this.width = 30.0,
+      this.height = 30.0,
+      AnchorPos anchor,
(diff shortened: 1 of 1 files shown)


johnpryan:
> Could be a potential fix for issue 88

@marianoarga did this fix your issue?

marianoarga:
@johnpryan Will try this out and let you know

marianoarga:
Nope, it does not solve my issue, and I dont quite understand what is this PR solving, since isStaticScale=false and zooming out is not adjusting the marker's size properly for me. It is making the adjustment even worse, plese revisit this PR.
By the way, AnchorPos solved my issue, I just checked.

alfanhui:
This PR adjust the size of the marker in relation to the zoom scale, so that when you zoom out, the size of the marker is retained in the realworld. Good for marking areas.

marianoarga:
I see. It is not working as spected for me, sorry.

gimox:
Adjusting market size on zoom is good for me. I manually scaling my markers based on zoom value. Add the code in the framework can be useful for all.

simongomez95:
Is this not getting merged?

johnpryan:
I would prefer to keep the Marker API in line with the [leaflet API](https://leafletjs.com/reference-1.4.0.html#marker)

You can still implement this feature as a new layer type using the [plugin api](https://github.com/johnpryan/flutter_map/blob/master/example/lib/pages/plugin_api.dart). Let me know if you have any questions or this doesn't work for you

PR #131, opened 2018-10-06
Prevent map layers excessive builds

tomwyr:
Fixes the issue where all layers get rebuilt, whenever only one of them change its bounds https://github.com/apptreesoftware/flutter_map/issues/130.


johnpryan:
thanks for the contribution!

PR #139, opened 2018-10-20
to close #138 detect if map move depends on gestures

### LINKED ISSUE #138: Gesture detection, a way to detect user gesture
i need to locate when user move the map, or when the map is moved by the controller.
So "onPositionChanged" is called in both situation.

The case is this:
The map is updated on location change, so i use controller to change the center replacing the user location. this is a basic navigation app..
now i need to stop map update only if user shift/touch the map.
Both resulting in onPositionChanged, there is a way to detect it?

gimox:
Added hasGesture property on MapPosition. OnPositionChanged user can check if map is moved by controller or by user gestures.

### DIFF
--- a/lib/flutter_map.dart
+++ b/lib/flutter_map.dart
@@ -69,7 +69,7 @@ abstract class MapController {
 }
 
 typedef TapCallback(LatLng point);
-typedef PositionCallback(MapPosition position);
+typedef PositionCallback(MapPosition position, bool hasGesture);
 
 class MapOptions {
   final Crs crs;
--- a/lib/flutter_map.dart
+++ b/lib/flutter_map.dart
@@ -136,5 +136,6 @@ class MapPosition {
   final LatLng center;
   final LatLngBounds bounds;
   final double zoom;
-  MapPosition({this.center, this.bounds, this.zoom});
+  final bool hasGesture;
+  MapPosition({this.center, this.bounds, this.zoom, this.hasGesture = false});
 }
--- a/lib/src/map/map.dart
+++ b/lib/src/map/map.dart
@@ -20,8 +20,8 @@ class MapControllerImpl implements MapController {
     }
   }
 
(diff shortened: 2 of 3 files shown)


vinicentus:
+1

lib/flutter_map.dart
```diff
@@ -136,5 +136,6 @@ class MapPosition {
   final LatLng center;
   final LatLngBounds bounds;
   final double zoom;
-  MapPosition({this.center, this.bounds, this.zoom});
+  final bool hasGesture;
+  MapPosition({this.center, this.bounds, this.zoom, this.hasGesture = false});
```
johnpryan: a MapPosition is not an event, so I would prefer that the `onPositionChanged` stream emits a new object; `MapPositionChangedEvent` with two fields; the MapPosition and a bool representing whether or not the map has moved due to the current user.

gimox: ok now MapPositionChangedEvent emit two events

gimox:
new usage:

```
 onPositionChanged: (MapPosition mapPosition, bool hasGesture) {
                    if(hasGesture == true) {
                     ...
                    }
                  },
```

lib/src/map/map.dart
```diff
@@ -20,8 +20,8 @@ class MapControllerImpl implements MapController {
     }
   }
 
-  void move(LatLng center, double zoom) {
-    _state.move(center, zoom);
+  void move(LatLng center, double zoom, [bool hasGesture = false]) {
```
johnpryan: consider making this a named optional rather than an unnamed optional param

gimox: why?

johnpryan: just something to consider for readability

johnpryan:
Looks good, thanks for your patience!

AlexandreRoba:
This seems to be broken on version 0.5.0

PR #149, opened 2018-10-26
callback support for polyline, polygon & circle layers

LJaraCastillo:
* Added polygon, polyline & circle callback support for onTap & onLongPress events. **
* Removed "new" for instance creation in the project (the code looks better without them, they are optional anyway)
* Updated Gradle plugin and build.tools to 28.0.3 in the example project.
* Added pages in the example project for polygon & circles. (I used them to test the callback)

** Added multiple functions for this matter and part of them are Dart implementation of functions from android-maps-utils (https://github.com/googlemaps/android-maps-utils) in github, this functions are originally written in Java. The functions transcribed have a legend in the documentation of each functions. I don't know if this is a problem (license related, android-maps-utils have an Apache-2.0 license) but i hope not.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,7 @@
+# CHANGELOG
+
 ## [0.2.0] - 10/25/2018
+
 - Use NetworkImageWithRetry for tile layers (#145)
 - Add rebuild capability to LayerOptions (#144)
 - Added Circle layer (#137)
@@ -7,21 +10,24 @@
 Thanks to @kengu, @mortenboye, and @tomwyr for this release!
 
 ## [0.1.4] - 9/24/2018
+
 - Polygon Support (#118)
 
 Thanks to @JulianBerger for this release!
 
 ## [0.1.3] - 9/18/2018
+
 - fix identical map position callbacks (#111)
 - Prune tiles bug fix (#112)
 
 Thanks to @IhorKlimov and @tomwyr for this release!
 
 ## [0.1.2] - 8/21/2018
-- Added polyline customisation options (#94)
+
+- Added polyline customization options (#94)
 - Expose map bounds (#99)
 - Added onTap example (#103)
-- route bugfix (#104)
+- route bug fix (#104)
 - options is now required (#105)
 - Project refactor and changes to offline map #85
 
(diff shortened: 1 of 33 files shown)


kengu:
@LJaraCastillo, do you want to cooperate on finishing this feature? I'm in need of the same features you have implemented, and I don't wont to make a competing pull-request which essentially solves the same problem. The ultimate goal is to add  geometry editing support, which depends on gesture detection on geometries.

It is however, a bit hard to separate new features from refactoration and code cleanup since everything is in one single commit. You need to update the pull-request also to remove the current conflict with master branch. I guess @johnpryan will ask the same to be done before the pull-request is accepted.

LJaraCastillo:
i'm fine with a change on that matter because i wasn't so sure about my approach too. When i wrote the code i thought it will be more precise to use geodesic coordinates (i could be wrong) instead of screen coordinates. If you say is the same and more efficient i am up to change it.

kengu:
Great, so if it is OK by you, I can rework the solution into several commits, omitting some of the code cleanup and unrelated features like the new geometry identifier `key`? This makes it easier for @johnpryan to assess the primary contribution which is geometry gesture detection, without mentally performing `separation of concerns`.

kengu:
Adding support for geometry identifiers (aka `key`) looks to me like a separate pull-request. But I could be wrong, it depends on what @johnpryan prefers?

kengu:
@LJaraCastillo, my local branch [issue/geometry_gesture](https://github.com/DISCOOS/flutter_map/compare/master...DISCOOS:issue/geometry_gestures?expand=1) contains an alternative and stripped down version of you solution. Hit-tests on geometries are as proposed done on screen-coordinates (`offsets`), using the [Liang-Barsky algorithm](https://en.wikipedia.org/wiki/Liang%E2%80%93Barsky_algorithm) (adapted from [this gist](https://gist.github.com/ChickenProp/3194723))

LJaraCastillo:
i was checking it and is pretty similar but with offsets instead of geodesic coordinates. Everything looks good to me but I have doubts with all this, you are gonna make a PR with this change? should i close this PR in that case?

kengu:
Yes, if that's a solution that works for you, I can make a new pull-request and you can close off this one. 

Note that the only features that are missing in the new PR are

- Circle border fix (you added offsets in CircleLayer)
- PolygonLayer option `closeFigure`
- PolygonLayer and PolylineLayer option `displayPoints`

which all are nice additions, but off-topic and should be added as an separate PR.

LJaraCastillo:
I can make a PR with that missing stuff later, after your PR is merged. Those changes are not that important as the callback support.

kengu:
Great, I'll create the pull-request soon.

Issue #113, opened 2018-09-18
Support asymmetric bounds padding


csjames:
FitBoundsOptions currently takes a Point, perhaps an EdgeInsets would be more appropriate, especially given the parameters name is padding. 

Asymmetric padding is supported by Google maps and I believe it is a useful feature

I'll take a look at doing it myself if the boss things the padding looks crap,

Thanks, James :)

csjames:
Bump

johnpryan:
thanks for filing an issue. I don't currently have plans to work on this but if you would like to make a PR I agree this would make more sense.

FitBounds options I believe was part of the port from Leaflet

csjames:
I'll get it done, just need a bit of time  :)

Thanks! 
James

johnpryan:
Thanks for filing an issue. I went through the PR and nitpicked a few things. If you have time to fix them up I can include them in the next release.

Thanks!

johnpryan:
This is fixed now

PR #155, opened 2018-10-31
FitBoundsOptions EdgeInsets padding

csjames:
#113 

Asymetric padding has many use cases. EdgeInsets is idiomatic flutter way to provide this. The existing code was pretty much ready, I really only changed types

### DIFF
--- a/example/lib/pages/animated_map_controller.dart
+++ b/example/lib/pages/animated_map_controller.dart
@@ -12,7 +12,8 @@ class AnimatedMapControllerPage extends StatefulWidget {
   }
 }
 
-class AnimatedMapControllerPageState extends State<AnimatedMapControllerPage> with TickerProviderStateMixin {
+class AnimatedMapControllerPageState extends State<AnimatedMapControllerPage>
+    with TickerProviderStateMixin {
   // Note the addition of the TickerProviderStateMixin here. If you are getting an error like
   // 'The class 'TickerProviderStateMixin' can't be used as a mixin because it extends a class other than Object.'
   // in your IDE, you can probably fix it by adding an analysis_options.yaml file to your project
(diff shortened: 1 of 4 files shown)


lib/src/map/map.dart
```diff
@@ -154,10 +156,20 @@ class MapState {
...
-    var paddingTL = options.padding;
-    var paddingBR = options.padding;
+    var paddingTL = Point<double>(options.padding.left, options.padding.top);
+    var paddingBR =
+        Point<double>(options.padding.right, options.padding.bottom);
+
+    print(options.padding.left);
+    print(options.padding.right);
+    print(options.padding.top);
+    print(options.padding.bottom);
```
johnpryan: remove prints :)

lib/src/map/map.dart
```diff
@@ -154,10 +156,20 @@ class MapState {
...
+
+    var paddingTotalXY = paddingTL + paddingBR;
+
+    print(paddingTotalXY);
```
johnpryan: remove print :)

example/lib/pages/animated_map_controller.dart
```diff
@@ -34,23 +35,31 @@ class AnimatedMapControllerPageState extends State<AnimatedMapControllerPage> wi
...
 
     controller.addListener(() {
       // Note that the mapController.move doesn't seem to like the zoom animation. This may be a bug in flutter_map.
-      mapController.move(LatLng(_latTween.evaluate(animation), _lngTween.evaluate(animation)), _zoomTween.evaluate(animation));
-      print("Location (${_latTween.evaluate(animation)} , ${_lngTween.evaluate(animation)}) @ zoom ${_zoomTween.evaluate(animation)}");
+      mapController.move(
+          LatLng(_latTween.evaluate(animation), _lngTween.evaluate(animation)),
+          _zoomTween.evaluate(animation));
+      print(
+          "Location (${_latTween.evaluate(animation)} , ${_lngTween.evaluate(animation)}) @ zoom ${_zoomTween.evaluate(animation)}");
```
johnpryan: This is left over from another contribution, but in any case we should remove it

johnpryan:
ok I will take a look at this after the break - if you would like to file an issue to revert that would be a big help. thanks

PR #156, opened 2018-11-02
Fix interactive option, add decoration option, add epsg4326

JulianBerger:
Fixed the interactive option, I hope the way I did it is okay?!

Added a decoration option for the map container.

Added EPSG4326 projection the second time since the last time it conflicted with another pull request.

### DIFF
--- a/lib/src/layer/tile_layer.dart
+++ b/lib/src/layer/tile_layer.dart
@@ -52,6 +52,9 @@ class TileLayerOptions extends LayerOptions {
   ///Color shown behind the tiles.
   final Color backgroundColor;
 
+  ///Container Widget decoration
+  final Decoration decoration;
+
   ///Turns on/off the offlineMode.
   ///
   ///Requires the urlTemplate to target assets or a filesystem path.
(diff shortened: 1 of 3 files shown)


lib/src/layer/tile_layer.dart
```diff
@@ -53,6 +53,9 @@ class TileLayerOptions extends LayerOptions {
   ///Color shown behind the tiles.
   final Color backgroundColor;
 
+  ///Container Widget decoration
+  final Decoration decoration;
```
johnpryan: Why is this required? If you wanted decoration around the map, couldn't the map widget be wrapped in a Container?

JulianBerger: I added it to test some styling but you are right, its probably unnecessary.

lib/src/map/flutter_map_state.dart
```diff
@@ -52,11 +52,11 @@ class FlutterMapState extends MapGestureMixin {
           .map((layer) => _createLayer(layer, widget.options.plugins))
           .toList();
       return new GestureDetector(
-        onScaleStart: handleScaleStart,
-        onScaleUpdate: handleScaleUpdate,
-        onScaleEnd: handleScaleEnd,
+        onScaleStart: options.interactive ? handleScaleStart : null,
+        onScaleUpdate: options.interactive ? handleScaleUpdate : null,
```
johnpryan: Can you help me understand why this is required? I thought that if `handleScaleStart` is null the map will behave normally. I don't see why `interactive` is required for the API.

johnpryan:
fixed by #207 - can you make a separate PR for epsg4326? It's easier to get things merged when they are smaller changes.

PR #160, opened 2018-11-07
Adding callbacks for gestures on layer elements

kengu:
Adds the following features for Polyline, Polygon and CircleMarker elements:

- `onTap` and `onLongPress` gesture detection
- Refactored geometry build methods
- Conversion methods from LatLng to Offset and vice-versa

### DIFF
--- a/example/android/gradle/wrapper/gradle-wrapper.properties
+++ b/example/android/gradle/wrapper/gradle-wrapper.properties
@@ -3,4 +3,4 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-3.3-all.zip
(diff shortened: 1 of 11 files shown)


kengu:
@LJaraCastillo, here it the pull-request.

LJaraCastillo:
Nice

johnpryan:
hang tight, I'm trying to fix some merge conflicts here.

kengu:
@johnpryan, I'll look into #174. The merge conflicts are now resolved. The code is refactored into using the `ositioned_tap_detector` package introduced by the `zoom-focal-point` pull-request.

johnpryan:
I would like to close this and try to go with an approach like https://github.com/johnpryan/flutter_map/issues/174 that allows all layers to handle their own gesture detection. This will ultimately be the most flexible way to allow custom layers (like polygons) to decide if they've been tapped and whether or not to forward the gesture to the underlying gesture handler provided by the MapGestureMixin.

johnpryan:
Thanks for all the feedback and pull requests, please let me know if this doesn't work for you. I've assigned #174 to myself but if you would like to take a stab at it let me know!

PR #182, opened 2018-12-08
Updated README.md to have newest package

mattbajorek:
To make it easier for people to get up and running with the examples since they won't work with the older package.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ Add flutter_map to your pubspec:
 
 ```yaml
 dependencies:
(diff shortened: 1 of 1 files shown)


artob:
(I submitted a similar update previously in #122, for the same reasons.)

Given that this PR wasn't merged and the PR's version number update is already out of date with regards to the actually published [package](https://pub.dartlang.org/packages/flutter_map) on Pub, perhaps the thing to do here would be to remove version number information from the README altogether?

Issue #203, opened 2019-01-08
Use image cache for tiles images


SamuelRioTz:
Every tiles are downloaded again after every restart the app, it will be good if we use CachedNetworkImageProvider instead NetworkImageWithRetry.

CachedNetworkImageProvider is part of:
[https://pub.dartlang.org/packages/cached_network_image](url)

I already have an example working with both options:

> if(options.cachedTiles){
        return new CachedNetworkImageProvider(url);
      }else{        
        return new NetworkImageWithRetry(url);
      }

lnalex:
Feature released in 0.5.0 : https://github.com/johnpryan/flutter_map/pull/204

This issue can be closed.

PR #204, opened 2019-01-08
adding cached network image dependency for tiles

SamuelRioTz:
#203 
Add option( `cachedTiles=true` ) for use CachedNetworkImageProvider instead NetworkImageWithRetry.
Storage image tiles in cache memory, for keep the images tiles after restart the application even if there is no internet connection.


cetorres:
When do you think we're going to have this released?
Thanks.

PR #218, opened 2019-01-30
TileProviders instead of bools (offline/assets/filesystem)

avbk:
I needed a custom tile provider because I wanted to use a mbtiles file directly. Because it looked quite hacky to have another option, which overrides the other bools, I created a TileProvider and provided some default implementations based on the existing code.

I hope you like it :)

I could also provide the MBTileProvider, but since mbtiles is basically sqlite, it would add the sqflite dependency...


johnpryan:
Solid - thanks! I'll take a look when I can

avbk:
@cetorres sure, but I need to refactor it first, because currently is has dependencies to other parts in the app. 

@johnpryan do you want such a MBTile feature? If so, I would directly refactor it so that one can use it as a default tile provider

cetorres:
@avbk thank you! Please send it to me (cetorres@cetorres.com) or let me know the repo you put it.
I was also thinking on a tool, that could be apart, to allow downloading map tiles for offline use, directly from the app, with boundaries and zoom levels defined by the user in the app. It would download the tiles and directly store in the filesystem, on the normal z,x,y directory structure to use directly from the plugin. What do you think?

johnpryan:
@avbk can you provide an example repo to test this change with?

avbk:
@johnpryan just fixed the conflicts. I migrated the `cacheTiles` bool into a `CachedNetworkTileProvider`, which is used by default. I think I can provide an example in a separate branch.

avbk:
Done. See #240

avbk:
I was just thinking, since the change breaks public api by removing the bools -- should we upgrade the version from to 0.6.0 ? If you'd like I can do so and also try to update the dependencies. Thereby #234 would be fixed as well... What do you think?

cetorres:
Hi @avbk, can you send me the mbtiles provider code, please?

avbk:
@cetorres it's already linked above (#240)

cetorres:
Sorry, @avbk I didn't see before. Thank you!!

avbk:
@johnpryan have you found any time to review the feature yet? Do you need any further support?

johnpryan:
Is there a way to test this change? maybe an example repo?

avbk:
I've replaced the offline page in the example with the new approach. In #240 I've also added an example with an offline MBTile Map of Berlin. Do you need any other examples?

PR #232, opened 2019-02-21
Fix for setting map interactive. 

wmcshane:


johnpryan:
fixed by #207 - thanks for the contribution!

PR #237, opened 2019-02-28
Is user gesture

wmcshane:
added `isUserGesture`

There may be a better term for this, I wanted a way to distinguish between the user panning and zooming in and out on the map vs movement triggered via buttons or a listener.
 
The use case for this is to allow the my location following as seen in google maps. When you click the button to goto your location it highlights the icon on the button to indicate that it is fixed to your location. When the user pans or zooms the location following turns off. However, you don't want it to turn off when the map controller calls move.

Distinguishes between direct map actions and external controller movements.

I have added an example page to demonstrate.


example/lib/pages/follow_my_location.dart
```diff
@@ -0,0 +1,370 @@
...
+import 'package:flutter/services.dart';
+
+//////////////////////////////
+// To set emulator location //
+//////////////////////////////
+// goto terminal
+// > adb devices
+//
+// telnet to device using the port associated with the device listing from the previous command
+// > telnet localhost <port>
```
johnpryan: Can we remove this example? If you like, I can link to this example in the README. It's a little complicated to run for folks just wanted to see some use-cases.

There's probably a better way to demonstrate this with some fake interactions using streams: https://dart.dev/articles/libraries/creating-streams

wmcshane: That's fair, I'll remove the example and if i get a chance at some point i'll put together a cleaner one. The main thing is having the isUserGesture distinction. 

Thank you for the feedback. :)

wmcshane: Removed and rebased.

PR #243, opened 2019-03-05
make sw and ne private

vinicentus:
It is unnecessary to both expose the variables (`sw` and `ne`) and have separate getters (`southWest` and `northEast`) for them. It might even be confusing since someone might wonder what the difference between them is. Therefore I suggest that we make the variables sw and ne private.


vinicentus:
...This doesn't really belong here but I noticed that my username was written incorrectly in the changelog. Can you fix it in the next release @johnpryan?

johnpryan:
fixed the mispelling :) - I'll take a look at this when I get a chance

PR #265, opened 2019-04-13
Updated README to run the example.

manhluong:
### Why this PR?

Executing `flutter run` in the example folder, with a running iPhone XR simulator, causes the following error:

```
Launching lib/main.dart on iPhone XR in debug mode...

Oops; flutter has exited unexpectedly.
Sending crash report to Google.
```

From the log, I can read:

```
FileSystemException: FileSystemException: Getting current working directory failed, path = '' (OS Error: Too many open files, errno = 24)
```

Which is the problem reported here: https://github.com/trentpiercy/trace/issues/1

### How this PR fixes the issue?

Updated the README with the instructions I found in the link.

### What are the expected outcomes after the merge?

It is possible to run the example app in the iOS simulator.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,8 +55,13 @@ Configure the map using `MapOptions` and layer options:
   }
 ```
 
-See the `flutter_map_example/` folder for a working example app.
+### Run the example
 
+See the `example/` folder for a working example app.
+
+To run it, in a terminal cd into the folder.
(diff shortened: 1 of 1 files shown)


johnpryan:
thanks!  I think this is happening because the example uses a lot of tiles... added an issue to track https://github.com/johnpryan/flutter_map/issues/269

PR #270, opened 2019-04-24
Linter for exemple project

lsaudon:


johnpryan:
thanks! Out of curiosity, were you seeing the linter rules defined in the package applied to the example?

lsaudon:
Yes, I didn't add the rules in the example it's the project rules.
It is possible to exclude folders.

PR #275, opened 2019-05-09
added minimal zoom level for marker

Isolus:
I have implemented a minimum zoom level for markers.
A marker only becomes visible at minZoom or higher.  This is useful if you have many markers that are only important on detailed maps and zoom out.
Maybe someone else finds the function useful.


johnpryan:
I don't think this is something we want to add to the API. Instead, you can listen to onPositionChanged on MapOptions, and change the markers you are rendering based on the MapPosition's `zoom` field

PR #284, opened 2019-05-20
Update .gitignore for the example

smasato:
Ignore files generated when running example on iOS.


johnpryan:
thanks for the contribution!

Issue #230, opened 2019-02-18
When the anchor point is out  of bounds, the whole marker widget disappears


aytunch:
![simulator screen shot - iphone xr - 2019-02-18 at 23 45 19](https://user-images.githubusercontent.com/6442915/52975740-b211b180-33d7-11e9-91c9-df4e711547db.png)
![simulator screen shot - iphone xr - 2019-02-18 at 23 45 28](https://user-images.githubusercontent.com/6442915/52975743-b63dcf00-33d7-11e9-9e2e-35fc467cbf52.png)

You can see from the second image that left marker is lost with the slightest pan. The moment the little triangle of marker leaves out of any side of the screen, the whole widget disappears. Is there a way to keep it on the map like a safe area or setting a margin for marker?

aytunch:
I had a chance to experience this issue in a real device using hand gestures instead of the mouse and it really bothers me alot to see a bunch of disappearing markers on the screen. Very buggy feeling. I hope there is already a solution or a quick fix. The reason I switched from flutter google maps to flutter maps and mapbox was mainly the power of using custom widget markers(many thanks). And i assume most people will use markers that has a good amount of width like the ones i have instead of default location markers.

rayrrr:
A much easier approach would be to know the width of the marker widget, divide it in half, and then use that number (marker widget * 1/2) as a "padding width" around all 4 edges of the visible area. Then it will behave exactly as we all want, without some complicated mastery of knowing which marker widget is where at all times.

johnpryan:
I think this is an easy fix, thanks for the feedback.

aytunch:
@johnpryan John, were you able to look into this problem? It would be great to have that working:)

aytunch:
> I think this is an easy fix, thanks for the feedback.

@johnpryan Sorry for reminding. I am sure you are very busy. I am reminding again since you said this requires an easy fix. I want my big markers not to disappear:D

aytunch:
@johnpryan Do you think wrapping `FlutterMap` inside of an `SizedBox` which is larger than the screen size or just wrapping it in a container with margin/padding would solve this issue? And also would it have any side effects in terms of for i.e. performance?

johnpryan:
I have not found time to think about the approach. If you have any ideas I would be happy to review a pull request.

aytunch:
[...]

PR #306, opened 2019-06-05
expose centerzoom, public getBoundsCenterZoom and Anchor.forPos

lpongetti:
Hi,
i'm creating a marker cluster plugin with all feature like animation and spiderfy, and this two edits can help me for don't rewrite you code.

### DIFF
--- a/lib/src/map/map.dart
+++ b/lib/src/map/map.dart
@@ -134,7 +134,7 @@ class MapState {
     if (!bounds.isValid) {
       throw Exception('Bounds are not valid.');
     }
-    var target = _getBoundsCenterZoom(bounds, options);
+    var target = getBoundsCenterZoom(bounds, options);
     move(target.center, target.zoom);
   }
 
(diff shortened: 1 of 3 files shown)


lib/src/map/map.dart
```diff
@@ -134,7 +134,7 @@ class MapState {
     if (!bounds.isValid) {
       throw Exception('Bounds are not valid.');
     }
-    var target = _getBoundsCenterZoom(bounds, options);
+    var target = getBoundsCenterZoom(bounds, options);
```
johnpryan: I think a better way would be to expose this method on `MapControllerImpl` and `MapController` instead. Their purpose is to expose the methods on `MapState`

lpongetti: Right, my bad. now its ok

lpongetti:
Ok i added getBoundsCenterZoom to mapcontroller

lpongetti:
You have release version 0.6.0 from dev branch... dev branch doens't support channel stable!!

lpongetti:
Ok, i correct some things..
1) polygon with border didn't close the polygon
2) #230 marker will removed only if totally disappear from screen

i saw better what i need and i need getBoundsCenterZoom public  from mapstate public because in my plugin i have mapstate and not mapcontroller.

lib/src/map/map.dart
johnpryan: I see, thanks for the clarification.

PR #309, opened 2019-06-06
Add plugin reference flutter_map_marker_cluster in readme

lpongetti:
Hi, i created plugin flutter_map_marker_cluster.
I think that it will be usefull for many people, so i would add reference.


johnpryan:
splendid!

PR #322, opened 2019-06-10
Useful for using custom assets based on coords

quantosapplications:
This provider can be useful when working with local assets which does not follow x, y, z naming.


quantosapplications:
I'm trying to extend this creating a new kind of layer which can render a custom widget based on coords.

For doing this I need the same logic that tile_layer expose except for the build function that is binded to render an image widget in the _createTileWidget method

`final Widget content = Container(
      child: FadeInImage(
        fadeInDuration: const Duration(milliseconds: 100),
        key: Key(_tileCoordsToKey(coords)),
        placeholder: options.placeholderImage != null
            ? options.placeholderImage
            : MemoryImage(kTransparentImage),
        image: options.tileProvider.getImage(coords, options),
        fit: BoxFit.fill,
      ),
    );`

Instead of this I would like to render a custom widget which can be built passing a function within the options like this

`final content = widget.options.customWidget(coords);`

Do you think that I can override someway the existing tile_layer or do I have to create a new layer for this?
Do you think it can be useful for the library?

johnpryan:
lgtm, thanks!

quantosapplications:
Thanks to you!
Any hint for making a widget layer?

johnpryan:
have you taken a look at the plugin example?

PR #328, opened 2019-06-12
Fix for type '_BroadcastStream<void>' is not a subtype of type 'Stream<int>'

ZeBrody:
Fix for this issue:
https://github.com/johnpryan/flutter_map/issues/311


edufolly:
Waiting for the release.

PR #341, opened 2019-06-25
Changed Example Overlay Image Appbar title to match Drawer title

joshwalker9115:
Minor cosmetic update:
I noticed the the Appbar title on the Overlay Image page was 'Circle' and updated it to 'Overlay Image' so it matches its title in the Drawer widget.  

Thanks for the awesome Repo!


johnpryan:
thanks!

PR #358, opened 2019-07-14
Update pubspec.yaml

whiteshadoww:
upgrade cached_network_image


johnpryan:
Are there breaking changes in 1.1.0? Technically, this change will break folks who want to stay at 1.0.0 and don't want to opt-in to 1.1.0.

Issue #116, opened 2018-09-20
Implement map rotation


vinicentus:
It would be nice to be able to rotate the map.

Take a look at https://github.com/flutter/flutter/pull/17345.

The `mapOptions` could also take an optional argument `allowMapRotation`, that would default to true. Maybe everyone doesn't need or want map rotation...

johnpryan:
I think the challenge here will be to determine what "box" the map tiles need to render in. (e.g. rotating the map may result in tiles in the viewport not being fetched)

kinda similiar to this problem: https://stackoverflow.com/questions/10392658/calculate-the-bounding-boxs-x-y-height-and-width-of-a-rotated-element-via-jav/11771163 except the blue rectangle needs to be large enough such that the red rectangle (the viewport) is covered.

i think Flutter supports rotation so i don't see that piece being a huge effort.

vinicentus:
Ok, I see. What info do we get? The coordinates of the viewport and the rotation?

Then we probably also want to filter out the tiles that are colored in the picture, since they are not needed.

![img_20181001_130744](https://user-images.githubusercontent.com/17357504/46282920-7e92cf80-c57b-11e8-9795-6e4fccf0e832.jpg)
(apologies for the large and ugly picture)

johnpryan:
Yes exactly. Thanks for making an illustration. I'm going to mark this as "help wanted" since this is a slightly challenging problem!

ibrierley:
Just thinking about this, rotation would be nice, but what about things like text on a tile, wouldn't that mean it's now upside down if we rotate 180, would that be a bad experience ? Would it only be useful for tiles without text, or vector tiles or something ?

vinicentus:
@ibrierley I think it would still be useful even with text on tiles. Humans are pretty good at reading stuff that is rotated.

I have noticed that google maps automatically rotate street names (maybe it's a separate layer or something?). That is of course the optimal solution but I don't think many free map providers offer this? (Correct me if I'm wrong)

johnpryan:
Yes I suspect that is one of the reasons Leaflet does not support rotation.

escamoteur:
We will need this functionality in our project. So I will have a closer look at this.

escamoteur:
Not sure if that is enough for you needs. It only allows to rotate the map from the App side.

johnpryan:
This issue still isn't fully resolved AFAICT. There's still work to be done to make sure that tiles on the edge of the viewport are displayed

[...]

PR #359, opened 2019-07-16
rotation added

escamoteur:
Hi, 
I added rotation functionality to the map.
Not exactly like asked in https://github.com/johnpryan/flutter_map/issues/116

Because I don't see a lot of benefit of enabling the user to rotate the map but to rotate the map from the app side e.g. connected to the compass.
So the mapController got a rotate function and the map option an initial rotation.
The rotation is done at the outmost scope so I rotate the whole map with all it's layer because it don't make sense for a single layer not to rotate.

I did not change the version, I let that to you when publishing :-)

I hope you like it.

### DIFF
--- a/example/lib/pages/map_controller.dart
+++ b/example/lib/pages/map_controller.dart
@@ -20,6 +20,7 @@ class MapControllerPageState extends State<MapControllerPage> {
   static LatLng dublin = LatLng(53.3498, -6.2603);
 
   MapController mapController;
+  double rotation = 0.0;
 
   @override
   void initState() {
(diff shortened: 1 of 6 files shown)


escamoteur:
Special thanks to  @slightfoot for the idea using an Overflowbox

johnpryan:
This is really cool - thanks!

I think there's some usability issues that need to be fixed before this can be an official feature. The map jumps as you pan around when there's rotation. The tiles that are fetched for a rotated map can still lead to blank spots (as discussed in #116) since the visible area needs to take into account the rotation now.

escamoteur:
Hi John,
I tried it here and had no such problems. The visible area does take this into account. By this:

`
        angle = degToRadian(rotation);
        final rangle90 = sin(_rad90 - angle).abs();
        final sinangle = sin(angle).abs();
        // to make sure that the whole screen is filled with the map after rotation
        // we enlarge the drawing area over the available screen size
        width = (constraints.maxWidth * rangle90) +
            (constraints.maxHeight * sinangle);
        height = (constraints.maxHeight * rangle90) +
            (constraints.maxWidth * sinangle);

        mapState.size = CustomPoint<double>(width, height);
`

escamoteur:
Ah, I now saw what you meant with the jumping around. it has to do with the fling that isn't rotated.

PR #378, opened 2019-07-29
Update and remove packages

lsaudon:
Hello,

I noticed that some packages were not being used.
I also updated the packages that could be updated.

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -13,17 +13,12 @@ dependencies:
     sdk: flutter
   tuple: ^1.0.2
   latlong: ^0.6.1
-  positioned_tap_detector: ^1.0.2
+  positioned_tap_detector: ^1.0.3
   transparent_image: ^1.0.0
-  async: ^2.1.0
   flutter_image: ^2.0.0
-  cached_network_image: ^1.0.0
-  sqflite: ^1.1.5
-  path_provider: ^1.1.0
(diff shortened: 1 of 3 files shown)


pubspec.yaml
```diff
@@ -13,17 +13,12 @@ dependencies:
...
   latlong: ^0.6.1
-  positioned_tap_detector: ^1.0.2
+  positioned_tap_detector: ^1.0.3
   transparent_image: ^1.0.0
-  async: ^2.1.0
   flutter_image: ^2.0.0
-  cached_network_image: ^1.0.0
-  sqflite: ^1.1.5
-  path_provider: ^1.1.0
-  vector_math: ^2.0.0
```
johnpryan: I don't think these are unused:

- `path_provider` and `sqlflite` are used in `src/layer/mbtiles/mbtiles_image_provider.dart` 
- `vector_math` is used in `src/gestures/gestures.dart`
- `package:async` is used by `src/map/flutter_map_state.dart`

Could you add these back?

lsaudon: Sorry, they are used in other packages so present. async is also used.
My PR is not relevant. 

Thank you for your feedback.

lsaudon:
Sorry, they are used in other packages so present. async is also used.
My PR is not relevant. 

Thank you for your feedback.

PR #379, opened 2019-07-30
fixed exception during initialization

escamoteur:
Hi,

I observed that when you have enabled 'break on all exceptions' that flutter_map threw an exception while initializing due to that `_lastCenter` was not initialized. I changed a little bit the sequence and made sure that in this case `move`  is executed which initializes `_lastCenter` correctly.

IMHO it would be worth to have a closer look over the whole initialization process sometimes.

### DIFF
--- a/lib/src/map/map.dart
+++ b/lib/src/map/map.dart
@@ -88,11 +88,11 @@ class MapState {
 
   set size(CustomPoint s) {
     _size = s;
-    _pixelOrigin = getNewPixelOrigin(_lastCenter);
     if (!_initialized) {
       _init();
       _initialized = true;
     }
+    _pixelOrigin = getNewPixelOrigin(_lastCenter);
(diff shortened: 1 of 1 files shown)


lib/src/map/map.dart
```diff
@@ -115,7 +115,7 @@ class MapState {
     zoom = _fitZoomToBounds(zoom);
     final mapMoved = center != _lastCenter || zoom != _zoom;
 
-    if (!mapMoved || options.isOutOfBounds(center) || !bounds.isValid) {
+    if (_lastCenter != null && (!mapMoved || options.isOutOfBounds(center) || !bounds.isValid)) {
```
johnpryan: I think this file needs to be formatted - can you run dartfmt?

escamoteur:
Done

Issue #381, opened 2019-07-31
OverlayImage with transparency has no transparency since last update


connybruhn:
Hi,
many thanks for the great library. I'm working on showing precipitation radar images on a map using OverlayImages.

The previous version was able to display radar images (PNG with transparent areas), but, as we know, it was throwing some errors.

I was very happy to see the bug fix for https://github.com/johnpryan/flutter_map/issues/294. 

But now, after updating flutter_map, the image has no more transparency.

I attach 2 screenshots to show the difference:

- flutter_map: 0.5.6 (you see transparency on areas without precipitation)
- flutter_map: 0.7.0 (no transparency, you can only see a little bit of the background, because opacity of overlay image is 0.9)

![flutter_map_0 5 6](https://user-images.githubusercontent.com/24451312/62217144-631a7000-b3aa-11e9-8278-f85feabe3d8d.png)
![flutter_map_0 7 0](https://user-images.githubusercontent.com/24451312/62217163-6d3c6e80-b3aa-11e9-808b-26a1a7c2e133.png)

![radar_overlay](https://user-images.githubusercontent.com/24451312/62217291-b096dd00-b3aa-11e9-95ba-5a51daf16288.png)

Do you need more information?

Conny

johnpryan:
Looking at the second screenshot, it  seems like the images might be getting stacked up on top of each other over and over. Can you check the number of values being passed in to `OverlayImageLayerOptions.overlayImages`? This could certainly be a but in OverlayImageLayer, but it's hard to tell

johnpryan:
Thanks @4kssoft for the fix - it been published to Pub. Please re-open if you are still seeing the issue

connybruhn:
Many thanks for the very fast bug fix. With the newest version it is working perfect.

PR #382, opened 2019-07-31
Fix OverlayImage with transparency #381

4kssoft:


johnpryan:
fixes https://github.com/johnpryan/flutter_map/issues/381

PR #400, opened 2019-08-22
chores: presentation and documentation.

GregorySech:


GregorySech:
This is still a work in progress but I do not mind opening other PRs if early merges are preferable or for any other reason.

lib/flutter_map.dart
```diff
@@ -90,13 +92,22 @@ typedef void TapCallback(LatLng point);
...
+/// Through [crs] the Coordinate Reference System can be
+/// defined, it defaults to [Epsg3857].
 class MapOptions {
   final Crs crs;
   final double zoom;
   final double rotation;
   final double minZoom;
   final double maxZoom;
-  final bool debug;
+  final bool debug; // TODO no usage outside of constructor. Marked for removal?
```
johnpryan: yes we can either remove this or mark it as `@deprecated`

johnpryan:
I think Travis is failing because`lib/src/layer/tile_layer.dart` needs to be formatted with `dartfmt`

GregorySech:
If it is alright with you I was thinking of moving around some classes, maybe create a `lib/src/layer/tile_providers` directory and place MBTilesImageProvider and the ones that are in tile_layer.dart there.
Exporting them into tile_layer.dart might be a good way to mitigate imports breakage, not enough if someone is importing `lib/src/layer/mbtiles/mbtiles_image_provider.dart` directly.

lib/flutter_map.dart
GregorySech: i was feeling a little pedantic and deprecated it. Maybe it could be removed at the first API change for MapOptions.

johnpryan:
sounds good to me

johnpryan:
LGTM - ready to merge if you are

GregorySech:
I'm trying to understand the MapPlugin API. 
MapPlugin has the supportsLayer method that takes LayerOptions and returns true if the plugin supports it.
Its other method is createLayer that returns the layer's Widget. This method takes the MapState, a Stream<Null> stream and LayerOptions.
I do not get why LayerOptions contains a rebuild stream and what is the stream given to createLayer is for.
My confusion comes from the _merge method that is used to produce createLayer.stream argument. 
I get that it that it is supposed to "fallback" to be the mapState.onMoved method but isn't it always creating new controllers every time _merge is called?

GregorySech:
Ah sorry didn't read the previous comment. For me it's ok.

GregorySech:
My pleasure.
I'll open a new PR for docs and small changes. 
By looking around the issues seems that the plugin API could be potentially revisited and that you might need help doing that.
Should we get a new issue to discuss what to focus on for the next release (0.8?) and what to postpone? 
The kanban is a good strategy once planning is done but some kind of roadmap will be needed. 
Deciding what to do in the "near" term could be a step in the right direction.

johnpryan:
Yeah the [Kanban board](https://github.com/johnpryan/flutter_map/projects/1) needs to be adjusted a little (and is only editable by maintainers). I like the idea of an issue do discuss priorities for the project.

PR #401, opened 2019-08-22
Make it easier to extend TileProvider

kengu:
CustomTileProvider only gives allows for custom url-scheme. This PR make it possible to extend TileProvider to make your own TileProvider implementations without the use of internal apis.

### DIFF
--- a/example/.gitignore
+++ b/example/.gitignore
@@ -9,3 +9,4 @@ ios/.symlinks/
 packages
 pubspec.lock
 .flutter-plugins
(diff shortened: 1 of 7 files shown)


kengu:
Sure @johnpryan, I'll lokk into it shortly

kengu:
There, @johnpryan - all done.

johnpryan:
Can you double check if these changes are correct? I don't see any changes to TileProvider

kengu:
There, take two. This time with changes.

PR #407, opened 2019-08-31
Fix/transparent polyline

### LINKED ISSUE #347: Polyline transparency denser at each point.
Love this widget -- thanks so much.
When drawing a polyline with transparency each point it twice as dense as the line - I think the transparency should be uniform along the length of the line.  I suspect that the line is been drawn from point A to Point B then from Point B to Point C -- so that Point B gets drawn over twice.

See image, I think this line should not have the darker red circles.

Thanks :)

Polyline(
          color: Color.fromRGBO(255, 0, 0, 0.5),
         strokeWidth: 20,
          isDotted: false,
             points: [array]
Screenshot_1561765845

2ZeroSix:
fix #347

preview:
image

### DIFF
--- a/lib/src/layer/polyline_layer.dart
+++ b/lib/src/layer/polyline_layer.dart
@@ -1,5 +1,5 @@
 import 'dart:math';
-import 'dart:ui';
+import 'dart:ui' as ui;
 
 import 'package:flutter/widgets.dart';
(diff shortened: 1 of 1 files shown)


2ZeroSix:
@johnpryan Hi, I noticed that the work on pull requests slowed down a bit. Can I do something to make this pull request merged?

PR #414, opened 2019-09-08
Add more conservative boundary calculation

Lootwig:
In order to work with locally stored maps, I needed a way to prevent tiles outside of the boundaries to be loaded. The current calculation for isOutOfBounds() is too simple here, because it only checks the center - this ignores that with the center still "inside", there can still be up to 3/4 of the screen outside the specified bounds.

The new extension of `MapOptions` is supposed to fix this - BUT: There is still an error somewhere in the calculations of screen width/height in degrees that I have yet to figure out. Currently, the `corners` calculated are way outside of the actual displayed area. For the width, line 196 seems to require a division by 4 rather than 2 in order to yield the correct coordinates at the left/right screen sides. For height, it seems to be something like `/ 3` instead of `/ 2`, at least on an iPhone Xr simulator.

I'm still trying to figure out where I took a wrong turn, maybe someone can help me out here.

### DIFF
--- a/lib/flutter_map.dart
+++ b/lib/flutter_map.dart
@@ -1,6 +1,7 @@
 library flutter_map;
 
 import 'dart:async';
+import 'dart:math';
 
 import 'package:flutter/material.dart';
 import 'package:flutter/widgets.dart';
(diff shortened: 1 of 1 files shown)


Lootwig:
I have absolutely no idea why the screen height calculation needs the additional "*4/3" factor.

Also, pipeline seems to be broken.

PR #424, opened 2019-09-17
Fix error when unproject bottomLeft or topRight  and lat are < -90 or > 90 or lng are < -180 or > 180

lpongetti:
I'have a tablet and when i put FlutterMap with zoom: 1 and minzoom: 1, debug catch a throw in LatLng class because latitude or longitude are invalid ( lat < -90 or > 90 or lng < -180 or > 180). So i inserted a control before, when i unproject the point in pixel. With this fix it doesn't give error when i drag the map outside to the screen.

### DIFF
--- a/lib/src/geo/crs/crs.dart
+++ b/lib/src/geo/crs/crs.dart
@@ -1,5 +1,7 @@
 import 'dart:math' as math;
 
+import 'package:flutter/cupertino.dart';
+import 'package:flutter/rendering.dart';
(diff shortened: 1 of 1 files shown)


lpongetti:
This fix is a very importat for my project, so please to check and if it's all ok create new version.

lpongetti:
Hi @johnpryan, are you still here?
Can you check this pull request pls? It's very important fix

PR #427, opened 2019-09-21
Resolve UI overflow

Varuni-Punchihewa:
Resolve the UI overflow occurred at "Marker Anchor Points" screen.

Screenshot_20190921-220452


Varuni-Punchihewa:
reference: https://github.com/johnpryan/flutter_map/issues/428

PR #430, opened 2019-09-23
Slide map along map boundaries

Lootwig:
This gives users the option to replace the behavior where, on reaching the boundaries while panning, the movement is simply canceled. Instead, if there is a free movement axis, the pan movement vector's component along that axis is still used.

Boundaries calculations are additionally extended to account for the actual screen boundary rather than just checking whether the center is inside/outside, in order to simplify working with downloaded tiles. This ensures that, as long as the currently visible map slice is smaller than the boundaries, there will never be a need for tiles outside those boundaries.

Caching a few tiles outside the boundaries is recommended for cases where the user is allowed to zoom out so much that the entire area fits on the screen.


johnpryan:
Thanks for the PR - would you mind making this an additional option on MapOptions instead of making a new subclass? I would also like to see an example and maybe some tests.

Lootwig:
@johnpryan I did add an example for the boundary sliding feature,

However I am not sure if another one for the adaptive boundary calculation makes sense. I built the feature in order to allow tile downloading for offline maps using only tiles actually covering area _inside_ the specified boundaries (without the need for buffer tiles that are attempted to load when zooming out). This is hard to demonstrate inside an example in my opinion, do you disagree?

johnpryan:
Yeah it's potentially confusing for users, so if there's a way to exemplify it I think it helps give an explanation.

Lootwig:
The example would differ from the "offline map" one exclusively by the number of tiles in the assets folder 🤔 I don't see how this would be more useful than elaborating in the parameter documentation?

johnpryan:
I see your point about the sample, thanks for adding it anyway 😆

Issue #290, opened 2019-05-22
Add example showing current location


johnpryan:
use https://pub.dev/packages/location to display the user's current location as a Marker

aytunch:
@johnpryan This would be perfect as to make the library more of a "whole" solution. Personally, I use [geolocator](https://pub.dev/packages/geolocator) and I am very glad with that.

johnpryan:
I think this is something we can include as an example but not part of the library

maRci002:
Solved, see #447

PR #447, opened 2019-10-16
Issue 290 : Add example of showing current location

raacker:
#290  Add example page that use [flutterlocation](https://pub.dev/packages/location) to achieve device's location and update it in live. 

* I need a confirm that upgrading sample app's requirements is okay or not. Because I had really hard time with debugging "location" library and finally got a reason why it was not running okay with flutter_map sample app.

https://pub.dev/packages/location#breaking-changes

as described on README page, location version 2 uses AndroidX and SDK 28 to run it. 

If you guys think it's too unnecessary, I might can try "location" version 1 to integrate.

I will include build.gradle commits for now.

### DIFF
--- a/example/android/app/build.gradle
+++ b/example/android/app/build.gradle
@@ -15,8 +15,7 @@ apply plugin: 'com.android.application'
 apply from: "$flutterRoot/packages/flutter_tools/gradle/flutter.gradle"
 
 android {
-    compileSdkVersion 25
-    buildToolsVersion '25.0.3'
+    compileSdkVersion 28
 
     lintOptions {
         disable 'InvalidPackage'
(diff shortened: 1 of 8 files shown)


FrankenApps:
Does the map get centered again when the position changes?
Or is the first location the center of the map forever (and only the marker moves)?

raacker:
@FrankenApps I will double check how I wrote the code but it will center to marker position if you see the code. Thanks for mentioning that.

raacker:
@FrankenApps I will double check how I wrote the code but it will center to marker position if you see the code. Thanks for appointing that.

johnpryan:
Thanks!

Issue #428, opened 2019-09-21
"Marker Anchor Points" screen overflows


Varuni-Punchihewa:
![Screenshot_20190921-220452](https://user-images.githubusercontent.com/27306185/65376542-97c9ea00-dcbe-11e9-85cc-c1b76c5be107.png)

PR #448, opened 2019-10-16
Issue 428 : Marker Anchor Sample Page's right is overflowed

raacker:
#428   Make them to be placed in screen width range.

### DIFF
--- a/example/lib/pages/marker_anchor.dart
+++ b/example/lib/pages/marker_anchor.dart
@@ -67,6 +67,9 @@ class MarkerAnchorPageState extends State<MarkerAnchorPage> {
       ),
     ];
 
+    double screenWidth = MediaQuery.of(context).size.width;
+    double minWidth = (screenWidth - 16) / 4;
+
     return Scaffold(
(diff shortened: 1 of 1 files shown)


example/lib/pages/marker_anchor.dart
```diff
@@ -67,6 +67,9 @@ class MarkerAnchorPageState extends State<MarkerAnchorPage> {
       ),
     ];
 
+    double screenWidth = MediaQuery.of(context).size.width;
+    double minWidth = (screenWidth - 16) / 4;
```
johnpryan: Have you tried replacing the `Row` widget with a [Wrap](https://api.flutter.dev/flutter/widgets/Wrap-class.html)?

raacker: image

Without minWidth, Wrap container will generate this layout. Well we might can just move "Bottom" to 2nd row. Or just keep on with current fix?

raacker: @johnpryan Oh and I'm very sorry about duplicated PR. #427 already tried on this issue. Will close this PR as soon as confirmed

PR #449, opened 2019-10-17
Polygon culling

wmcshane:
Added Polygon culling based on screen bounds. 

If you set polygonCulling to true on PolygonLayerOptions then any polygons outside of the screen’s bounds won't be drawn, therefore, saving rendering time and improving performance when displaying a large set of polygons.

Additionally, bounding boxes have been added to the Polygon object, this makes the logic for determining if a polygon is overlapping the screen very quick.


johnpryan:
Hi @wmcshane, can you add a short description and fix the formatting issues? thanks!

wmcshane:
Hi @johnpryan, i run dartfmt, however, it still seems to fail. I'm new to Travis CI, if you could point me in the right direction so i can determine the formatting issues that'd be appreciated.

PR #452, opened 2019-10-21
polyline with gradient

### LINKED ISSUE #486: Does this package support gradient lines for polygons
I was thinking of this between 3 points **A, B & C**

From *A(Red) to B(green) to C(blue)*

Screenshot from 2019-12-17 22-21-07
Screenshot from 2019-12-17 22-23-58

Basically, there will be a source and destination. And multiple points might be in between them having different unit values for a specific quantity(say rainfall of yesterday)

Is it possible to show the gradient lines in between the points as a single polygon with appropriate color.

Or is it only possible to draw multiple lines from point to point with different color?

@johnpryan

SebWojd:
Add possibility to create polyline with gradient color.

closes #486

### DIFF
--- a/lib/src/layer/polyline_layer.dart
+++ b/lib/src/layer/polyline_layer.dart
@@ -9,7 +9,8 @@ import 'package:latlong/latlong.dart';
 class PolylineLayerOptions extends LayerOptions {
   final List<Polyline> polylines;
 
-  PolylineLayerOptions({this.polylines = const [], rebuild}) : super(rebuild: rebuild);
+  PolylineLayerOptions({this.polylines = const [], rebuild})
+      : super(rebuild: rebuild);
 }
 
 class Polyline {
@@ -61,10 +62,12 @@ class PolylineLayer extends StatelessWidget {
           var i = 0;
           for (var point in polylineOpt.points) {
             var pos = map.project(point);
-            pos = pos.multiplyBy(map.getZoomScale(map.zoom, map.zoom)) - map.getPixelOrigin();
+            pos = pos.multiplyBy(map.getZoomScale(map.zoom, map.zoom)) -
+                map.getPixelOrigin();
             polylineOpt.offsets.add(Offset(pos.x.toDouble(), pos.y.toDouble()));
             if (i > 0 && i < polylineOpt.points.length) {
-              polylineOpt.offsets.add(Offset(pos.x.toDouble(), pos.y.toDouble()));
+              polylineOpt.offsets
+                  .add(Offset(pos.x.toDouble(), pos.y.toDouble()));
             }
             i++;
           }
@@ -107,7 +110,9 @@ class PolylinePainter extends CustomPainter {
     if (polylineOpt.gradientColors == null) {
       paint.color = polylineOpt.color;
     } else {
-      polylineOpt.gradientColors.isNotEmpty ? paint.shader = _paintGradient() : paint.color = polylineOpt.color;
+      polylineOpt.gradientColors.isNotEmpty
+          ? paint.shader = _paintGradient()
+          : paint.color = polylineOpt.color;
     }
 
     final filterPaint = Paint()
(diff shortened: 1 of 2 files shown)


lib/src/layer/polyline_layer.dart
```diff
@@ -95,12 +100,17 @@ class PolylinePainter extends CustomPainter {
     final rect = Offset.zero & size;
     canvas.clipRect(rect);
     final paint = Paint()
-      ..color = polylineOpt.color
       ..strokeWidth = polylineOpt.strokeWidth
       ..strokeCap = StrokeCap.round
       ..strokeJoin = StrokeJoin.round
       ..blendMode = BlendMode.src;
 
+    if (polylineOpt.gradientColors == null) {
```
johnpryan: need to also handle the isEmpty case

johnpryan:
closes #486

johnpryan (approved):

FilipeLarga:
@SebWojd Does this pr support using a gradient of only two colors over the whole line, instead of defining a color for every point? What does stop colors do? It is undocumented and not used in the example.

SebWojd:
@FilipeLarga This feature support as many colors as you wish. You have to play with gradientColors and colorsStop. I implemented here the standard way how Dart use linear gradient. 

Here you have a snippet:

```
                 PolylineLayerOptions(
                    polylines: [
                      Polyline(
                        points: pointsGradient,
                        strokeWidth: 4.0,
                        gradientColors: [
                          Color(0xffE40203),
                          Color(0xffFEED00),
                          Color(0xff007E2D),
                          Color(0xff0000FF),
                        ],
                        colorsStop: [0.1, 0.3, 0.5, 0.95]
                      ),
                    ],
                  ),
```

and the result is:

Zrzut ekranu 2020-06-21 o 13 08 18

moovida:
Hi @SebWojd , sorry for resurrecting this here. I have an issue I am describing on issue #772
Is it possible that this works properly only for straight lines? Or is it supposed to work also for complex polylines?

Vincent7140:
Hi @SebWojd , sorry for resurrecting this here. I have the same issue you had describing on issue https://github.com/fleaflet/flutter_map/issues/772

Did you manage to correct the issue since then ? thank you

Vincent7140:
Hi @moovida  , sorry for resurrecting this here. I have the same issue you had describing on issue https://github.com/fleaflet/flutter_map/issues/772

Did you manage to correct the issue since then ? thank you

PooruTorie:
Hi, i have the same error like in #772
In a line with more than 5 points

Jamesm-98:
I'm also facing a similar issue here - suffering from the gradient being referenced to a straight line from first to last points. This makes colouring polylines by arbitrary variable difficult/innacurate when the polyline is not a straight line (or close to one). If the polyline curves back towards the first point it's colour goes "backwards" up the colour gradient.

Vincent7140:
Hi, I've done this a while ago and I have never coded in dart again but looking into my code it seems I had replace this :


gradientColors: _generateGradientColors(colors),

by this : 

      List<Color> gradient = _generateGradientColors(colors);
      List<dynamic> result = _generateColorStops(gradient);
      List<Color> newColors = result[0];
      List<double> colorStops = result[1];


and then this :

return Polyline(
        points: points, 
        strokeWidth: 6,
        gradientColors: _generateGradientColors(colors),
      );
    }).toList();
    
by this    
 
return Polyline(
        points: points, 
        strokeWidth: 6,
        gradientColors: newColors,
        colorsStop : colorStops,
      );
    }).toList();
 


Maybe it will help good luck !

JaffaKetchup:
From what I can tell, only workarounds of splitting up the line into smaller lines will work. Maybe a shader could work but I don't think so with the basic shader support Flutter has (it would need flutter_gpu).

I've looked into this quite a bit, and actually it doesn't seem to be implemented in many places except in editing software. Even on the web it's not possible. See for example https://github.com/cereallarceny/gradient-path.

PR #469, opened 2019-12-04
Tidyup

domesticmouse:
Hey John,

Looks like there are some changes inbound in `ImageProvider.load`. This change is enough to make it compile and work against head, but I'm unsure if it is correct. Do you want to chat with the peeps involved with https://github.com/flutter/flutter/issues/26194 and make sure I'm doing the right thing here?

brett


domesticmouse:
PTAL

domesticmouse:
@johnpryan: This has turned into a cleanup PR. I'm curious if this is interesting or if I should just close it out.

johnpryan:
merci!

PR #478, opened 2019-12-15
upgrade imageloader support for Flutter 1.12

kimlet:


kengu:
@johnpryan - flutter_map is now broken on flutter/stable from version 1.12. 

Could you please accept this PR if possible?

PR #479, opened 2019-12-15
Update pubspec.yaml

itsMatoosh:
Make it compatible with the latest cached_network_image: 2.0.0-rc.1


johnpryan:
This should be fixed now

PR #505, opened 2020-02-02
Controller position stream

porfirioribeiro:
Sometimes we need to do something when the map position changes.
It can be done at map level using `onPositionChanged` or on plugins with the `rebuild` stream.
But i think it should be possible to be notified of map changes with the controller.

I thought about reusing the existing `_onMoveSink` and changing its type from `Null` to `MapPosition` and that might be a breaking change?

_The example will not run because it depends on androidX that i added on PR #504_


johnpryan:
I don't think this an API that flutter_map needs to provide. Can't users choose to use `package:location` if the want to?

johnpryan:
There's now a sample that shows how to do this with package:location: https://github.com/johnpryan/flutter_map/pull/447

porfirioribeiro:
Its not exactly the same problem.
That sample is to move the map to the current GPS location

My PR was just to be able to subscribe the map position, was able to do it on my side, but i think was a nice feature to add

johnpryan:
I see, thanks for the clarification

johnpryan:
Can you fix the merge conflicts?

PR #520, opened 2020-02-15
Keep old tiles

ibrierley:
This is quite a big update. There are 3 main components.

1 Try and keep tiles when there is an outstanding one (especially for bad connections etc)
2 Keep tiles for a small amount of time anyway
3 I also added an option for greedy tiles, you can set this (default 1) to look ahead in the direction you are panning, so it's a smoother transition when wanting the next file.
4 I also reversed the order of tiles in the tilelist sort method, so may want to double check that, but it seemed to help with some display issues, where tiles didn't update until moving and some flashing.

1,2,3 are all options, as in different use cases, some may want to have less tiles loaded for some reason.

It may be worth testing more on any older devices to see if the experience is improved, I've tested on those where I can.


porfirioribeiro:
I will test it in my in development app next Monday and give you feedback
This is one of the most missing things!!

porfirioribeiro:
Nice work, i have been testing on my app and it gives a better navigation experience on the map.
I have tiles with transparency and sometimes i can see tiles overlaping like this:
imagem

Most times it disappears fast, but sometimes it gets stuck until i move the map again.

And i can feel it a little bit less performant on my old test device, but could be perception

But still, thanks for your effort, i hope this gets released some time soon!

ibrierley:
Thanks for the feedback, that's really useful.

I hadn't considered transparent tiles, when having a backup tile....I
wonder at some point if the solution could be re-engineered to have backups
as the fallback tile (this was one earlier suggestion, but there were some
issues when I originally looked at that method). I will have a ponder for a
future amendment.

One option could be to have the defaults as disabled in the mean time.

I think there will be a slight loss in performance as typically there will
have to be more tiles loaded (setting greedytiles to 0 may help if not much
preloading/panning is needed). But it may be a tradeoff between slightly
less performance vs nicer experience.

Ian

ibrierley:
Actually, I don't think we could use the fallback tile method anyway, as they would have the wrong scaling on it, plus if zooming out, you need the previous 4 tiles for each tile, as a fallback.

ibrierley:
I've just added some extra bits that may help with the transparent tiles if you wanted to test that.

porfirioribeiro:
I will test again tomorrow morning!

ibrierley:
Thanks, I'm just exploring to see if there's a slightly tighter option, as I think this is an improvement, there's a few bit's that help with the odd underlying bug I think eg in the original, a tile is missing from
LatLng(-21.831126, 18.419449) zoom 13.623521933239402 and so it may be useful to track down some of these oddities at the same time. This I think is related one of the other reported issues on a missing tile at the edges when you move it appears/disappears.

If it's felt it's needed to get this live sooner than later though, I'll work on some of those amendments listed first.

ristiisa:
so...  what happened?

ibrierley:
iirc there was some other reworking of tiles that superceded it really so wasn't worth it.

ibrierley:
I can't remember if the new code  marci did had something similar. I know it kept old tiles around a certain time, but maybe that was just for transitions. Feel free to take any relevant code from this which is absent and create a new pull request though.

PR #525, opened 2020-02-21
fixes tiles at edge not being loaded, due to using the same tilezoom …

ibrierley:
I think there was a typo here, as it's just comparing itself to get a ratio of map zoom to tile zoom. So I think it needs to use a this.zoom instead. This hopefully fixes https://github.com/johnpryan/flutter_map/issues/517

### DIFF
--- a/lib/src/map/map.dart
+++ b/lib/src/map/map.dart
@@ -272,7 +272,7 @@ class MapState {
   }
 
   Bounds getPixelBounds(double zoom) {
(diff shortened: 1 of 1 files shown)


ibrierley:
Just a note, happy for this to be closed, as the original problem is fixed (I think by not using this function now), but I think the original code that may be used elsewhere is still wrong, so may still be worth a ponder, as I think this would fix it.

PR #529, opened 2020-02-27
Support custom CRS

maRci002:
Hello guys, me and @fegyi001 made avaible to use custom CRS in this project with help of [proj4dart](https://pub.dev/packages/proj4dart) project (we are the authors of it).

`Proj4Crs` and `Proj4Projection` classes based on [Proj4Leaflet](https://github.com/kartena/Proj4Leaflet) project and it has [BSD 2-Clause "Simplified" License](https://github.com/kartena/Proj4Leaflet/blob/master/LICENSE) so we need to make sure we can use their codebase.

We also added some support for multiple origins based on this [commit](https://github.com/kartena/Proj4Leaflet/pull/171) (which is still not reviewed but I tested this feature on this map and the map isn't bouncing unlike in Leaflet's map 👌).

We made an example page which uses [EPSG:3413](http://epsg.io/3413) crs.
map_custom_crs

Closes #514

PR #541, opened 2020-03-06
Proj4dart update

fegyi001:
Updated **proj4dart** to its latest version (1.0.1) and did some documenting.

**Version update**
I'm the other main contributor of proj4dart and we recently published a new version which also supports WKT CRS definitions (OGC & ESRI).

**Documenting**
Since I think using custom CRS will be used frequently in the future I did a "Custom CRS" section in "Usage". But explaining fully how it can be used would be too much in the main readme. Therefore I added a dedicated Readme to the Custom CRS example page itself.


maRci002:
This PR adds the forgotten readme section which was missed from #529. I think these changes helps Google SEO engine if some wants to find for instance `flutter map custom crs` to put this repo in first place.

PR #542, opened 2020-03-06
feat(*): add layers widgets

ruizalexandre:
:warning: __BREAKING CHANGES__
- No more layer options definition (REMOVED)
- No more plugin API (REMOVED)

:fire: __All layers become flutter widgets !__

---

* Add shared mapState widget
* Improve rebuild layers
* mapState API available on your custom widget with `var mapState = MapStateInheritedWidget.of(context).mapState;`
* Update example with new strategy
* Add simultation of user location


ruizalexandre:
@johnpryan @synw

Can close : 
- https://github.com/johnpryan/flutter_map/issues/535
- https://github.com/johnpryan/flutter_map/issues/459 
- https://github.com/johnpryan/flutter_map/issues/482

porfirioribeiro:
This seems very interesting, will be watching this!

ruizalexandre:
Hi @ibrierley 
Moving your plugin to this new logic can take much time ?
https://github.com/ibrierley/flutter_map_dragmarker

ibrierley:
This should do it I think...

https://github.com/ibrierley/flutter_map_dragmarker/blob/master/lib/dragmarker_widget_layer.dart

There's 2 files in there, dragmarker.dart for the old plugin format, and dragmarker_widget_layer.dart for the referenced one here.

It would be good to hear whether this format will likely become the new standard, as I'll convert my tile layer code as well to it at some point if so.

lehno:
I would use that

ruizalexandre:
@lehno I'm rebasing on master ! Updated PR today :)

ruizalexandre:
@johnpryan I updated file names on my PR, that's why the build failed

ruizalexandre:
> First of all - I love using Widgets instead of declarative options for the layers. Great job.
> 
> I think `MapStateInheritedWidget` is a mouthful and it can be moved to `MapState` class itself, so you can do `MapState.of(context)` and get the state instance. Because of that this static method should directly lookup the inherited widget, then return the `mapState` instead of the widget itself:
> 
> My suggestion:
> 
> ```dart
> class MapState {
>   ...
>   static MapState of(BuildContext context, {bool nullOk = false}) {
>     assert(context != null);
>     assert(nullOk != null);
>     final widget = context.dependOnInheritedWidgetOfExactType<MapStateInheritedWidget>();
>     if (nullOk || widget != null) {
>       return widget?.mapState;
>     }
>     throw FlutterError('MapState.of() called with a context that does not contain a FlutterMap.');
>   }
> }
> ```
> 
> Not necessary, but given this `InheritedWidget` won't be useful by itself it can become a hidden widget - `_MapStateInheritedWidget`.
> 
> Second suggestion - to get this to be even considered to be merged with `flutter_map` there should be a way for the two approaches to coexist (at least initially):
> 
> * `layers` should remain a list of declarative options;
> * introduce `children` to the `FlutterMap` which can use the widgets;
> * ensure `layers` or `children` is null - less complexity is better;
> * if the `layers` are present and `children` are `null` - then create the `children` automatically;
> * I think `plugins` can only work with `layers`, but am not 100% sure.

Thanks @avioli ! 
I'll change my PR to take care about your return. 🔥

Sly2024:
Any update on this PR?
The proposed approach to work with widget is way more efficient and easier to use

ruizalexandre:
@Sly2024 I need to finish my last update
Build pipeline is blocked because I've updated file names !

PR #546, opened 2020-03-11
Fix Bug 545 stacked MBTileImageProvider

### LINKED ISSUE #545: Impossible to stack two transparent MBTilesImageProviders on top of each other.
I have tried to layer two transparent TileLayers, both using an MBTilesImageProvider. Only the first defined layer was displayed. While investigating the problem, I noticed that the MBTileImage overwrites the == operator. 

```
  @override
  bool operator ==(other) {
    return other is MBTileImage && coords == other.coords;
  }
```

But it only checks if the image has the same coordinates. Because both layers have the same coordinates, the top layer is always loaded. I tested the file path for comparison and then both layers were displayed correctly.

```
  @override
  bool operator ==(other) {
    return other is MBTileImage && coords == other.coords && fileUrl == other.fileUrl;
  }
```

felixjunghans:
Added a third parameter to the MBTileImage to distinguish two different MBTileImages with the same coordinates.

Fixes #545

### DIFF
--- a/lib/src/layer/tile_provider/mbtiles_image_provider.dart
+++ b/lib/src/layer/tile_provider/mbtiles_image_provider.dart
@@ -76,6 +76,7 @@ class MBTilesImageProvider extends TileProvider {
     return MBTileImage(
       database,
       Coords<int>(x, y)..z = z,
+      mbtilesFile?.path ?? asset,
     );
   }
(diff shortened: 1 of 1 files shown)


johnpryan:
Looks good, thanks!

PR #553, opened 2020-03-18
Add extras property to markers

HugoHeneault:
Using https://github.com/lpongetti/flutter_map_marker_cluster I needed to add extra properties to markers (so I can sum this props to show a total number on the cluster). 

Doing so requires to update the `Marker` class here and I think it can be helpful in many other cases. 

Please tell me if I need to change anything. 

Thanks!

### DIFF
--- a/lib/src/layer/marker_layer.dart
+++ b/lib/src/layer/marker_layer.dart
@@ -77,12 +77,14 @@ class Marker {
   final double width;
   final double height;
   final Anchor anchor;
+  final Map<String, dynamic> extras;
 
   Marker({
     this.point,
     this.builder,
     this.width = 30.0,
     this.height = 30.0,
+    this.extras,
     AnchorPos anchorPos,
   }) : anchor = Anchor.forPos(anchorPos, width, height);
 }


johnpryan:
This seems like something that can be done in a separate data structure. If you can help me understand why this is necessary we might be able to come up with a better solution.

HugoHeneault:
Hey @johnpryan 

So I have a cluster of markers from a `MarkerClusterLayerOptions` layer.
The cluster marker need to show sum of dynamic data (not only total number of markers clustered). So I pass this data in my marker `extras` field.

I build them like this:

```dart
builder: (BuildContext context,
    List<Marker> markers) {
  int hikesCount = 0;

   // take data from extra field
  markers.forEach((marker) {
    if (marker.extras != null &&
        marker.extras['hikes'] != null) {
      hikesCount += marker.extras['hikes'];
    }
  });

  if (hikesCount == 0) {
    return MarkerIconWidget(
        icon: 'assets/icons/flag.svg',
        text: (hikesCount)
            .toString());
  } else {
    return MarkerIconWidget(
        icon: 'assets/icons/home.svg'
    );
  }
}),
```

Do you have a better way? :)

The PR got merged on https://github.com/lpongetti/flutter_map_marker_cluster/pull/35 even if it wasn't merged here, so it will break next `flutter_map_marker_cluster` unless his maintainer revert it or this PR is accepted. 🤷

johnpryan:
Markers aren't meant to store any custom data, they're a type that's used to display information that's passed in. You probably want to sum `hikesCount` outside the builder method.

HugoHeneault:
Hey @johnpryan 

Clustering is dynamic (depends on cluster settings, screen size, zoom level, number of markers to cluster), so you the only place where you can know which markers are in the cluster is inside the builder method. 

How would you do it?

Thanks! 👌

Maistho:
It should be possible to store that information in another map outside, similar to the first example below.

If you need to store custom data on the markers, you should also be able to subclass the marker class, like the second example.

Reservations for the examples being sloppy, it's quite late and I got a bit tired because your PR https://github.com/lpongetti/flutter_map_marker_cluster/pull/35 had been accepted and released which broke my build 🙃 

```dart
class MapExample extends StatelessWidget {
  final List<Map<String, dynamic>> markerSource = [
    <String, dynamic>{'pos': LatLng(10, 10), 'hikes': 0},
    <String, dynamic>{'pos': LatLng(11, 11), 'hikes': 1},
    <String, dynamic>{'pos': LatLng(12, 12), 'hikes': 2},
    <String, dynamic>{'pos': LatLng(13, 13), 'hikes': 3},
    <String, dynamic>{'pos': LatLng(14, 14), 'hikes': 4},
  ];
  @override
    // Probably move this part to initState or something similar depending on how you actually get the data
    final List<Marker> markers = [];
    Map<Marker, int> hikes;
    for (var s in markerSource) {
      final m = Marker(point: s['pos']);
      hikes[m] = s['hikes'];
    }
    return FlutterMap(
      options: MapOptions(),
      layers: [
        MarkerClusterLayerOptions(
          markers: markers,
          builder: (context, markers) {
            int hikesCount = 0;

            for (final marker in markers) {
              hikesCount += hikes[marker] ?? 0;
            }

            if (hikesCount == 0) {
              return MarkerIconWidget(
                  icon: 'assets/icons/flag.svg', text: (hikesCount).toString());
            } else {
              return MarkerIconWidget(
                icon: 'assets/icons/home.svg',
              );
            }
          },
        ),
      ],
    );
  }
}
```

```dart
class HikingDataMarker extends Marker {
  HikingDataMarker({
    this.hikes,
    LatLng point,
    WidgetBuilder builder,
    double width = 30.0,
    double height = 30.0,
    AnchorPos anchorPos,
  }) : super(
          point: point,
          builder: builder,
          width: width,
          height: height,
          anchorPos: anchorPos,
        );
  final int hikes;
}

class MapExample2 extends StatelessWidget {
  final List<Map<String, dynamic>> markerSource = [
    <String, Object>{'pos': LatLng(10, 10), 'hikes': 0},
    <String, dynamic>{'pos': LatLng(11, 11), 'hikes': 1},
    <String, dynamic>{'pos': LatLng(12, 12), 'hikes': 2},
    <String, dynamic>{'pos': LatLng(13, 13), 'hikes': 3},
    <String, dynamic>{'pos': LatLng(14, 14), 'hikes': 4},
  ];
  @override
    // Probably move this part to initState or something similar depending on how you actually get the data

    final List<Marker> markers = [];
    for (var s in markerSource) {
      markers.add(HikingDataMarker(
        hikes: s['hikes'],
        point: s['pos'],
      ));
    }
    return FlutterMap(
      options: MapOptions(),
      layers: [
        MarkerClusterLayerOptions(
          markers: markers,
          builder: (context, markers) {
            int hikesCount = 0;

            for (final marker in markers) {
              final HikingDataMarker m = marker;
              hikesCount += m.hikes;
            }

            if (hikesCount == 0) {
              return MarkerIconWidget(
                  icon: 'assets/icons/flag.svg', text: (hikesCount).toString());
            } else {
              return MarkerIconWidget(
                icon: 'assets/icons/home.svg',
              );
            }
          },
        ),
      ],
    );
  }
}
```

TheLastGimbus:
Hi there!
What is the state of this change?

@johnpryan let me show you what kind of problem I have:

In our app, markers display information about air pollution - if it's good, it's a green circle - if bad, red circle

We wanted to do a clusterisation, and clustered circles would have an averge color of markers inside them - problem is, cluster builder gives us *only markers* - and we have no good way tell what data they represent, and how to build the cluster!

What we ended up doing, is wrapping **everything** in a `HashMap<Marker, PolutionData>` - so that we can get data from marker object as a key - but this really seems hacky, and may affect perfomence - and could all be avoided, if a `Marker` could contain *any kind of data* - some `String tag` or **anything** :pray:

lib/src/layer/marker_layer.dart
```diff
@@ -77,12 +77,14 @@ class Marker {
   final double width;
   final double height;
   final Anchor anchor;
+  final Map<String, dynamic> extras;
```
TheLastGimbus, line 80: This could just be changed to `dynamic` (thank you Dart :heart:) - and we could decide what kind of data we want to put there ourselves :dart:

```suggestion
  /// Any kind of extra data that you want to pass with the marker
  /// This may be helpful when using clustering
  final dynamic extras;
```

TheLastGimbus (approved):

Maistho:
@TheLastGimbus is there any reason why you cannot extend Marker with a custom subclass and include any data you want?

I'm using a subclass in my app, it works, doesn't need any changes and has better type safety.

```dart
class PollutionMarker extends Marker {
  final PollutionData data;

  PollutionMarker({
    LatLng point,
    WidgetBuilder builder,
    double width = 30.0,
    double height = 30.0,
    AnchorPos anchorPos,
    this.data,
  }) : super(
          point: point,
          builder: builder,
          width: width,
          anchorPos: anchorPos,
        );
}
```

TheLastGimbus:
Oh, I... I didn't realise I can do this :laughing: I will try it out, and if it works, we can pretty much close this :sweat_smile: 

I mean, we could always add some kind of `int id` to original `Marker`, but yeah, we won't need that if we can use `extends`

HugoHeneault:
Closing at it seems to be a [solution](https://github.com/fleaflet/flutter_map/pull/553#issuecomment-782951948).

TheLastGimbus:
Yes, `extends` works really nice :ok_hand: 

Thank you @Maistho !

PR #570, opened 2020-04-04
show old tiles until new ones are downloaded

### LINKED ISSUE #79: Show old tiles until new ones are downloaded
When a user zooms in enough, new, more zoomed in tiles are loaded and displayed. Would it be possible to show the images of the current layer until the new pictures are loaded instead of just showing a grey map while the new tiles are loading?

obiwanzenobi:
Hello, this PR fixes #79 

By adding separate widget `image_tile` we are able to observe tile download status.
It does basically the same as FedeInImage, but with AnimatedSwitcher.

There is new part in a build method that checks if all of the tiles on the screen are downloaded and adds  a background layer of tiles if some of them are still loading.

### DIFF
--- a/lib/src/layer/image_tile.dart
+++ b/lib/src/layer/image_tile.dart
@@ -1,0 +1,65 @@
+import 'package:flutter/foundation.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter_map/flutter_map.dart';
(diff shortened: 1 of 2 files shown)


obiwanzenobi:
Demo:
ezgif com-optimize

PR #572, opened 2020-04-04
Improve tile management

### LINKED ISSUE #59: Zooming out very far causes null pointers
When zooming out very far and moving the map around, you are greeted with the following error.

https://pastebin.com/VxQ5jKhX

Setting the pan boundary to -90, -180 and 90, 180 did not resolve the issue.

### LINKED ISSUE #79: Show old tiles until new ones are downloaded

### LINKED ISSUE #125: map not fit perfect screen on ios
Map view not fit flexible tag on ios device < X.

look at grey border.
It work on iphone X.

error

```
return new Padding(
      padding: new EdgeInsets.all(0.0),
      child: new Column(
        children: [
          new Flexible(
            child: new FlutterMap(
              options: new MapOptions(
                  center: new LatLng(39.213695, 9.121050),
                  interactive: false,
                  minZoom: 12.0,
                  maxZoom: 19.0,
                  zoom: 16.5,
                  onTap: (latLng) {
                    print('tap callback  $latLng');
                  },
                  onPositionChanged: (MapPosition mapPos) {
                    print("pos changed $mapPos");
                  }),
              layers: [
                new TileLayerOptions(
                    offlineMode: false,
                    urlTemplate:
                        "https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png",
                    //  urlTemplate: "assets/offlinemaps/cagliari/{z}/{x}/{y}.png"),
                    subdomains: ['a', 'b', 'c']),
                new MarkerLayerOptions(markers: markers)
              ],
            ),
          ),
        ],
      ),
    );
```

### LINKED ISSUE #227: Consider consolidating `maxZoom` on MapOptions and TileLayerOptions
```
child: new FlutterMap(
                options: new MapOptions(
                  center: new LatLng(51.5, -0.09),
                  zoom: 5.0,
                ),
                layers: [
                  new TileLayerOptions(
                      urlTemplate:
                          "https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png",
                      subdomains: ['a', 'b', 'c'],
                      maxZoom: 3),
                  new PolylineLayerOptions(
                    polylines: [
                      new Polyline(
                          points: points,
                          strokeWidth: 4.0,
                          color: Colors.purple),
                    ],
                  )
                ],
              ),
```

### LINKED ISSUE #339: Using mapsforge offline maps in flutter_map
Hi there. While mbtiles offline maps are great for some cases, for me it is necessary to have large offline maps on my device. For this I have used the mapsforge project, which has nicely compressed OSM data files in vector format, that can be converted in tiles. Just as an info, files of huge OSM info data are around 2 Giga, so they fit on any device.

That said, recently a port of mapsforge to flutter was done: https://github.com/mikes222/mapsforge_flutter

I tried to implementa tileprovider for that (I am prototyping a port of the geopaparazzi project) and the first results look good:
image

I found a couple of issues and was wondering if I could get some information:

1. everytime I zoom in, the first result is a blurred map. When I then pan slightly (really just to say that it did pan), the map get's sharper. Here the same region before and after:

image

2. On low zoom values, the mapsforge engine get's very slow. This makes the map view freeze completely. I wonder if there is a way to have the tiles loaded only once they are ready. Also I am not sure how to return nothing where there are no tiles. Should I create an empty image for that?

Thanks for any info you might give me and sure thanks for this great project.

### LINKED ISSUE #345: Grab current map image with all layers to image/file- enhancement ?
Thanks for great plugin !

I am trying to save current map image with all overlays to the file. Unfortunately widget is not updated until I finish on pressed event. Inside onPressed function I move the map and then want to grab image. So the image I grab is current (old) map view before whole tree is refreshed.
I am looking for a way to get current map view (with all layers) before is it actually rendered. Any ideas ?

### LINKED ISSUE #377: Add callback when map is done painting tiles
See #355 for more context

### LINKED ISSUE #386: Zoom Levels and Process for Offline Maps
When using offline tiles, you obvious get the tiles you want at certain zoom levels.

When using say openstreetmaps online, once you had got to max zoom levels, the pngs of the tiles you have 'scale' up, which keeps the mapping on screen (abit with degraded quality of the images) but allows closer zoom levels, where tiles do not exist.

We would love to impliment this in flutter map - does anyone know how this could be developed, unless it is already a thing?

There are similar issues where people have asked for the same thing to happen inbetween loading of tiles (when you zoom the tile images enlarge until new tiles have loaded in) and thus this could be used when no more offline tiles are found at higher zoom levels

We are developers and happy to help, but will need a bit of direction on this one

Thanks

### LINKED ISSUE #429: Get lag when zoom in map
I have 2 titlemap: 1. base map, 2. wms from my geoserver. I feel my app load too slow when zoom in map. How can i  increase performances my app? Thanks you

### LINKED ISSUE #444: Exception caught by image resource service - not handled.
If the phone has slow data connection sometimes a tiles are not downloaded before timeout.
As I understand cached image provided raises exception which is not handled inside plugin.  
The debug console shows:
`
════════ Exception caught by image resource service ════════════════════════════
Couldn't download or retrieve file.
════════════════════════════════════════════════════════════════════════════════` 

In release version I have general handler for all exceptions and this is  coming up  all the time.

### LINKED ISSUE #457: Too slow for practical use?
Hi
I'm doing a research to see if we can move from a Cordova based app to Flutter. First thing I tried is this Leaflet map plugin (used the example under "Markers" here https://tphangout.com/flutter-maps-and-markers/).
I'm kind of hoping I'm missing something, because running that on my Huawei P20 Lite is simple too slow to be of any practical use. On that very same phone, our Leaflet map in our Cordova app is just flying.

Am I the only one having performance problems with this plugin?

### LINKED ISSUE #463: Not all necessary tiles are loaded after a zoom change (demo video)
Good job with this library @johnpryan and all contributors!

I'm experiencing a bug where after zooming in to a new level, not all tiles that should appear on the screen are loaded. Moving the camera is needed to make them load.

Video demonstrating a zoom increase with not all tiles (bottom left in this case) loading until i start scrolling on the map again. https://imgur.com/a/3kfI3qM

The tiles provider used was AssetTileProvider().

Here is the code but I'm not sure how relevant it is:

```dart

FlutterMap(
  mapController: mapController,
  options: MapOptions(
    center: lisbon,
    zoom: 13.0,
    minZoom: 13.0,
    maxZoom: 16.0,
    swPanBoundary: LatLng(38.358, -9.634),
    nePanBoundary: LatLng(39.15, -8.458),
  ),
  layers: [
    TileLayerOptions(
      keepBuffer: 4,
      tileProvider: AssetTileProvider(),
      urlTemplate: "assets/map/{z}/{x}/{y}.png",
      maxZoom: 16,
      tms: true,
      backgroundColor: Colors.white,
     ),
     MarkerLayerOptions(
       markers: markers,
     ),
   ],
),

```

I would love to contribute myself but I'm fairly novice and looking at the source code seems very complex. Hoping to see this and the _keep the old tiles until the new ones are loaded_ improvements for 0.8 as I truly believe that would make this the best flutter map library!

### LINKED ISSUE #492: Drag is not fluid enough
I compared flutter_map to the normal google maps. Don't get me wrong, flutter_map's drag works fine, but it's as not smooth as the one in google maps. Do you have any idea what's the difference?

### LINKED ISSUE #517: Tiles are not loading on the edges of the map
I'm having an issue where the tiles on the edges of the map are not loading after the map is moved.

Steps to reproduce:
1. run the example project
2. zoom into the map until street level
3. pan the map slowly until unloaded tiles appear and stop
4. the unloaded tiles will not finish loading

Any ideas on how to solve this?

Might be related to #463

I tested this with:
note8:
Flutter (Channel master, v1.15.3-pre.37, on Microsoft Windows [Version 10.0.18362.86], locale et-EE)
Android SDK version 29.0.2

iphone 6s:
Flutter (Channel master, v1.14.4-pre.28, on Mac OS X 10.14.6 18G103, locale en-EE)
Xcode - develop for iOS and macOS (Xcode 11.3.1)

### LINKED ISSUE #523: zoomOffset and zoomReverse parameters aren't used
The parameters aren't used in any calculations. This means that if one tries to use 512px tiles with a zoomOffset of -1 (as suggested [here](https://digitransit.fi/en/developers/apis/3-map-api/background-map/) under "Display map using Leaflet") it just won't work.

https://github.com/johnpryan/flutter_map/blob/a639b6d72fd69d7c4918e37cca0e12b79c4334eb/lib/src/layer/tile_layer.dart#L45-L46

### LINKED ISSUE #536: handling tileProvider errors
sometime tileProvider cause stopping of the app because of the remote server cant return a response or the cachingNetwork cant find a file but i didn't find a way to handle these errors to avoid the crash of the app

### LINKED ISSUE #538: MinZoom and MaxZoom not working

maRci002:
Hello, I made some changes mainly in [tile_layer.dart](/johnpryan/flutter_map/blob/master/lib/src/layer/tile_layer.dart) in order to keep the old tiles on the screen when zooming if new ones not avaible. These changes based on current Leaflet's tile fetching /retaining strategy see [TileLayer.js](https://github.com/Leaflet/Leaflet/blob/master/src/layer/tile/TileLayer.js) / [GridLayer.js](https://github.com/Leaflet/Leaflet/blob/master/src/layer/tile/GridLayer.js).

I also added some missing features like minZoom / minNativeZoom / maxNativeZoom / zoomReverse / zoomOffset / errorImage and fixed / tracked down some other problems which I will mention in closes section.

Normal use case (fast internet connection)
normal use

Simulating slow internet connection
slow internet2

Add support for Error tiles
error image

Add support for minNativeZoom / maxNativeZoom (this demo shows locked minNativeZoom: 16 while real zoom is about 12/10 <- it will be a real trial for mobile phones bandwidth / render time)
minNativeZoom test1

Support for remove tiles if we are out of zoom limits (we don't call any setState if we are out of bounds, however we pay attention for zoom lvl changes) -- this can happen if forinstance MapOptions has maxZoom: 18, but TileLayerOptions has maxZoom: 16; works with minZoom too, and fixed #59 overscale problems (helps if MapOptions / TileLayerOptions has same maxZoom, however without this fix overscale could go beyond limits which could have caused remove all tiles).
remove tiles when out of zoom bounds

closes #59
closes #79
closes #125
closes #227 (we don't need remove maxZoom because we may have multiple layers and if current tilelayer is more than maxZoom then _removeAllTiles called also applies to minZoom; we stop fetching new tiles when not inside min /maxZoom moreover we skip setState calls for performance reasons)
closes #339 (performance -- maybe this PR resolves this because build() method was designed to run as fast as possible (_update is [throttled](https://github.com/Leaflet/Leaflet/blob/master/src/layer/tile/GridLayer.js#L248) and _pruneTiles / _setView called only when necessary))
closes #386 
closes #429
closes #444
closes #457
closes #463
closes #492 (performance -- maybe this PR resolves this because build() method was designed to run as fast as possible (_update is [throttled](https://github.com/Leaflet/Leaflet/blob/master/src/layer/tile/GridLayer.js#L248) and _pruneTiles / _setView called only when necessary))
closes #517
closes #523
closes #536
closes #538

PR will be closed also:
closes #525
closes #570

PR also gives opportunity to finish these issues easily (we can add some callbacks to TileLayerOptions - when a Tile is downloaded / become active (when loaded + and fade ends) / Tile loads with an error / or when all tiles loaded (_noTilesToLoad internal function can fire callback when the grid layer loaded all visible tiles)):
see #345
see #377

### DIFF
--- a/lib/src/layer/tile_layer.dart
+++ b/lib/src/layer/tile_layer.dart
@@ -141,11 +141,15 @@ class TileLayerOptions extends LayerOptions {
   ///
   final Map<String, String> additionalOptions;
 
-  // Tiles will not update more than once every `updateInterval` when panning.
+  // Tiles will not update more than once every `updateInterval` milliseconds
+  // (default 200) when panning.
+  // It can be 0 (but it will calculating for loading tiles every frame when panning / zooming, flutter is fast)
   // This can save some fps and even bandwidth
   // (ie. when fast panning / animating between long distances in short time)
   final Duration updateInterval;
 
+  // Tiles fade in duration in milliseconds (default 100),
+  // it can 0 to avoid fade in
   final Duration tileFadeInDuration;
 
   TileLayerOptions(
@@ -168,11 +172,21 @@ class TileLayerOptions extends LayerOptions {
       // ignore: avoid_init_to_null
       this.wmsOptions = null,
       this.opacity = 1.0,
-      this.updateInterval = const Duration(milliseconds: 150),
-      this.tileFadeInDuration = const Duration(milliseconds: 100),
+      // Tiles will not update more than once every `updateInterval` milliseconds
+      // (default 200) when panning.
+      // It can be 0 (but it will calculating for loading tiles every frame when panning / zooming, flutter is fast)
+      // This can save some fps and even bandwidth
+      // (ie. when fast panning / animating between long distances in short time)
+      int updateInterval = 200,
+      // Tiles fade in duration in milliseconds (default 100),
+      // it can 0 to avoid fade in
+      int tileFadeInDuration = 100,
       rebuild})
-      : assert(tileFadeInDuration != null && !tileFadeInDuration.isNegative),
-        assert(updateInterval != null && !updateInterval.isNegative),
+      : updateInterval =
+            updateInterval <= 0 ? null : Duration(milliseconds: updateInterval),
+        tileFadeInDuration = tileFadeInDuration <= 0
+            ? null
+            : Duration(milliseconds: tileFadeInDuration),
         super(rebuild: rebuild);
 }
 
@@ -302,7 +316,7 @@ class _TileLayerState extends State<TileLayer> with TickerProviderStateMixin {
     _update(null);
     _moveSub = widget.stream.listen((_) => _handleMove());
 
-    if (options.updateInterval.inMicroseconds == 0) {
+    if (options.updateInterval == null) {
       _throttleUpdate = null;
     } else {
       _throttleUpdate = StreamController<LatLng>(sync: true);
@@ -851,7 +865,7 @@ class _TileLayerState extends State<TileLayer> with TickerProviderStateMixin {
     }
 
     tile.loaded = DateTime.now();
-    if (options.tileFadeInDuration.inMicroseconds == 0 ||
+    if (options.tileFadeInDuration == null ||
         (tile.loadError && null == options.errorImage)) {
       tile.active = true;
     } else {
@@ -864,7 +878,9 @@ class _TileLayerState extends State<TileLayer> with TickerProviderStateMixin {
       // Wait a bit more than tileFadeInDuration (the duration of the tile fade-in)
       // to trigger a pruning.
       Future.delayed(
-        options.tileFadeInDuration + const Duration(milliseconds: 50),
+        options.tileFadeInDuration != null
+            ? options.tileFadeInDuration + const Duration(milliseconds: 50)
+            : const Duration(milliseconds: 50),
         () {
           setState(_pruneTiles);
         },
--- a/lib/src/layer/tile_layer.dart
+++ b/lib/src/layer/tile_layer.dart
@@ -299,31 +299,7 @@ class _TileLayerState extends State<TileLayer> {
 
   @override
   Widget build(BuildContext context) {
-    // var pixelBounds = _getTiledPixelBounds(map.center);
-    // var tileRange = _pxBoundsToTileRange(pixelBounds);
-
-    // var tileCenter = tileRange.getCenter();
-    var tilesToRender = <Tile>[
-      for (var tile in _tiles.values)
-        // if ((tile.coords.z - _level.zoom).abs() <= 1) tile
-        // if (tile.imageInfo != null)
-        tile
-    ];
-
-    tilesToRender.sort((aTile, bTile) {
-      final a = aTile.coords;
-      final b = bTile.coords;
-
-      var zIndexA = _levels[a.z].zIndex;
-      var zIndexB = _levels[b.z].zIndex;
-
-      if (zIndexA == zIndexB) {
-        // (a.distanceTo(tileCenter) - b.distanceTo(tileCenter)).toInt(); <-- no matters since Positioned don't care about this, however we use this when sorting tiles to download
-        return 0;
-      } else {
-        return zIndexB.compareTo(zIndexA);
-      }
-    });
+    var tilesToRender = _tiles.values.toList()..sort();
 
     var tileWidgets = <Widget>[
       for (var tile in tilesToRender) _createTileWidget(tile)
@@ -341,10 +317,8 @@ class _TileLayerState extends State<TileLayer> {
   }
 
   Widget _createTileWidget(Tile tile) {
-    var coords = tile.coords;
-
-    var tilePos = tile.tilePos; // _getTilePos(coords);
-    var level = _levels[coords.z];
+    var tilePos = tile.tilePos;
+    var level = tile.level;
     var tileSize = getTileSize();
     var pos = (tilePos).multiplyBy(level.scale) + level.translatePoint;
     var width = tileSize.x * level.scale;
@@ -437,6 +411,7 @@ class _TileLayerState extends State<TileLayer> {
 
     for (var z in toRemove) {
       _removeTilesAtZoom(z);
+      _levels.remove(z);
     }
 
     var level = _levels[zoom];
@@ -444,9 +419,7 @@ class _TileLayerState extends State<TileLayer> {
 
     if (level == null) {
       level = _levels[zoom] = Level();
-
       level.zIndex = maxZoom;
-
       level.origin = map.project(map.unproject(map.getPixelOrigin()), zoom) ??
           CustomPoint(0.0, 0.0);
       level.zoom = zoom;
@@ -511,7 +484,7 @@ class _TileLayerState extends State<TileLayer> {
       if (entry.value.coords.z != zoom) {
         continue;
       }
-      _removeTile(entry.key);
+      toRemove.add(entry.key);
     }
 
     for (var key in toRemove) {
@@ -814,6 +787,7 @@ class _TileLayerState extends State<TileLayer> {
       coordsKey: tileCoordsToKey,
       tilePos: _getTilePos(coords),
       current: true,
+      level: _levels[coords.z],
       imageProvider:
           options.tileProvider.getImage(_wrapCoords(coords), options),
       tileReady: _tileReady,
@@ -888,11 +862,12 @@ class _TileLayerState extends State<TileLayer> {
 
 typedef void TileReady(Coords<double> coords, dynamic error, Tile tile);
 
-class Tile {
+class Tile implements Comparable<Tile> {
   final String coordsKey;
   final Coords<double> coords;
   final CustomPoint<num> tilePos;
   final ImageProvider imageProvider;
+  final Level level;
 
   bool current;
   bool retain;
@@ -914,6 +889,7 @@ class Tile {
     this.tilePos,
     this.imageProvider,
     this.tileReady,
+    this.level,
     this.current = false,
     this.active = false,
     this.retain = false,
@@ -949,6 +925,19 @@ class Tile {
   void _tileOnError(dynamic exception, StackTrace stackTrace) {
     if (null != tileReady) {
       tileReady(coords, exception, this);
+    }
+  }
+
+  @override
+  int compareTo(Tile other) {
+    var zIndexA = level.zIndex;
+    var zIndexB = other.level.zIndex;
+
+    if (zIndexA == zIndexB) {
+      // (a.distanceTo(tileCenter) - b.distanceTo(tileCenter)).toInt(); <-- no matters since Positioned don't care about this, however we use this when sorting tiles to download
+      return 0;
+    } else {
+      return zIndexB.compareTo(zIndexA);
     }
   }
 }
--- a/lib/src/layer/tile_layer.dart
+++ b/lib/src/layer/tile_layer.dart
@@ -120,6 +120,10 @@ class TileLayerOptions extends LayerOptions {
   ///
   Map<String, String> additionalOptions;
 
+  // Tiles will not update more than once every `updateInterval` milliseconds when panning.
+  // This can save some bandwidth (fast panning / fast animate move) / fps
+  final int updateInterval;
+
   TileLayerOptions(
       {this.urlTemplate,
       this.tileSize = 256.0,
@@ -136,8 +140,10 @@ class TileLayerOptions extends LayerOptions {
       // ignore: avoid_init_to_null
       this.wmsOptions = null,
       this.opacity = 1.0,
+      this.updateInterval = 50,
       rebuild})
-      : super(rebuild: rebuild);
+      : assert(updateInterval >= 0),
+        super(rebuild: rebuild);
 }
 
 class WMSTileLayerOptions {
@@ -251,7 +257,7 @@ class _TileLayerState extends State<TileLayer> {
   double _tileZoom;
   Level _level;
   StreamSubscription _moveSub;
-  final _throttleUpdate = PublishSubject<void>();
+  PublishSubject<void> _throttleUpdate;
 
   final Map<String, Tile> _tiles = {};
   final Map<double, Level> _levels = {};
@@ -263,9 +269,11 @@ class _TileLayerState extends State<TileLayer> {
     _update(null);
     _moveSub = widget.stream.listen((_) => _handleMove());
 
-    _throttleUpdate
-        .throttleTime(Duration(milliseconds: 50))
-        .listen((_) => _update(null));
+    _throttleUpdate = widget.options.updateInterval == 0
+        ? null
+        : (PublishSubject<void>()
+          ..throttleTime(Duration(milliseconds: 50))
+          ..listen((_) => _update(null)));
   }
 
   @override
@@ -653,9 +661,20 @@ class _TileLayerState extends State<TileLayer> {
 
   void _handleMove() {
     setState(() {
-      // _update(null);
-      _throttleUpdate.add(null);
-      _setZoomTransforms(map.center, map.zoom);
+      var zoom = _clampZoom(map.zoom);
+
+      if ((zoom - _tileZoom).abs() >= 1) {
+        // It was a zoom lvl change
+        _setView(map.center, zoom);
+      } else {
+        if (null == _throttleUpdate) {
+          _update(null);
+        } else {
+          _throttleUpdate.add(null);
+        }
+
+        _setZoomTransforms(map.center, map.zoom);
+      }
     });
   }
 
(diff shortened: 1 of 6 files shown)


edihasaj:
Good job man, this is very close to what I was thinking to achieve.
Good job.

maRci002:
`ImageProvider placeholderImage;` become deprecated, anyway we can still use it at lowest retained / active level (lowest avaible level based on lowest zIndex).

If someone wants to observe tile fetching / pruning / aborting then revert [remove diagnostic prints](https://github.com/johnpryan/flutter_map/pull/572/commits/380d650948361f67addade76f89577185d52e5c6) commit.

Maybe fade in can be improved (Tile has an animationController), but user is able to increase default 100ms.

maRci002:
If maintainer plans to merge this PR then pls wait a bit because I am going to add some documentation and making some performance refactors (ie. making local function call from setState instead of creating closure-s).

porfirioribeiro:
Good job! I hope this gets merged some time soon!
This tiles bug was stressing a little bit

obiwanzenobi:
Nice! It looks very much like leaflet now.

I was working on same problem and one of my consideration was what to do with panning - should the tiles be retained?

ezgif com-optimize (1)
panningWithRetain

aytunch:
@maRci002 Thanks for this PR, I hope it gets merged very soon
@obiwanzenobi Thanks for the illustration, I definitely think the tiles should retain. Especially when we have markers on the map, not retaining and seeing a marker on empty container looks very bad imho

pento:
I've been trying out this PR, it looks really good!

I noticed a regression with transparent tiles: the old tiles are still visible until the new tiles finish rendering. Here's an example:



This map is rendered in three layers: the base tile layer showing the ground/water, the second layer is a transparent overlay image, the top layer is another tile layer, with just the place names, and the rest is transparent. You can see there's a brief moment when zooming that the place names are being shown at the old and new zoom level.

Here's [the code that generates this map](https://github.com/pento/TidyWeather/blob/fc60ebd975e140536af8e518bb0e5f6363063951/lib/cards/radar.dart#L67-L103), if that helps.

maRci002:
@pento I will take a deeper look soon, but what happens if you provide `tileFadeInDuration = 0` to `TileLayerOptions`?

pento:
Not much difference that I could observe. It possibly takes slightly less time for the old tiles to disappear, but there's still a crossover where both the old and new tiles are visible at the same time.

johnpryan:
Woohoo! I'm taking a look now, but this is looking great so far!

johnpryan:
I think this looks great, if anyone sees their issue closed and think it was a mistake, please leave a comment. That said, I went through the issues and it looks like this closes them all, with the exception of maybe #492 and #339.

@maRci002 thanks a TON for this contribution!

lib/src/layer/tile_layer.dart
```diff
@@ -401,55 +640,101 @@ class _TileLayerState extends State<TileLayer> {
...
+          _setView(map.center, tileZoom);
+        });
+      }
+    } else {
+      setState(() {
+        if ((tileZoom - _tileZoom).abs() >= 1) {
+          // It was a zoom lvl change
+          _setView(map.center, tileZoom);
+        } else {
+          if (null == _throttleUpdate) {
```
johnpryan, line 681: @maRci002 do you always put null on the left hand side of an equality check? :P

maRci002: Sometimes I like using [yoda style](https://en.wikipedia.org/wiki/Yoda_conditions) :smile:

It has it own advantage; forinstance in JS this is a valid expression statement (but user wanted to check if `i` equals to `7`, like `i == 7` or `i === 7`)
```
var i = 0;
if (i=7){
  // this block will be executed
}
console.log(i); // prints 7
```

Or in java:
```
String i = null;
if (i.equals("CONSTANT")) {} // NullPointerException
```

It has it's own benefits but not in Dart :sweat_smile:

maRci002:
@johnpryan you are welcome.

maRci002:
@pento so you are using multiple `TileLayerOptions`, it can be a problem if base layer loads faster than the transparent layer or vice versa because they are fully independent (even in Leaflet) however this is not your case.

I think this will help your problem just change this:
```
    tile.loaded = DateTime.now();
    if (options.tileFadeInDuration == null ||
        (tile.loadError && null == options.errorImage)) {
      tile.active = true;
    } else {
      tile.startFadeInAnimation(options.tileFadeInDuration, this);
    }

    setState(() {});

    if (_noTilesToLoad()) {
      // Wait a bit more than tileFadeInDuration (the duration of the tile fade-in)
      // to trigger a pruning.
      Future.delayed(
        options.tileFadeInDuration != null
            ? options.tileFadeInDuration + const Duration(milliseconds: 50)
            : const Duration(milliseconds: 50),
        () {
          setState(_pruneTiles);
        },
      );
    }
```

To this:
```
    setState(() {
      tile.loaded = DateTime.now();
      if (options.tileFadeInDuration == null ||
          (tile.loadError && null == options.errorImage)) {
        tile.active = true;
      } else {
        tile.startFadeInAnimation(options.tileFadeInDuration, this);
      }

      if (_noTilesToLoad()) {
        _pruneTiles();
      }
    });
```

maRci002:
@obiwanzenobi tiles which are on the same zoom level are retained based on margin.
```
    var margin = options.keepBuffer;
    var noPruneRange = Bounds(
      tileRange.bottomLeft - CustomPoint(margin, -margin),
      tileRange.topRight + CustomPoint(margin, -margin),
    );
```

edihasaj:
> Nice! It looks very much like leaflet now.
> 
> I was working on same problem and one of my consideration was what to do with panning - should the tiles be retained?
> Preview without:
> ezgif com-optimize (1)
> Preview with retain on panning:
> panningWithRetain

I think this is very important, hence should be merged. Can you pull a new request with this?

lib/src/layer/tile_layer.dart
```diff
@@ -571,21 +870,199 @@ class _TileLayerState extends State<TileLayer> {
...
+    this.tilePos,
+    this.imageProvider,
+    this.tileReady,
+    this.level,
+    this.current = false,
+    this.active = false,
+    this.retain = false,
+    this.loadError = false,
+  }) {
+    try {
```
kengu, line 930: @maRci002 - I think this constructor body introduced error reported in #608. The reason why PR https://github.com/johnpryan/flutter_map/pull/609 works is that it replaces the state in the widget-tree, forcing a fresh _TileLayerState to be constructed. 


In it self, the code is not wrong, the problem occurs in `_addTile`, where assignment to array `_tiles[tileCoordsToKey]` is done after the body is executed. 
```
_tiles[tileCoordsToKey] = Tile(
      coords: coords,
      coordsKey: tileCoordsToKey,
      tilePos: _getTilePos(coords),
      current: true,
      level: _levels[coords.z],
      imageProvider: options.tileProvider.getImage(_wrapCoords(coords), options),
      tileReady: _tileReady,
    );
```
The callback method `tileReady` is called indirectly from the method  `_tileOnLoad` when the image has been loaded. If the tile is already cached, the method will return immediately, invoking `tileReady` before the assignment to `_tiles[tileCoordsToKey]` is completed. In the actual callback method `_tileReady`, [the assignment is evaluated](https://github.com/johnpryan/flutter_map/pull/572/files#diff-d0fc1fe821e34169c8068a76a40415f3R826).
```
var key = _tileCoordsToKey(coords);
    tile = _tiles[key];
    if (null == tile) {
      return;
    }
```  
and as you can see, the method will return before it is rendered to the canvas. 

**Working solution**

- Removing the code from constructor body
- using the new method in `loadTileImage` in PR https://github.com/johnpryan/flutter_map/pull/584/files#diff-d0fc1fe821e34169c8068a76a40415f3R1007 to load image expliclity in `_addTile`

I've done this change in https://github.com/DISCOOS/flutter_map/blob/track_change_and_evict_errors/lib/src/layer/tile_layer.dart#L887 and it solves the problem:
```
void _addTile(Coords<double> coords) {
    var tileCoordsToKey = _tileCoordsToKey(coords);
    _tiles[tileCoordsToKey] = Tile(
      coords: coords,
      coordsKey: tileCoordsToKey,
      tilePos: _getTilePos(coords),
      current: true,
      level: _levels[coords.z],
      imageProvider: options.tileProvider.getImage(_wrapCoords(coords), options),
      tileReady: _tileReady,
    );
    _tiles[tileCoordsToKey].loadTileImage();
  }
```

I could make an PR with the change to #584 if you're interested?

maRci002: I'll checkout your suggestions soon. However #609 works because if drawer detects if user wants to navigate to the same menu then it just closes drawer instead of reopen same menu. You can test it if you pan away then if you try to reopen same menu then map current location stays the same as before.

kengu: Yes, of course. My bad.

velval:
Hi @maRci002 thanks for this. I am using flutter_map 0.10.1+1 and I can see the map is retaining the previous tiles when zooming if the network connection is slow. However, I am testing the case where I manually set the device to flight mode and switch off all internet connections. In this case when trying to zoom the map shows a grey tile instead of keep displaying the previous zoom tiles.  
Was wondering if there is an option on the TileLayerOptions that we need to switch or if this is just the expected behaviour. Thanks in advance.

maRci002:
@velval Leaflet does the same thing so this is the intended behaviour if a Tile loads with error then we show empty Tile on screen because it's loaded so we can throw away old level Tiles.

https://leafletjs.com/ on PC I turned off my internet and gray Tiles appear when zooming in or out. You can use `errorImage:` which will show your placeholder Image if a tile loads with error.
However there is a little difference if you turn back your internet on Leaflet then it won't reload your Tiles (we do the same thing) however when you zoom out and back to error Tiles then they will be refetched. With PR #577 this behavior can be achived, note: I'm going to add network change detection to it and maybe a version where a little reload Icon will be on error Tiles and if user taps it then Tile will be reloaded (manual version).

And [582](https://github.com/fleaflet/flutter_map/pull/582#issuecomment-693403638) will allow to do anything with Tiles including error Tiles.

rorystephenson:
@maRci002 I'm working on refactoring the tile layer code again and I've noticed that the z coordinate of tiles (the zoom level of tiles) is always a whole number but the minNativeZoom and maxNativeZoom are doubles. I wondered if there is a use case for this? If it is possible to change them to ints instead that would allow me to simplify some of the tile related code and cache some values for various zoom levels.

maRci002:
> @maRci002 I'm working on refactoring the tile layer code again and I've noticed that the z coordinate of tiles (the zoom level of tiles) is always a whole number but the minNativeZoom and maxNativeZoom are doubles. I wondered if there is a use case for this?

If I read the doc it suggest me this should be integer.
```dart
  /// Maximum zoom number the tile source has available. If it is specified, the
  /// tiles on all zoom levels higher than maxNativeZoom will be loaded from
  /// maxNativeZoom level and auto-scaled.
  final double? maxNativeZoom;
```

I am not fully aware of the project at the moment but here you should test what happens for instance when `_tileZoom` is `19` and `zoom` is `20.1` (_`zoom` is calculated by `_clampZoom` which uses both `minNativeZoom` and `maxNativeZoom`_). I think it will call `_setView` which will reset `_tileZoom`
https://github.com/fleaflet/flutter_map/blob/3e8368d85d45e1431ed78488f0df102cef81b925/lib/src/layer/tile_layer/tile_layer.dart#L616-L621

rorystephenson:
Thanks for the response @maRci002! I think your right that in practice it is used like an integer anyway 👍 .

rorystephenson:
@maRci002 I just noticed that the tile provider rounds the tile numbers anyway so it looks like an int is fine 🎉 

https://github.com/fleaflet/flutter_map/blob/3e8368d85d45e1431ed78488f0df102cef81b925/lib/src/layer/tile_layer/tile_provider/base_tile_provider.dart#L29-L31

maRci002:
@rorystephenson I think `_tileZoom` should be an int? `_clampZoom` should accept int value and return int value and `maxNativeZoom` / `minNativeZoom` should be also int? values.

`maxZoom` / `minZoom` seems fine as double values for instance if `maxZoom` is 3.2 then 3.200001 will remove the tiles (_by setting: `tileZoom = null`_):
https://github.com/fleaflet/flutter_map/blob/3e8368d85d45e1431ed78488f0df102cef81b925/lib/src/layer/tile_layer/tile_layer.dart#L518-L522

PR #504, opened 2020-02-02
Updated example to androidX

porfirioribeiro:
Update example to use androidX.
No more androidX warning when running the example

PR #575, opened 2020-04-07
Update example app (use AndroidX)

vinicentus:
Use kotlin and swift as languages. Support androidx and web. The name was updated to flutter_map_example since it varied across files, and a version number was added.

I regenerated the app and copied over the dart files etc. Related to #504.

Tested on android. I would appreciate it if someone could test this on ios (and possibly web). I am not completely sure if the assets work on ios...

If you have any suggestions or find any mistakes here, feel free to point them out :)


vinicentus:
@dpatrongomez have you tested this?

dpatrongomez:
> @dpatrongomez have you tested this?

Yes, It goes nice

RichLewis007:
You've made a lot of nice changes here that will help people who are new to flutter_map to use, understand and modify the Example. Thanks!

johnpryan:
Some permissions needed to be modified to show a `package:location` sample, could you see if it's possible to fix the merge conflicts or re-do these steps?

vinicentus:
I can take a look at it tomorrow.

vinicentus:
I will create a new pull request, as it is easier for me.

Before I can do that I'd like to know why this line was included:
https://github.com/fleaflet/flutter_map/blob/eb0cf9cd2cad4823a1902d3db5231e59d801e519/example/android/build.gradle#L11
It seems to me as if it is completely unnecessary, as I don't think this library uses any [google services](https://developers.google.com/android/guides/google-services-plugin), but I could be wrong.
Could @raacker maybe elaborate?

And then there's this commit, which seems a bit strange to me. Shouldn't this file be automatically generated? Why are these changes needed, and can this be solved by just regenerating the project?
https://github.com/fleaflet/flutter_map/commit/60591815794b96ee76427bbfadcb58f181ee9fa0

johnpryan:
Regenerating the XCode project makes sense to me. I'm not sure why `google-services` is necessary, since `package:location` doesn't require any Android specific configuration: https://pub.dev/packages/location

vinicentus:
Okay. I will remove ‘google-services‘, regenerate the example, and make a new PR.

PR #577, opened 2020-04-07
WIP: Evict error tiles

### LINKED ISSUE #576: Tiles will not load after usage without a network
When the connectivity was lost and is restored the tiles that failed to load when the connectivity was lost will continue to fail to load:
`SocketException: Failed host lookup: 'a.tile.openstreetmap.org' (OS Error: No address associated with hostname, errno = 7)`

Video of the issue:
https://streamable.com/9kaguw

Some time after the connectivity is restored an exception is caught:
image

Tested with ac52f28716f799b2265da280e79ff56dc4950e1a
Flutter version 1.15.17
Framework revision 2294d75bfa (5 weeks ago), 2020-03-07 00:28:38 +0900
Dart version 2.8.0 (build 2.8.0-dev.12.0 9983424a3c)

maRci002:
Hello, this PR resolves #576 issue, this problem existed even in 0.8.2 release. Fortunately #572 made it possible to support Tile evict calls. `TileLayerOptions` got a `EvictErrorTileStrategy` enum which is currently `none` by default!

Maybe every `TileProvider` should define a default one if user not provides explicit `evictErrorTileStrategy` enum, forinstace #564 (-- @bugDim88 anyway what does storage tile caching do with errors? --) may want default `none` and `others not`, or `MBTilesImageProvider` my use `none` by default because if a Tile not exists on file system then maybe we shouldn't evict Tile image.

0.8.2 preview problem:
0 8 2 release

0.9.0 preview problem:
9 0 release

Preview the fix with `EvictErrorTileStrategy.dispose` this behaves like Leaflet if a Tile not retained anymore, and some time later Tile should be refeched this option simulates it:
9 0 release fix

Preview the fix with `EvictErrorTileStrategy.notVisible`:
evict not visible

closes #576


maRci002:
Ok, I think this is a little bit over engineered, maybe I should just listen for connection status (note:  it only provides information if there is a network connection, but not if the network is connected to the Internet) and it should redownload the currently visible error tiles too (note: this is not true if `TileProvider` using filesystem). However it won't fix siituations like this #580.

maRci002:
Don't merge this PR I'm going to add support for network changes soon.

johnpryan:
@maRci002 what's the status on this?

kengu (approved):

PR #584, opened 2020-04-11
Observe TileLayerOptions changes

### LINKED ISSUE #583: Issues when change map tiles URL dynamically
First thing, congratulation! The v0.9.0 release include really nice features!

Since this new update, when I try to change the map tiles URL while running the app, it takes a couple seconds, after to start navigating the mapo with the finger, until all tiles are updated. Before, this change was instantly performed.

Here's how I change the URL dynamically:
```
FlutterMap(
  options: MapOptions(
    ...
  ),
  layers: <LayerOptions>[
    TileLayerOptions(
      urlTemplate: context.watch<MapProvider>().mapUrl,
      backgroundColor: Theme.of(context).canvasColor,
    ),
    MarkerLayerOptions(
      ...
    ),
  ],
),
```

Is this a bug, or is there a better method to update the map URL, for example using the `MapController` class? Thank!

maRci002:
Hello, this PR adds support for detecting URL changes / tileSize changes / updateInterval changes and reloads Tiles if necessary.

change layer

closes #583

### DIFF
--- a/lib/src/layer/tile_layer.dart
+++ b/lib/src/layer/tile_layer.dart
@@ -884,6 +884,8 @@ class _TileLayerState extends State<TileLayer> with TickerProviderStateMixin {
       print(error);
 
       tile.loadError = true;
+    } else {
+      tile.loadError = false;
     }
 
     var key = _tileCoordsToKey(coords);
--- a/lib/src/layer/tile_layer.dart
+++ b/lib/src/layer/tile_layer.dart
@@ -317,6 +317,42 @@ class _TileLayerState extends State<TileLayer> with TickerProviderStateMixin {
     _update(null);
     _moveSub = widget.stream.listen((_) => _handleMove());
 
+    _initThrottleUpdate();
+  }
+
+  @override
+  void didUpdateWidget(TileLayer oldWidget) {
+    super.didUpdateWidget(oldWidget);
+    var reloadTiles = false;
+
+    final oldUrl = oldWidget.options.urlTemplate ??
+        oldWidget.options?.wmsOptions?._encodedBaseUrl;
+    final newUrl = options.urlTemplate ?? options?.wmsOptions?._encodedBaseUrl;
+    if (oldUrl != newUrl) {
+      // URL has been changed drop all Tiles and reload them
+      reloadTiles = true;
+    }
+
+    if (oldWidget.options.tileSize != options.tileSize) {
+      // tileSize has been changed drop all Tiles and reload them
+      _tileSize = CustomPoint(options.tileSize, options.tileSize);
+      reloadTiles = true;
+    }
+
+    if (oldWidget.options.updateInterval != options.updateInterval) {
+      // updateInterval has been changed
+      _throttleUpdate?.close();
+      _initThrottleUpdate();
+    }
+
+    if (reloadTiles) {
+      _removeAllTiles();
+      _resetView();
+      _update(null);
+    }
+  }
+
+  void _initThrottleUpdate() {
     if (options.updateInterval == null) {
       _throttleUpdate = null;
     } else {
@@ -331,12 +367,12 @@ class _TileLayerState extends State<TileLayer> with TickerProviderStateMixin {
 
   @override
   void dispose() {
-    super.dispose();
-
     _removeAllTiles();
     _moveSub?.cancel();
     options.tileProvider.dispose();
     _throttleUpdate?.close();
+
+    super.dispose();
   }
 
   @override
--- a/lib/src/layer/tile_layer.dart
+++ b/lib/src/layer/tile_layer.dart
@@ -141,16 +141,29 @@ class TileLayerOptions extends LayerOptions {
   ///
   final Map<String, String> additionalOptions;
 
-  // Tiles will not update more than once every `updateInterval` milliseconds
-  // (default 200) when panning.
-  // It can be 0 (but it will calculating for loading tiles every frame when panning / zooming, flutter is fast)
-  // This can save some fps and even bandwidth
-  // (ie. when fast panning / animating between long distances in short time)
+  /// Tiles will not update more than once every `updateInterval`
+  /// (default 200 milliseconds) when panning.
+  /// It can be null (but it will calculating for loading tiles every frame when panning / zooming, flutter is fast)
+  /// This can save some fps and even bandwidth
+  /// (ie. when fast panning / animating between long distances in short time)
   final Duration updateInterval;
 
-  // Tiles fade in duration in milliseconds (default 100),
-  // it can 0 to avoid fade in
+  /// Tiles fade in duration in milliseconds (default 100),
+  /// it can be null to avoid fade in
   final Duration tileFadeInDuration;
+
+  /// Opacity start value when Tile starts fade in (0.0 - 1.0)
+  /// Takes effect if `tileFadeInDuration` is not null
+  final double tileFadeInStart;
+
+  /// Opacity start value when an exists Tile starts fade in with different Url (0.0 - 1.0)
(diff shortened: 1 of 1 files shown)


lib/src/layer/tile_layer.dart
```diff
@@ -317,6 +317,42 @@ class _TileLayerState extends State<TileLayer> with TickerProviderStateMixin {
...
+    if (oldWidget.options.tileSize != options.tileSize) {
+      // tileSize has been changed drop all Tiles and reload them
+      _tileSize = CustomPoint(options.tileSize, options.tileSize);
+      reloadTiles = true;
+    }
+
+    if (oldWidget.options.updateInterval != options.updateInterval) {
+      // updateInterval has been changed
+      _throttleUpdate?.close();
+      _initThrottleUpdate();
```
porfirioribeiro, line 345: Shouldn't also set `reloadTiles = true;`?

maRci002: I think not because `updateInterval` is just milliseconds which allows you define how often do you want search for new Tiles (_by default on mobile Leaflet it only loads new tiles when panning ends_).

So basicly if you configure it to `2000ms (2sec)` by default and you change some time later to `200ms (0.2 sec)` then this shouldn't affect already loaded Tiles however it will just run `_update` method (searching) more often.

porfirioribeiro: Ohh sure, my bad :p

Good job @maRci002

lib/src/layer/tile_layer.dart
```diff
@@ -317,6 +317,42 @@ class _TileLayerState extends State<TileLayer> with TickerProviderStateMixin {
...
+    _initThrottleUpdate();
+  }
+
+  @override
+  void didUpdateWidget(TileLayer oldWidget) {
+    super.didUpdateWidget(oldWidget);
+    var reloadTiles = false;
+
+    final oldUrl = oldWidget.options.urlTemplate ??
+        oldWidget.options?.wmsOptions?._encodedBaseUrl;
```
avioli, line 329: Given that `options.wmsOptions` takes precedence once set - that should be the first to check, shouldn't it?

maRci002: If `wmsOptions` is present then `urlTemplate` should be `null` if is `not null` then programmer made a mistake. I will change code what you suggested however this shouldn't be fixed at this side.

maRci002:
Old tiles can be overwritten too with new ones without reload see #583.

kengu:
@johnpryan - I think this PR should be prioritized, together with #577. It solves some real problems that blocks actual usage in the wild. We are using flutter_map in an [open source app](https://github.com/DISCOOS/sarsys-app) for search and rescue, which demands good behavior when network conditions are poor. These PRs resolves some long-standing issues for us. 

@maRci002 - this is awesome work, thanks!

kengu:
@maRci002 - I think there are some additional use-cases to cover, which I'm currently looking into. Most importantly; tracking connectivity changes from offline to online. Currently, there is no way of externally forcing a invalidation of tiles with `Tile.loadError=true`. One solution could be to add the method `MapController.evictErrorTiles()`. This could be used to evict all error tiles when the parent widget of FlutterMap detects an offline to online transitions has occurred. I think there are probably more use-cases for this type of capability. This solutions allows for handling of connectivity issue decoupled from flutter_map.

maRci002:
@kengu this PR is ready to merge.

#577 is still under development however you are right `TileLayer` shouldn't detect connection issues because an app may have multiple `TileLayers` but is is still possible one or more layer can be indenpendent from internet forinstance `AssetTileProvider`. `abstract class TileProvider` may has a method which return `true` if it depends on internet connection so `Flutter Map` can check each layer when connection restores.

kengu:
> @kengu this PR is ready to merge.
> 
> #577 is still under development however you are right `TileLayer` shouldn't detect connection issues because an app may have multiple `TileLayers` but is is still possible one or more layer can be indenpendent from internet forinstance `AssetTileProvider`. `abstract class TileProvider` may has a method which return `true` if it depends on internet connection so `Flutter Map` can check each layer when connection restores.

Yes, that my line of thinking also. Did a quick test locally using the same approach as you have in this PR. Just adding a "offline" member to the options and checking if it is changed and false, which evicts all error tiles works like a charm when combined with the connectivity pacakge.

maRci002:
Connectivity package will trigger even if you just connect to router but it may has not connection to internet.

If you just play with `offline` flag trick without setting any `evictErrorTileStrategy` then this could happen:

1. Open map, it loads 1,2,3,4 tiles
2. Internet gone
3. User pans map to 5,6,7,8 tiles, they will be error tiles (don't forget image provider cache these errors: `NetworkImage docs:` _The image will be cached regardless of cache headers from the server._)
4. User pans map to like 101, 102, 103, 104 they will be error tiles <- `_tiles` won't hold 1,2,3,4 and 5,6,7,8 tiles anymore because those tiles aren't current or retained anyomore because user panned far away.
5. Internet comes back
6. 101, 102, 103, 104 tiles will be loaded
7. User pans back to 5,6,7,8 tiles these will be loaded as error since `_tiles` did not held any reference to these tiles anymore when connection restored so there wasn't any canche to evict them. <- can be fixed if you use forinstance `notVisible` stategy mixed with `offline trick`.

kengu:
@maRci002 - that's right. I'm actually using `EvictErrorTileStrategy.dispose` and added what's probably missing in #577 which is disposing in `_update` method on tiles that are current (visible). This works like a charm together.  

Although, the impact of not setting `EvictErrorTileStrategy` is minimized when combining [connectivity](https://pub.dev/packages/connectivity) package with [data_connection_checker](https://pub.dev/packages/data_connection_checker), which is what I've opted for. Made a singleton class of the combination named [ConnectivityService](https://github.com/DISCOOS/sarsys-app/blob/master/lib/services/connectivity_service.dart), and wrapped it in a [NetworkSensitive](https://github.com/DISCOOS/sarsys-app/blob/master/lib/widgets/network_sensitive.dart) widget which rebuild it's children each time connection type or status changes . In the parent widget of FlutterMap just `TileLayerOptions.offline = Provider.of<ConnectivityStatus>(context).status` in the build method for FlutterMap and tiles will reload automatically. The only remaining "error situation" now it a real error from the tile-server, which is normal operation.

S-Man42:
@maRci002 Could you please provide the source code of your example above. I find it very interesting.

maRci002:
@S-Man42 I don't have the example code but it ist just a `FloationActionButton` which changes `WMSTileLayerOptions' baseUrl` via `setState`.

However since #585 needs this PR I made there [an example](https://github.com/johnpryan/flutter_map/issues/613#issuecomment-626572293) just register your `String url;` and write your own `setState()` logic which replaces `url`

S-Man42:
That works for me very well! Great! Unfortunately I cannot use in my app currently, because I am using the Marker Popup Plugin which depends on flutter_map 1.4.0 and has problems with your git commit checkout. We need to get this merged!

fryette:
Any chance to merge it? Same issue as for @S-Man42

lib/src/layer/tile_layer.dart
```diff
@@ -952,11 +1034,14 @@ class Tile implements Comparable<Tile> {
     _imageStream?.removeListener(_listener);
```
fryette, line 1034: I think we also need to dispose of the animation controller, that was a reason for crash

fryette:
I used this commit for about month and found that we need also to dispose Animation Controller
@maRci002 could you please push change?

Lidgett:
Any chance of merging this?  Having the option to reset/reload the tiles is pretty essential in my opinion.

JanikoNaber:
Please we need one additional approval. Could anyone have a look at this PR?
This function is great!

fryette:
@maRci002 please approve proposed change

PR #585, opened 2020-04-13
Support retina mode

### LINKED ISSUE #212: add support for detectRetina display
there is an option in original leaflet js library which is used for retina display(high pixel density) however I couldn't find such feature in FlutterMap widget. this feature will prevent map images render blurry in devices with high pixel density 
here is the github issue about this problem [Blurry Maps](https://github.com/Leaflet/Leaflet/issues/6068#issuecomment-368256572)
here is leaflet document description about this option

> If true and user is on a retina display, it will request four tiles of half the specified size and a bigger zoom level in place of one to utilize the high resolution.

maRci002:
Hello, I made some changes to support simulating retina mode (_note: If geoserver supports retina `@2` tiles then it it advised to use them instead of simulating_).

It won't detect if device supports retina because I don't want force user to use `Material app` because of `MediaQuery.of(context)`.

However it is documented how to use it:
```
  /// If `true`, it will request four tiles of half the specified size and a
  /// bigger zoom level in place of one to utilize the high resolution.
  ///
  /// If `true` then MapOptions's `maxZoom` should be `maxZoom - 1` since retinaMode
  /// just simulates retina display by playing with `zoomOffset`.
  /// If geoserver supports retina `@2` tiles then it it advised to use them
  /// instead of simulating it (use {r} in the [urlTemplate])
  ///
  /// It is advised to use retinaMode if display supports it, write code like this:
  /// TileLayerOptions(
  ///     retinaMode: true && MediaQuery.of(context).devicePixelRatio > 1.0,
  /// ),
  final bool retinaMode;

```
Moreover if retina is enabled then `MapOptions`'s  `maxZoom` should be `maxZoom - 1`, currently `MapState` doesn't detect any `maxZoom` based on Layers like Leaflet _(If not specified and at least one `GridLayer` or `TileLayer` is in the map, the highest of their `maxZoom` options will be used instead.)_.

retina

Note: this PR needs #584 so merge this only when it is accepted.

closes #212

### DIFF
--- a/lib/src/layer/tile_provider/tile_provider.dart
+++ b/lib/src/layer/tile_provider/tile_provider.dart
@@ -28,7 +28,7 @@ abstract class TileProvider {
       'y': coords.y.round().toString(),
       'z': z.round().toString(),
       's': getSubdomain(coords, options),
-      // 'r': options.retinaMode ? '@2x' : '',
+      'r': '@2x',
     };
     if (options.tms) {
       data['y'] = invertY(coords.y.round(), z.round()).toString();
--- a/lib/src/layer/tile_provider/tile_provider.dart
+++ b/lib/src/layer/tile_provider/tile_provider.dart
@@ -17,7 +17,8 @@ abstract class TileProvider {
 
   String getTileUrl(Coords coords, TileLayerOptions options) {
     if (options.wmsOptions != null) {
-      return options.wmsOptions.getUrl(coords, options.tileSize.toInt());
+      return options.wmsOptions
+          .getUrl(coords, options.tileSize.toInt(), options.retinaMode);
     }
 
     var z = _getZoomForUrl(coords, options);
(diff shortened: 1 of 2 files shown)


lib/src/layer/tile_provider/tile_provider.dart
```diff
@@ -26,7 +27,8 @@ abstract class TileProvider {
       'x': coords.x.round().toString(),
       'y': coords.y.round().toString(),
       'z': z.round().toString(),
-      's': getSubdomain(coords, options)
+      's': getSubdomain(coords, options),
+      // 'r': options.retinaMode ? '@2x' : '',
```
HugoHeneault, line 31: Why is this commented? It would be nice to download the @2x tiles?

mat8854:
@maRci002 @HugoHeneault I compared rendering quality of this Flutter lib to the native `osmdroid` lib a few weeks. 

The map looked better and more crisp on `osmdroid` (same bounding box). Is this related to this density fix?

maRci002:
#584 updated so I merged that into this to avoid conflicts whenever they got approved.

Issue #593, opened 2020-04-19
Handle exception on move without internet connection


beerline:
Hi flutter_map team! 

How I can handling exception while user move or zoom map but internet connect is lost?

Steps to do:
Open app with interne

1. t connection. As result we see the map
2. enable "Airplane mode"
3. tray to zoom the map
4. getting console errors

Code the same as in [this example](https://github.com/johnpryan/flutter_map/blob/master/example/lib/pages/zoombuttons_plugin_option.dart) 

console errors
```
I/flutter (15443): SocketException: Failed host lookup: 'c.tile.openstreetmap.org' (OS Error: No address associated with hostname, errno = 7)
I/flutter (15443): SocketException: Failed host lookup: 'a.tile.openstreetmap.org' (OS Error: No address associated with hostname, errno = 7)
I/flutter (15443): SocketException: Failed host lookup: 'c.tile.openstreetmap.org' (OS Error: No address associated with hostname, errno = 7)
I/flutter (15443): SocketException: Failed host lookup: 'b.tile.openstreetmap.org' (OS Error: No address associated with hostname, errno = 7)
I/flutter (15443): SocketException: Failed host lookup: 'c.tile.openstreetmap.org' (OS Error: No address associated with hostname, errno = 7)
I/flutter (15443): SocketException: Failed host lookup: 'a.tile.openstreetmap.org' (OS Error: No address associated with hostname, errno = 7)
I/flutter (15443): SocketException: Failed host lookup: 'b.tile.openstreetmap.org' (OS Error: No address associated with hostname, errno = 7)
I/flutter (15443): SocketException: Failed host lookup: 'b.tile.openstreetmap.org' (OS Error: No address associated with hostname, errno = 7)
I/flutter (15443): SocketException: Failed host lookup: 'a.tile.openstreetmap.org' (OS Error: No address associated with hostname, errno = 7)
```

Thank you for your help!

avioli:
The only way I can think of handling this is using a custom `TileProvider`, that keeps track of the current connectivity state and only loads tiles if connected to the internet or if a socket exception happened then await connectivity change to retry. It might have to use a custom `ImageProvider` since `getImage` is not async.

A plugin, paired with a tile provider might be a better fit, since it has access to the map state, thus will be able to invoke a reload on the visible tiles.

maRci002:
[...]

PR #600, opened 2020-04-24
Handle exception on move without internet connection

beerline:
Hi **flutter_map** team!

Here is desigion for issue #593 

- add errorTileCallback for handling errors while no tiles
- example how to use errorTileCallback

Main goal to catch this exception, is to notify user about internet connection lost.

### DIFF
--- a/example/lib/pages/tile_loading_error_handle.dart
+++ b/example/lib/pages/tile_loading_error_handle.dart
@@ -1,0 +1,76 @@
+import 'package:flutter/material.dart';
+import 'package:flutter_map/flutter_map.dart';
+import 'package:latlong/latlong.dart';
+
+import '../widgets/drawer.dart';
+
+class TileLoadingErrorHandle extends StatefulWidget {
+  static const String route = '/tile_loading_error_handle';
+
+  @override
+  _TileLoadingErrorHandleState createState() => _TileLoadingErrorHandleState();
+}
+
+class _TileLoadingErrorHandleState extends State<TileLoadingErrorHandle> {
+  @override
+  Widget build(BuildContext context) {
+    bool _needLoadingError = true;
+
+    return Scaffold(
+      appBar: AppBar(title: Text('Tile Loading Error Handle')),
+      drawer: buildDrawer(context, TileLoadingErrorHandle.route),
+      body: Padding(
+        padding: EdgeInsets.all(8.0),
+        child: Column(
+          children: [
+            Padding(
+              padding: EdgeInsets.only(top: 8.0, bottom: 8.0),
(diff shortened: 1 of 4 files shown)


example/lib/widgets/drawer.dart
```diff
@@ -156,6 +157,13 @@ Drawer buildDrawer(BuildContext context, String currentRoute) {
             Navigator.pushReplacementNamed(context, OverlayImagePage.route);
           },
         ),
+        ListTile(
+          title: const Text('Tile loading error handle'),
+          selected: currentRoute == OverlayImagePage.route,
```
avioli, line 162: Shouldn’t this be `TileLoadingErrorHandle.route`?

example/lib/pages/tile_loading_error_handle.dart
```diff
@@ -0,0 +1,75 @@
...
+        child: Column(
+          children: [
+            Padding(
+              padding: EdgeInsets.only(top: 8.0, bottom: 8.0),
+              child: Text('Turn on Airplane mode and try to move or zoom map'),
+            ),
+            Flexible(
+              child: Builder(builder: (BuildContext context) {
+                return FlutterMap(
+                  mapController: MapController(),
```
avioli, line 33: This is a bad example use `MapController` use. A map controller is not necessary.

avioli, line 33: Sorry for nitpicking - this is a bad example use MapController use. A map controller is not necessary.

lib/src/layer/tile_layer.dart
```diff
@@ -181,7 +186,8 @@ class TileLayerOptions extends LayerOptions {
       // Tiles fade in duration in milliseconds (default 100),
       // it can 0 to avoid fade in
       int tileFadeInDuration = 100,
-      rebuild})
+      rebuild,
+      this.errorTileCallback,})
```
avioli, line 190: Another nitpick - usually `rebuild` is last.

beerline:
i fixed all issues

avioli:
@beerline I’m sorry if I gave you the idea I’m somehow able to merge this pull request into this repo. I was just pointing the issues I saw.

Apparently there is a formatting issue in `drawer.dart` that needs fixing.

@johnpryan can merge it, after reviewing if for himself, but it is up to him to find the time to do so and make a choice if the project requires this feature at all.

johnpryan:
This is exposing some weird quirks of the API. Tile is meant to be an internal class, so this callback's API might be subject to change in the future. But I think overall it makes sense to provide a way to catch these problems in the app.

johnpryan:
Thanks for reviewing @avioli

PR #611, opened 2020-05-08
Polyline culling

### LINKED ISSUE #610: Drawing thousands of lines causes the program to crash
I needed to draw thousands of roads on the map, but it was very slow at a hundred roads, and when I drew a thousand roads, the program quit and crashed.
Is there a way to fulfill this requirement?

maRci002:
Added Polyline culling based on screen bounds. (This PR is the polyline version of #449)

If you set `polylineCulling` to `true` on `PolylineLayerOptions` then any polylines outside of the screen’s bounds won't be drawn, therefore, saving rendering time and improving performance when displaying a large set of polylines.

I also updated `PolygonLayerOptions` to calculate boundingBox only when necessary.

closes #610

### DIFF
--- a/lib/src/layer/polygon_layer.dart
+++ b/lib/src/layer/polygon_layer.dart
@@ -11,9 +11,17 @@ class PolygonLayerOptions extends LayerOptions {
   final bool polygonCulling;
 
   /// screen space culling of polygons based on bounding box
-  PolygonLayerOptions(
-      {this.polygons = const [], this.polygonCulling = false, rebuild})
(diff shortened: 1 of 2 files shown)


maRci002:
I think I can also make `PolylinePainter` a bit faster, I'll look through it's code later.

ibrierley:
Sounds good, was looking at doing something like this myself. Lots of good work you've done.

AlecADAS:
Quite late to this, but is there some sort of culling option for flutter googlemaps?

PR #619, opened 2020-05-16
feat(*): widget layers

### LINKED ISSUE #459: How to rebuild a layer
Hi,
I need some help to force to rebuild a layer. If it's not possible, how can I rebuild all the map ?
Could you provide me any example ?
Another possibility is to disable cache on a layer. 
Thanks

### LINKED ISSUE #482: Trigger rebuilds from MapState in plugins
Hi. I am currently porting a package to a plugin ([this one](https://github.com/synw/map_controller)). It needs to be able to the control map rebuilds (it manages the map assets). The problem is that the `move` method from `MapState` checks if the center or zoom has changed, which is not the case if I add a marker or something.

Is there a way to trigger a rebuild from a plugin, did I miss something? If not would it be possible to remove the check and make `move` to run even if no zoom or center has changed? The code that blocks this is [this line](https://github.com/johnpryan/flutter_map/blob/master/lib/src/map/map.dart#L113).

### LINKED ISSUE #535: Performance issue when using setState()
Hi, I am using your plugin on version flutter_map: ^0.8.2 and Flutter 1.12.13 hotfix 7 and I observe some performance issue. 

I want to display user's current location on the map using location plugin but when I listen to the stream and new value appears, only thing to reflect changes on the map is use setState, but there is the issue - when I rebuild whole map, it starts to be slow and I observe some performance issue (when I click on marker and open new page using Navigator etc.) 

Interesting is that I notice performance drop even if I use only one time location Future - so map is rebuilded only once. 

Thanks for your time!

ruizalexandre:
@johnpryan 

> I rewrite all my last PR 
> Compatible with plugins and can combine last way to use layers (with LayerOptions).

---

🎸 **No breaking changes**

---

🤖 _What's inside?_
- Adding `MapState.of(context)` to get the current MapState
- Adding `forceRebuild` method on MapState for rebuild all map
- Can use simple widget with FlutterMap now with the new property `children`
- Fix initState/dispose lifecyle -> called after TileLayer code

📜 _More explanation_

__What the sort in stack with children and layers?__
`children` property is before `layers` property in stack of Widget.

__I want create a simple plugin, how?__
Use simple Flutter Widget, example `Text("I'm a simple plugin now")`

__I want create a plugin who can access to mapState, how?__
Create a Flutter Widget with `MapState.of(context)` to get the current mapState.

🆘  _Linked issues_
Close #535
Close #459
Close #482

### DIFF
--- a/example/lib/pages/widgets.dart
+++ b/example/lib/pages/widgets.dart
@@ -1,0 +1,130 @@
+import 'dart:async';
+
+import 'package:flutter/material.dart';
+import 'package:flutter_map/plugin_api.dart';
+import 'package:latlong/latlong.dart';
+
+import '../pages/zoombuttons_plugin_option.dart';
+import '../widgets/drawer.dart';
+
+class WidgetsPage extends StatelessWidget {
+  static const String route = 'widgets';
+
+  @override
+  Widget build(BuildContext context) {
+    return Scaffold(
+      appBar: AppBar(title: Text('Widgets')),
+      drawer: buildDrawer(context, WidgetsPage.route),
+      body: Padding(
+        padding: EdgeInsets.all(8.0),
+        child: Column(
+          children: [
+            Flexible(
+              child: FlutterMap(
+                options: MapOptions(
+                  center: LatLng(51.5, -0.09),
+                  zoom: 5.0,
+                  plugins: [
+                    ZoomButtonsPlugin(),
+                  ],
+                ),
+                layers: [
+                  ZoomButtonsPluginOption(
+                    minZoom: 4,
+                    maxZoom: 19,
+                    mini: true,
+                    padding: 10,
+                    alignment: Alignment.bottomLeft,
+                  )
+                ],
+                children: <Widget>[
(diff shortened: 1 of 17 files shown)


example/lib/pages/widgets.dart
```diff
@@ -0,0 +1,130 @@
...
+        _marker = _markers[_markerIndex];
+        _markerIndex = (_markerIndex + 1) % _markers.length;
+      });
+    });
+  }
+
+  @override
+  void dispose() {
+    super.dispose();
+    _timer.cancel();
```
S-Man42, line 94: Could this lead into an issue: canceling the timer after disposing?

ruizalexandre: Fixed :) thx @S-Man42

fryette:
Any news on it?

johnpryan:
Looks good to me, The next step would be to make the `options` classes go away eventually. For example, `MarkerLayerWidget` could just have the parameters in `MarkerLayerOptions` in it's constructor.

PR #637, opened 2020-05-29
fix(*): tile display

### LINKED ISSUE #608: Empty map tails when return to same route
Hi flutter_map team! 

### LINKED ISSUE #636: Tiles are not shown if used in another instance of FlutterMap

video-2020-05-28-16-56-41

Best Regards

ruizalexandre:
This PR fix **ONLY** fetch images !

-> Fetch image after adding in tiles dict

**I found an error on rotation present on version 9. 
Tile layer logic is broken, fix here. => https://github.com/johnpryan/flutter_map/pull/641**

🚀 Works fine on iOS/Android
🔄 Race condition fixed

Fix #636
Fix #608

### DIFF
--- a/lib/src/layer/tile_layer.dart
+++ b/lib/src/layer/tile_layer.dart
@@ -428,7 +428,7 @@ class _TileLayerState extends State<TileLayer> with TickerProviderStateMixin {
           for (var tile in _tiles.values) {
             tile.imageProvider = options.tileProvider
                 .getImage(_wrapCoords(tile.coords), options);
-            tile.loadTileImage();
+            tile.fetchImage();
           }
         } else {
           reloadTiles = true;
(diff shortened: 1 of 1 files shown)


lib/src/layer/tile_layer.dart
```diff
@@ -327,16 +325,18 @@ class _TileLayerState extends State<TileLayer> with TickerProviderStateMixin {
         ),
       )..listen(_update);
     }
+
+    super.initState();
```
avioli, line 329: Why is `super.initState` called last? Only `super.dispose` is advised to be called last.

ruizalexandre: I'm following instruction from Dart/Flutter Plugin for VSCode only

```@override
  void initState() {
    // TODO: implement initState
    super.initState();
  }
```

ruizalexandre: I’lol update this line to first line :)

kuhnroyal:
Can you rebase this so this can get into a minor patch version.
The last version had a lot of fixes but this is till missing.

catmorte:
hey! please do something!)

ruizalexandre:
This week I’ll rebase all of this and update all my PRs

ruizalexandre:
> Can you rebase this so this can get into a minor patch version.
> The last version had a lot of fixes but this is till missing.

Done :)

PR #641, opened 2020-06-01
WIP: fix(*): tile display on rotation

### LINKED ISSUE #608: Empty map tails when return to same route

### LINKED ISSUE #636: Tiles are not shown if used in another instance of FlutterMap

ruizalexandre:
@johnpryan 
Since the version 9, rotation and navigate into many instance of Flutter Map was broken.
I rewrite a part of tile layer logic. 

Tile layer now use a simple grid and each case is a simple tile widget.
`TileWidget` fetch image if no `ImageInfo` is available from `_tiles` at the right zoom.

Best regards ! 

Fix #636
Fix #608

### DIFF
--- a/lib/src/layer/tile_layer.dart
+++ b/lib/src/layer/tile_layer.dart
@@ -372,15 +372,14 @@ class TileLayer extends StatefulWidget {
   }
 }
 
-class _TileLayerState extends State<TileLayer> with TickerProviderStateMixin {
+class _TileLayerState extends State<TileLayer> {
   MapState get map => widget.mapState;
-
   TileLayerOptions get options => widget.options;
+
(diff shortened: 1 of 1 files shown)


ruizalexandre:
Need to rework my PR

maRci002:
#715 fixed multiple map instance / navigaton gray tiles problem
#695 fixed rotation / interactive flag on fly change problem

ruizalexandre:
@maRci002 rotation fixed ?

maRci002:
> @maRci002 rotation fixed ?

Yes it is fixed.

PR #655, opened 2020-06-17
binded -> bound

saibotma:
"binded" is an incorrect conjugation of "bind". "bound" is the correct conjugation.


johnpryan:
good catch, thanks!

PR #662, opened 2020-06-21
Fix #446: Polyline rendering on web

### LINKED ISSUE #446: [web] disappearing bits when loading PolylineLayerOptions
I hope it is ok to open this issue derived from #440, even if for the web version of flutter.
Currently I am finding this the only really blocking issue for the web and would like to share my findings and ask for suggestions or fix ideas.

Right now I am wondering if the issues is something in the Polyline painter. 

For example, this is a map with background and markers (one of which an image... nifty :-)):
image

This is the same but lines instead of markers (background disappears):

image

This is background-lines-markers (background disappears, line is behind image):

image

And last but not least, same as before, but lines on top of markers, i.e. background-markers-lines:

image

As you can see, the last one seems perfect...
but if I zoom in on the line:

image

The image marker disappears, whereas the background remains.

If I then zoom even more, which makes the line to get out of the boundary of the image marker (which is fixed) and therefore overlay on the background map:

image

... the background disappears again.

The feeling is that whatever the line overlays directly, is "killed" by it.
I am sorry that I do not have a more code related report.

Any idea what might cause the issue?

hlin079g6:
This fixes Polylines failing to render on web, as described in #446 with --dart-define=FLUTTER_WEB_USE_SKIA=false.

The true root cause is that canvas.saveLayer does not work, see [flutter issue #48417](https://github.com/flutter/flutter/issues/48417). This works around that problem by using different blend modes, which works in most cases, but it is not a perfect fix. Overlapping translucent polylines with borders render incorrectly at the point of overlap. This seems rare enough to me.

On platforms where canvas.saveLayer does work, the fix has no effect. So overlapping translucent polylines with borders will continue to work correctly (tested with Android).

While I was in there, I fixed the border size for dotted lines and changed them to use paths. Large border sizes can cause the dots to overlap, and if they are translucent that results in stronger coloring where the circles overlap, which doesn't look right. Paths handle these kinds of overlaps.

### DIFF
--- a/lib/src/layer/polyline_layer.dart
+++ b/lib/src/layer/polyline_layer.dart
@@ -127,7 +127,7 @@ class PolylinePainter extends CustomPainter {
       ..strokeWidth = polylineOpt.strokeWidth
       ..strokeCap = StrokeCap.round
       ..strokeJoin = StrokeJoin.round
-      ..blendMode = BlendMode.src;
+      ..blendMode = BlendMode.srcOver;
 
     if (polylineOpt.gradientColors == null) {
       paint.color = polylineOpt.color;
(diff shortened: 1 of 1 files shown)


moovida:
I can confirm that this works (below a small extract of gps tracks with many many points) and removes big blocker for web:

image

Thanks @hlin079g6 !

johnpryan:
Looks good, thanks for the explanation!

PR #684, opened 2020-07-09
add Drawer to live location page

vinicentus:
simple and quick fix


vinicentus:
I see that this problem has already been taken care of in https://github.com/fleaflet/flutter_map/commit/d6b365ed7228da1e85bdbe3b19b20b313af20b81. Closing PR.

PR #702, opened 2020-07-25
Add scrolling support

### LINKED ISSUE #660: Support scroll to zoom and pinch to zoom for desktop and web
There is pinch to zoom functionality on mobile as far as i'm concirned, but i could not find scroll to zoom and pinch to zoom functionality for desktop and web.

Would be a great feature to be on par with e.g. Google Maps or OSM online versions.

### LINKED ISSUE #820: Zoom on scroll flutter web
Hi,

Is there a way to zoom in/zoom out on scroll up/scroll down on flutter web ?

All the best

jld3103:
A problem is that if scroll and the scroll again during the animation it first animates to the first location, but then animates to the second location which was during the animation.
I also noticed that the zoom and move animations have a little offset, but I think it's a flutter performance problem rather than one of this package.

fixes #660
fixes #820

### DIFF
--- a/lib/src/gestures/gestures.dart
+++ b/lib/src/gestures/gestures.dart
@@ -1,6 +1,7 @@
 import 'dart:async';
 import 'dart:math' as math;
 
+import 'package:flutter/gestures.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_map/flutter_map.dart';
 import 'package:flutter_map/src/gestures/interactive_flag.dart';
@@ -22,6 +23,15 @@ abstract class MapGestureMixin extends State<FlutterMap>
   void savePointer(PointerEvent event) => ++_pointerCounter;
 
   void removePointer(PointerEvent event) => --_pointerCounter;
+
+  void handlePointerSignal(PointerSignalEvent pointerSignal) {
+    if (pointerSignal is PointerScrollEvent) {
+      handleScroll(
+        pointerSignal.localPosition,
+        pointerSignal.scrollDelta.dy,
+      );
+    }
+  }
 
   var _rotationStarted = false;
   var _pinchZoomStarted = false;
@@ -48,6 +58,10 @@ abstract class MapGestureMixin extends State<FlutterMap>
   Animation _doubleTapZoomAnimation;
   Animation _doubleTapCenterAnimation;
 
+  AnimationController _scrollController;
+  Animation _scrollZoomAnimation;
+  Animation _scrollCenterAnimation;
+
   int _tapUpCounter = 0;
   Timer _doubleTapHoldMaxDelay;
 
@@ -68,6 +82,9 @@ abstract class MapGestureMixin extends State<FlutterMap>
         AnimationController(vsync: this, duration: Duration(milliseconds: 200))
           ..addListener(_handleDoubleTapZoomAnimation)
           ..addStatusListener(_doubleTapZoomStatusListener);
+    _scrollController =
+        AnimationController(vsync: this, duration: Duration(milliseconds: 200))
+          ..addListener(_handleScrollZoomAnimation);
   }
 
   @override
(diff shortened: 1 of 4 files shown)


kengu:
@jld3103 Thanks for the the contribution! Can you resolve the conflicts and update this PR? I will review it when this is done.

jld3103:
I'll try. The implementations has some glitches that happen when you scroll to fast etc., but it works

jld3103:
Example does work, not sure if other configs are broken. I didn't have a look at the code since I wrote it so it might be buggy.

nickoe:
I guess this is an attempt to address https://github.com/fleaflet/flutter_map/issues/660 ?

jld3103:
I didn't see those issue, but it does address them, yes

jld3103:
@kengu what's up with review?

kengu:
@jld3103 I've tested it locally. I works somewhat as expected, but when min/max zoom is not not given it breaks down because the user can easily zoom too zoom-levels far beyond the actual zoom-bounds supported by the layer source. It then becomes difficult to return to the correct zoom layer effectively "breaking it" (the user only sees the background and believes that flutter_map is in an unrecoverable state). I think the problem is worsened by the fact that the scroll-to-zoom ratio is a bit high, and should probably be dynamic depending on how fast the scrolling occurs (mouse wheels that can scroll without resistance yields high scroll rates as one example).

jld3103:
Yes I know it's not smooth at all. It was just enough for what I needed it. I currently have no project using it, so I'll probably not work on this. Maybe someone else interested in this can rework this.

lib/src/gestures/gestures.dart
```diff
@@ -566,6 +583,32 @@ abstract class MapGestureMixin extends State<FlutterMap>
     }
   }
 
+  void handleScroll(Offset position, double delta) {
+    // Makes scrolling slower when zoomed in
+    final scale =
+        3 - mapState.zoom / (mapState.options.maxZoom ?? double.infinity) * 1.5;
```
kengu, line 589: The scroll-to-zoom ratio is to high. We need to take into account that high scroll rates could result in too rapid and unbounded zoom outside actual bounds supported by map sources. We need to take into account minZoom to ensure that both bounds are honoured if given, at that min zoom is 0 regardless of much the user zooms out. Maximum zoom should also have a practical bound to prevent the user from zooming too long inwards (zoom above 20 is not very usable, see for instance the [general zoom-level calculations made by Mapbox](https://docs.mapbox.com/help/glossary/zoom-level/))

lib/src/gestures/gestures.dart
```diff
@@ -624,6 +667,16 @@ abstract class MapGestureMixin extends State<FlutterMap>
     );
   }
 
+  void _handleScrollZoomAnimation() {
+    setState(() {
```
kengu, line 671: `setState` is not needed. Calling `mapState.move` will redraw the widget

lib/src/gestures/gestures.dart
```diff
@@ -706,7 +759,7 @@ abstract class MapGestureMixin extends State<FlutterMap>
 
   double _getZoomForScale(double startZoom, double scale) {
     var resultZoom =
-        scale == 1.0 ? startZoom : startZoom + math.log(scale) / math.ln2;
+        startZoom + ((scale > 0 ? 1 : -1) * (math.log(scale.abs()) / math.ln2));
```
kengu, line 762: You could add practical bounds here or inside `mapState.fitZoomToBounds`, for instance `max(0,min(zoomResult, 25)`

kengu (commented):
> Yes I know it's not smooth at all. It was just enough for what I needed it. I currently have no project using it, so I'll probably not work on this. Maybe someone else interested in this can rework this.

I've added some suggestions that shouldn't take that long to fix. If you could find some time to look at it would be much appreciated!

jld3103:
Ah I forgot this! Will apply your suggestions now

jld3103:
@kengu I used another method for selecting the scroll scale which should be better and more easy to understand (I already didn't know how the code I wrote worked).

jld3103:
I see someone else tried this too with #914. I tried it and I like how simple it is and also it doesn't have the issues I was facing, but there is no animation. Maybe we should use that implementation? @kengu @lucavinci

lucavinci:
I like this PR because of the animation, I could add animation to my PR though. My zoom code is much simpler. I could either simplify this PR or add animations to mine.

jld3103:
If you try to add the animation and it works I'd be happy to close this PR and let you do it.

aytunch:
It doesn't matter which PR is merged. Both are cool. But as a user, the animation is a must :)

jld3103:
But it's buggy :shrug:

aytunch:
any update for this PR?

kengu:
Resolved by #914. Closing

PR #715, opened 2020-08-04
Fix gray tiles when tile image is already avaible

### LINKED ISSUE #608: Empty map tails when return to same route

### LINKED ISSUE #636: Tiles are not shown if used in another instance of FlutterMap

### LINKED ISSUE #707: Tiles not appearing when using FlutterMap on multiple screens
Steps to reproduce:

1. Create two screens with FlutterMap (both maps should have identical center and zoom). Alternatively, you can run [this](https://bitbucket.org/maro24/tile_bug_example/src/master/) example.
2. Navigate from the first screen to the other.
3. Tiles displayed on the first screen never appear on the second one.
4. Panning in any direction will show tiles but the original area will remain unloaded.

This used to work on 0.8.2 but does not work since the changes to tile management on 0.9.0.

Removing the indicated line (965) from the _tileReady method in tile_layer.dart solves the issue. 

    var key = _tileCoordsToKey(coords);
    tile = _tiles[key]; //Removing this solves the issue
    if (null == tile) {
      return;
    }

I do not understand the package enough to know what purpose does keeping tiles in the _tiles hash map serve but so far I have not noticed any issues since deleting it so am quite curious about this.

### LINKED ISSUE #713: Tail class  _tileReady issue
hi,when used CachedNetworkTileProvider to load same url tile,loadTileImage method can case a error, it cause tile image not load.
for example:
first load:  "_imageStream.addListener(_listener);" will async exec ,so  _addTile method can put tile in _tiles.when _tileReady is ready.
    var key = _tileCoordsToKey(coords);
    tile = _tiles[key]; 
    if (null == tile) { //it ok ,not entry  here,exec load image
      return;
    };
when widget rebuild again:
because some url, _imageStream.addListener(_listener); already have cache image,so _tileReady  run now,but _tiles list is empty,so in _tileReady method,
var key = _tileCoordsToKey(coords);
    tile = _tiles[key]; //_tiles is empty,
    if (null == tile) { //entry  here,exit ,so no load image.
      return;
    };

am i have some config error?or this is a issue?
please tell me.

maRci002:
Hello, this PR make sure the `loadTileImage`'s listeners runs after the `tile` is in `_tiles` map .

```dart
  /// Adds a listener callback that is called whenever a new concrete [ImageInfo]
  /// object is available or an error is reported. If a concrete image is
  /// already available, or if an error has been already reported, this object
  /// will notify the listener synchronously.
  ///
  /// If the [ImageStreamCompleter] completes multiple images over its lifetime,
  /// this listener's [ImageStreamListener.onImage] will fire multiple times.
  ///
  /// {@macro flutter.painting.imageStream.addListener}
  void addListener(ImageStreamListener listener) {
```

> If a concrete image is already available, or if an error has been already reported, this object will notify the listener synchronously.

Since this code:
```dart
var key = _tileCoordsToKey(coords);
tile = _tiles[key];
if (null == tile) {
  return;
}
```
skipped fade-in part, the tiles remained gray. This only happaned if `image object` was in phone's ram so there wasn't an async callback since `image` was instantly avaible, however this fix doesn't assume `async image load` anymore.

closes #707
closes #713
closes #636
closes #608

### DIFF
--- a/lib/src/layer/tile_layer.dart
+++ b/lib/src/layer/tile_layer.dart
@@ -936,7 +936,7 @@ class _TileLayerState extends State<TileLayer> with TickerProviderStateMixin {
 
   void _addTile(Coords<double> coords) {
     var tileCoordsToKey = _tileCoordsToKey(coords);
-    _tiles[tileCoordsToKey] = Tile(
+    var tile = _tiles[tileCoordsToKey] = Tile(
       coords: coords,
       coordsKey: tileCoordsToKey,
       tilePos: _getTilePos(coords),
@@ -946,6 +946,8 @@ class _TileLayerState extends State<TileLayer> with TickerProviderStateMixin {
           options.tileProvider.getImage(_wrapCoords(coords), options),
       tileReady: _tileReady,
     );
+
+    tile.loadTileImage();
   }
 
   void _tileReady(Coords<double> coords, dynamic error, Tile tile) {
@@ -1078,9 +1080,7 @@ class Tile implements Comparable<Tile> {
     this.active = false,
     this.retain = false,
     this.loadError = false,
-  }) {
-    loadTileImage();
-  }
+  });
 
   void loadTileImage() {
     try {
@@ -1132,14 +1132,14 @@ class Tile implements Comparable<Tile> {
   void _tileOnLoad(ImageInfo imageInfo, bool synchronousCall) {
     if (null != tileReady) {
       this.imageInfo = imageInfo;
-      Timer.run(() => tileReady(coords, null, this));
+      tileReady(coords, null, this);
     }
   }
 
   void _tileOnError(dynamic exception, StackTrace stackTrace) {
     if (null != tileReady) {
-      Timer.run(() => tileReady(
-          coords, exception ?? 'Unknown exception during loadTileImage', this));
+      tileReady(
+          coords, exception ?? 'Unknown exception during loadTileImage', this);
     }
   }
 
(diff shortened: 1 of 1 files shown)


maRci002:
How to test, just change `flutter_map` dependecy to:
```
  flutter_map:
    git:
      url: https://github.com/fleaflet/flutter_map.git
      ref: 7fb7fdd6210837921dd7a4ca4aad9a7e436415ea
```

lib/src/layer/tile_layer.dart
```diff
@@ -946,6 +946,8 @@ class _TileLayerState extends State<TileLayer> with TickerProviderStateMixin {
           options.tileProvider.getImage(_wrapCoords(coords), options),
       tileReady: _tileReady,
     );
+
+    tile.loadTileImage();
```
johnpryan, line 950: (nitpick) This code could be use the cascade operator to simplify:

```
_tiles[tileCoordsToKey] = Tile(
  // ...
)..loadTileImage();
```

maRci002: I tried it but didn't work. I'll give it another try maybe hotreload bugged.

maRci002: I retried it didn't work. I think this happens under hood when cascade operator is used.
1. Tile object created
2. `loadTileImage` method called on Tile
3. Tile added to `_tiles[tileCoordsToKey]`

With temp variable this happens:
1. Tile object created
2. Tile added to `_tiles[tileCoordsToKey]`
3. `loadTileImage` method called on Tile

So even if `loadTileImage`'s callback is synchronous we are sure it is already in `_tiles` map.

PR #717, opened 2020-08-04
bugfix: Prevent parent ListView widgets from snatching scrolling gestures

### LINKED ISSUE #253: ListView snatches scroll gestures
I put my map in a ListView.  When dragging left and right on the map, the maps scrolls around, when going up and down (the direction of the ListView), the ListView gets those gestures rather than the map.  How do I get the map to use those gestures instead?

ritikmishra:
Closes #253.

The issue was fixed by using a `GestureArenaTeam` in conjunction with extra gesture recognizers to ensure that the `ScalingGestureRecognizer` wins when we would expect it to do so.

### DIFF
--- a/lib/src/map/flutter_map_state.dart
+++ b/lib/src/map/flutter_map_state.dart
@@ -1,6 +1,7 @@
 import 'dart:async';
 
 import 'package:async/async.dart';
+import 'package:flutter/gestures.dart';
 import 'package:flutter/widgets.dart';
 import 'package:flutter_map/flutter_map.dart';
 import 'package:flutter_map/src/gestures/gestures.dart';
@@ -75,6 +76,42 @@ class FlutterMapState extends MapGestureMixin {
       mapState.setOriginalSize(constraints.maxWidth, constraints.maxHeight);
       var size = mapState.size;
 
+      var scaleGestureTeam = GestureArenaTeam();
+
+      var scaleGestureDetector = ({Widget child}) => RawGestureDetector(
+            gestures: <Type, GestureRecognizerFactory>{
+              ScaleGestureRecognizer:
+                  GestureRecognizerFactoryWithHandlers<ScaleGestureRecognizer>(
+                      () => ScaleGestureRecognizer(),
+                      (ScaleGestureRecognizer instance) {
+                scaleGestureTeam.captain = instance;
+                instance.team ??= scaleGestureTeam;
+                instance
+                  ..onStart = handleScaleStart
+                  ..onUpdate = handleScaleUpdate
+                  ..onEnd = handleScaleEnd;
+              }),
+              VerticalDragGestureRecognizer:
+                  GestureRecognizerFactoryWithHandlers<
+                          VerticalDragGestureRecognizer>(
+                      () => VerticalDragGestureRecognizer(),
+                      (VerticalDragGestureRecognizer instance) {
+                instance.team ??= scaleGestureTeam;
+                // these empty lambdas are necessary to activate this gesture recognizer
+                instance.onUpdate = (_) {};
+              }),
+              HorizontalDragGestureRecognizer:
+                  GestureRecognizerFactoryWithHandlers<
+                          HorizontalDragGestureRecognizer>(
+                      () => HorizontalDragGestureRecognizer(),
+                      (HorizontalDragGestureRecognizer instance) {
+                instance.team ??= scaleGestureTeam;
+                instance.onUpdate = (_) {};
+              })
+            },
+            child: child,
+          );
+
       return MapStateInheritedWidget(
         mapState: mapState,
         child: Listener(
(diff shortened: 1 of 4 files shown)


pqtrong17:
How to add it into my project? Do you have a example code?

ritikmishra:
If you want to use the fork, put the following in your pubspec.yaml

```
  flutter_map:
    git:
      url: https://github.com/ritikmishra/flutter_map.git
      ref: fix-listview-scrolling
```

pqtrong17:
> If you want to use the fork, put the following in your pubspec.yaml
> 
> ```
>   flutter_map:
>     git:
>       url: https://github.com/ritikmishra/flutter_map.git
>       ref: fix-listview-scrolling
> ```

Okie. Thank you so much. Where did the file's location do this? I want to understand more about it

pqtrong17:
It's worked for me. Can you explain how to it works?

jakobkuehne:
Sadly I can't use your git version because of flutter_map_marker_cluster.

`Because flutter_map_marker_cluster 0.2.9 depends on flutter_map ^0.10.1+1 and no versions of flutter_map_marker_cluster match >0.2.9 <0.3.0, flutter_map_marker_cluster ^0.2.9 requires flutter_map from hosted.`

Or am I doing something wrong? So far, I've never added a package directly from Git. But apart from the possibility to add your great customization via Git, this would be a very important feature and I can't understand why this is still not officially available in flutter_map...

> If you want to use the fork, put the following in your pubspec.yaml
> 
> ```
>   flutter_map:
>     git:
>       url: https://github.com/ritikmishra/flutter_map.git
>       ref: fix-listview-scrolling
> ```

romainpurchla (approved):
Work for me

sm2017:
> It's worked for me. Can you explain how to it works?

@ritikmishra I'm interested to know that how it works

@johnpryan Please accept the PR

ritikmishra:
The mechanism through which Flutter recognizes gestures is called the [gesture arena](https://flutter.dev/docs/development/ui/advanced/gestures#gesture-disambiguation). Different gesture recognizers compete in the arena, and the **first** recognizer to accept the gesture, or the last recognizer remaining in the arena (i.e the other recognizers rejected the gesture), is responsible for handling it. 

Previously, the map was implicitly using 3 gesture recognizers through `GestureDetector`: `ScaleGestureRecognizer` (handling panning and zoom), `TapGestureRecognizer`, and `LongPressGestureRecognizer`. If the map is inside a widget that scrolls (i.e a `ListView`), a fourth gesture recognizer will be in the arena: either a `HorizontalDragGestureRecognizer` or a `VerticalDragGestureRecognizer` depending on the direction that it scrolls.

The problem was caused by the `ScaleGestureRecognizer` taking too long to accept the gesture, resulting in the `HorizontalDragGestureRecognizer`/`VerticalDragGestureRecognizer` winning in the gesture arena.

The solution was to use a [`GestureArenaTeam`](https://api.flutter.dev/flutter/gestures/GestureArenaTeam-class.html) to utilize a new `HorizontalDragGestureRecognizer` and a new `VerticalDragGestureRecognizer` to help the original `ScaleGestureRecognizer` win faster. A `GestureArenaTeam` has a captain and members -- whenever any gesture recognizer in the team accepts the gesture, the captain is put in charge of handling it. This way, a user can pan the map when it is inside a `ListView`/other scrollable widget.

sm2017:
@ritikmishra as the PR is not merged, is there alternative solution , for example wrapping map with another widget?

ritikmishra:
I couldn't find a workaround, which is why I made this PR. If you can find some way to stop the ListView from being so greedy in the gesture arena & ensure that the pointer events from gestures that don't go to the ListView _do_ go to the FlutterMap, that should fix the problem.

gsimko:
@johnpryan can you please take a look?

kengu:
@ritikmishra Can you resolve the conflict and update this PR? I can review it when this is done.

ritikmishra:
@kengu i have resolved the conflict

lib/src/map/flutter_map_state.dart
```diff
@@ -87,50 +124,50 @@ class FlutterMapState extends MapGestureMixin {
...
-                            if (widget.children != null &&
-                                widget.children.isNotEmpty)
-                              ...widget.children,
-                            if (widget.layers != null &&
-                                widget.layers.isNotEmpty)
-                              ...widget.layers.map(
-                                (layer) => _createLayer(layer, options.plugins),
-                              )
-                          ],
+              child: scaleGestureDetector(
```
kengu, line 131: If I understand this correctly @ritikmishra, the TeamGestureRecognizer will always ensure that flutter_map takes precedence over ListView scoll? I don't think we can conclude that this is the only behaviour user expect. I think we need to make this behavior configurable, just as it is for [scollables](https://api.flutter.dev/flutter/widgets/Scrollable-class.html) in general. I think we should add something similar (or equal) to [scrollPhysics](https://api.flutter.dev/flutter/widgets/ScrollPhysics-class.html) to MapOptions constructor.

@johnpryan? What do you think?

kengu (changes requested):

BenBitDesign:
Hi guys

I think @kengu makes a good point to make this configurable.

Is anyone looking into this?

kengu:
Closing, fixed by #994.

PR #726, opened 2020-08-11
On Android: map loading in release build works now.

wolfc01:
I wanted to 'feel' the performance of the flutter_map on android, by release building the example. 
Then I discovered no maps were loaded due to the release app not having INTERNET permission. I added the permission in the AndroidManifest.xml of the release configuration.

Br
Carl.


johnpryan:
Thanks!

PR #740, opened 2020-09-05
Fix #629 | Reload map if options changed for overrideTilesWhenUrlChanges

### LINKED ISSUE #629: How to switch map layers
Hi,

I have two map layers: OSM and MapBox Satellite:

```
 layers: [
   TileLayerOptions(
        subdomains: ['a', 'b', 'c'],
        tileProvider: CachedNetworkTileProvider()
    ),
    TileLayerOptions(
        urlTemplate: 'https://api.mapbox.com/v4/mapbox.satellite/{z}/{x}/{y}@2x.jpg90?access_token={accessToken}',
        additionalOptions: {
          'accessToken': '...'
        },
        tileProvider: CachedNetworkTileProvider()
    )
  ]
```

Is there a way to add a layer switch to switch between both layers?

As workaround I tried to add a button with `setState()` in ` onPressed()`. However, althought the `build()` is called, the underlying map is not changing:
```

_showSatellite = false;
...
layers: [
  _showSatellite ? TileLayerOptions(/*mapbox*/) : TileLayerOptions(/*osm*/)
]

...

Button(
  onPressed: () {
    setState() {
      _showSatellite = !_showSatellite;
    }
  }
)
```

Any help is appreciated :) Thanks in advance!

Edit: It seems that the layer changes. When I pan or zoom, the switch layer is shown. So, I need a way to refresh the currently shown tiles anyway...

Xennis:
The option `overrideTilesWhenUrlChanges` currently only works if the URL
template changes but not if the options change.

For example:

```dart
TileLayerOptions(
  urlTemplate: 'https://api.mapbox.com/styles/v1/mapbox/{id}/tiles/{z}/{x}/{y}',
  additionalOptions: <String, String>{
    'id': mapboxStyle,
  },
```

If the value of `id` (here the Mapbox layer style, e.g. streets, satellite, etc.)
changes the option `overrideTilesWhenUrlChanges` should override the tiles as well.

### DIFF
--- a/lib/src/layer/tile_layer.dart
+++ b/lib/src/layer/tile_layer.dart
@@ -1,6 +1,7 @@
 import 'dart:async';
 import 'dart:math' as math;
 
+import 'package:collection/collection.dart' show MapEquality;
 import 'package:flutter/material.dart';
 import 'package:flutter/widgets.dart';
 import 'package:flutter_map/src/core/bounds.dart';
@@ -199,7 +200,7 @@ class TileLayerOptions extends LayerOptions {
     this.maxNativeZoom,
     this.zoomReverse = false,
     double zoomOffset = 0.0,
-    this.additionalOptions = const <String, String>{},
+    Map<String, String> additionalOptions,
     this.subdomains = const <String>[],
     this.keepBuffer = 2,
     this.backgroundColor = const Color(0xFFE0E0E0),
@@ -247,6 +248,11 @@ class TileLayerOptions extends LayerOptions {
         tileSize = wmsOptions == null && retinaMode && maxZoom > 0.0
             ? (tileSize / 2.0).floorToDouble()
             : tileSize,
+        // copy additionalOptions Map if not null, so we can safely compare old
+        // and new Map inside didUpdateWidget with MapEquality.
+        additionalOptions = additionalOptions == null
+            ? const <String, String>{}
+            : Map.from(additionalOptions),
         super(key: key, rebuild: rebuild);
 }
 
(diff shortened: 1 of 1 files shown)


maRci002:
This isn't perfect:

```dart
      final oldOptions = oldWidget.options.additionalOptions;
      final newOptions = options.additionalOptions;

      if (oldUrl != newUrl || oldOptions != newOptions) {
```

When you decalre every time a new `HashMap` they won't be equal, see
```dart
  additionalOptions: <String, String>{
    'id': mapboxStyle,
  },
```

Just see this: `print({'x': 1} != {'x': 1} );` you would think this will print `false` because they are equals, however it will print `true`, so in this case `oldOptions != newOptions` will always evaluate to `true` which will reload tiles unnecessarily.

On the other hand if you cache `additionalOptions` in a `Stateful widget` and later you put / change a parameter in it then I would assume `TileLayer` would detects changes and map will reload tiles, however this won't happen, see:
```dart
class _HomePageState extends State<HomePage> {
  final params = <String, String>{
    'id': 'myStyle',
  };

  void onEvent() {
    params['id'] = 'myStyle2';
  }

...
                  TileLayerOptions(
                    urlTemplate: 'https://api.mapbox.com/styles/v1/mapbox/{id}/tiles/{z}/{x}/{y}',
                    additionalOptions: params,
                  )
...
```
If an event calls `onEvent()` method, then it will change param, however TileLayer has the same instance so replace will affect that HashMap too, so they will be always equals, see:
```dart
  var a = {'x': 1};
  a['y'] = 2;
  
  print(a != a);
```
It will print `false`, however I wanted to reload tiles because one param changed.

Xennis:
Dear @maRci002 thanks for your detailed feedback. What do you recommend to fix this issue? Would using [MapEquality](https://api.flutter.dev/flutter/package-collection_collection/MapEquality-class.html) help here?

maRci002:
Yes you can use it, however there are still problems if programmer using explicit `null` value for `additionalOptions`.
```dart
  print((const MapEquality()).equals({'x': 1}, {'x': 1})); // true
  print((const MapEquality()).equals(null, {'x': 1})); // false !
```

So in `TileLayerOptions` change this line `this.additionalOptions = const <String, String>{},` to `Map<String, String> additionalOptions,` and add this line to initializer list: `additionalOptions = additionalOptions ?? const <String, String>{},` so now you make sure there isn't any `null` values and `MapEquality` will be fine. Note, this won't handle if you cache HashMap and change value inside it.

You have to handle updates on Map, so this case:
In initializer list you should use this line:
```dart
additionalOptions = additionalOptions == null
    ? const <String, String>{}
    : Map.from(additionalOptions), // copy Map, so we can safely compare old and new Map inside didUpdateWidget with MapEquality
```

Xennis:
Hey @maRci002 great thanks for your help. I adopted the MR accordingly.

maRci002 (approved):
LGTM

johnpryan (approved):

Issue #727, opened 2020-08-14
Replace "latlong: ^0.6.1" package as it is deprecated and not maintained.


Pinkal7600:
We should replace the `latlong: ^0.6.1` package as it is not maintained and have some limitation for latitude and longitude value. Like we can only provide latitude value between -90 to 90 and longitude value between -180 to 180. So this makes the limitation for flutter_map compare to leaflet as leaflet does not have this kind of restrictions. 

I personally face this issue(https://github.com/fleaflet/flutter_map/issues/725) in my one use case so I replace this package with `maps_toolkit: ^1.1.0+2` and make some modification in that as well to use the `math_util.dart` for flutter_map like `toRadians()`, `toDegrees()`, etc...

pdenert:
Could you make PR with your working solution? I think that author of this package will be glad of it, as this is a side project for him and he call for maintainers #680

johnpryan:
Duplicate of https://github.com/fleaflet/flutter_map/issues/485

PR #750, opened 2020-09-21
Feature #727 | Replace latlong with maps_toolkit

yakanet:
As mentioned in #727 , replace latlong dependency with maps_toolkit.

The change is minimal, but it might create some **breaking changes** with other libs based on flutter_map, as the package name for the LatLong class is not the same, and the class LatLong is now immutable.

In most of the cases just replace the package name with the following:
```diff
- import 'package:latlong/latlong.dart';
+ import 'package:maps_toolkit/maps_toolkit.dart';
```


pdenert (approved):

pdenert (approved):

kengu:
@yakanet Could you fix the conflicts before we look into your proposed switch to maps_toolkit?

maRci002:
This package seems good, just resolve conflicts. I kinda like this package LatLng's `hashCode` override version.

danilocubos:
#854 Fixes this problem.

1075dn:
I would prefer switching to `maps_toolkit` as well.

ibrierley:
Not sure if this is still necessary now, would like to hear if people think it is and why, as there were conflicts and maps_toolkit can be used standalone if I'm understanding it correctly ?

JaffaKetchup:
Would agree. Not sure if this is necessary now the new version of 'latlong2' is being used?

JaffaKetchup:
I'm going to close this as 'latlong2' is now being used, which already has some of the same extra features as 'maps_toolkit'. Besides, if those extra functions are needed, you can just require that library separately?

Thanks for your contribution!

JaffaKetchup:
Ok, for now, I'm going to close this PR. I don't think it's necessary anymore, and might cause more problems than create benefits.
Thank you for your contribution, and feel free to submit more in the future should you feel the need to. Thanks!

PR #769, opened 2020-11-04
Add pad function to LatLngBounds

legarland:
This PR is to add a pad function to the LatLngBounds class. The logic is based on Leaflet's pad function. This is useful when showing markers on the map with a bit of padding around the edges. Below are Leaflet's comments on the function.

// Returns bounds created by extending or retracting the current bounds by a given ratio in each direction.
// For example, a ratio of 0.5 extends the bounds by 50% in each direction.
// Negative values will retract the bounds.

### DIFF
--- a/lib/src/geo/latlng_bounds.dart
+++ b/lib/src/geo/latlng_bounds.dart
@@ -111,4 +111,12 @@ class LatLngBounds {
     }
     return true;
   }
+
+  void pad(double bufferRatio) {
+    var heightBuffer = (_sw.latitude - _ne.latitude).abs() * bufferRatio;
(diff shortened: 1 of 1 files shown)


kengu (approved):
LGTM

kengu:
@legarland The build fails because dartfmt is not run (the test does this to check if changes are formatted correctly, which is seems like it's not). Can you format change med dartfmt and update the PR?

kengu (changes requested):
@legarland The build fails because dartfmt is not run (the test does this to check if changes are formatted correctly, which is seems like it's not). Can you format the changes with dartfmt and update the PR?

kengu:
#844 will fix formatting error

kengu (approved):

PR #781, opened 2020-11-29
Add a fastReplace option to TileLayerOptions.

pento:
Based on [this comment](https://github.com/fleaflet/flutter_map/pull/572#issuecomment-610269536), this PR adds a `fastReplace` option to `TileLayerOptions`.

The purpose of this option is to force tiles to be removed as soon as their replacement tile is ready, skipping `tileFadeIn*` options, as well as the hardcoded delay in `_tileReady()`.

## Screenshots

These screenshots are based on [this config](https://github.com/pento/TidyWeather/blob/try/map-improvements/lib/cards/radar.dart#L71-L113): the top layer (the place names) uses transparent tiles.

### Before

Note how different zoom levels of the top layer are visible at the same time while zooming.

### After

It no longer shows multiple zoom levels at the same time, though there is sometimes a small flicker where no tile is visible at all, even after the tiles have been cached. I'm happy to make adjustments if there's a way to avoid this, but otherwise, I think this is an improvement.

### DIFF
--- a/lib/src/layer/tile_layer.dart
+++ b/lib/src/layer/tile_layer.dart
@@ -232,6 +232,11 @@ class TileLayerOptions extends LayerOptions {
   /// being temporarily visible), the old layer is removed as quickly as
   /// possible when this is set to `true` (default `false`).
   ///
+  /// This option is likely to cause some flickering of the transparent layer,
+  /// most noticeable when using pinch-to-zoom. It's best used with maps that
+  /// have `interactive` set to `false`, and zoom using buttons that call
+  /// `MapController.move()`.
+  ///
   /// When set to `true`, the `tileFadeIn*` options will be ignored.
   final bool fastReplace;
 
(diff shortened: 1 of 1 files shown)


ibrierley:
Haven't had chance to test yet, but my main test would be how it acts on pinch/zoom. The old flutter_map flickered a lot due to the problem, so I think it may be a step back if that is impacted, but there again, if it's just an option and clearly marked that it may cause flicker, I guess it feels ok.

I did some testing with this stuff a while back and it felt like you had to have at least one of..
Fading old image in.
Displaying old image and having 2 overlayed for a very short period.
A possible flash during transition.

pento:
Thanks for the feedback, @ibrierley! You're right, there is noticeable flickering when using pinch-to-zoom, particuarly when the layer is not yet cached: the GIF below shows this when zooming in, then the flickering does become less noticeable when zooming back out again.

I've updated the option documentation to note this caveat.

pento:
I've rebased this PR, I'm still using this change in my own fork, but I'd love if it could be merged here, instead. 🙂

pento:
Okay stalebot.

kengu (approved):
LGTM

kengu:
@pento Thanks for the contribution!

PR #683, opened 2020-07-09
Regenerated flutter_map and example

vinicentus:
- regenerated flutter_map
- regenerated example, using flutter create --org me.jpryan flutter_map_example
- generated web files in example after switching to beta channel

Noteworthy consequences:
- updated example package name to be consistent among files
- bumped required sdk
- removed minimum required flutter version
- updated gitignore files
- removed androidx module flag, this isn't a module
- removed other unnessecary or incorrect code
- formatted some files and added comments to make things clearer

Tested on chrome and android

PR #793, opened 2020-12-24
Upgraded flutter + package versions

jopmiddelkamp:
In this PR I've upgraded the packages used in the `flutter_map` plugin and the updated the example project to match the latest version of `flutter create`.


kengu:
@jopmiddelkamp PR #683 was merged yesterday, which I believe does the same thing. The only significant difference is that you have upgraded the dependencies also. Would you like to update this PR to align it with #683?

JaffaKetchup:
Hi @jopmiddelkamp,
What needs to be done here to get this ready for merging? I'm referring to https://github.com/fleaflet/flutter_map/pull/793#issuecomment-802761776 and maybe that the dependencies can be even newer now?

jopmiddelkamp:
> @jopmiddelkamp PR #683 was merged yesterday, which I believe does the same thing. The only significant difference is that you have upgraded the dependencies also. Would you like to update this PR to align it with #683?

Yes seems like it did the same thing.

JaffaKetchup:
Ok, thanks for your contribution! Will close now.

Issue #795, opened 2021-01-04
Download Tiles & Store Them For Offline Use


JaffaKetchup:
Hello,
I have been trying to use `flutter_map` to create an app where a specific 'Downloads Only' mode can be enabled. The aim is to have tiles loaded whilst not in this mode downloaded to the users device, like caching (except I do not want these tiles to be deleted by the system). Then, once the user is offline or in 'Downloads Only' mode, I'd like to use the images stored from before. 

I also have setup multiple map tile servers and have a global variable holding the app's document directory '`globals.saveDir`'. I have started doing this using the `network_to_file_image` package like so:
```
class AdvancedFileCacheProvider extends TileProvider {
  @override
  ImageProvider getImage(Coords<num> coords, TileLayerOptions options) {
    return NetworkToFileImage(
      url: getTileUrl(coords, options),
      file: File(
        p.join(
            globals.saveDir.path,
            getTileUrl(coords, options)
                .replaceAll('https://', '')
                .replaceAll('http://', '')),
      ),
      debug: true,
    );
  }
}
...
TileLayerOptions(
          urlTemplate: Uri.decodeQueryComponent(urlForMap),
          subdomains: ['a', 'b', 'c'],
          tileProvider: AdvancedFileCacheProvider(),
),
...
```

This setup prints:
```
I/flutter (14390): Fetching image from: http://mt.google.com/vt/lyrs=m&x=16324&y=10894&z=15
I/flutter (14390): Fetching image from: http://mt.google.com/vt/lyrs=m&x=16325&y=10894&z=15
I/flutter (14390): Saving image to file: /data/user/0/[appname]/app_flutter/mt.google.com/vt/lyrs=m&x=16324&y=10894&z=15
I/flutter (14390): Saving image to file: /data/user/0/[appname]/app_flutter/mt.google.com/vt/lyrs=m&x=16325&y=10894&z=15
etc...
```
, which means the downloading of images are working as intended.

However, once I go offline (there is no special mode yet, but I think the `network_to_file_image` package is supposed to work like this...), the images still try to be fetched from the internet, resulting in a `SocketException` error.

Has anyone else implemented a similar logic? I understand that this package will cache these tiles, but anything in the cache is vulnerable to the system clearing it, and in my circumstance, this would be quite a large problem.

Thank you very much!
Luka S.

JaffaKetchup:
Any help would be much appreciated!
Luka S

ibrierley:
[...]

PR #798, opened 2021-01-06
Better Offline Mapping

JaffaKetchup:
Hello,
As a follow up to issue #795, I have created this pull request. Please note the differentiation between existing caching solutions and this new solution: caches can be cleared without warning by the system, app storage can't.

I have added a new tile provider to be used as below:

> Use `PersistentAdvancedCacheTileProvider()` as the `TileLayerOptions > tileProvider` option, and set `TileLayerOptions > saveDir` to a variable containing (await) `getApplicationDocumentsDirectory()` from library [`path_provider`](https://pub.dev/packages/path_provider).
> 
> As the user scrolls around the map, tiles will be saved to `[saveDir]/tiles`. Then, when requested again, tiles will be taken from that directory if available, else a network tile request will be made. Thus, this is a file first solution.
> 
> This configuration should be used where offline functionality is required because cached tiles from other tile providers can be cleaned by the system without user notification at any point. With this setup, only the user (through App Settings > Storage > Clear Storage) or the app itself can clear the tiles. It is up to the app developer (you) to manage these tiles...
> 
> If your user wants to get the latest tiles from the server for tiles already loaded, you must clear the tiles folder. If the user wishes to remove tiles altogether, you must provide that functionality. This library will not handle that for you.
> 
> A recommended (however untested) integration is to use another tile provider (such as the default) whilst not in 'Offline Mode' so that the newest tiles are always loaded, then switch to this provider once the user enters 'Download & Offline Mode'. Then, when the user switches back out, clear the tiles folder.

This pull request adds one dependency. This PR has only been tested on an Android Pixel 3a emulator.

Thanks!
Luka S

### DIFF
--- a/README.md
+++ b/README.md
@@ -238,6 +238,17 @@ Then execute `ulimit -S -n 2048` ([ref](https://github.com/trentpiercy/trace/iss
 Then execute `flutter run` with a running emulator.
 
 ## Offline maps
+Full offline map functionality has been created by JaffaKetchup. Use `PersistentAdvancedCacheTileProvider()` as the `TileLayerOptions > tileProvider` option, and set `TileLayerOptions > saveDir` to a variable containing (await) `getApplicationDocumentsDirectory()` from library [`path_provider`](https://pub.dev/packages/path_provider).
+
+As the user scrolls around the map, tiles will be saved to `[saveDir]/tiles`. Then, when requested again, tiles will be taken from that directory if available, else a network tile request will be made. Thus, this is a file first solution.
+
+This configuration should be used where offline functionality is required because cached tiles from other tile providers can be cleaned by the system without user notification at any point. With this setup, only the user (through App Settings > Storage > Clear Storage) or the app itself can clear the tiles. It is up to the app developer (you) to manage these tiles...
+
+If your user wants to get the latest tiles from the server for tiles already loaded, you must clear the tiles folder. If the user wishes to remove tiles altogether, you must provide that functionality. This library will not handle that for you.
+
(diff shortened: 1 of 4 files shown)


JaffaKetchup:
I'm not sure why this build fails, it works fine with my testing.

Luka S

johnpryan:
Any custom tile providers aren't likely to be added at this point, since you can build your own package that provides them. This keeps the core package small. Feel free to suggest any packages you would like mentioned in the README.

stevenv-synap:
I see that some people have struggled with caching tiles/improving tile load timings on the web platform. My solution was to introduce a custom TileProvider (Allows configuration of the HTTP requests, if required), and then usage of the extended_image package allowed for caching of images at the browser level (same sort of behaviour you would get with Cache-Control response headers on any website).

Image Package:
https://pub.dev/packages/extended_image
The ExtendedNetworkImageProvider class provides the caching capabilities.

```
class DioTileProvider extends TileProvider {
  final Dio dio;

  DioTileProvider({
    Map<String, String> headers = const {},
    Dio? httpClient,
  })  : dio = httpClient ?? Dio(),
        super(
          headers: {
            ...headers,
            'User-Agent': headers['User-Agent'] == null
                ? 'flutter_map_tile_caching for flutter_map (unknown)'
                : 'flutter_map_tile_caching for ${headers['User-Agent']}',
          },
        );

  @override
  ImageProvider<Object> getImage(
      TileCoordinates coordinates, TileLayer options) {
    var tileUrl = getTileUrl(coordinates, options);
    return ExtendedNetworkImageProvider(tileUrl,
        cache: true, cacheMaxAge: const Duration(days: 15));
  }

  @override
  void dispose() {
    dio.close();
    super.dispose();
  }
}
```


Hope this helps. Cheers

JaffaKetchup:
Hey @stevenv-synap,

Session caching will come back in v6, so hopefully this shouldn't be necessary. There are also some other improvements to load time!

PR #805, opened 2021-01-25
Added support for custom templates in URL

govorilka:
I added support for sprintf templates inside urls. For example, you can use templates like this:
`http://example.com/{x:%08i}/{y:%08i}/{z:%02i} `
When replacing the number 1,2,3, the sample will be parsed as:
`http://example.com/00000001/00000002/03`
I need functionality to work with our corporate tile server, but I think it might be useful for someone else as well.

### DIFF
--- a/lib/src/core/util.dart
+++ b/lib/src/core/util.dart
@@ -7,13 +7,13 @@ var _templateRe = RegExp(r'\{ *([\w_-]+) *\}');
 /// Replaces the templating placeholders with the provided data map.
 ///
 /// Throws an [Exception] if any placeholder remains unresolved.
-String template(String str, Map<String, String> data) {
+String template(String str, Map<String, dynamic> data) {
   return str.replaceAllMapped(_templateRe, (Match match) {
     var value = data[match.group(1)];
     if (value == null) {
       throw Exception('No value provided for variable ${match.group(1)}');
     } else {
-      return value;
+      return value.toString();
     }
   });
 }
--- a/lib/src/layer/tile_layer.dart
+++ b/lib/src/layer/tile_layer.dart
@@ -16,7 +16,7 @@ import 'package:tuple/tuple.dart';
 import 'layer.dart';
 
 typedef TemplateFunction = String Function(
-    String str, Map<String, String> data);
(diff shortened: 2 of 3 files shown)


johnpryan:
Rather than change the template function, I think it would be better to allow users to customize it: 74a4d2b72b79dd83dcb8748079f7bc5a7477dff6 - does that solve your issue?

govorilka:
In a template function, you can replace the "data" parameter with map <String, dynamic>.  Then we can write our own number-to-string conversion.30 янв. 2021 г. 0:12 пользователь John Ryan <notifications@github.com> написал:
Rather than change the template function, I think it would be better to allow users to customize it: 74a4d2b - does that solve your issue?

johnpryan:
Why not just convert the number to a string?

govorilka:
Our tile server only works if the number in the URL has a fixed length with leading zeros. For example, "1" should look like "00000001". If we use toString(), we don't need templateFunction, but we need to rework the server for that.
Adding templateFunction to TileLayerOptions is a good idea, but we can make it more flexible.

johnpryan:
Can you help me understand why the current Map needs to be changed into a `Map<String, dynamic>`?

govorilka:
The current implementation uses toString(). If x == 10, y == 20, z == 3, then x.toString() == "10", y.toString() == "20", z.toString() == "3". We need outConvert(x) == "00000010", ourConvert(y) == "00000020", ourConvert(z) == "03". We using numbers convertation with leading zeroes filling from package https://pub.dev/packages/sprintf. If we use Map<String, String> we lose information about the type of parameters "x", "y", "z" and have to do an extra conversion from string to number inside implementation of templateFunction(). Using Map<String, dynamic> also allows us to pass parameters through additionalOptions without converting each value to String and doing some calculations with them inside implementation of templateFunction().

We also fixed unit tests in #810 because build on Travis doesn't works now.

JaffaKetchup:
Hi there,
Bit out of sync here. What needs to be done?
Thanks.

JaffaKetchup:
Closing due to inactivity. Feel free to ping me to reopen!

PR #808, opened 2021-01-30
Listing flutter_map_floating_marker_titles under Plugins

androidseb:
Hello,

Thanks for the great project! I have built this floating marker titles library for flutter, and flutter_map's code helped me a lot with that.

I'm hoping to give back by bringing value to your library by adding this feature, and I thought the best way to introduce this was with this PR, please let me know if there is a better way.

I hope this helps :-)


kengu (changes requested):
LGTM, and nice work! You need to update this PR with master. That will resolve tests that are failing.

kengu (approved):

PR #813, opened 2021-02-03
Add deprecated flags to constructor parameters

meszarosdezso:
Added two missing `@deprecated` flags to the `MapOptions` constructor (debug and interactive props) to warn the developers in their source code about the deprecation.


kengu (changes requested):
LGTM. You need to update this PR with master. That will resolve the tests failing.

kengu (approved):

Issue #836, opened 2021-03-08
Loading tiles outside min/maxZoom when changing BaseMap


Anders55:
STR:
1. Set up a BaseMap with maxZoom = 20
2. Set up a BaseMap with maxZoom = 19 (OpenStreetMap)
3. Load the BaseMap from step 1
4. Zoom to zoom=20. 
5. Change BaseMap to the map in step 2

Problem: flutter_map will now in it's didUpdateWidget method of the _TileLayerState class  iterate through the old tiles (from the Basemap in step 1) and try to replace them with new corresponding tiles (from the BaseMap in step 2). This causes an error as the new BaseMap does not have tiles for the same coords and zoom-level (zoom=20) as the BaseMap in step 1.

Anders55:
PR here: https://github.com/fleaflet/flutter_map/pull/837

kengu:
Closed by #837.

PR #810, opened 2021-02-01
Fixed unit test

dynaevdv:
Fixed unit tests via MockHttpServer. 
Now test don`t required internet connection. Test folder now contains map.png(30 kb) which provided for test via Mock server. 

All tests passed!

PR #837, opened 2021-03-08
Fix for Zoom-value outside min/maxZoom when changing BaseMap

Anders55:
The fix will iterate through the current tiles and check if any of them are outside the minZoom and maxZoom-values of the new BaseMap-layer. If so, it will stop flutter_map from trying to replace old tiles with the corresponding tiles from the new BaseMap.

### DIFF
--- a/lib/src/layer/tile_layer.dart
+++ b/lib/src/layer/tile_layer.dart
@@ -460,19 +460,8 @@ class _TileLayerState extends State<TileLayer> with TickerProviderStateMixin {
       reloadTiles = true;
     }
 
-    var maxTileZoomValue = 1.0;
-    var minTileZoomValue = 20.0;
-    for (var tile in _tiles.values) {
-      if (tile.level.zoom > maxTileZoomValue) {
-        maxTileZoomValue = tile.level.zoom;
-      }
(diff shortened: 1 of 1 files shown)


Anders55:
Issue description with STR here: https://github.com/fleaflet/flutter_map/issues/836

kengu:
@johnpryan / @maRci002 - builds are consistently breaking and PR #810 fixes this.

lib/src/layer/tile_layer.dart
```diff
@@ -429,6 +429,20 @@ class _TileLayerState extends State<TileLayer> with TickerProviderStateMixin {
       reloadTiles = true;
     }
 
+    var maxTileZoomValue = 1.0;
```
kengu, line 432: if `reloadTiles == true` already, this section is unnecessary. Should be moved to `if(!reloadTiles) {...` on line 451

kengu (changes requested):

kengu:
@Anders55 PR #810 fixes the unit-test that fails. You should pull master and merge/rebase it into this PR. This will resolve the travis-build.

kengu (approved):

Issue #631, opened 2020-05-26
Show tiles from both offline and cached provider


Swepilot:
First, thanks to @johnpryan for a great package.

I have created png tiles zoom level 1-6 for the entire planet and using them as assets with `AssetTileProvider`. I will consider packaging the png files and then unpacking into the file system to use `FileTileProvider` because it seems I have to enter every subdirectory separately in pubspec.yaml and that seems to slow down hot reloads (if anyone has a better suggestions, please...)

Anyway, the zoom 1-6 is good enough for the offline use, but when the user is online I of course want to be able to zoom in more using a normal CachedNetworkImageProvider so I choose provider depending on whether the user is online or offline.

Now, what I really want is to enable the offline user to zoom in more and show the tiles from the CachedNetworkImageProvider if those tiles are in the cache (if not in the cache perhaps show a default png saying something like "NOT AVAILABLE"). Is there any way of doing that already or would that require a new Provider?

PR #841, opened 2021-03-15
Add Caching Implementations to README

### LINKED ISSUE #402: Increase Caching Period
Is there a way i can increase caching period of map tiles.

### LINKED ISSUE #576: Tiles will not load after usage without a network

### LINKED ISSUE #614: flutter: CacheManager: Failed to download file
TileLayerOptions(
                    urlTemplate: 'http://mt{s}.google.com/vt/lyrs=m&x={x}&y={y}&z={z}',
                    subdomains: ['0','1','2','3'],
                    tileProvider: CachedNetworkTileProvider(),
                  ),

flutter: CacheManager: Failed to download file from http://mt3.google.com/vt/lyrs=m&x=32752&y=21795&z=16 with error:
SocketException: OS Error: Operation timed out, errno = 60, address = mt3.google.com, port = 52998

### LINKED ISSUE #649: Periodically clear the images from cache
I understand the purpose of using the `CachedNetworkTileProvider` is to have the tiles, well, cached. But what if I want to get fresh tiles from the server? After how long does the cache expire (if at all) and is it possible to expel the tiles manually?

My case is that I have a layer of weather data (not WMS) that I want to update periodically, but I don't want to use a non-caching provider.

I know the cached images are in the libCachedImageData directory but there is no way of telling which images belong to the layer in question. I rather not delete the whole cache. 

Is there a way of doing this already or what would the best practice be?

### LINKED ISSUE #748: Don't catching on errorTileCallback
Hi, 
I would appreciate if you can give me a help.
I have a "hello world" app and I copied all the code from the following template to my app project for test errors. The only thing I commented was the drawer:
https://github.com/fleaflet/flutter_map/blob/master/example/lib/pages/tile_loading_error_handle.dart

Steps I did: 
1- I enter the page of the tile_loading_error_handle template.
2- I entered airplane mode
3- Zoom on the map
4- The app crashes and the vscode shows:
> SocketException (SocketException: Failed host lookup: 'a.tile.openstreetmap.org' (OS Error: No address associated with hostname, errno = 7)

The error ocurred in _network_image_io.dart
https://i.imgur.com/qme2uwn.png
It should be noted that the app does not reach errorTileCallback, but rather stays in the network_image catch

I made a video:
https://youtu.be/zMzoznvhJMk

Thanks for your help

More date:
environment:
  sdk: ">=2.7.0 <3.0.0"

dependencies:
...
flutter_map: ^0.10.1+1

Flutter 1.20.4 • channel stable • https://github.com/flutter/flutter.git
Framework • revision fba99f6cf9 (6 days ago) • 2020-09-14 15:32:52 -0700
Engine • revision d1bc06f032
Tools • Dart 2.9.2

### LINKED ISSUE #752: Warning database has been locked for 0:00:10.000000.
When loading several TileLayerOptions at the same time, using the default `CachedNetworkTileProvider` I get the warning "Warning database has been locked for 0:00:10.000000." I suspect this is caused by `cached_network_image` or rather `flutter_cache_manager`. Loading only a few layers at the same time does not produce this problem.

Anyone had this and able to provide a solution/workaround (except not loading several layers...)

### LINKED ISSUE #832: How to replace CachedNetworkTileProvider
Hi,

currently I am using the CachedNetworkTileProvider, which was removed here: https://github.com/fleaflet/flutter_map/commit/1a6b495554f2e61e841874aff21621852cbeed9b

So, when updating to 0.11.0, how can I replace the CachedNetworkTileProvider properly to keep the caching functionality?

Kind regards!

MooNag:
Added recommended caching implementations for developers to use in readme (since flutter_map does not natively provide caching to reduce dependency requirements and since caching is such a use-case specific thing); **however** caching can be easily implemented with flutter_map using any Flutter ImageProvider that caches. This gives the app developer more control and reduces flutter_map maintainer load. Adding this to the readme will also lower the amount of Issues and PRs that add caching in many different ways when it should be handled by the developer.

Issues that this should close:
Fixes #832
Fixes #649
Fixes #576 (This issue will not be relevant to this repo and might still exist depending on ImageProvider)
Fixes #752 
Fixes #748 (Issue will not be relevant to this repo and still might exist depending on ImageProvider)
Fixes #614 (I think they just did not set up their URL correctly; but this should be closed)
Fixes #402 (Not relevant anymore)

Issues that this could close:
#631 (The issue should be modified or closed)

Here are PRs: that should close:
Fixes #564 (PR adds sqlite dependency and attempts to add caching TileProvider. This should be done as a plugin or developers should use an existing caching image provider)

### DIFF
--- a/README.md
+++ b/README.md
@@ -229,6 +229,45 @@ child: FlutterMap(
 
 For more details visit [Custom CRS demo page](./example/lib/pages/custom_crs/Readme.md).
 
+### How to Cache Tiles (Custom TileProvider)
+
+1. Using Flutter's NetworkImage:
+Flutter has a built in ImageProvider (NetworkImage) that caches images in memory until an app restart.
+```dart
+class CachedTileProvider extends TileProvider {
+  const CachedTileProvider();
+  @override
+  ImageProvider getImage(Coords<num> coords, TileLayerOptions options) {
+    return NetworkImage(getTileUrl(coords, options));
+  }
+}
+```
+
+2. Using the `cached_network_image` dependency:
+This dependency has an ImageProvider that caches to disk which means the cache persists through an app restart.
+```dart
+import 'package:cached_network_image/cached_network_image.dart';
+
+class CachedTileProvider extends TileProvider {
+  const CachedTileProvider();
+  @override
+  ImageProvider getImage(Coords<num> coords, TileLayerOptions options) {
+    return CachedNetworkImageProvider(
+      getTileUrl(coords, options),
+      //Now you can set options that determine how the image gets cached via whichever plugin you use.
+    );
+  }
+}
+```
+
+Lastly we need to add the CachedTileProvider TileProvider to TileLayerOptions
+```dart
+TileLayerOptions(
+  urlTemplate: 'https://example.com/{x}/{y}/{z}',
+  tileProvider: const CachedTileProvider()
+)
+```
+
 ## Run the example
 
 See the `example/` folder for a working example app.


johnpryan:
This looks good, I think many of these caching solutions could be implemented using a separate package too.

johnpryan (approved):

Issue #829, opened 2021-02-27
Migrate to null safety


AyushBherwani1998:
One of our side project uses this library, would be happy to help to migrate to null safety @johnpryan

AyushBherwani1998:
There are few blockers 

flutter_image  is deprecated and no longer maintained under Flutter org.                
latlong is archived and is no longer maintained.                              
positioned_tap_detector has been discontinued                
proj4dart is not yet migrated to sound null safety.

andreystavitsky:
For latlong here is active pull request
https://github.com/fleaflet/flutter_map/pull/750

aytunch:
in order for migrating to Flutter 2.0 we will need this library to be null safe. This is a one of a kind package which lets Flutter community to use their beloved widgets as markers. @johnpryan do you have any strategy for this? @AyushBherwani1998 thanks for listing the deprecated dependencies. It might be a good starting point before nnbd migration. I think flutter_image and latlong are not going to be hard to replace. I don't know what proj4dart is for and if there is a replacement for positioned_tap_detector.

kevin-lot:
For "latlong".
What about "latlng" package already null safety compatible ?

nb: I didn't check the code https://github.com/xclud/flutter_latlng.

JAICHANGPARK:
```

Error: Cannot run with sound null safety, because the following dependencies
don't support null safety:

 - package:flutter_map
 - package:latlong
 - package:logging
 - package:validate
 - package:proj4dart
 - package:tuple
 - package:positioned_tap_detector
 - package:flutter_image
 - package:quiver
 - package:mgrs_dart
 - package:wkt_parser
 - package:unicode
 - package:lists


```

jpeiffer:
It looks like `latlong` was abandoned but someone has forked that and added Null Safety:
https://pub.dev/packages/latlong2

attex:
logging has been updated to null-safety 
tuple has a prerelease for null-safety
quiver has been updated to null-safety
unicode has been updated to null-safety

arsamme:
I think this repo is archived and I gonna fork and publish NS version of flutter_map.

aytunch:
@arsamme thanks. But I do hope it is not the case for flutter_map. I see the owner @johnpryan active from time to time. I would love to hear his opinion about the future of this great package

arsamme:
@aytunch I hope too. But he didn't respond to any issue for a long time.

o1dnik:
[...]

PR #843, opened 2021-03-17
Replace PositionedTapDetector with new NNBD package

arsamme:
Old package was marked as discontinued so I've published null-safety version as new package.
I hope this help migrating flutter_map to NNBD asap.
#829


kengu (approved):
LGTM. Does not impact consumers of flutter_map and positioned_tap_detector is retired anyway.

PR #851, opened 2021-03-22
Nullsafety

escamoteur:
This is a migration to unsound null safety. So far I couldn't see problems, so I expect as soon as the missing packages are available that should be fine.
At some places I wasn't sure if previous nullable values should still be nullable, I left the null checks despite having changed the types to nonnullables and added a `TODO` comment.

PR #860, opened 2021-04-03
update dependency of tuple package from ^1.0.2 to ^2.0.0

supermuesli:
```tuple: ^1.0.2``` is causing dependency conflicts in my application which can be solved by updating to ```tuple: ^2.0.0```.
The [changelogs](https://pub.dev/packages/tuple/changelog) of the tuple package suggest that there were no breaking changes, so I decided to updated the package in ```flutter_map``` to ```tuple: ^2.0.0```.

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -10,7 +10,7 @@ environment:
 dependencies:
   flutter:
     sdk: flutter
-  tuple: ^1.0.2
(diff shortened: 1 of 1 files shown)


kengu:
@supermuesli PR #851 will resolve this when all dependencies have a null-safe version available. Closing.

PR #868, opened 2021-04-07
Added onMapCreated callback

### LINKED ISSUE #867: [Feature request] Add a callback after map is created
Sometimes it is helpful to return the mapController after it was created.
Especially because there is now way of knowing if a mapController was disposed.

tlueder:
This fixes #867 by adding an onMapCreated callback to the map options.


kengu (approved):
LGTM

kengu:
@tlueder Thanks for the contribution!

PR #870, opened 2021-04-14
Upgrade to sound null-safety

kengu:
Moves @escamoteur [nullsafety-branch](https://github.com/escamoteur/flutter_map/tree/nullsafety) in PR #851 so that the community can contribute to it without blocking progress.


lib/src/map/map.dart
```diff
@@ -338,17 +341,17 @@ class MapState {
...
-  LatLng getCenter() {
+  LatLng? getCenter() {
     if (_lastCenter != null) {
       return _lastCenter;
     }
     return layerPointToLatLng(_centerLayerPoint);
   }
 
-  LatLngBounds getBounds() {
+  LatLngBounds? getBounds() {
```
josxha, line 354: Is there a reason getBounds() returns a `LatLngBounds?` instead of a `LatLngBounds` object? As far as I can see, null cannot be returned here because `_calculateBounds()` already returns a `LatLngBounds` object.

escamoteur: Agreed, don't know why I out it there. Same with the `bounds` property but I wonder why do we have a get function and a property?

josxha: Right.  The same goes for line 164:
```dart
LatLng get center => getCenter() ?? options.center;
```
getCenter() can also never return null. 

It seems that the get properties are generally used and the getBounds() and getCenter() functions are more internal functions. So it's really worth considering if they really make sense.

kengu: > Agreed, don't know why I out it there. Same with the `bounds` property but I wonder why do we have a get function and a property?

Changing it from function to property would be a breaking. I guess this is a good time to to some breaking improvements as any.

kengu: @josxha The dart analyzer does not allow `LatLongBounds` as return-type because `_lastBounds` is defined as nullable with `LatLngBounds? _lastBounds`. I don't know if this is an error on the dart-side since the if-check above clearly removes the possibility of returning null. I guess there is something else going on here (like me not understanding every aspect of sound null-safety and thus how to handle cases like these) 

Screenshot 2021-04-25 at 14 18 49

Marthijs-Berfelo: @kengu regarding the null check, despite the not null check you need to "Cast away the null". So the

kengu: > @kengu regarding the null check, despite the not null check you need to "Cast away the null". So the
> `return _lastBounds;` should be `return _lastBounds!;`
> 
> also see: https://stackoverflow.com/questions/63253015/what-does-the-exclamation-mark-mean-before-a-function-call
> the official null safety documentation does suggest it's not necessary, however I've encountered similar issues and resolved them similarly.

Check - will look into it

kengu: Seems reasonable enough to use `!`, event though static safety is lost. See [null assertion operator](https://dart.dev/null-safety/understanding-null-safety#null-assertion-operator)

> Of course, like any cast, using ! comes with a loss of static safety. The cast must be checked at runtime to preserve soundness and it may fail and throw an exception. But you have control over where these casts are inserted, and 
 you can always see them by looking through your code.

I'll change `getCenter()` and `getBounds()`it soon

kengu: Changes pushed now

kengu:
> I created the pull request #894 to make flutter_map independent of [flutter_image](https://pub.dev/packages/flutter_image) and use the http package instead. After that, as far as I know, this package would be ready for a release with sound null safety. Since http is probably already integrated in most apps, it would also be a nice little dependency clean up.

Thank you @josxha - I think this is a reasonable way forward, see my [rfc described here](https://github.com/fleaflet/flutter_map/pull/894#pullrequestreview-655533983)

kengu:
@johnpryan We are ready to go!

ThexXTURBOXx:
@kengu Maybe we should wait for @josxha to merge [my PR](https://github.com/josxha/flutter_map/pull/1) into his branch. After that you can merge his (and my) changes, such that these are in this PR as well then?

example/lib/pages/network_tile_provider.dart
```diff
@@ -0,0 +1,87 @@
...
+      drawer: buildDrawer(context, route),
+      body: Padding(
+        padding: EdgeInsets.all(8.0),
+        child: Column(
+          children: [
+            Padding(
+              padding: EdgeInsets.only(top: 8.0, bottom: 8.0),
+              child: Wrap(children: [
+                Text('This is a map that is using NetworkTileProvider.'),
+                Text('Turn on flight mode to see that images are cached.'),
```
ThexXTURBOXx, line 59: We should remove this line, right? This doesn't seem to be true by documentation, right? If that's true, we should rename the provider also.

I think I missed part of the discussion in #894, so just ignore that comment if what I'm saying is wrong....

ThexXTURBOXx (changes requested):

ThexXTURBOXx:
This is everything else I could find. Just some missing TODOs and one last thing from the last discussion :)

Of course, this is not my project or anything, but maybe this helps sorting out some last things before @johnpryan can give his final review

JaffaKetchup:
@kengu also asked me to change the base (merge to) branch on my PR for offline stuff (documentation only) (https://github.com/fleaflet/flutter_map/pull/913) to this branch, so can that be merged as well please :)!
Looking forward to the release!

aytunch:
@johnpryan Hi, do you have a time frame to review this PR? We have a very long pubspec.yaml in our project and only `flutter_map` is not converted to null safety yet. And it is blocking us for updating our other packages. Reading the comments, I see there are others in the same situation too. Would appreciate it a lot if you could review this, thanks.

johnpryan:
Yes, I can take a look today.

example/lib/pages/network_tile_provider.dart
andreandersson: Based on the comments in the code here: https://github.com/fleaflet/flutter_map/blob/issues/829-nullsafety/lib/src/layer/tile_layer.dart#L116

I would say you are correct, and that the opposite is true - it's not cached. Adding the word "not" in the example would probably be enough?

andreandersson:
I've resolved the comments from @ThexXTURBOXx in #919

JaffaKetchup:
Hi,
Glad and excited to see we're almost there!
Can we also [merge this](https://github.com/fleaflet/flutter_map/pull/913) as part of this release?
Thanks for the great work you guys have done :)

example/lib/pages/network_tile_provider.dart
ThexXTURBOXx: @andreandersson Is the `CachedNetworkTileProvider` still a thing, i.e. does it still exist? If not, we should also remove that comment in the code. Thanks for spotting that! :)

danielocampo2:
Thanks for all the work you're doing, is there any update on this review? Eagerly waiting for this and the other flutter_map libs to be updated.

Thanks!

JaffaKetchup:
Hi,
Just to let you guys know, I've made some more changes to the caching section that I added in my PR #913. It might now be quite complicated, however the naming of some tile providers have made it difficult, besides the fact that Flutter's own documentation contradicts itself!
I have also slightly renewed the old offline maps section to let the user know that the guide will be using TileMill which is difficult to install and has some question marks hanging over it's repository.

lib/src/layer/group_layer.dart
```diff
@@ -7,20 +7,20 @@ class GroupLayerOptions extends LayerOptions {
   List<LayerOptions> group = <LayerOptions>[];
 
   GroupLayerOptions({
-    Key key,
```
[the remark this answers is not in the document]
johnpryan: merged, thanks

JaffaKetchup:
Um, my #913 PR was just overwritten.

johnpryan:
@JaffaKetchup There were merge conflicts, so I'm re-adding that manually.

JaffaKetchup:
Ok, I've linked the exact changes I made above (click Raw to copy paste) :)

andreandersson:
Same goes with #919 , not sure if all those have been solved yey

JaffaKetchup:
Hooray!
Thanks for the work everyone!

aytunch:
Wow :) 
I am so happy. When will it be available on pub.dev?
Thanks to everyone putting in some great work.

JaffaKetchup:
Yeah same, great if this could be published to pub.dev sooner rather than later.

johnpryan:
Published, thanks everyone!

Issue #271, opened 2019-04-27
Fling animation can be more fluid


aytunch:
I am not talking about dragging(pan). While dragging, the map is fluid. However while swiping(pan), like a fast flicker of finger, the map moves in an unnatural way. Seems too much acceleration. We can do it like Apple Maps or Google maps swipe gesture. If I am not wrong, this could be solved by altering the swipe gesture. Damping and Curves.

johnpryan:
Yep, I think I agree that this is slightly unnatural, at least on iOS. (Android users might feel differently). The fling animation is here: https://github.com/johnpryan/flutter_map/blob/1d048608ab0ab6daf716b3c4b20db7a62cc44536/lib/src/gestures/gestures.dart#L20 and is using the AnimationController.fling() function from the flutter framework.

aytunch:
@johnpryan @lpongetti There has been some great contributions in the past days to flutter_map by you two. I really think that the Fling gesture is very un natural at the moment. It is as if the animation accelerates instead of decelerating. I use Google maps, Apple maps, yandex maps on iOS and all of them behave the same way when it comes to fling(swipe/throw) gesture animation. But Flutter map behaves differently. Thanks

lpongetti:
I'll see when i'll have some time

lpongetti:
I read the code and at first look i think that can be the duration... its never setted. It have to be proportioned by velocity. but i have to try.. monday

aytunch:
@lpongetti Sounds great, thanks. Now that you pointed out a hint about duration, I looked at the code and it seems the author records the magnitude of the fling gesture and calculates the new center according to it. And then makes a linear tween to that new center. In other gestures like double tap author used Curves.fastOutSlowIn. But I couldn't see something like that for fling animation. Maybe that is what causes the unnatural animation along with no duration. Good luck since it seems too much math:D

lpongetti:
Surely the distance and duration are wrong. Fling animation sounds good.. but i dont know... Monday i will look for a solution

lpongetti:
i tried to remove fling animation and i added CurvedAnimation with Curves.decelerate.
i have pixel per seconds and i have to calculate  the duration and end offset.
with the right duration and offset the animation will have at the start the same speed of touch.

unfortunately I'm not a mathematician and i don't know how to compute them.

if you want you can try

`
var magnitude = details.velocity.pixelsPerSecond.distance;
    if (magnitude < _kMinFlingVelocity) {
      return;
    }

[...]

PR #881, opened 2021-04-25
Update map fling spring physics parameters to reduce initial acceleration

joeldsparks:
This change fixes the fling acceleration issue reported in #271.

The initial rapid acceleration appears to be due to the spring physics being critically dampened. I changed the default to be over damped and tweaked it by manually comparing it to google and apple maps and now it's smoother.

Reference: [SpringDescription.withDampingRatio constructor
](https://api.flutter.dev/flutter/physics/SpringDescription/SpringDescription.withDampingRatio.html)

### DIFF
--- a/lib/src/gestures/gestures.dart
+++ b/lib/src/gestures/gestures.dart
@@ -491,7 +491,13 @@ abstract class MapGestureMixin extends State<FlutterMap>
 
     _flingController
       ..value = 0.0
-      ..fling(velocity: magnitude / 1000.0);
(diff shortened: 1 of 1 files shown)


kengu (approved):
Tested on physical device. LGTM.

kengu:
@joeldsparks Thanks for the contribution!

aytunch:
Thanks for this fix, it is much better now.
When can we expect this to be in stable?

PR #869, opened 2021-04-08
Added new plugin to README

JaffaKetchup:
Changed the following things in the README.md file:
- Added new plugin (flutter_map_tile_caching) to the plugins list.
- Removed obsolete tile caching guide.
- Clarified difference between new plugin and preconfigured offline maps.

PR #787, opened 2020-12-04
remove mbtiles tile provider + sqflite dependency

johnpryan:
fixes #786

@avbk The dependency on sqlflite can cause problems for users who want to use flutter_map without the mbtiles provider.  Since this is a map plugin, I think this code should be written as a separate package. That way, you would include flutter_map along with the new package, (e.g. flutter_map_mbtiles) and then use the MBTilesImageProvider from that package. 

Would you like to make that package since you were the original author?

PR #894, opened 2021-05-10
Replace the flutter_image package with RetryClient from the http package 

josxha:
Since [flutter_image](https://pub.dev/packages/flutter_image) seems to have a very low priority from the flutter development team, I followed [this tip](https://github.com/flutter/flutter_image/pull/22#issuecomment-831169101) and replaced the package with the [RetryClient](https://pub.dev/documentation/http/latest/retry/RetryClient-class.html) from the [http package](https://pub.dev/packages/http). I tested my pull request with the example app and it worked fine.

The PR would also make flutter_map independent of flutter_image in the future.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## [0.13.0] - x/xx/2021
+## [0.13.0] - xx/xx/2021
 This version has support for sound null safety. For this purpose, some inactive packages were exchanged with active forks.
 
 - Sound null safety migration (#851, #870)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,14 @@
+## [0.13.0] - x/xx/2021
+This version has support for sound null safety. For this purpose, some inactive packages were exchanged with active forks.
+
+- Sound null safety migration (#851, #870)
+  - requires flutter version 2.0.0 or higher
+  - latlong is replaced with latlong2
+- Remove the package flutter_image and add http instead (#894)
+  - http has to be version 0.13.2 or higher for this package (#894)
+
+Thanks to escamoteur, ThexXTURBOXx, Sata51, tazik561, kengu, passsy, Ahmed-gubara, johnpryan and josxha for this release!
+
 ## [0.12.0] - 3/16/2021
 TileLayerOptions now takes some additional options, templateFunction,
 tileBuilder, tilesContainerBuilder, and evictErrorTileStrategy
--- a/README.md
+++ b/README.md
@@ -229,45 +229,6 @@ child: FlutterMap(
 
 For more details visit [Custom CRS demo page](./example/lib/pages/custom_crs/Readme.md).
 
-### How to Cache Tiles (Custom TileProvider)
-
-1. Using Flutter's NetworkImage:
-Flutter has a built in ImageProvider (NetworkImage) that caches images in memory until an app restart.
-```dart
-class CachedTileProvider extends TileProvider {
-  const CachedTileProvider();
-  @override
-  ImageProvider getImage(Coords<num> coords, TileLayerOptions options) {
-    return NetworkImage(getTileUrl(coords, options));
-  }
-}
-```
-
-2. Using the `cached_network_image` dependency:
-This dependency has an ImageProvider that caches to disk which means the cache persists through an app restart.
-```dart
-import 'package:cached_network_image/cached_network_image.dart';
-
-class CachedTileProvider extends TileProvider {
-  const CachedTileProvider();
-  @override
-  ImageProvider getImage(Coords<num> coords, TileLayerOptions options) {
-    return CachedNetworkImageProvider(
-      getTileUrl(coords, options),
-      //Now you can set options that determine how the image gets cached via whichever plugin you use.
-    );
-  }
-}
-```
-
-Lastly we need to add the CachedTileProvider TileProvider to TileLayerOptions
-```dart
-TileLayerOptions(
-  urlTemplate: 'https://example.com/{x}/{y}/{z}',
-  tileProvider: const CachedTileProvider()
-)
-```
-
 ## Run the example
 
 See the `example/` folder for a working example app.
@@ -276,9 +237,11 @@ To run it, in a terminal cd into the folder.
 Then execute `ulimit -S -n 2048` ([ref](https://github.com/trentpiercy/trace/issues/1#issuecomment-404494469)).
 Then execute `flutter run` with a running emulator.
 
-## Offline maps
+## Preconfigured offline maps
 
-[Follow this guide to grab offline tiles](https://tilemill-project.github.io/tilemill/docs/guides/osm-bright-mac-quickstart/)<br>
+This method is only to use preconfigured and prepackaged offline maps. For advanced caching and ability to dynamically download an area, see the [flutter_map_tile_caching](https://github.com/JaffaKetchup/flutter_map_tile_caching) plugin.
+
+First, [follow this guide to grab the tiles](https://tilemill-project.github.io/tilemill/docs/guides/osm-bright-mac-quickstart/).<br>
 Once you have your map exported to `.mbtiles`, you can use [mbtilesToPng](https://github.com/alfanhui/mbtilesToPngs) to unpack into `/{z}/{x}/{y}.png`.
 Move this to Assets folder and add asset directories to `pubspec.yaml`. Minimum required fields for offline maps are:
 
@@ -308,6 +271,7 @@ Note that there is also `FileTileProvider()`, which you can use to load tiles fr
 
 ## Plugins
 
+- [flutter_map_tile_caching](https://github.com/JaffaKetchup/flutter_map_tile_caching): Provides ability to properly cache tiles for offline use & easily download a region of a map for later offline use
 - [flutter_map_marker_cluster](https://github.com/lpongetti/flutter_map_marker_cluster): Provides Beautiful Animated Marker Clustering functionality
 - [user_location](https://github.com/igaurab/user_location_plugin): A plugin to handle and plot the current user location in FlutterMap
 - [flutter_map_location](https://github.com/Xennis/flutter_map_location): A plugin to request and display the users location and heading on the map
--- a/README.md
+++ b/README.md
@@ -286,6 +286,6 @@ Note that there is also `FileTileProvider()`, which you can use to load tiles fr
 
 For the latest roadmap, please see the [Issue Tracker]
 
-[Leaflet]: http://leafletjs.com/
+[Leaflet]: https://leafletjs.com/
 [Mapbox]: https://www.mapbox.com/
 [Issue Tracker]: https://github.com/johnpryan/flutter_map/issues
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -36,6 +36,7 @@ linter:
     - prefer_null_aware_operators
     - prefer_single_quotes
     - prefer_typing_uninitialized_variables
+    - sort_pub_dependencies
     - test_types_in_equals
     - throw_in_finally
     - unnecessary_brace_in_string_interps
--- a/example/lib/main.dart
+++ b/example/lib/main.dart
@@ -1,4 +1,5 @@
 import 'package:flutter/material.dart';
+import 'package:flutter_map_example/pages/network_tile_provider.dart';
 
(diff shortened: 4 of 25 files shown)


kengu (changes requested):
@johnpryan I think this is a good approach to the lack of maintenance of `flutter_image`. Removing a dependency is good practice when the usage is limited to a single use-case (here NetworkImageWithRetry). It increases the code footprint slightly, but not to a degree that is concerning. 

It would introduce a breaking change that need to be noted in the change log @josxha

josxha:
@kengu Yes you are right. I haven't touched the change log because the package version hasn't been raised and no newsection exists in the change log yet. 
I can look over the pull requests https://github.com/fleaflet/flutter_map/pull/870 and https://github.com/fleaflet/flutter_map/pull/851 and start collecting release notes.

josxha:
@kengu How should I proceed with the mentions? Are only committers counted or also code reviewers? 
Last release notes: ```Thanks to gr4yscale, maRci002, MooNag, tlserver, 6y```

kengu:
> @kengu How should I proceed with the mentions? Are only committers counted or also code reviewers?
> The mentions in the last release notes:
> `Thanks to gr4yscale, maRci002, MooNag, tlserver, 6y`

Both collaborators and code authors I guess.

josxha:
I hope these are all the changes.
The release date is currently blank and has to be set when the time comes.

kengu (approved):
LGTM

MooNag:
I think this is a good change. Removing dependencies that are use-case specific is important (a feature that has a hanging dependency). I think this is a good middle ground for this since most things use http we are simplifying the dependency chain.

kengu:
@josxha / @MooNag I've done some testing the example app and there are significant difference in behaviour between `NetworkImageWithRetry` implemented here and in `flutter_image`. More specific, caching is much poorer in the new implementation. See below. 

Behaviour **without** `flutter_image`
https://user-images.githubusercontent.com/866528/118354562-eac77380-b56b-11eb-9329-e2d1399a7e2f.mp4

Behaviour _with_ `flutter_image`
https://user-images.githubusercontent.com/866528/118354582-fca91680-b56b-11eb-9e61-22a8a9e32a48.mp4

I think we should stick to flutter_image now that it is migrated to null-safety. We still need to wait for them to release a new version before we can release a new version of `flutter_map`.

MooNag:
@kengu caching *should* be handled by your own implementation and have nothing to do with the image provider. Check the Readme section of flutter map for how to implement caching. Ping me if you would like me to help you (I handled that section) or make a sample project with and without flutter_image. If that does not make the caching consistent then ping me and I will look into it because then it would be a bug that I need to fix.

kengu:
@MooNag I'm not able to any changes to a PR, OP @josxha needs to do that. 

My point it that changing to this new implementation of `NetworkImageWithRetry` regress to a less usable behaviour compared to `NetworkImageWithRetry` implemented in `flutter_image` (play and compare the two videos above). The issue is not about caching per se, it's about a less usable solution compared to what we are replacing.

Now that `flutter_image` is part of the core plugins maintained by the Flutter Team, I don't see the need to replace it any longer.

ThexXTURBOXx:
Nice to see that my suggestion has been implemented here already ^^

I just wanted to do that myself, so thanks @josxha for doing this yourself and sparing my time here :)

josxha:
I find @kengu's cache test very interesting. **flutter_image** clearly offers the better "out of the box" solution here. 
However, I also find @ThexXTURBOXx's view plausible that the user should be given complete freedom over caching. 
I would therefore find it difficult to choose between the two. 

I have merged the pull request https://github.com/josxha/flutter_map/pull/1 so that a potential release is not stuck with me. It looks good to me.

ThexXTURBOXx:
I also find the cache test indeed very interesting. I also think, this should really be handled by the user. For this very reason, there already exist some libraries that cover this behavior. See https://github.com/JaffaKetchup/flutter_map_tile_caching for example.

ibrierley:
I think whatever is chosen, it's very key to make sure there's some documentation about it in the readme (and ideally suggest the alternatives).

kengu:
@ThexXTURBOXx My main reservation it that this change results in unexpected behaviour for current users of NetworkTileProvider. Event with proper warnings in changelog and readme, it is highly probable that someone is going to file a bug.

ThexXTURBOXx:
@kengu Actually, I just tried on my device. The exact same scenario as you: Turn off internet connection and try to zoom out and in. For me, the tiles are still cached and everything seems to be the same as before. Even the error messages are the same:
```
I/flutter (26429): SocketException: Failed host lookup: 'c.tile.openstreetmap.org' (OS Error: No address associated with hostname, errno = 7)
I/flutter (26429): SocketException: Failed host lookup: 'a.tile.openstreetmap.org' (OS Error: No address associated with hostname, errno = 7)
I/flutter (26429): SocketException: Failed host lookup: 'a.tile.openstreetmap.org' (OS Error: No address associated with hostname, errno = 7)
```
I am not sure what the difference is between our setups. I also just used the `example` app on Android (real Android device, Xiaomi Mi 9T Pro). Did you change anything else besides that?

MooNag:
I just wanted to chime in because I am the one to ping about caching.

For this package caching is supposed to be handled by the user because every person needs a different solution. I have documented in the Readme turn-key disk/memory caching solutions.

I suspect that the flutter_image package is providing memory caching. I think that is the package I use in the Readme as the memory caching solution. 

When using this pr please use one of the caching solutions in the Readme (which involves creating your own tile provider) to see if the issue persists. If it does persist it is an issue with this pr.

It is likely people will see that as a bug and report it. If they do tell them to read the Readme. We can alleviate it by including the breaking change in the change log instructing people to read the caching section of the Readme.

ThexXTURBOXx:
@MooNag Actually, after some research, I found out that the `flutter_image` does not explicitly cache stuff. The problem is that the `http` package's client does not cache, unlike the `io` package's client, which does cache. There are already several issues opened for `http` that suppose a cache manager or similar to enable caching by default, but for `io` it seems to be even undocumented behavior, which is unwanted in many cases (according to many issue reports).

Since it is undocumented behavior, it would be even better to move away from this old behavior to implementing this caching by ourselves using `flutter_cache_manager` for example, such that this behavior isn't fixed in a future release of Flutter and _then_ people will complain.

JaffaKetchup:
Hi all,
If you want to remove `flutter_image`, but still need good caching, please check out the plugin: https://github.com/JaffaKetchup/flutter_map_tile_caching

ThexXTURBOXx:
@JaffaKetchup I already mentioned your plugin here :)

MooNag:
That plug-in implements an even more turn key solution. It came as a result of us deciding any caching is going to be handled outside of flutter_map. That plugin was created and I added documentation to the Readme. To think of it I should also add mention to that plug-in in the Readme.

Removal of flutter_image from flutter_map is the right decision. The only value it provides is NetworkImageWithRetry and it also exhibits unexpected caching behavior as it is caching where it maybe shouldn't be. Users of flutter_map should use a proper caching image provider. The change in behavior will probably break some people's projects but it will be for the better; and they will know about the breaking change in a change log entry with an explanation of the change and how to fix it by reading the Readme and or using a flutter_map plugin or using a caching image provider they create in house or on pub.dev

JaffaKetchup:
@ThexXTURBOXx It seems like you did indeed! GitHub was hiding some comments from me, so I couldn't see it without clicking to show more. :)

@MooNag I made a pull request (https://github.com/fleaflet/flutter_map/pull/869) a few weeks back adding my plugin and removing the information about self-implementation of caching that you added in (in PR https://github.com/fleaflet/flutter_map/pull/841) because it seemed less useful with the new plugin. If you want to add your guide back in, just let me I know and I can open a new PR if you don't want to do it.

However, I also agree that `http` is a better package choice than `flutter_image` for exactly the reason @MooNag said (https://github.com/fleaflet/flutter_map/pull/894#issuecomment-841428126).

MooNag:
@JaffaKetchup I think we should keep the docs for self-implemented caching and the little paragraph I had explaining how the chain of ImageProvider->TileProvider-> what gets displayed on the map works to make sure that caching makes sense to people and isn't just a black box. Or if people want to use flutter_image, or some other image provider they are already using, have made in house, or prefer more than the black box plugin. It is just a few lines of code and it will still work as long even if the caching plugin for flutter_map isn't maintained anymore. My personal preference for dependencies is less is more. Hanging features and things that are user dependent or already have a huge ecosystem (like caching ImageProviders in flutter) CAN have a black box solution for flutter_map but it should be clearly documented that you can cache without an extra black_box dependency or use an image provider you already are using.

I think options are good but just telling people to use a black box solution to something like caching is dangerous when it is can be easily handled by 5 lines of code and an existing well maintained image provider.

josxha:
> @MooNag Since it is undocumented behavior, it would be even better to move away from this old behavior to implementing this caching by ourselves using `flutter_cache_manager` for example, such that this behavior isn't fixed in a future release of Flutter and _then_ people will complain.

As far as I know `flutter_cache_manager` has been a dependency in earlier versions of `flutter_map`.
> [0.11.0] - 01/29/2021
This version removes various tile providers that depend on plugins. This helps simplify the flutter_map release process. Tile providers can be implemented in your app or in a separate package.
remove mbtiles tile provider + sqlflite dependency (#787)

Glad to hear about `flutter_map_tile_caching`! 😀

kengu:
@ThexXTURBOXx Nothing more than switching between the implementations. 

@MooNag If we are going this route, the broader question then becomes what the real value of having two network providers? The naming of NoCachingNetworkProvider which indicates that NetworkTileProvider is caching implicitly.  

Screenshot 2021-05-21 at 20 38 30


I think the new implementation of NetworkTileProvider chould be renamed to something like NoCachingNetworkRetryProvider to clarify and maybe help the community to not expect it to cache (anymore)?

MooNag:
> > @MooNag Since it is undocumented behavior, it would be even better to move away from this old behavior to implementing this caching by ourselves using `flutter_cache_manager` for example, such that this behavior isn't fixed in a future release of Flutter and _then_ people will complain.
> 
> As far as I know `flutter_cache_manager` has been a dependency in earlier versions of `flutter_map`.
> 
> > [0.11.0] - 01/29/2021
> > This version removes various tile providers that depend on plugins. This helps simplify the flutter_map release process. Tile providers can be implemented in your app or in a separate package.
> > remove mbtiles tile provider + sqlflite dependency (#787)
> 
> Glad to hear about `flutter_map_tile_caching`! 😀


Yeah we removed it because caching is such a user specific thing we wanted to allow for the user to choose how or if they cache their image tiles. Also that dependency was heavy and relied on a lot of other dependencies. Removing it gave a choice of how people want to cache and made flutter_map significantly more lightweight. I added the docs on how to create a tile provider with whatever image provider (that caches or doesn't cache) you want at that time to document the change.

MooNag:
@kengu The expectation should be that NO built in tile providers provide caching. Some might still provide memory caching. There is a built in flutter image provider that might be used in one of the tile providers that does memory caching.

kengu:
> providers

I don't disagree with that, less is more here, and one less dependency is a good thing. But I want to make sure we help the community to understand this through consistent naming of code, not readme alone (which many do not read that carefully).

JaffaKetchup:
@MooNag, did you want me to add back your instructions then, perhaps as a 'if you don't want another package in your project, then you can do this...'? If so, I'll open a new PR at some point soon.
And I think the naming is important, it needs to be clear what caches and what doesn't.

MooNag:
@JaffaKetchup yeah let's open a PR to add the docs back and also mention that there are no built in tile providers that provide caching and also look at modifying this PR to include changes to the built in tile providers naming to make it not sound like they cache.

JaffaKetchup:
Ok, I'm making a new PR now.

JaffaKetchup:
https://github.com/fleaflet/flutter_map/pull/913 is the new PR for what @MooNag, @kengu, and I have just been talking about.

JaffaKetchup:
Am I right in thinking that `NetworkTileProvider()` does actually cache images, and if so, should I update my PR to mention this in the docs? Or shall I just leave it out and say there aren't any providers that cache.
Overall, I'm not sure the point of two TileProviders that are so similar.

JaffaKetchup:
I have further changes my PR to reflect the new information I have found looking through the source code.

Issue #885, opened 2021-05-01
Stateful markers state are recreated even when markers still visible


eric-taix:
## Current behaviour

When using stateful markers, their states are/may be recreated even when markers are on screen. This happens when another marker goes off screen.

In the video below, I've created an example with colored containers as markers.
The color's state is initialized in the `initState` of the marker widget:
```dart
  @override
  void initState() {
    super.initState();
    color = _ColorGenerator.getColor();
  }
```

As you can see in the video below, when all markers are on screen everything is ok. But when one marker goes off screen, some (not all) markers are recreated. 

https://user-images.githubusercontent.com/712517/116788875-5b6e8a80-aaac-11eb-8697-efcbf36e956e.mp4

## Expected behaviour 

This is fine that a colored marker, which goes off screen and then goes on screen, changes its color as the marker was not more in the widget tree, but I would expect that markers which still on screen keep their state (in my real application the state is an animation, and I don't want the animation to be played again each time I move the map).

## Analysis

This issue is typically a `Key` issue but adding a key to `ColoredMarker` does not solve the issue.

In the current code the `MarkerLayer`  widget uses a `Stack` (https://github.com/fleaflet/flutter_map/blob/master/lib/src/layer/marker_layer.dart#L252) to display markers and using a key in `ColoredMarker` should be enough but markers are wrapped into a `Positioned` (https://github.com/fleaflet/flutter_map/blob/master/lib/src/layer/marker_layer.dart#L241) widget to set marker position according to their anchor and position. 

Emily Fortuna mentioned these kind of dragons (https://medium.com/flutter/keys-what-are-they-good-for-13cb51742e7d)
> Where do I put ‘em?
Short answer: if you need to add keys to your app, you should add them at the top of the widget subtree with the state you need to preserve.
A common mistake I’ve seen is people thinking they only need to put a key on the first stateful widget, but there be dragons. Don’t believe me? To show what sort of trouble we can get into, I wrapped my colorfulTile widgets with padding widgets, but I left the keys on the tiles.

## How to fix

The key of a stateful marker should be put in the top of the widget subtree so it should be put in the `Positioned` widget. One solution is to add a `Key key` attribut to the `Marker` class and use it as key for the `Positioned` widget:
```dart
          markers.add(
            Positioned(
              key: marker.key, // Use the key of the marker
              width: marker.width,
              height: marker.height,
              left: pos.x - width,
              top: pos.y - height,
              child: markerWidget,
            ),
          );
```

[...]

PR #901, opened 2021-05-13
Add a key for markers to preserve state

eric-taix:
Fix issue https://github.com/fleaflet/flutter_map/issues/885 which prevent from preserving markers state.

A key was added to the `Marker` and used for the top level widget in the markers layer (the `Positioned`). This ensure that even if some markers are added / removed from the map, the `Stack` element is able to find the corresponding widget and then preserve its state instead of recreating a new one.


kengu (approved):
LGTM

kengu:
@ eric-taix Thanks for the contribution! Close #885.

PR #913, opened 2021-05-21
Clarification of all options available for offline mapping

JaffaKetchup:
There are now 4 different ways of caching/downloading/preconfiguring offline mapping, so this PR aims to describe them all, and say which option is best for which use case.
I also fixed some capitalisation inconsistencies.

PR #919, opened 2021-06-02
Review comments from issues/829-nullsafety

andreandersson:


example/lib/pages/network_tile_provider.dart
```diff
@@ -56,7 +56,7 @@ class NetworkTileProviderPage extends StatelessWidget {
               padding: EdgeInsets.only(top: 8.0, bottom: 8.0),
               child: Wrap(children: [
                 Text('This is a map that is using NetworkTileProvider.'),
-                Text('Turn on flight mode to see that images are cached.'),
+                Text('Turn on flight mode to see that images are not cached.'),
```
ThexXTURBOXx, line 59: I am unsure if this is the right way. If there is no caching, why should we tell the user about it? I would still suggest removing that, but let's @kengu or @johnpryan decide about that

lib/src/core/bounds.dart
```diff
@@ -32,8 +32,7 @@ class Bounds<T extends num> {
   }
 
   /// This [Bounds] central point.
-  CustomPoint<double> getCenter() {
-    //TODO should this be a getter?
+  CustomPoint<double> get center {
```
ThexXTURBOXx, line 35: Could be shorter/cleaner (I think... Don't have an IDE right now):
```dart
CustomPoint<double> get center => CustomPoint<double>(
    (min.x + max.x) / 2,
    (min.y + max.y) / 2,
);
```

ThexXTURBOXx (changes requested):

example/lib/pages/network_tile_provider.dart
JaffaKetchup: Yeah there's a whole shenanigans about caching. In the end I think it was decided that caching would not be mentioned anywhere apart from a dedicated caching section in the docs because the main provider doesn't cache properly and we didn't want things to be too confusing. That's what [this PR](https://github.com/fleaflet/flutter_map/pull/913) is for.
So I think it should be removed all together.

ThexXTURBOXx: @JaffaKetchup Okay, then in that case I propose also removing [that reference](https://github.com/fleaflet/flutter_map/blob/issues/829-nullsafety/lib/src/layer/tile_layer.dart#L116) that @andreandersson spotted

andreandersson: I haven't checked that far into the NetworkTileProvider, so I don't know at all what it does. What I do know is that images are not cached, and that _might be interesting to developers?_ If anyone has a good description I'm very open to changing it...

But I'll most likely not take the time in the near future to look deeper into it and find a better description though.

ThexXTURBOXx: @andreandersson No problem! :)
I would propose something like

```
This Provider does not provide caching.\n
For further options about that, check flutter_map's README on GitHub.
```

JaffaKetchup: > @andreandersson No problem! :)
> I would propose something like
> 
> ```
> This Provider does not provide caching.\n
> For further options about that, check flutter_map's README on GitHub.
> ```

That seems a good solution when the PR mentioned above is merged

andreandersson: I've assumed #913 will be merged as well, and updated the description here and in tile_layer.dart ;)

JaffaKetchup (changes requested):
`NonCachingNetworkTileProvider()` is the default provider, not `CachedNetworkTileProvider()`! In fact `CachedNetworkTileProvider()` doesn't actually exist. Honestly, I think the tile providers need a rename! `NetworkTileProvider()` actually uses `NetworkImageWithRetry()` which (as far as I understand) doesn't cache tiles, and `NonCachingNetworkTileProvider()` actually uses `NetworkImage()` which (as far as I understand) also does cache tiles, but not very well and not very flexibly.

I'd change it to:
``` dart
/// Provider to load the tiles. The default is `NonCachingNetworkTileProvider()` which
/// doesn't cache tiles and won't retry the HTTP request. Use `NetworkTileProvider()` for
/// a provider which will retry requests. For the best caching implementations, see the
/// flutter_map readme.

andreandersson:
> `NonCachingNetworkTileProvider()` is the default provider, not `CachedNetworkTileProvider()`! In fact `CachedNetworkTileProvider()` doesn't actually exist. Honestly, I think the tile providers need a rename! `NetworkTileProvider()` actually uses `NetworkImageWithRetry()` which (as far as I understand) doesn't cache tiles, and `NonCachingNetworkTileProvider()` actually uses `NetworkImage()` which (as far as I understand) also does cache tiles, but not very well and not very flexibly.
> 
> I'd change it to:
> 
> ```dart
> /// Provider to load the tiles. The default is `NonCachingNetworkTileProvider()` which
> /// doesn't cache tiles and won't retry the HTTP request. Use `NetworkTileProvider()` for
> /// a provider which will retry requests. For the best caching implementations, see the
> /// flutter_map readme. 
> ```

Fixed! I believe we should create a ticket for looking further into the documentation around caching and fix it. And I don't think we have to fix it for the null-safety release (which hopefully would happen soon...)

JaffaKetchup:
Thanks, I think that's a bit better. I'm aiming to clarify the documentation around caching with #913!

PR #928, opened 2021-06-08
add vector_map_tiles plugin to readme

greensopinion:
I'd like to add vector_map_tiles to the readme so that it's more discoverable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -347,6 +347,7 @@ See also `FileTileProvider()`, which loads tiles from the filesystem.
 - [flutter_map_marker_popup](https://github.com/rorystephenson/flutter_map_marker_popup): A plugin to show customisable popups for markers.
 - [map_elevation](https://github.com/OwnWeb/map_elevation): A widget to display elevation of a track (polyline) like Leaflet.Elevation
 - [flutter_map_floating_marker_titles](https://github.com/androidseb/flutter_map_floating_marker_titles): Displaying floating marker titles on the map view
+- [vector_map_tiles](https://pub.dev/packages/vector_map_tiles): A plugin that enables the use of vector tiles.
 
 ## Roadmap
 


JaffaKetchup:
This looks very useful and would definitely widen the possibilities!
Although, I would like to see a way to use existing tile providers, so that this plugin could be used with another plugin that provides a provider (let's say my offline one for example), but I'm not sure if this is possible?

greensopinion:
@JaffaKetchup thanks for the suggestion. I did look at that however [`TileProvider`](https://github.com/fleaflet/flutter_map/blob/master/lib/src/layer/tile_provider/tile_provider.dart#L11) is all about tiles as images, whereas this is all about vector tiles, which are not images.

ibrierley:
I concur with that, I took at look at using the existing tile providers for my vector stuff, and it felt like banging a square peg in a round hole (was looking at a separate codec decode and things like that). The alternative felt like trying to change some of the core code, which felt like a bad idea.

JaffaKetchup:
Yeah, fair enough!

ibrierley (commented):
Fine with me, it shows it's current status on the readme, and I've tested it, so I think should be included.

greensopinion:
bumping this so it doesn't get automatically closed - would be great to get this in!

JaffaKetchup:
Definitely!

________________________________

JaffaKetchup:
@greensopinion Please see https://github.com/greensopinion/flutter-vector-map-tiles/issues/11

kengu (approved):

greensopinion:
nice, thanks!

JaffaKetchup:
@greensopinion Your plugin has now also been added to the WIP documentation as discussed:
https://flutter-map.vercel.app/plugins/list#:~:text=vector_map_tiles%20by%20greensopinion

greensopinion:
great!

PR #936, opened 2021-06-15
fix : mapController.moveAndRotate with null id

spyke555:
The id variable in the mapController moveAndRotate function is declared as nullable, but a null check is applied. This causes the function to not work if it is called without the optional id field.

Suggest casting the String? to a String type to allow the null id value to be used.


tlueder:
This PR only circumvents the issue but doesn't fix the root problem and should be closed in favour of https://github.com/fleaflet/flutter_map/pull/1003

PR #937, opened 2021-06-15
Expose method for fitBounds calculation

JankoLancer:
I need this calculation in my project, so I can animate markers move and zoom.
I think this can be useful for others too.

### DIFF
--- a/lib/flutter_map.dart
+++ b/lib/flutter_map.dart
@@ -5,6 +5,7 @@ import 'dart:math';
 
 import 'package:flutter/widgets.dart';
 import 'package:flutter_map/flutter_map.dart';
+import 'package:flutter_map/src/core/center_zoom.dart';
 import 'package:flutter_map/src/geo/crs/crs.dart';
 import 'package:flutter_map/src/gestures/interactive_flag.dart';
 import 'package:flutter_map/src/gestures/map_events.dart';
@@ -121,6 +122,11 @@ abstract class MapController {
   /// Fits the map bounds. Optional constraints can be defined
   /// through the [options] parameter.
   void fitBounds(LatLngBounds bounds, {FitBoundsOptions? options});
+
+  /// Calculate map location and zoom that fit given [bounds]
+  /// Optional constraints can be defined through the [options] parameter.
+  CenterZoom getBoundsCenterZoom(LatLngBounds bounds,
+      {FitBoundsOptions? options});
 
   Future<Null> get onReady;
 
--- a/lib/src/map/map.dart
+++ b/lib/src/map/map.dart
@@ -52,6 +52,12 @@ class MapControllerImpl implements MapController {
   }
 
   @override
+  CenterZoom getBoundsCenterZoom(LatLngBounds bounds,
+      {FitBoundsOptions? options}) {
+    return _state.getBoundsCenterZoom(bounds, options!);
+  }
+
+  @override
   LatLng get center => _state.center;
 
   @override


comatory:
This would be useful to have!

kengu:
@JankoLancer - If you fix the conflicts, I'll review this PR soon

raycastsant:
It would be magnificent to have this option in the official version. Since I come with problems in my project, trying that the map is shown in the container where I am showing it.
For the work I am doing, it is necessary to show data from some buildings planes in **Geojson**, which are dynamic, and add a **rotation** to show the map in the desired way.
What I´m doing at this moment is the first one I call `MapController.FitBounds (...)` with the boundaries where all the polygons are, and then I call:
```
   MapController.Movendrotate (
            Mapcontroller.center,
            mapcontroller.zoom + mybuilding.initzoom,
            Mybuilding.Rotation,
          )
```
As you can see, I have to add a zoom value (`mybuilding.initzoom`) to obtain the desired frame.
But this brings me problems when the size of the screen is changed (on the web) or when the application is used on a mobile device, since the box where the map is painted changes and, sometimes, the zoom is still too large or small in some building planes.

raycastsant:
Maybe this image can give you a better idea of what is happening to me:

ibrierley:
I'm a little unclear on this, what exactly is needed...as fitBounds and centerZoomFitBounds are both exposed to the mapController aren't they ?

Isn't it more that maybe they don't take into account rotation, so should be fixed, or am I totally missing the point :D (very possible, as only just looked at this! so probably just needs a bit of clarity from the start).

raycastsant:
> I'm a little unclear on this, what exactly is needed...as fitBounds and centerZoomFitBounds are both exposed to the mapController aren't they ?
> 
> Isn't it more that maybe they don't take into account rotation, so should be fixed, or am I totally missing the point :D (very possible, as only just looked at this! so probably just needs a bit of clarity from the start).

Exactly, it seems to me that the problem is that fitBounds and centerZoomFitBounds are not working correctly with rotation.

ibrierley:
I'm still a bit uncertain about this with the examples (i.e small screen size vs large screen size), and then with rotation.

Probably best bet is to create a minimal example for anyone to test which highlights the problem, and can be tested for a correct solution.

OE-Ahmed (approved):

OE-Ahmed (approved):

OE-Ahmed (approved):

OE-Ahmed (approved):
I have approved from my side.

ibrierley (changes requested):
I'm a bit unclear on this one. I just get errors whenever I try and use it. It may be related to various bits not being initialized when I'm trying though (maybe related to related problems where if FlutterMap hasn't got a screen size when called it can't retrieve the info), but either way I think maybe it needs an example page. Even then I'm a bit uncertain if I can cause exceptions with it.

JaffaKetchup:
@JankoLancer Can you resolve the issues @ibrierley mentioned, and resolve the conflicts, then we can merge this? Thanks :)

JaffaKetchup:
@JankoLancer, hope you're doing well.
Please [resolve the changes that @ibrierley requested](https://github.com/fleaflet/flutter_map/pull/937#pullrequestreview-1010521860) as soon as possible, otherwise I will have to close this PR.
Thanks for your contributions so far!

JaffaKetchup:
Unfortunately, I'm going to have to close this due to inactivity. Please don't hesitate to contribute again!

PR #944, opened 2021-06-21
Implemented equality operators for LatLngBounds and MapPosition

TheOneWithTheBraid:
In some cases it might be useful to know whether e.g. LatLngBound *are* just some other LatLngBounds.

I created a proper operator implementation making it easyer to determine and making all user's implementations cleaner.

### DIFF
--- a/lib/flutter_map.dart
+++ b/lib/flutter_map.dart
@@ -375,6 +375,16 @@ class MapPosition {
   final bool hasGesture;
 
   MapPosition({this.center, this.bounds, this.zoom, this.hasGesture = false});
+
+  @override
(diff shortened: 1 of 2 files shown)


mehdigharibi:
hello , 
is there any method to link array of markers with polyline ???
thanks

JaffaKetchup:
Nooo, don't go stale, this is useful functionality! 😅

> hello , 
> is there any method to link array of markers with polyline ???
> thanks

I'm not sure I understand, but can't you just make a PolylineLayerOptions() and pass in the same points as the markers?

kengu (approved):

PR #989, opened 2021-08-05
Use default values from MapOptions before MapController is ready

kengu:
This PR removes a class of LateInitializationErrors when passing MapController as argument to FlutterMap widget. See #986 for context. The change is breaking for clients that create their own MapController.

**flutter_map.dart** (breaking change)
```dart
factory MapController(MapOptions options) => MapControllerImpl(options);
```

### DIFF
--- a/example/ios/Podfile
+++ b/example/ios/Podfile
@@ -1,0 +1,41 @@
+# Uncomment this line to define a global platform for your project
+# platform :ios, '9.0'
+
+# CocoaPods analytics sends network stats synchronously affecting flutter build latency.
+ENV['COCOAPODS_DISABLE_STATS'] = 'true'
+
+project 'Runner', {
+  'Debug' => :debug,
+  'Profile' => :release,
+  'Release' => :release,
+}
+
+def flutter_root
+  generated_xcode_build_settings_path = File.expand_path(File.join('..', 'Flutter', 'Generated.xcconfig'), __FILE__)
+  unless File.exist?(generated_xcode_build_settings_path)
+    raise "#{generated_xcode_build_settings_path} must exist. If you're running pod install manually, make sure flutter pub get is executed first"
+  end
+
+  File.foreach(generated_xcode_build_settings_path) do |line|
+    matches = line.match(/FLUTTER_ROOT\=(.*)/)
+    return matches[1].strip if matches
+  end
+  raise "FLUTTER_ROOT not found in #{generated_xcode_build_settings_path}. Try deleting Generated.xcconfig, then run flutter pub get"
+end
+
+require File.expand_path(File.join('packages', 'flutter_tools', 'bin', 'podhelper'), flutter_root)
+
+flutter_ios_podfile_setup
+
+target 'Runner' do
+  use_frameworks!
+  use_modular_headers!
+
+  flutter_install_all_ios_pods File.dirname(File.realpath(__FILE__))
+end
+
+post_install do |installer|
+  installer.pods_project.targets.each do |target|
+    flutter_additional_ios_build_settings(target)
+  end
+end
--- a/example/ios/Podfile.lock
+++ b/example/ios/Podfile.lock
@@ -1,0 +1,22 @@
+PODS:
+  - Flutter (1.0.0)
+  - location (0.0.1):
+    - Flutter
+
+DEPENDENCIES:
+  - Flutter (from `Flutter`)
+  - location (from `.symlinks/plugins/location/ios`)
+
+EXTERNAL SOURCES:
+  Flutter:
+    :path: Flutter
+  location:
+    :path: ".symlinks/plugins/location/ios"
+
+SPEC CHECKSUMS:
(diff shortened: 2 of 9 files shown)


JaffaKetchup:
Looks good, what happens if you still declare the options in the map widget rather than in the controller?

kengu:
> Looks good, what happens if you still declare the options in the map widget rather than in the controller?

It is passed to the MapController factory. Essentially, MapController is now the owner of MapOptions. If you pass a MapOptions instance to FlutterMap together with mapController, the MapController.options is updated with it.

JaffaKetchup:
Ok, and is MapController.options a setter? ie. Can you update the options in anyway?

kengu:
> Ok, and is MapController.options a setter? ie. Can you update the options in anyway?

Yes. I don't think that is a problem, but I'm not completely sure.

JaffaKetchup:
> > Ok, and is MapController.options a setter? ie. Can you update the options in anyway?
> 
> Yes. I don't think that is a problem, but I'm not completely sure.

Well, if it works, then I think I'm all for it. As long as you definitely can still change the options.

JaffaKetchup (approved):
> > ~Ok, and is MapController.options a setter? ie.~ Can you update the options in anyway, either setting the state of the old method, or using the controller for the new method? This is important for things like changing layers and my plugin.
> 
> Yes. I don't think that is a problem, but I'm not completely sure.

Well, if it works, then I think I'm all for it. As long as you definitely can still change the options, this is absolutely vital to me.

ibrierley:
This works ok for me, but I'm a little unsure if anyone changes MapOptions, or if it's just a startup state only (I don't think you can change MapOptions later, but I'm not sure if you can originally either).

So when JaffaKetchup says "as long as you can still change the options"...just double checking, what options are we talking about that you currently change ? May just need a bit of clarity on this, as I'm not confident options (or at least certain options) can be changed, so don't feel right approving just yet.

kengu:
Yes, you can change options in the build methods by passing MapOptions as before, see https://github.com/fleaflet/flutter_map/pull/989/files#diff-9e38b372180867f0bbd262304705977db87f5689ebfa9085d1dc766de9f19ccaL172

JaffaKetchup:
@ibrierley was talking about things like `tileProvider:` and `urlTemplate:` two very important things to me and my plugin.
@kengu, if you're absolutely sure that won't cause any issues when it gets reassigned (or whatever happens), then great!

kengu:
> This works ok for me, but I'm a little unsure if anyone changes MapOptions, or if it's just a startup state only (I don't think you can change MapOptions later, but I'm not sure if you can originally either).
> 
> So when JaffaKetchup says "as long as you can still change the options"...just double checking, what options are we talking about that you currently change ? May just need a bit of clarity on this, as I'm not confident options (or at least certain options) can be changed, so don't feel right approving just yet.

MapOptions, not LayerOptions, handling of these are unchanged.

JaffaKetchup:
OH WAIT! I'm getting things mixed up! SORRY. Well in that case, have my double approval.

JaffaKetchup (approved):
Here's my double approval! 🤣

JaffaKetchup:
Meanwhile, I'm going to promote this yet again for @kengu and other new visitors in the future (helo from the past), very sorry again, but I feel like soon is a good time to do this if we're doing breaking changes: https://github.com/fleaflet/flutter_map/discussions/927#discussioncomment-1134586

JaffaKetchup:
#990 looks like it's having an error similar. Will it be solved by this?

kengu:
> #990 looks like it's having an error similar. Will it be solved by this?
Maybe. When the OP does not supply any stacktrace or other means to determine where the error occurred, it is very hard to know without attempting to reproduce (which I do not have time or energy to do). This fixes only read-operations to be "safe" until FlutterMapState is initialised. Any attempt to perform a command like move or zoom on the controller will still fail. One workaround is to check if the controller is ready and silently skip the command if not. But that is pretty hacky.

JaffaKetchup:
Hi there,
Any news of this PR being merged (I assume we're waiting for @johnpryan) or of the removal of async creation of `MapState`?
Many thanks.

JaffaKetchup:
@ibrierley Maybe we need to have a look at this sooner rather than later, it's been quite inactive for a while.

ibrierley (commented):
Might be a few days before I can dig into the code and test properly. Couple of ponderings in the meantime...
Do we have an example that definitely causes current issues we can test with...
Do we need the ios/Podfile stuff...
Do we need to force mapController to take any options (is this the breaking change bit?)...can they be optional (eg Mapoptions? options, and if not passed, mapController just uses the maps MapOptions ?  (I may well be misunderstanding this bit without running through the code and testing). I'd rather not force any specific basic usage if we can get around it (as it means every single example out there in the wild no longer works...)

JaffaKetchup:
Closed in favour of #1293, partially due to inactivity.

PR #994, opened 2021-08-11
bugfix: Prevent parent ListView widgets from snatching scrolling gestures

Binabh:
This is basically copy of https://github.com/fleaflet/flutter_map/pull/717 (since the PR creator does not seem intrested in implementing the commented option and PR is now stale). I saw that the PR is not  being merged because there is no option to developer if they want previous behavior. I have tried to add a bool in mapoptions to determine the type of behavior in the map.

Let me know if anything else is required. I want this feature to be implemented.
Thank You!

### DIFF
--- a/lib/src/map/flutter_map_state.dart
+++ b/lib/src/map/flutter_map_state.dart
@@ -137,43 +137,7 @@ class FlutterMapState extends MapGestureMixin {
                     onTapDown: _positionedTapController.onTapDown,
                     onTapUp: handleOnTapUp,
                     child: scaleGestureDetector(
-                      child: ClipRect(
-                        child: Stack(
-                          children: [
-                            OverflowBox(
-                              minWidth: size.x as double?,
-                              maxWidth: size.x as double?,
-                              minHeight: size.y as double?,
-                              maxHeight: size.y as double?,
-                              child: Transform.rotate(
-                                angle: mapState.rotationRad,
-                                child: Stack(
-                                  children: [
-                                    if (widget.children.isNotEmpty)
-                                      ...widget.children,
-                                    if (widget.layers.isNotEmpty)
-                                      ...widget.layers.map(
-                                        (layer) => _createLayer(
-                                            layer, options.plugins),
-                                      )
-                                  ],
-                                ),
-                              ),
-                            ),
-                            Stack(
-                              children: [
-                                if (widget.nonRotatedChildren.isNotEmpty)
-                                  ...widget.nonRotatedChildren,
-                                if (widget.nonRotatedLayers.isNotEmpty)
-                                  ...widget.nonRotatedLayers.map(
-                                    (layer) =>
-                                        _createLayer(layer, options.plugins),
-                                  )
-                              ],
-                            ),
-                          ],
-                        ),
-                      ),
+                      child: _buildMap(size)
                     ),
                   )
                 : GestureDetector(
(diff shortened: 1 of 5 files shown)


rknell:
I've just tested this out and it works as advertised without any noticeable issues.

However the name of the flag wasn't very intuitive and its implemented as being on by default so it would be a breaking change.

I would suggest something like "captureGesturesInScrollview: true" and have it off by default. But im not a package maintainer so just my 2c.

Otherwise great work, I was just about to go about implementing it myself for a project due out tomorrow.

Binabh:
> However the name of the flag wasn't very intuitive and its implemented as being on by default so it would be a breaking change.
> 
> I would suggest something like "captureGesturesInScrollview: true" and have it off by default. But im not a package maintainer so just my 2c.

Thank You for your comment. I think It is not a big issue to change that. I can do it. I will wait for review from maintainers and make all the changes along with this if they also suggest any.

rknell:
Yeah absolutely! Their opinion matters, I really just wanted to let them know its working for me in the wild which gives them some confidence to merge the pr in any form.

kengu:
@Binabh Can you apply formatting and push the changes? Just run `dart format .` from repo-root.

Binabh:
@kengu done. let me know if anything else is needs to be modified

lib/src/map/flutter_map_state.dart
```diff
@@ -92,49 +130,98 @@ class FlutterMapState extends MapGestureMixin {
...
+                                ...widget.nonRotatedChildren,
+                              if (widget.nonRotatedLayers.isNotEmpty)
+                                ...widget.nonRotatedLayers.map(
+                                  (layer) =>
+                                      _createLayer(layer, options.plugins),
+                                )
+                            ],
+                          ),
+                        ],
+                      ),
```
kengu, line 222: This is just duplicated code, which needs to be moved to a new method that can be called from this builder method. This reduce technical dept and makes the code more maintainable. 

```dart
Widget _buildMap() { ... }

...
     return MapStateInheritedWidget(
        mapState: mapState,
        child: Listener(
            onTap: handleTap,
            onLongPress: handleLongPress,
            onDoubleTap: handleDoubleTap,
            child: options.allowPanningOnScrollingParent  
              ? GestureDetector(
                   onTap: _positionedTapController.onTap,
                   onLongPress: _positionedTapController.onLongPress,
                   onTapDown: _positionedTapController.onTapDown,
                   onTapUp: handleOnTapUp,
                   child: _buildMap(),
                 )
              :  GestureDetector(
                    onScaleStart: handleScaleStart,
                    onScaleUpdate: handleScaleUpdate,
                    onScaleEnd: handleScaleEnd,
                    onTap: _positionedTapController.onTap,
                    onLongPress: _positionedTapController.onLongPress,
                    onTapDown: _positionedTapController.onTapDown,
                    onTapUp: handleOnTapUp,
                    child: _buildMap())),
...
```

kengu (changes requested):
LGTM. See my comment on code-duplication. When this is fixed, I'll approve and merge.

lib/src/map/flutter_map_state.dart
Binabh: sure. I will do this

Binabh: @kengu my latest commit should fix the code duplication issue. Please review and let me know if any further changes required.

Binabh:
@kengu fixed the code duplication issue. let me know if anything else required

kengu (approved):

Issue #1002, opened 2021-08-29
Inconsistent null-aware behaviour in moveAndRotate


JonIsAmazingYa:
Noticed this minor issue that I have patched locally -- `id` is marked (I believe correctly) as an optional parameter, however the null-aware `!` operator is then used to perform a null-check.

In the underlying `move` and `rotate` functions, `id` is not mandatory, and as a keyword arg it seems sensible that this operator were not used in this context.

Surely in this case, we should either enforce `id` as a required parameter everywhere, or, in my view preferably, remove the null-safe check.

https://github.com/fleaflet/flutter_map/blob/f19cce5c3a2a3d0d9691547fd3bd9e7d8c1d7823/lib/src/map/map.dart#L36

Will open a PR, but opening this issue also for discussion, as not entirely familiar with the codebase.

JaffaKetchup:
This has been causing issues for me too ("null check operator used on null value"), and as `id` should be optional, your PR is the right way of doing things, I think.

tlueder:
Same for me, this broke my app and it took me a while to find the cause.

PR #1003, opened 2021-08-29
Remove inappropriate null-aware checking to moveAndRotate

JonIsAmazingYa:
Proposed fix for issue mentioned in #1002


JaffaKetchup:
This also covers the same issue as #936.

tlueder:
But this one is an actual fix, and https://github.com/fleaflet/flutter_map/pull/936 is not.

johnpryan (approved):

Issue #965, opened 2021-07-09
Animate camera fit bounds


vrman:
is this possible to do this with flutter_map ? you can do this in google map flutter with animateCamera function like this :
`controller.animateCamera(CameraUpdate.newLatLngBounds(_bounds(markers), 50));`

cant find something similar in flutter_map plugin

bonfry:
I'm trying doing it with 4 Tween but the effect doesn't look good
```
setCenterOnMap(
    LatLng latLng, [
    double zoom = 8,
  ]) async {
    final boundsFactor = 0.64 / zoom;

    final _boundNordTween = Tween<double>(
      begin: _mapController.bounds!.north,
      end: latLng.latitude + boundsFactor,
    );
    final _boundWestTween = Tween<double>(
      begin: _mapController.bounds!.west,
      end: latLng.longitude - boundsFactor,
    );
    final _boundEastTween = Tween<double>(
      begin: _mapController.bounds!.east,
      end: latLng.longitude + boundsFactor,
    );
    final _boundSouthTween = Tween<double>(
      begin: _mapController.bounds!.south,
      end: latLng.latitude - boundsFactor,
    );

    // Create a animation controller that has a duration and a TickerProvider.
    var controller = AnimationController(
      duration: const Duration(milliseconds: 500),
      vsync: this,
    );
  
    Animation<double> animation = CurvedAnimation(
      parent: controller,
      curve: Curves.easeIn,
    );

    controller.addListener(() {
      _mapController.fitBounds(
        LatLngBounds(
          LatLng(
            _boundNordTween.evaluate(animation),
            _boundWestTween.evaluate(animation),
          ),
          LatLng(
            _boundSouthTween.evaluate(animation),
            _boundEastTween.evaluate(animation),
          ),
        ),
        options: FitBoundsOptions(padding: EdgeInsets.only(bottom: 240)),
      );
    });

    animation.addStatusListener((status) {
      if (status == AnimationStatus.completed) {
        controller.dispose();
      } else if (status == AnimationStatus.dismissed) {
        controller.dispose();
      }
    });

    controller.forward();
  }
```

PR #1011, opened 2021-09-02
Added centerZoomFitBounds to MapController

tlueder:
This PR adds a method to calculate the center of a given map bound.
Which is needed for animated camera transitions from one bound to the next.

This should help with #965


kengu (approved):

PR #1022, opened 2021-09-10
Prepared for release of 0.14.0

### LINKED ISSUE #1021: Release 0.14.0
@johnpryan - Can you release a new version to pub.dev?

kengu:
Closed #1021

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,19 @@
+## [0.14.0] - 6/7/2021
+This version contains the following changes
+- Added scroll wheel zoom support for web
+- Added TapPosition to TapCallback
+- Added center to LatLngBounds
+- Added equality operators for LatLngBounds and MapPosition
+- Added support for resetting TileLayer cache
+- Added attribution builder to TileLayer
+- Added 'inside' parameter to FitBoundsOptions
+- Added centerZoomFitBounds to MapController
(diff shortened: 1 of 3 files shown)


JaffaKetchup (changes requested):
Some pub.dev dependencies can be updated, including but not limited to `async` and `http`. If you're publishing a major version, it would be nice for the dependencies to be updated?

Also resolving major bug #1002 by #1003 would be good.

kengu:
@JaffaKetchup Updated dependencies now. I'm a bit worried about #1002, #1003 and #989. All of them seems a bit hairy. In the case of #989, I think the correct thing to do would is to refactor how the MapState is bootstrapped to remove the async nature of member initialisations. I short, I think we should release what we have, and do a 0.14.1 bugfix release later on. There is a lot of good stuff that shouldn't be held back anymore.

JaffaKetchup (approved):
OK, I understand that, and I agree. These feature additions are really cool and who knows how long the aforementioned PR's will take.

tlueder:
Hi,

https://github.com/fleaflet/flutter_map/issues/1002 Should be fixed before the release.

As this is a real blocker for me. And it seem that it is a bug from the null safety migration.
Because moveAndRotate uses the id only to call move and rotate, which both don't require a non null id.
Only moveAndRotate does.

johnpryan (approved):

PR #1120, opened 2022-01-09
Fix examples ZoomButtonsPluginOption to check min and max properly

chriscant:
In this example, when zooming in or out, the test against the int min and max zoom fails as the cast "as double" fails.
In addition the min and max tests are the wrong way round.
Finally, if the zoom hits a min or max, the min or max zoom isn't applied.
The fix includes changing the type of minZoom and maxZoom to double which does not effect other code that references it.


JaffaKetchup (approved):
LGTM

JaffaKetchup:
@chriscant Please pull from upstream, then this is ready for merge. Thanks!

PR #1132, opened 2022-01-17
Encode boolean values as uppercase string in WMS request.

stou:
Handle WMS server that only accepts boolean values as uppercase strings.

### DIFF
--- a/lib/src/layer/tile_layer.dart
+++ b/lib/src/layer/tile_layer.dart
@@ -349,6 +349,9 @@ class WMSTileLayerOptions {
   /// tile transparency flag
   final bool transparent;
 
+  /// Encode boolean values as uppercase in request
+  final bool uppercaseBoolValue;
+
   // TODO find a way to implicit pass of current map [Crs]
   final Crs crs;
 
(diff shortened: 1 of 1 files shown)


ibrierley (approved):
Feels ok to me.

JaffaKetchup:
Forgive me if I'm wrong, but couldn't you just pass in the boolean string as uppercase anyway?

ibrierley:
I think currently it would have to be a bool rather than a string, and hence toString would be lowercase if I'm not mistaken (quite possible).

lib/src/layer/tile_layer.dart
```diff
@@ -349,6 +349,9 @@ class WMSTileLayerOptions {
   /// tile transparency flag
   final bool transparent;
```
JaffaKetchup, line 350: @ibrierley @stou Ah yes, I see now :)

JaffaKetchup (approved):
LGTM. Just waiting on maintainers to return or people to be given write access. See #1117.

JaffaKetchup:
@stou Please pull upstream when possible, then I (or another maintainer) can merge.

stou:
@JaffaKetchup I have merged in the latest. Looks like the analysis and tests passes now :-)

PR #1135, opened 2022-01-19
Fix EPSG4326 parameter difference with Leaflet.js

yeleibo:
image


ibrierley:
Can you describe what the difference this may make to users that they would notice when testing or using ?

yeleibo:
when map option crs set with EPSG4326,the map load error

JaffaKetchup:
@yeleibo Can you please pull upstream, so I can re-run checks and review (and merge)? Thanks,

yeleibo:
> @yeleibo你能拉到上游，所以我可以重新运行检查和审查（和合并）？谢谢，
 
yes,I already pull upstream！

JaffaKetchup (approved):
LGTM, new example page seems to work. Bit of an odd projection, never used it before, but I guess this has fixed something. Will merge now...

PR #992, opened 2021-08-06
Documentation Improvements

JaffaKetchup:
As discussed in thread https://github.com/fleaflet/flutter_map/discussions/927#discussioncomment-1002834, this PR will merge a whole new documentation for `flutter_map`, including:

 - Improved out-of-code documentation
 - Improved 'dartdoc' (inside code) documentation

PR #1143, opened 2022-01-29
Add specs for utility functions

comatory:


JaffaKetchup:
Looks like the first two commits are also present in #1144.

JaffaKetchup (approved):
Changes to internal documentation. Changes to tests. LGTM.

Not sure about what will happen in regards to the first two commits also being present in #1144, which also needs to be merged? @comatory or @johnpryan or any other maintainer, do you have an idea?

comatory:
It doesn't matter which one we merge first, I can rebase & deal with conflicts and push again.

JaffaKetchup:
Ok, will merge this first.

Note, this is also present in the 'Documentation Improvements' (#992), but is being merged here now as it does not necessarily need to wait for the rest of the improvements.

comatory:
On Sat Jan 29, 2022 at 4:09 PM CET, Luka S wrote:
> Ok, will merge this first.
>
> Note, this is also present in the 'Documentation Improvements' (#992), but is being merged here now as it does not necessarily need to wait for the rest of the improvements.
>

Should be OK I think, if this gets to `master`, the other PR's branch
will get updated (either via web interface or locally with git).

Unless there would be some conflicts.

PR #1147, opened 2022-02-02
Added option to use pxCache and also length check on Markers to avoid crash

ibrierley:
Don't merge this without testing/checking. I'd like some thoughts first on whether we want an option to turn caching off, and if using map.var rather than widget.map.var is preferred or not. 

There's recently been a crash on MarkerLayer where the pxCache and the Markers didn't match up, after adding a marker after a tap.
This is a suggestion that looks to prevent the crash from happening, making sure we update if the lengths mismatch. Also added the option to turn off pxCaching as it "could" get out of date I think. Default is on.

There may be a better solution.

### DIFF
--- a/lib/src/layer/marker_layer.dart
+++ b/lib/src/layer/marker_layer.dart
@@ -8,6 +8,10 @@ import 'package:latlong2/latlong.dart';
 class MarkerLayerOptions extends LayerOptions {
   final List<Marker> markers;
 
+  /// Do we want to cache projections for performance, caching can
+  /// introduce inaccuracies if the cache is out of date
+  final bool usePxCache;
+
   /// If true markers will be counter rotated to the map rotation
   final bool? rotate;
 
@@ -38,6 +42,7 @@ class MarkerLayerOptions extends LayerOptions {
     this.rotate = false,
     this.rotateOrigin,
     this.rotateAlignment = Alignment.center,
+    this.usePxCache = true,
     Stream<Null>? rebuild,
   }) : super(key: key, rebuild: rebuild);
 }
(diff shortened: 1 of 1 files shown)


ibrierley:
I'm happier with this now, as I think it does help fix crashes when adding markers and the cache hasn't updated. Would be good to get a 2nd glance from someone.

JaffaKetchup:
Will see what I can do in the way of testing.

lib/src/layer/marker_layer.dart
```diff
@@ -8,6 +8,10 @@ import 'package:latlong2/latlong.dart';
 class MarkerLayerOptions extends LayerOptions {
   final List<Marker> markers;
 
+  /// Do we want to cache projections for performance, caching can
+  /// introduce inaccuracies if the cache is out of date
```
JaffaKetchup, line 12: Consider using more 'direct' documentation?

```suggestion
  /// Toggle marker position caching. Enabling will improve performance, but may introduce
  /// errors when adding/removing markers. Default is enabled (`true`).
```

lib/src/layer/marker_layer.dart
```diff
@@ -191,10 +196,30 @@ class _MarkerLayerState extends State<MarkerLayer> {
   var _pxCache = <CustomPoint>[];
 
   // Calling this every time markerOpts change should guarantee proper length
```
JaffaKetchup, line 198: Should this be a documentation comment (use '///')?

```suggestion
  /// Calling this every time markerOpts change should guarantee proper length
```

JaffaKetchup (changes requested):
Can't see anything glaringly wrong with the code (haven't tested it yet). Just a few documentation tweaks may be necessary?

Before merging, needs to be correctly formatted to pass the automated checks, using `dart format --fix`.

ibrierley:
Thanks, good points! I'm just wondering if it would be more correct (at the expense of optimisation), to store an array of the previous set of markers or similar), check each one to see if its different, if so, update (maybe with an option of cache on/off).

If we have a list of X markers (lets say they are taxis moving around a city) where the number of markers doesn't change, I'm not sure the markers would still read an old cached value and be wrong. This fixes additions to marker lists, but not mutations...

Or is there a neat way to know if a List has been modified without having to traverse. I suspect this may be microoptimisation though, as I think the original problem was not having to recalculate the projections which is heavier.

JaffaKetchup:
That is an interesting point. I'd probably merge this first and look at that later though, because (as you said) that is future optimisation.

ibrierley:
Yes, I think I agree. I'm just tied up today, but should get chance to redo the bits you suggested over the weekend. I may also just do a bit of benchmarking with a new option, see how much added time checking the whole list would be vs a length check or redoing all projections.

PR #1167, opened 2022-02-19
Treat layers as internal plugins and refactor code

grahamsmith:
This PR moves the implementation of layers e.g. TileLayer, CircleLayer to be treated as plugins. The internal original implementation remains unchanged.

Benefits
- "Dog fooding" the Plugin implementation
- Removes lots of IF's when dealing with layers, now we just iterate over them
- 3rd party plugins should work with group layers

I have kept the same API in MapPlugin however it would have been good if it took a generic of `<T extends LayerOptions>` to avoid later casting... but doing so would mean if this was accepted then all plugins would need to perform an update, which seems a lot of overhead. This change can still be made but would need some community coordination etc.

### DIFF
--- a/lib/src/plugins/internal/group_layer.dart
+++ b/lib/src/plugins/internal/group_layer.dart
@@ -2,6 +2,17 @@ import 'package:flutter/widgets.dart';
 import 'package:flutter_map/flutter_map.dart';
 import 'package:flutter_map/src/map/map.dart';
 
+class GroupLayerPlugin extends MapPlugin {
+  @override
+  Widget createLayer(
+      LayerOptions options, MapState mapState, Stream<Null> stream) {
+    return GroupLayer(options as GroupLayerOptions, mapState, stream);
+  }
+
+  @override
+  bool supportsLayer(LayerOptions options) => options is GroupLayerOptions;
+}
+
 /// [LayerOptions] that describe a layer composed by multiple built-in layers.
 class GroupLayerOptions extends LayerOptions {
   List<LayerOptions> group = <LayerOptions>[];
@@ -27,10 +38,11 @@ class GroupLayerWidget extends StatelessWidget {
 
 class GroupLayer extends StatelessWidget {
   final GroupLayerOptions groupOpts;
-  final MapState map;
+  final MapState mapState;
   final Stream<Null> stream;
 
-  GroupLayer(this.groupOpts, this.map, this.stream) : super(key: groupOpts.key);
+  GroupLayer(this.groupOpts, this.mapState, this.stream)
+      : super(key: groupOpts.key);
 
   @override
   Widget build(BuildContext context) {
@@ -38,7 +50,8 @@ class GroupLayer extends StatelessWidget {
       stream: stream,
       builder: (BuildContext context, _) {
         var layers = <Widget>[
-          for (var options in groupOpts.group) _createLayer(options)
+          for (var options in groupOpts.group)
+            _createLayer(options, mapState.options.plugins)
         ];
 
         return Container(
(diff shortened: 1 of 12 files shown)


lib/src/plugins/internal/group_layer.dart
```diff
@@ -50,21 +58,13 @@ class GroupLayer extends StatelessWidget {
...
-    if (options is PolylineLayerOptions) {
-      return PolylineLayer(options, map, options.rebuild);
-    }
-    if (options is PolygonLayerOptions) {
-      return PolygonLayer(options, map, options.rebuild);
-    }
-    if (options is OverlayImageLayerOptions) {
-      return OverlayImageLayer(options, map, options.rebuild);
+  Widget _createLayer(LayerOptions options, List<MapPlugin> plugins) {
+    final stream = options.rebuild ?? Stream.empty();
```
grahamsmith, line 62: This might be a smell but without refactoring much wider it seemed the least disruptive option.

grahamsmith: The MapPlugin contract states that a layer is promised a non null stream however the internal layers take an optional one. Seems like a good future piece of work to align this (hopefully to not null) and then remove any checks or force unwraps.

grahamsmith:
I've fixed the formatting "errors" - feel free to give it another run

ibrierley:
Just at a quick glance..you've been busy :D. 

I feel like this should be split into separate PRs, one for formatting, another for Plugins ? It's kind of interesting as I was mentioning the idea of moving layers to plugins yesterday (even pondering external only for a future iteration).

Just double checking, I think your code would make external Plugins preferred to internal plugins (i.e if someone wants to override a plugin it would work?).

Are there any breaking changes with this ?

Some of the formatting I prefer visually, but others may disagree (not sure if dartfmt has anything to say about it if run through?), I'm unsure what any original decisions were made on this (I could care less personally, just aware if it affects others, or overall code style).

grahamsmith:
**I feel like this should be split into separate PRs, one for formatting, another for Plugins ?**
- I have reverted the formatting as I run at 120 chars over 80.

 **It's kind of interesting as I was mentioning the idea of moving layers to plugins yesterday (even pondering external only for a future iteration).**
- Great minds etc
- Can make that the next move using these changes, however it'll need some comms. It's quite nice having some basic ones built in :+1:
 

**Just double checking, I think your code would make external Plugins preferred to internal plugins (i.e if someone wants to override a plugin it would work?).**

Ah good point. We could just default to the list of internal ones if a list of plugins is not specified and then if it is people can enable the ones they want with changes or extensions? 

**Are there any breaking changes with this ?**
- I purposefully kept the same API with MapPlugin even though changes would make all this better. See original comments.

JaffaKetchup:
> I purposefully kept the same API with MapPlugin even though changes would make all this better. See original comments.

Maybe we should be doing breaking changes, if it can make things less convoluted. Just my thoughts.

ibrierley:
I don't have a problem with breaking changes if needed for different versions I think. Just needs a little extra thought is all.

grahamsmith:
-why not both gif-

Big believer of incremental change. If the changes are positive then awesome I'd vote lgtm.

Happy to make the next PR too where there can be breaking changes to make it better.

ibrierley:
Took a quick test on this...I get the error

```
 Bad state: Can't find correct layer for Instance of 'TileLayerOptions'. Perhaps when you create your FlutterMap you need something like this:
I/flutter (31567):     options: new MapOptions(plugins: [MyFlutterMapPlugin()])
I/flutter (31567): #0      FlutterMapState._createLayer (package:flutter_map/src/map/flutter_map_state.dart:199)
I/flutter (31567): #1      FlutterMapState._buildMap.<anonymous closure> (package:flutter_map/src/map/flutter_map_state.dart:172)
```

When trying the Home example in examples folder. I tried also adding TileLayerPlugin() to the plugins list, but didn't make any difference. So I'm a bit unclear on this. I'm guessing related to your earlier point about community coordination, all of the examples would need to be rewritten also, and would probably break others code...so can't really merge this at the moment (unless I'm missing something!), but as a general concept I like the idea.

grahamsmith:
Weird, I ran this locally before opening the PR. I'll give it another blend tomorrow.

ibrierley:
In fact, just thinking out loud more....we can now use WidgetLayers, so in the future, if all plugins would have to change anyway, one can do...

```
class MarkerLayerWidget extends StatelessWidget {
  final MarkerLayerOptions options;

  MarkerLayerWidget({Key? key, required this.options}) : super(key: key);

  @override
    final mapState = MapState.maybeOf(context)!;
    return MarkerLayer(options, mapState, mapState.onMoved);
  }
}
```
so would there even be any point in plugins, just use Widget layers (I may be missing some other reason for using plugins I forget about). Just trying to think if there's a catch-22 behind redoing the layers, if everyone needed to rewrite them, there would be no need for that part of the code....

JaffaKetchup:
Sorry, bit out of the loop with this one. What's happening here?

grahamsmith:
I am going to close this and roll my own attempt to be honest.

Everyone's been lovely and engaging but feels like there's too many different opinions and my time with a baby on the way is limited. 👶

Hope to speak soon! 👍

JaffaKetchup:
No problem @grahamsmith, thanks for all your hard work!
Maybe we can take this somewhere in the future when all the bugs have been resolved and there's more time to focus on one thing.

PR #1168, opened 2022-02-19
Refactor MapState Widgets

grahamsmith:
A minor PR

I *think* we weren't using the OverflowBox and after testing I could not see a difference compared using a SizedBox as the min and max took the same values.

The ClipRect *seemed* redundant too.

I have ran the changes on the example in the project and could not spot the difference but its not to say there isn't one.

EDIT - also noticed the IF checks were redundant when creating the stack children. If the lists are empty then the combined list will be too. Same for the map, mapping on the empty list will just exit with an empty output.

### DIFF
--- a/lib/src/map/flutter_map_state.dart
+++ b/lib/src/map/flutter_map_state.dart
@@ -154,39 +154,32 @@ class FlutterMapState extends MapGestureMixin {
   }
 
   Widget _buildMap(var size) {
-    return ClipRect(
-      child: Stack(
-        children: [
-          OverflowBox(
-            minWidth: size.x as double?,
-            maxWidth: size.x as double?,
-            minHeight: size.y as double?,
-            maxHeight: size.y as double?,
-            child: Transform.rotate(
-              angle: mapState.rotationRad,
-              child: Stack(
-                children: [
-                  if (widget.children.isNotEmpty) ...widget.children,
-                  if (widget.layers.isNotEmpty)
-                    ...widget.layers.map(
-                      (layer) => _createLayer(layer, options.plugins),
-                    )
-                ],
-              ),
-            ),
-          ),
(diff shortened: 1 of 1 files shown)


ibrierley (commented):
I think this needs a bit of care...now I'm not sure exactly, but I think clipRect was introduced when rotated layers were introduced (as ever I may be wrong!). I'm not sure if that was for performance or function related reasons, but I think needs good testing around that area first to see if there's any implications...

grahamsmith:
How do I test this?

Could we add something to the example so we can check easily next time etc?

It would enhance the example with extra demo and at least give a place to check if future us breaks something. 🤔

ibrierley:
I think probably the first one is to try and gauge performance and any visual issues on rotated maps (maybe on different orientations), especially with polylines/gons. I think the default is having rotation on...it's a bit fiddly to have a specific test when not sure the exact reason it was added in the first place though and no code documentation about it's need...but it would need rotated tiles and polys involved I suspect...

@escamoteur I think this was originally added in your rotation code, not sure if you can recall the purpose of the cliprect here if still here and can recall ?

ibrierley:
Managed to do a bit of testing. This breaks being a rectangular clipped widget with inner rotated tiles, and I think messes with the rotation center. So if you try the example home page, you would find the rotated tiles overlap the outer text, and looks a bit odd.

escamoteur:
you ned the overflowbox so that when you rotate the edges aren't empty. then you need the ClipRect to mask the overflow when you aren't rotating. you could add an `if` that checks if rotation is a clean multiple of 90degrees and omit the Overflow and clip

ibrierley:
Thanks for the reply, that pretty much matches with what I saw when testing! I think it's best to close this for the time.

Issue #1156, opened 2022-02-08
Flicker grid between map tiles on zoom


tobias-ullerich-init:
I have recently experienced a flicker grid between the map tiles when zooming on a map with rotation disabled. This bug seems to be depended on the Flutter SDK version and present since Flutter SDK 2.8.x. I tested this with the 0.14.0 version of the library, as well as the master branch. Running the application on Flutter 2.5.x will not lead to this bug. The bug will only appear with disabled map rotation interaction.
The attached video shows the graphical bug more precise. There is a line from north to south near London and a line from east to west at the height of the Label "United Kingdom".

https://user-images.githubusercontent.com/81169686/152992764-b0ee3241-c334-4eac-8f86-05600d7d2c5a.mp4

To reproduce the bug, simply use this code and run the application with the Flutter SDK 2.8.x or 2.10.x: 

```dart
FlutterMap(
  options: MapOptions(
    center: LatLng(51.5, -0.09),
    zoom: 5.0,
    interactiveFlags: InteractiveFlag.drag |
        InteractiveFlag.flingAnimation |
        InteractiveFlag.pinchMove |
        InteractiveFlag.pinchZoom |
        InteractiveFlag.doubleTapZoom,
  ),
  layers: [
    TileLayerOptions(
      urlTemplate:
          'https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png',
      subdomains: ['a', 'b', 'c'],
      // For example purposes. It is recommended to use
      // TileProvider with a caching and retry strategy, like
      // NetworkTileProvider or CachedNetworkTileProvider
      tileProvider: NonCachingNetworkTileProvider(),
    ),
    MarkerLayerOptions(markers: markers)
  ],
)
```

ibrierley:
This sounds similar to another problem, where you get a hairline between times sometimes. They found that if they removed the Opacity widget the hairlines were fixed.

I'm wondering if there's a change in Flutter where there's a rounding error or something on the Opacity widget (iirc there were some changes to Opacity in the latest flutter ver). It would be interesting if you get time to test removing it (apologies, I'll be away for over a week, so won't have time to do any debugging).

JaffaKetchup:
Related to #1126 perhaps?

tobias-ullerich-init:
> This sounds similar to another problem, where you get a hairline between times sometimes. They found that if they removed the Opacity widget the hairlines were fixed.

[...]

Issue #1126, opened 2022-01-12
Tile boundaries visible at certain zoom levels


pmjobin:
I built the demo app using Flutter 2.8.1 (desktop linux target) and I noticed that tile boundaries pop out as white lines at certain zoom levels while they are seamless at others.

![Screenshot from 2022-01-12 15-35-32](https://user-images.githubusercontent.com/1387411/149209616-a19bc571-f627-46cc-beaf-5a9ea51015a2.png)

jiazeh:
Same issue, and this issue only happens on Flutter 2.8.x and above..

mohammedX6:
+1

DrNorton:
+1

ibrierley:
Does this happen on mobile too, or just on desktop ?

DrNorton:
> Does this happen on mobile too, or just on desktop ?

On mobile too, but only with disabled rotation

JaffaKetchup:
This seems strange, and would be probably Flutter 2.8's fault, because I haven't seen this before.
Maybe the changelog will hold the answers?

DrNorton:
I think that located problem, if I delete Opacity widget in _AnimatedTileState

**Before**:

`   return Opacity(
      opacity: widget.tile.opacity,
      child: widget.tileBuilder == null
          ? tileWidget
          : widget.tileBuilder!(context, tileWidget, widget.tile),
    );`

**After**:

` return widget.tileBuilder == null
        ? tileWidget
        : widget.tileBuilder!(context, tileWidget, widget.tile);`

The problem is gone. There are no white lines.

And there are many opacity issues that were merged in [release notes](https://docs.flutter.dev/development/tools/sdk/release-notes/release-notes-2.8.

ibrierley:
Good find, this url should work if anyone has problems with above url https://docs.flutter.dev/development/tools/sdk/release-notes/release-notes-2.8.0

JaffaKetchup:
Duplicate of #1156

mostafaac30:
same

PR #1169, opened 2022-02-20
possible fix for opacity widget causing hairlines and possible flicker

ibrierley:
This is a possible workaround for the flicker and hairline issues https://github.com/fleaflet/flutter_map/issues/1126 & https://github.com/fleaflet/flutter_map/issues/1156

It removes the Opacity widget (I'm not sure this is going to get fixed easily by Flutter, so think we will have to create our own), and introduces the opacity to the RawImage instead.

This feels possibly logical, however, I'm not confident on the Animated Tile stuff, so this may not be exactly the same.... I assume the opacity at this position is for fading in new tilesets/zooms ? If anyone knows this part of the code, it would be useful to give some thoughts, as I'm a little unsure how to test it properly.

It would also be useful for those with the initial problems to see if it helps.

Rotation needs to be disabled for the original hairline type issues to appear, i.e interactiveFlags: ~InteractiveFlag.rotate in MapOptions

### DIFF
--- a/lib/src/layer/tile_layer.dart
+++ b/lib/src/layer/tile_layer.dart
@@ -1381,14 +1381,11 @@ class _AnimatedTileState extends State<AnimatedTile> {
         : RawImage(
             image: widget.tile.imageInfo?.image,
             fit: BoxFit.fill,
-          );
+            opacity: widget.tile.animationController);
 
-    return Opacity(
-      opacity: widget.tile.opacity,
-      child: widget.tileBuilder == null
-          ? tileWidget
-          : widget.tileBuilder!(context, tileWidget, widget.tile),
-    );
+    return widget.tileBuilder == null
+        ? tileWidget
+        : widget.tileBuilder!(context, tileWidget, widget.tile);
   }
 
   @override


JaffaKetchup:
I'll try to test soon

JaffaKetchup (commented):
@ibrierley
Unfortunately I can't test this. I've tried running the latest 'master' branch example app on the Home page, and I just can't see the white lines that I've seen before. Therefore I wouldn't be able to know if this has done anything?

Maybe this has fixed itself in Flutter updates? Alternatively (and I'm not sure why this would make a difference), I had to update the `compileSdkVersion` & `targetSdkVersion` to '31', and the `ext.kotlin_version` to '1.6.20'?

Can you try on your side, to see if you still get the issue?

ibrierley:
Did you try with rotation disabled ? i.e  interactiveFlags: InteractiveFlag.all & ~InteractiveFlag.rotate in MapOptions.

JaffaKetchup:
Ah no probably not - I just used the main example. My mistake: I'll try again tomorrow :)

stx:
I've tested this and it works great. Nice work @ibrierley!

JaffaKetchup (approved):
LGTM. Seems to fix the issue, but not sure what knock on impacts this might have. Will merge now...

PR #1170, opened 2022-02-21
Fix FileTileProvider to work on the web

mo-ah-dawood:
This pr create two files `file_tile_provider_io.dart` and `file_tile_provider_web.dart`
 and export the right one

### DIFF
--- a/example/ios/Podfile
+++ b/example/ios/Podfile
@@ -1,0 +1,41 @@
+# Uncomment this line to define a global platform for your project
+# platform :ios, '9.0'
+
+# CocoaPods analytics sends network stats synchronously affecting flutter build latency.
+ENV['COCOAPODS_DISABLE_STATS'] = 'true'
+
+project 'Runner', {
+  'Debug' => :debug,
+  'Profile' => :release,
+  'Release' => :release,
+}
+
+def flutter_root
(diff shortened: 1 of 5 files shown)


JaffaKetchup:
Thanks for this - I saw your note on the other PR. Just running checks now, should be able to review tomorrow or soon.

JaffaKetchup (approved):
LGTM. Haven't fully tested, but it doesn't look like it should cause any problems. If it's not too much of a hassle, can you remove the 'Podfile', as it's unrelated to this commit (I think)? Otherwise I'm happy to merge.

ibrierley (commented):
I'm a bit undecided on this one, as not enough experience on the flutter web side to know what's best. It doesn't feel like the Podfile is needed as previously mentioned, so I think that could go.

Regarding the other bits, something "feels" odd about having a FileTileProvider that uses a NetworkImage...but I can understand why the web makes this a bit messy. Wouldn't one simply not use it, or does it create an error when building for the web (again, I'm not so familiar with that)?

mo-ah-dawood:
The browser can not access file system directly it can accept blobs

so in flutter web you have to load the image using `NetworkImage` instead of `FileImage`

ibrierley:
All good on Android and Web for me.

PR #1181, opened 2022-03-09
improve performance

aitinsan:


ibrierley (commented):
Out of interest, is the moveTo still needed if using a polygon ?

JaffaKetchup:
@aitinsan
Thanks for your contribution! Performance improvements are certainly welcome at the moment.

Please consider changing the title and body of your PR to include a little more specific information, and maybe the statistics we can expect to see?

I will try to review soon, but I'm not confident with this part of the library, so I'm not sure of any side effects that might happen.

ibrierley:
I think the logic of using an unclosed poly makes sense, I've done that before ok. It's worth checking things like dotted lines, gradient colors, strokeCaps etc just to make sure they still work the same visually, but I think they should off the top of my head.

aitinsan:
I am using polylines and that think poorly improves performance but, can improve. Working more on it

PR #1182, opened 2022-03-10
trigger event on scroll wheel

a14n:
This PR adds a `MapEventScrollWheelZoom` event that is fired when a scroll wheel occurs. Currently no event is triggered on scroll wheel (and zoom change)

### DIFF
--- a/lib/src/gestures/gestures.dart
+++ b/lib/src/gestures/gestures.dart
@@ -37,7 +37,8 @@ abstract class MapGestureMixin extends State<FlutterMap>
             : max(mapState.options.minZoom ?? 0, mapState.zoom + delta);
         final newZoom = mapState.fitZoomToBounds(zoom);
         // Move the map to the new zoom level
-        mapState.move(mapState.center, newZoom, source: MapEventSource.custom);
+        mapState.move(mapState.center, newZoom,
+            source: MapEventSource.scrollWheel);
       }
     }
   }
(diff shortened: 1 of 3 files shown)


JaffaKetchup (approved):
LGTM - haven't tested, but only small changes and doesn't appear to have any breaking consequences.

ibrierley (commented):
I think this looks good looking at the code, but I won't have chance to test it in the flesh until probably next weekend, but I'm happy to merge if someone else has.

JaffaKetchup:
@a14n @pmjobin How will #1182 and #1191 work together? Will they interfere?

a14n:
I don't expect any issue between the 2 PR (except the merge conflict because there's a line that is changed on the 2 PR)

ibrierley:
happy to merge this I think, have tested it doesn't seem to break anything obvious...noting the question about PR 1191, not so familiar with merge conflicts (it currently says there aren't any...), does it matter which order they are merged in or anything ?

JaffaKetchup:
@ibrierley It does depend on whichever gets merged first will cause a conflict in the other.

JaffaKetchup:
Testing now...

JaffaKetchup (approved):
LGTM, works as expected when tested. Will merge this before #1191.

JaffaKetchup:
Many thanks for your contribution @a14n.

PR #1183, opened 2022-03-10
migrate from pedantic to flutter_lints

a14n:


ibrierley:
Interesting, this was far off my radar. I don't really know much on this side, but looking at https://stackoverflow.com/questions/67734486/pedantic-vs-flutter-lint-which-package-to-use-and-can-they-also-be-combined it says pedantic is deprecated.
Thanks for the work on this @a14n are there any consequences you can think of, is anything flutter version related for example that could crop up ?

a14n:
The semantics is almost the same and I don't think it will cause real issue. The only place where I change some code is around `MapController`. I added some methods to allow `FlutterMapState` to deal with `MapController` and not directly `MapControllerImpl`. But I do think there was a real problem about how `MapController` is handled (but I am not going to address that in this PR to avoid too complicated review)

lib/flutter_map.dart
```diff
@@ -137,6 +137,9 @@ abstract class MapController {
   double get rotation;
 
   Stream<MapEvent> get mapEventStream;
+  StreamSink<MapEvent> get mapEventSink;
+  set state(MapState state);
+  void dispose();
```
a14n, line 142: Here are the 3 added methods.

JaffaKetchup:
Thanks for your contribution!
To appease the checks, can you run: `dart format .` in the root? Many thanks again.

example/lib/pages/tile_loading_error_handle.dart
```diff
@@ -50,7 +50,7 @@ class _TileLoadingErrorHandleState extends State<TileLoadingErrorHandle> {
                       tileProvider: const NonCachingNetworkTileProvider(),
                       errorTileCallback: (Tile tile, error) {
                         if (_needLoadingError) {
-                          WidgetsBinding.instance.addPostFrameCallback((_) {
+                          WidgetsBinding.instance?.addPostFrameCallback((_) {
```
a14n, line 53: on master branch of flutter `WidgetsBinding.instance` is now not nullable (that's why this error occured)

pablojimpas:
Definitely a good move to migrate from `pedantic` as it is deprecated.
However, I will suggest choosing a stricter set of rules like the ones from the [`lint`](https://github.com/passsy/dart-lint) or [`very_good_analysis`](https://github.com/VeryGoodOpenSource/very_good_analysis) packages.

Here is a detailed comparison of linting packages https://rydmike.com/blog_flutter_linting.html

In the past I've done some tinkering trying to fix all "errors" that appear in this package using `very_good_analysis` and there are some interesting ones worth fixing, those regarding to avoiding dynamic calls or dynamic typing in general can have impact in performance for example.

a14n:
@pablojimpas I have no problem with enabling more lints but choosing the right set of rules is not the subject of this PR IMHO. For now I rely on the most consensual set for flutter packages (btw as author of several of the dartlang lints there are definitelly some of them I'll never enable on my code). Note that `flutter_lints` is constantly evolving and that more lints will be enable in the future.

pablojimpas:
> @pablojimpas I have no problem with enabling more lints but choosing the right set of rules is not the subject of this PR IMHO. For now I rely on the most consensual set for flutter packages (btw as author of several of the dartlang lints there are definitelly some of them I'll never enable on my code). Note that `flutter_lints` is constantly evolving and that more lints will be enable in the future.

Okey I see your point, I am fine with the choice of `flutter_lints` as it is the most used one. Later on we could evaluate if it is worth to enable some particular extra rule for this project.

Also, I would keep an eye on https://github.com/flutter/packages/pull/1165

JaffaKetchup:
Will try to test later today. Note it might take longer than usual, as there are a lot of changes.

a14n:
ping

JaffaKetchup:
Sorry for the long delay. Will try to handle this soon :)

ibrierley (approved):
Did a bit more testing on this. I like it, and appreciate the work. Happy for anyone else to do a bit more testing as well though first, as there's quite a few files, but most look straightforward.

JaffaKetchup:
@ibrierley if you're happy, don't worry about my review.
@a14n thanks for the great contribution, and sorry it took so long to review (I didnt even get round to it in the end!)

PR #1213, opened 2022-04-10
Update example data

Zverik:
Built the example app just now, and had to update some numbers. This pull request makes it compilable without errors or warnings.


JaffaKetchup:
Looks like it should work, as I've done similar things locally. Just need to test on my system to make sure.

JaffaKetchup (approved):
LGTM. Live location screen is still broken, but that's a separate issue (location permission not being requested at correct time due to asynchronous request). Will merge now...

PR #1216, opened 2022-04-14
Fix initial fit bounds

ibrierley:
Fix for problem where bounds sometimes breaks on an initial layoutBuilder that doesn't have a valid constraints size, and causes fixBounds to break. (sorry created a branch here, but didn't mean to!)

### DIFF
--- a/lib/src/map/flutter_map_state.dart
+++ b/lib/src/map/flutter_map_state.dart
@@ -77,14 +77,25 @@ class FlutterMapState extends MapGestureMixin {
   Widget build(BuildContext context) {
     _disposeStreamGroups();
     return LayoutBuilder(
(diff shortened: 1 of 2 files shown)


kengu (approved):
LGTM. Not tested locally thought

JaffaKetchup:
Will try to test today

JaffaKetchup:
Sorry I haven't been able to test today, I've been quite busy. I'll try to get it done ASAP.

JaffaKetchup (approved):
LGTM. Tested on web, but should be same on all platforms. Haven't used a broken test case from before, but this hasn't broken anything, and the code looks good.

As a side note, `maxBounds` has now completely broken, but this is not the cause of this.

Issue #1217, opened 2022-04-15
[BUG] Polylines lagging after upgrading from Flutter 2.5 to 2.10


MeixDev:
**Describe The Bug**
After upgrading a project from Flutter 2.5.2 to Flutter 2.10.4, Polylines induce a lot more lag on the map, especially on Android (it seems fine on iOS). I saw other issues already called for this, but were either closed or focusing on a different issue.
In our current usecase, we got a list of 764 lines, for a total of 4354 points. 
Zooming on the map obvisouly helps a bit, but isn't an acceptable solution. We do already use `polylineCulling: true` and didn't notice any significant amelioration.

**Expected Behavior**
To have the same amount of lag (which was a bit but not much when all the lines were displayed on-screen, and almost none with a bit of zooming in).

**Screenshots & Recordings**

Here is a video of the app in the 2.10.4 version of Flutter :
https://youtube.com/shorts/3O8CmaJ5cf4?feature=share

The same exact data, but this time in the 2.5.2 version of Flutter :
https://youtube.com/shorts/eyhn7JRGOtY?feature=share

**Additional Information**
The issue was also observed by my coworker @TesteurManiak. We tried to use both the pub.dev version and the git:master version of flutter_map, without any noticeable difference (Except the fact that the filling of polygons disappeared in the git version. Should we open another issue for that one ?)

---

**Doctors Report**

```shell
Doctor summary (to see all details, run flutter doctor -v):
[√] Flutter (Channel stable, 2.10.4, on Microsoft Windows [version 10.0.19044.1586], locale fr-FR)
[√] Android toolchain - develop for Android devices (Android SDK version 29.0.2)
[√] Chrome - develop for the web
[√] Visual Studio - develop for Windows (Visual Studio Professional 2019 16.11.3)
[√] Android Studio (version 2021.1)
[√] VS Code (version 1.66.2)
[√] Connected device (4 available)
[√] HTTP Host Availability

• No issues found!
```

**To Reproduce**
If necessary, it is possible for us to provide an extract of our List of Polylines as a json model, with a fromJson/toJson extension on the current model.

---

**Severity**
This will help us to label the issue quicker and decide what needs attention first. Only choose fatal if the entire app crashes, otherwise choose non-fatal.

- [x] Non-Fatal
- [ ] Fatal / App Crashes

**Frequency/Rarity**
This will help us to label the issue quicker and decide what needs attention first.

- [ ] Once
- [ ] Uncommon
- [ ] Common
[...]

Issue #1165, opened 2022-02-18
Tracking: Performance Issues


JaffaKetchup:
This issue aims to help track the multiple performance issues, so it's easier to handle them by linking them together, fixing them, or closing them.

 - [x] #1161
 - [x] #1154
 - [x] #958

There are likely to be more added in the future.

---

Fixed/ignored (closed) issues:

 - [x] #1107
 - [x] #1217
 - [x] #1245

grahamsmith:
Hello 👋 I am new to the repo so apologies if this is not the right place. Thanks for all the hard work thus far 👍 

I have been reading through the code and having a play and noticed there are some performance improvements that could be made. I then noticed this issue.

Unsure if you are accepting PRs but would be happy to help. For example I have noticed we work with `num` a lot when in practice we parse from `String` to `Double` and then treat is as a `num` in the middle and then cast it via `toDouble()` at the end. Whilst individually probably not that expensive it probably adds up over time. 

The WMS implementation looks like it has caused some ripple effects through the code too, for example when dealing with urlTemplate and then having to force unwrap it etc. I've seen notes in the code so I'm late to this party.

I think there's some general widget improvements (removing un-needed ones) and some newer language features to help out too.

Happy to help but didn't want to waste my time either if its not needed.

ibrierley:
Help is always wanted :), and all suggestions and PRs welcome ! So feel free and great to have others interested.

Whilst all optimisations welcome, I would also say as a general programming thing, I wouldn't get too hung up on microoptimisations and let it eat up your previous dev time, there are normally bigger fish to fry with performance.

mohammedX6:
I was experiencing very bad performance when I am showing some very basic lines (thickness only to 1), Then after researching I made this modification in the drawing method of the PolyLine layer

Instead of using `path.lineTo`
Use this :
```
//Assuming we have the offsets 
final path = ui.Path();
path.addPolygon(polyline.offsets, false); // Here is the magic
canvas.drawPath(path, paint);

polylines.add(CustomPaint(painter: PolylinePainter(polyline, false, false, path: path),size: size,)); 
```

ibrierley:
I think that makes sense, and if we were to rewrite it (or write a new plugin for polylines/gons then I'd be tempted to do some caching of the offsets (and likely paths) similar to the recent marker changes.

I'm a little surprised that its so different in performance though for basic lines, or are there a lot of them ?

mohammedX6:
[...]

PR #1219, opened 2022-04-19
refacto: added saveLayers parameter

TesteurManiak:
Solve #1217 

- Added saveLayers parameter to PolylineLayerOptions & PolylinePainter to enable or disable calls to canvas.saveLayer & canvas.restore (by default at `false` as advised in [ibrierley's comment](https://github.com/fleaflet/flutter_map/issues/1217#issuecomment-1101791862))
- Replaced calls to path.lineTo by path.addPolygon in PolylinePainter._paintLine (recommended here: https://github.com/fleaflet/flutter_map/issues/1165#issuecomment-1055342364)

### DIFF
--- a/lib/src/layer/polyline_layer.dart
+++ b/lib/src/layer/polyline_layer.dart
@@ -7,14 +7,23 @@ import 'package:flutter_map/src/map/map.dart';
 import 'package:latlong2/latlong.dart';
 
 class PolylineLayerOptions extends LayerOptions {
+  /// List of polylines to draw.
   final List<Polyline> polylines;
+
   final bool polylineCulling;
+
+  /// {@macro newPolylinePainter.saveLayers}
+  ///
+  /// By default, this is value is set to `false` to improve performance on
+  /// layers containing a lot of polylines.
+  final bool saveLayers;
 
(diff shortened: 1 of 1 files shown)


ibrierley:
Just thinking, I note the polygon layer also does a saveLayer...worth us doing the same there for consistency ?

TesteurManiak:
It depends, in my opinion you won't have the same performance issues with polygons as, most of the time, you will draw less polygons than polylines and you might need more often to manage opacity or gradients on polygons.

ibrierley (commented):
Tests fine for me...only thing I think would be an extra useful bit, is a comment to explain why someone would turn it on...eg turn this on if you get unwanted darker lines where they overlap, but there will be a performance hit.

TesteurManiak:
Yeah this is a good idea, I'm adding it.

JaffaKetchup:
I'll see what I can do to add this to the new documentation :)

ibrierley:
All good for me, fine for me to merge @JaffaKetchup or wanted to do some further testing ?

JaffaKetchup:
Its ok for you to merge if you're happy

PR #1225, opened 2022-05-05
Add polygon label placement options

beroso:
Saw [this PR](https://github.com/fleaflet/flutter_map/pull/1220) about centered labels for polygons today.

I think it can be useful to add an option to place the label using [Dart polylabel](https://github.com/beroso/dart_polylabel). It works well on both convex and concave polygons.

Example:

| Centroid (current approach)  | Polylabel |
| ------------- | ------------- |
| Screen Shot 2022-05-05 at 17 44 31 | Screen Shot 2022-05-05 at 17 44 07 |

### DIFF
--- a/lib/src/layer/label.dart
+++ b/lib/src/layer/label.dart
@@ -1,26 +1,30 @@
-import 'dart:math';
+import 'dart:math' as math;
 
+import 'package:collection/collection.dart';
 import 'package:flutter/material.dart';
+import 'package:flutter_map/plugin_api.dart';
+import 'package:polylabel/polylabel.dart';
 
 class Label {
   static void paintText(
     Canvas canvas,
     List<Offset> points,
     String? labelText,
-    TextStyle? labelStyle,
-  ) {
-    var maxDx = 0.0;
-    var minDx = double.infinity;
-    var dx = points.map((e) => e.dx).toList().fold<double>(0.0,
-            (previousValue, element) {
-          maxDx = max(maxDx, element);
-          minDx = min(minDx, element);
-          return previousValue + element;
-        }) /
-        points.length;
-    var dy = points.map((e) => e.dy).toList().fold<double>(
-            0.0, (previousValue, element) => previousValue + element) /
-        points.length;
+    TextStyle? labelStyle, {
+    PolygonLabelPlacement labelPlacement = PolygonLabelPlacement.centroid,
+  }) {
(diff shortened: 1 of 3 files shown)


JaffaKetchup:
Thanks for your PR! I'll have a proper look tomorrow :)
Just for now though, I can see that the enumerable names might need changing though, as they are a bit confusing.

lib/src/layer/label.dart
```diff
@@ -1,26 +1,30 @@
...
-          maxDx = max(maxDx, element);
-          minDx = min(minDx, element);
-          return previousValue + element;
-        }) /
-        points.length;
-    var dy = points.map((e) => e.dy).toList().fold<double>(
-            0.0, (previousValue, element) => previousValue + element) /
-        points.length;
+    TextStyle? labelStyle, {
+    PolygonLabelPlacement labelPlacement = PolygonLabelPlacement.centroid,
```
JaffaKetchup, line 14: I think we should change the default to this new method, but that's just my opinion.

lib/src/layer/polygon_layer.dart
```diff
@@ -26,6 +26,11 @@ class PolygonLayerOptions extends LayerOptions {
   }
 }
 
+enum PolygonLabelPlacement {
+  centroid,
+  polylabel,
+}
```
JaffaKetchup, line 32: I also think we might need to rename these options to make them clearer to other people. Perhaps:

```suggestion
enum PolygonLabelPlacement {
  squareCenter,
  polyCenter,
}
```

JaffaKetchup (changes requested):
This is working great! I've commented below on something(s) that might need changing, but I'm open to suggestions :)

ibrierley (approved):
I'm fine with this either way. It tests fine. It feels like each option in certain cases is preferable when I look at it visually, so no preference on which is default.

polylabel is maybe ambiguous as an parameter, but I guess it does what it says on the tin (uses whatever polylabel calculates, which could change and not be a center...). So no strong preference either way.

lib/src/layer/polygon_layer.dart
JaffaKetchup: Ok, @ibrierley made a good point on this one, so I won't change it...

JaffaKetchup (approved):
LGTM

Issue #1231, opened 2022-05-12
[BUG] Zoom change without center change does not emit move event if event source is custom


hschendel:
**Describe The Bug**
When the zoom is changed without a "real" move, and the event source is MapEventSource.custom, no move event is emitted by the map controller.

**Expected Behavior**
A zoom change changes the bounds and thus a move event is expected.

**To Reproduce**

To reproduce, we need to cause a zoom change with event source being custom.

On MacOS platform (most easy to reproduce just using pinch zoom, probably due to trackpad behaviour)

1. Instantiate a FlutterMap
2. Listen to mapEventStream
3. Pinch zoom out or in on the Mac's trackpad without moving
4. No event appears

Other platforms

1. Instantiate a FlutterMap
2. add the flutter_map_marker_cluster plugin
3. add enough markers to cause clustering depending on zoom
4. zoom out
5. tap a cluster to have the marker_cluster plugin zoom in (with event source being custom)
6. No event appears

---

**Severity**
This will help us to label the issue quicker and decide what needs attention first. Only choose fatal if the entire app crashes, otherwise choose non-fatal.

- [x] Non-Fatal
- [ ] Fatal / App Crashes

**Frequency/Rarity**
This will help us to label the issue quicker and decide what needs attention first.

- [ ] Once
- [x] Uncommon
- [ ] Common
- [ ] Always

**Applicable Platforms**
Only select those that you've tested on - one or more. If possible, test on a variety of platforms.

- [x] Android
- [ ] iOS
- [x] Web
- [ ] Windows
- [x] Others (beta platforms)

PR #1232, opened 2022-05-12
[BUG] Emit move event even when source is custom #1231

hschendel:
As outlined in #1231, and discussed in Discord on 2022-05-11, certain zoom changes do not cause a move event to be emitted. This appears to be a bug, as probably everything changing the bounds should cause such an event. I am using flutter_map together with the flutter_map_marker_cluster plugin which e.g. zooms in if you tap on a cluster and thus changes the bounds, and I need to be able to react to that.

Therefore, I propose the following change to MapState._handleMoveEmit():

If source is custom, and targetZoom or targetCenter are different than before, a move event is emitted.

Thank you for your time and for flutter_map in general 😃

### DIFF
--- a/example/lib/pages/tile_loading_error_handle.dart
+++ b/example/lib/pages/tile_loading_error_handle.dart
@@ -50,7 +50,7 @@ class _TileLoadingErrorHandleState extends State<TileLoadingErrorHandle> {
                       tileProvider: const NonCachingNetworkTileProvider(),
                       errorTileCallback: (Tile tile, error) {
                         if (_needLoadingError) {
-                          WidgetsBinding.instance?.addPostFrameCallback((_) {
+                          WidgetsBinding.instance.addPostFrameCallback((_) {
                             ScaffoldMessenger.of(context).showSnackBar(SnackBar(
                               duration: const Duration(seconds: 1),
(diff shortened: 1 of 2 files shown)


ibrierley (commented):
As an aside, that whole area of code (all the if statements), just makes me ponder why it wasn't written to just emit an event with a source and cut out loads of lines, but it would probably break usage to change all of that now. So I don't think I have a problem with this, it shouldn't affect anything else really.

hschendel:
I also tried out removing the condition that source must be custom, so that the event would be emitted for all event sources not covered before, whenever zoom or center changes. That broke the flutter_map test, so I removed it.

JaffaKetchup (approved):
LGTM as a fix, but maybe needs to be looked at in the long-term future. However, the analyser needs to be appeased before I can merge :)

hschendel:
I fixed the thing found by dart analyze in the example, so let's see how the pipeline likes this ;-)

JaffaKetchup:
Thanks, that seems to have fixed it. Will merge now :)

Many thanks for your contribution!

hschendel:
Thank you @ibrierley and @JaffaKetchup :)

PR #1237, opened 2022-05-15
Remove unused dependencies

pablojimpas:
I've notice that there are a couple of dependencies that aren't really being used.


JaffaKetchup (approved):
LGTM. Doesn't need to be tested as I can see no import statements have changed and the checks still pass.
Thanks for your contribution!

PR #1238, opened 2022-05-15
Enable stricter linter rules.

pablojimpas:
Mostly trivial code changes made by running `dart fix --apply`

The new most important rules enabled are:
  - `prefer_final_locals`: https://dart-lang.github.io/linter/lints/prefer_final_locals.html
  - `prefer_final_in_for_each`: https://dart-lang.github.io/linter/lints/prefer_final_in_for_each.html
  - `avoid_dynamic_calls`: https://dart-lang.github.io/linter/lints/avoid_dynamic_calls.html

All three rules take advantage of relatively recent improvements in the Dart SDK and thus "modernize" the codebase (eg. getting rid of all those `var`), they can potentially have a positive impact on performance as described in the reference links while also ensuring correctness long term.

I've also re-enabled `library_private_types_in_public_api` that is present in `flutter_lints` and was overridden locally.

EDIT:
  - I've now promoted `missing_required_param` and `missing_return` to errors in the analyzer and also enabled the "strong-mode" to avoid weak implicit casts and implicit dynamic.
  - A more cosmetic choice of preferring integer literals and using named constants to avoid copying.

### DIFF
--- a/lib/src/geo/crs/crs.dart
+++ b/lib/src/geo/crs/crs.dart
@@ -104,7 +104,7 @@ abstract class Earth extends Crs {
   bool get infinite => false;
 
   @override
-  final Tuple2<double, double> wrapLng = const Tuple2(-180.0, 180.0);
+  final Tuple2<double, double> wrapLng = const Tuple2(-180, 180);
 
   @override
   final Tuple2<double, double>? wrapLat = null;
--- a/lib/src/geo/crs/crs.dart
+++ b/lib/src/geo/crs/crs.dart
@@ -25,8 +25,8 @@ abstract class Crs {
   /// map point.
   CustomPoint latLngToPoint(LatLng latlng, double zoom) {
     try {
-      var projectedPoint = projection.project(latlng);
-      var scale = this.scale(zoom);
+      final projectedPoint = projection.project(latlng);
+      final scale = this.scale(zoom);
       return transformation.transform(projectedPoint, scale.toDouble());
     } catch (e) {
       return const CustomPoint(0.0, 0.0);
@@ -35,8 +35,8 @@ abstract class Crs {
 
   /// Converts a map point to the sphere coordinate (at a certain zoom).
   LatLng? pointToLatLng(CustomPoint point, double zoom) {
-    var scale = this.scale(zoom);
-    var untransformedPoint =
+    final scale = this.scale(zoom);
+    final untransformedPoint =
         transformation.untransform(point, scale.toDouble());
     try {
       return projection.unproject(untransformedPoint);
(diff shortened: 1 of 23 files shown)


pablojimpas:
I've also enabled `always_use_package_imports` since there were only a couple of occurrences of relative imports, this way we enforce consistency on one style going forward.

While I was looking at the imports I've also notice that `lib/flutter_map.dart` was importing itself (because exports a lot of things) instead of the individual dependencies, that caused a bunch of warnings on pub.dev that lowered the package score: 

image

ibrierley (approved):
Looks fine to me, and happy for it to be merged if ok with it. Thanks for the work.

lib/src/geo/crs/crs.dart
```diff
@@ -104,7 +104,7 @@ abstract class Earth extends Crs {
   bool get infinite => false;
 
   @override
-  final Tuple2<double, double> wrapLng = const Tuple2(-180.0, 180.0);
+  final Tuple2<double, double> wrapLng = const Tuple2(-180, 180);
```
JaffaKetchup, line 107: This should be of type <int, int> shouldn't it? And I'm not sure how it would work but maybe final and const isn't needed - haven't suggested this though as I'm not 100% sure.

```suggestion
  final Tuple2<int, int> wrapLng = const Tuple2(-180, 180);
```

lib/src/gestures/gestures.dart
```diff
@@ -12,7 +12,7 @@ import 'package:positioned_tap_detector_2/positioned_tap_detector_2.dart';
 
 abstract class MapGestureMixin extends State<FlutterMap>
     with TickerProviderStateMixin {
-  static const double _kMinFlingVelocity = 800.0;
+  static const double _kMinFlingVelocity = 800;
```
JaffaKetchup, line 15: Same here?


```suggestion
  static const int _kMinFlingVelocity = 800;
```

JaffaKetchup (changes requested):
Mostly looks good. A couple of things to maybe change though, suggested below. Happy to approve or change suggestions if you disagree! Many thanks, it's already looking much better.

lib/src/geo/crs/crs.dart
pablojimpas: I think this is wrong, they should be doubles because at the end LatLng still takes doubles. It's just a "cosmetic" change of not having unnecessary .0 enforce by the rule `prefer_int_literals`.

lib/src/gestures/gestures.dart
pablojimpas: In this case it's fine for this to be an int since it's only used in one place as a condition check.

JaffaKetchup (approved):
In that case, I'm happy to merge as well. Many thanks for your contributions!

Issue #1244, opened 2022-05-22
[BUG] placeholderImage doesn't work


Farrukh1198:
**Describe The Bug**
I want to show my custom tile image while tile is loading. For this purpose I use placeholderImage option of TileLayerOptions but I still get grey tiles instead of my image.

**Expected Behavior**
I expected my custom tiles to be shown while map tiles are loading

**Screenshots & Recordings**
I think it's clear without screenshots

---

**Doctors Report**

```shell
Doctor summary (to see all details, run flutter doctor -v):
[√] Flutter (Channel stable, 3.0.0, on Microsoft Windows [Version 10.0.14393], locale ru-RU)
[!] Android toolchain - develop for Android devices (Android SDK version 32.1.0-rc1)
    X cmdline-tools component is missing
      Run `path/to/sdkmanager --install "cmdline-tools;latest"`
      See https://developer.android.com/studio/command-line for more details.
    X Android license status unknown.
      Run `flutter doctor --android-licenses` to accept the SDK licenses.
      See https://flutter.dev/docs/get-started/install/windows#android-setup for more details.
[√] Chrome - develop for the web
[X] Visual Studio - develop for Windows
    X Visual Studio not installed; this is necessary for Windows development.
      Download at https://visualstudio.microsoft.com/downloads/.
      Please install the "Desktop development with C++" workload, including all of its default components
[√] Android Studio (version 2021.1)
[√] VS Code (version 1.67.2)
[√] Connected device (3 available)
[√] HTTP Host Availability
```

**To Reproduce**

My code is 
```
                TileLayerOptions(
                  updateInterval: 100,
                  backgroundColor: Colors.white,
                  placeholderImage: Image.asset('uploads/tile.png').image,
                  urlTemplate:
                      "https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png",
                  subdomains: ['a', 'b', 'c'],
                )
```

---

**Severity**
This will help us to label the issue quicker and decide what needs attention first. Only choose fatal if the entire app crashes, otherwise choose non-fatal.

- [ ] Non-Fatal

**Frequency/Rarity**
This will help us to label the issue quicker and decide what needs attention first.

- [ ] Always

**Applicable Platforms**
Only select those that you've tested on - one or more. If possible, test on a variety of platforms.

- [ ] Android

JaffaKetchup:
Hi there,

Do you get any errors trying to retrieve the image? Can you get the same image from other places in the app? 

[...]

Issue #1188, opened 2022-03-14
Release Version To 'pub.dev'


eliabieri:
There has been no release in 6 months. Any idea when a next version will be released?

ibrierley:
Personally I don't have any idea, we've mainly been working on existing bug fixes recently. Any specific reason for wondering, and any specific things desired ? May be an interesting discussion to have.

JaffaKetchup:
Currently, only the owner is able to publish to pub.dev, who hasn't been seen in a while.
Hopefully we can publish next time we see @johnpryan.

JaffaKetchup:
@johnpryan seems to be active elsewhere on GitHub, so hopefully he checks back here soon. I have the email address for him (from a GitHub commit), but I haven't contacted him on there yet.

JaffaKetchup:
For anyone interested, some internal discussion has taken place on this/similar issue - but without @johnpryan's input.
Rest assured, we are still considering the best route forward, and we won't stop trying to get the owner's attention :)

pablojimpas:
Is it possible to use the latest fixes & improvements of `flutter_map` using it directly from git alongside other plugins like `flutter_map_marker_cluster` that depend on a tagged hosted version of `flutter_map`?

JaffaKetchup:
Not sure. But try this:
1. Change flutter_map import back to normal (`flutter_map: `)
2. Move what you had there to underneath `dependency_overrides` (not sure what it's called, check it)
3. Import other plugins normally.

If that doesn't work, you'll have to fork each plugin and change its pubspec.yaml, then include your version.

pablojimpas:
Thanks @JaffaKetchup! I can confirm this approach works, no need to to fork each plugin.

eliabieri:
Any progress in reaching out to @johnpryan ?

JaffaKetchup:
Unfortunately not. I'll try to remember to send him an email tomorrow, but if that doesn't work, we'll have to figure out an alternative. Sorry for the delay!

JaffaKetchup:
@eliabieri and others interested. Pinging @ibrierley as well here.
I've sent @johnpryan an email, so hopefully he responds.

pablojimpas:
> I've sent @johnpryan an email, so hopefully he responds.

I guess he may have been busy with the launch of Flutter 3 and Google I/O stuff. Maybe not the best timing 😆

Solano-Furlan:
It looks like that this package in no longer maintained

[...]

PR #1248, opened 2022-05-28
Update Version

JaffaKetchup:
- Updated version (in pubspec.yaml) to 1.0.0 - semantic versioning
- Fixed README broken link

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,87 @@
+# Changelog
+
+## [1.0.0] - 2022
+
+Contains the following additions/removals:
+
+- Removed inappropriate null-aware checking from `moveAndRotate` - [#1003](https://github.com/fleaflet/flutter_map/pull/1003)
+- Removed unused dependencies from pubspec - [#1237](https://github.com/fleaflet/flutter_map/pull/1237)
+- Migrated to 'flutter_lints' from 'pedantic' - [#1183](https://github.com/fleaflet/flutter_map/pull/1183)
+- Made boolean values uppercase strings in WMS requests - [#1132](https://github.com/fleaflet/flutter_map/pull/1132)
+- Made pinch zoom use center of gesture for focus of zoom - [#1081](https://github.com/fleaflet/flutter_map/pull/1081)
+- Made scroll zoom use center of gesture for focus of zoom - [#1191](https://github.com/fleaflet/flutter_map/pull/1191)
+- Added stroke, cap, and join options to `Polyline` - [#1077](https://github.com/fleaflet/flutter_map/pull/1077)
+- Added option to use pixel cache and length check on `Markers` to avoid crash - [#1147](https://github.com/fleaflet/flutter_map/pull/1147)
+- Added `MapEventScrollWheelZoom` event when zooming using scroll wheel - [#1182](https://github.com/fleaflet/flutter_map/pull/1182)
+- Added `isFilled` parameter to `Polygon` - [#501](https://github.com/fleaflet/flutter_map/pull/501)
+- Added example page for `Polygon`s - [#501](https://github.com/fleaflet/flutter_map/pull/501)
+- Added `maxBounds` parameter to `MapOptions` - [#1211](https://github.com/fleaflet/flutter_map/pull/1211)
+- Added `tileBounds` parameter to `TileLayerOptions` - [#1212](https://github.com/fleaflet/flutter_map/pull/1212)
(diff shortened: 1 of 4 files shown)


ibrierley:
I'm fine with this unless any other objections from folks, especially other maintainers. It's not really a breaking change version, but now feels like a reasonable time where there has been a fair few changes, fixes, improvements etc and things feel stable.

I'd be happy for others to do a bit of testing on latest code and try and spot any flaws first, so maybe see if any objections within a week or so ?

pablojimpas:
The change in the plugin API from Stream<Null> to Stream<void> is in fact a breaking change, but I think that in semantic versioning all the versions below 1.0.0 do not have any guarantees in terms of API stability.

Marking a 1.0 release means that the package API it's stable and every breaking change down the road will mean increasing the major version number.

I'm not a fan of packages that say they're above stable (above 1.0.0) but break the API once or twice a year releasing new major versions, I'll suggest to make this move only if you're sure that no breaking changes will be needed to improve the package in the short/mid-term.

JaffaKetchup:
Hi @pablojimpas,

Thanks for your concerns, and you've made a good point about increasing major number for every breaking change.
The way flutter_map has been working is essentially to ignore the first number, use the second number as major and breaking, and third number as minor/patch. Therefore, technically flutter_map will be on 15.0.0, but I'm using 1.0.0 because you usually start from scratch when switching versioning systems.

What do you mean by Stream to Stream in the public API, I might have misunderstood something?

Thanks.

pablojimpas:
> What do you mean by Stream to Stream in the public API, I might have misunderstood something?

Sorry, the angle brackets broke formatting: `Stream<Null>` to `Stream<void>`

> Therefore, technically flutter_map will be on 15.0.0, but I'm using 1.0.0 because you usually start from scratch when switching versioning systems.

Okey, I like that.

ibrierley:
It's an interesting point about the Null vs Void Stream change. That one passed by by, because it was a lint change. One to watch out for in future.

ibrierley:
Just thinking a bit more on this, I feel like we should probably do a bit of work on the Changelog about previous changes first ? Maybe it's not so clear to others about recent changes that have happened, if it impacts them etc, what to watch out for.

JaffaKetchup:
Yeah, good idea as it hasn't been changed in ages. I'm happy to add it to this PR, and happy for you to add anything as well :)

JaffaKetchup:
Made those changes to the CHANGELOG. It was incredibly out of date: took over an hour to collate all the changes, authors, and PR/issue links! Let me know if anything needs changing.
I also updated the pubspec package description whilst I was at it, as it's very out of date compared to the latest one at the top of the README.

ibrierley:
Great stuff! I hadn't quite appreciated how many there were actually, they do stack up.

I'll try and take a better look over the weekend, but I suspect I won't have a lot to add.

pablojimpas:
Awesome work, actually making this changelog could help with https://github.com/fleaflet/flutter_map/issues/1188 since last time @johnpryan appeared to make a new release he asked for a detailed changelog to stay up to date with the recent activity https://github.com/fleaflet/flutter_map/issues/1021#issuecomment-917148158

JaffaKetchup:
If everyone here is happy to merge this, I think it's pretty much finished.

ibrierley:
Happy with the changelog, all looks good to me!

JaffaKetchup:
(@ibrierley Can't merge without a review if that's what you were intending :))

ibrierley (commented):
I think the changelog works is great, thanks for all of that! Just wondering actually before approving, should we mark placeholderImage as deprecated (as it doesn't work currently and seems to have been noted at that in the past without any actual note of it anywhere) https://github.com/fleaflet/flutter_map/issues/1244 ? I'm fine either way, but maybe if we decide that it's not going to be fixed, it would be a point to mention it here ?

JaffaKetchup:
Good point, just done it.

ibrierley (commented):
Hah being pedantic now :D, maybe we should add placeholderImage deprecation to the changelog as well ? (even though it was kinda broken a long while back)

JaffaKetchup:
I'm happy to do it, but I think we shouldn't as it's quite a minor (and now obvious to developers) deprecation and (as you mentioned) broken a long time ago. Happy to do it if you disagree though!

ibrierley:
No problem!

ibrierley (approved):
lgtm.

PR #1259, opened 2022-06-04
Fix for example with web failing for non-https link 

ibrierley:
Redoing PR for none https link in example page after Git issue, causing errors when using Flutter web. Mobile is fine normally, this only affects browser.


JaffaKetchup (approved):
LGTM, simple change.

PR #1267, opened 2022-06-09
Added line_animator plugin/package

ibrierley:
Package for interpolating using Haversine formula along LatLng (or other points) for gradual line drawing or moving markers along a path.


JaffaKetchup (approved):
LGTM. I'll add to new docs and plugins channel.

PR #1269, opened 2022-06-10
[PROTOTYPE] provide basic tilt and pitch

### LINKED ISSUE #1268: [FEATURE] tilt/pitch (map perspective)
It would be great to be able to render the map with pitch/tilt to give perspective.

I've been experimenting with tilt/pich and have created a prototype: https://github.com/greensopinion/flutter_map/tree/pitch

Prototype supports:
* rendering with pitch
* set the initial pitch via options
* update the map pitch using the controller
* two-finger gesture (push up/down) to change pitch

There's more to do before this could be included as a feature. Some things are top of mind:
* hit testing
* marker positioning
* support rendering layers without perspective (e.g. pop-up info bubbles on markers)

The prototype has a new example page which you can find in the menu named "Pitch". A two-finger push/pull gesture will tilt the map.

greensopinion:
Provide tilt/pitch gesture and visuals.
This is a proof-of-concept showing what
we could do, intended to solicit feedback.

Basic functionality:
* set the initial tilt via options
* update the map pitch using the controller
* two-finger gesture (push up/down) to change pitch

fixes #1268

### DIFF
--- a/example/ios/Flutter/AppFrameworkInfo.plist
+++ b/example/ios/Flutter/AppFrameworkInfo.plist
@@ -21,6 +21,6 @@
   <key>CFBundleVersion</key>
   <string>1.0</string>
   <key>MinimumOSVersion</key>
-  <string>8.0</string>
+  <string>9.0</string>
 </dict>
 </plist>
--- a/example/ios/Podfile
+++ b/example/ios/Podfile
@@ -1,0 +1,41 @@
+# Uncomment this line to define a global platform for your project
+# platform :ios, '9.0'
+
+# CocoaPods analytics sends network stats synchronously affecting flutter build latency.
+ENV['COCOAPODS_DISABLE_STATS'] = 'true'
+
+project 'Runner', {
+  'Debug' => :debug,
+  'Profile' => :release,
+  'Release' => :release,
+}
+
+def flutter_root
+  generated_xcode_build_settings_path = File.expand_path(File.join('..', 'Flutter', 'Generated.xcconfig'), __FILE__)
+  unless File.exist?(generated_xcode_build_settings_path)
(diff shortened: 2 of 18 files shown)


greensopinion:
This is a prototype, not intended to be merged

ibrierley (commented):
Few quick thoughts. (I can't test it on Android for some reason various kotlin errors, guessing sdk versions and compileSdkVersion doesn't help, but I can't see what requires that atm). 

I'm guessing some of these your aware of, but covering in case...

On the non-pitch example, the mousewheel zoom seems off (double click same as usual). 

If I select the pitch page, the top half is missing (that may just be on Web though, maybe needs height/width recalculating to take pitch/rotation into consideration for which tiles to load ?).

If I drag vertically, the rate of map movement is slower than the mouse. I'm guessing that's a perspective ratio thing, but feels a bit odd (in Web anyway). Horizontal is fine even when rotated.

Road labels look a bit blurry (unless high zoom level), I guess there's not a lot that can be done about that one, as some of the maps aren't the crispest at the best of times, let alone when shifted in perspective.

greensopinion:
Good feedback, thanks!

This is in no way ready. Gestures are a hack, and rotating the whole map is problematic (e.g. some markers will want pitch, others won't). Also with vector tiles there will be some labels that we will want to render without pitch (e.g. place names) whereas others should render in the plane of the map.

ibrierley:
Out of interest, is the primary aim to be able to enable a type of satnav mode for the vector tiles plugin ?

greensopinion:
replaced by #1273

greensopinion:
> Out of interest, is the primary aim to be able to enable a type of satnav mode for the vector tiles plugin ?

No, but that would be really cool.

Issue #1265, opened 2022-06-09
[BUG] Zoom issue on double click


JosefWN:
**Describe The Bug**
If I place the cursor on my home (or any other specific geographical location) and double click, the position of my cursor is moved geographically as the viewport zooms (this does not happen on scroll, or on Google Maps double click).

**Expected Behavior**
If I place the cursor on my home and scroll rather than double click, the position of my cursor remains on my home. It should be the same for double click.

**Additional Information**
`flutter_map` v1.0.0

**Severity**
This will help us to label the issue quicker and decide what needs attention first. Only choose fatal if the entire app crashes, otherwise choose non-fatal.

- [x] Non-Fatal
- [ ] Fatal / App Crashes

**Frequency/Rarity**
This will help us to label the issue quicker and decide what needs attention first.

- [ ] Once
- [ ] Uncommon
- [ ] Common
- [x] Always

**Applicable Platforms**
Only select those that you've tested on - one or more. If possible, test on a variety of platforms.

- [ ] Android
- [ ] iOS
- [x] Web
- [x] Windows
- [x] Mac <- not a beta platform?
- [ ] Others (beta platforms)

ibrierley:
I think I see what you mean..it moves on a double click, but kind of moves too far past the point where the event occured ?

JosefWN:
Yeah, I would expect the mouse pointer to remain stationary both geographically and on the screen. It works "correctly" for scroll, so if you try scroll vs double click I think you will see the difference right away.

ibrierley:
Can you test https://github.com/fleaflet/flutter_map/pull/1271 and see if it helps this (without breaking other stuff!)

PR #1271, opened 2022-06-10
fix for double click not keeping track with event position

ibrierley:
Don't merge this just yet, but is for testing currently. Fixes https://github.com/fleaflet/flutter_map/issues/1265

Previously double click wouldn't zoom to correct position. This should fix that, and refactors mousewheel zoom to use the same code now, in a separate method.

Also slight change is the animation easing on double click. There is a bug (existing before this change), where a zoom doesn't follow the correct linear path. The easing doesn't help here, so keeping it simple, so it's slightly less obvious, but ideally needs a fix in the future.


JosefWN:
Nice! Looks correct to me, I can zoom and double-tap my way in/out and still remain in the same geographical position relative to the cursor.

PR #1272, opened 2022-06-11
FlyTo method for zoom/panning between locations 

ibrierley:
Don't merge this yet, looking for testing feedback.

This is a conversion of Leaflets **flyTo** method. This follows the typical map zoom behaviour we would see between 2 distant locations, where it will zoom out first, pan, then zoom in, in a smooth curve style.

As it's experimental it's not too tightly woven into the normal gesture code, so currently it doesn't care about other gestures and cancelling them (it currently must complete its flyTo, it can't be interrupted). I also haven't added any custom events (I'm also a bit hesitant to have zillions of different types of events atm, so it currently will probably do a fling animation event if anyone cares, but that can be changed)

There's an example MapController FlyTo page in the examples folder.

I'm not sure of the best location for it, so currently it's sat on it's own in a map/movement.dart file, as it feels like it would pollute gestures.dart where other similar code is.

Use is ```mapController.flyTo(texas, zoom: 5.0, duration: 3);```

I've kept original leaflet license in there (not entirely sure this is needed, as I think it was copied original from some other math work some others did, but it feels like it's the right thing to do..but does that clash with flutter_map license ?).

Note, one thing I've been looking at, is flutter_maps slightly odd double tap animated/zoom pan behaviour where it always curves out before in. Ultimately, we could do a flyTo there instead of that (tested that, and I think it works better, but I think that's one step too far currently).

### DIFF
--- a/example/lib/main.dart
+++ b/example/lib/main.dart
@@ -57,7 +57,8 @@ class MyApp extends StatelessWidget {
         EsriPage.route: (context) => const EsriPage(),
         PolylinePage.route: (context) => const PolylinePage(),
         MapControllerPage.route: (context) => const MapControllerPage(),
-	MapControllerFlyToPage.route: (context) => const MapControllerFlyToPage(),
+        MapControllerFlyToPage.route: (context) =>
+            const MapControllerFlyToPage(),
         AnimatedMapControllerPage.route: (context) =>
             const AnimatedMapControllerPage(),
         MarkerAnchorPage.route: (context) => const MarkerAnchorPage(),
--- a/example/lib/main.dart
+++ b/example/lib/main.dart
@@ -13,6 +13,7 @@ import './pages/interactive_test_page.dart';
 import './pages/live_location.dart';
 import './pages/many_markers.dart';
 import './pages/map_controller.dart';
+import './pages/map_controller_flyto.dart';
 import './pages/marker_anchor.dart';
 import './pages/marker_rotate.dart';
 import './pages/max_bounds.dart';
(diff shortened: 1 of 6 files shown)


ibrierley:
It's a good point, and one I struggled with (for a good reason, but happy to explore more, but I think it's a difficult).

I think there's a bug/feature (not just with flutter_map) when you try and animate a scale and a translate separately (but at the same time), where it zooms the wrong slightly way at first. You can see this slightly with double click animations (i.e on the animated controller example, especially on web). If you double click far corner, it will move left slightly, before then panning right (even if the center issue is fixed). To me, that feels a bit buggy/cheap.

I think this is because at first, it compounds a scale with a translate (making the translate bigger when the scale is bigger, so the translate isn't actually linear any more), but I may be wrong on this cause (needs a maths person!). So actually I suspect it needs some complex easing math (I tried various separate easings, but I couldn't get there).

I also suspect this is the reason why flyTo exists. I think it may be possible also to get around it with keyframe animations, but again it's adding a layer of complexity. The gestures code is already quite messy with animations, so quite torn on this one.

I may end up just making it as a plugin if it's felt its too polluting.

JaffaKetchup:
@ibrierley You may be able to convert this to draft PR if you want it to look WIP.

ibrierley:
I think I'm going to take a look at making a plugin just for the moment, so it can be tested without as many issues or possible changes, and then ponder.

JosefWN:
It doesn't seem like Mapbox does it much differently from Leaflet:
https://github.com/mapbox/mapbox-gl-js/blob/19ec68b2ec417e51a06b62f065ebb569fcbeb987/src/ui/camera.js#L1448
https://github.com/Leaflet/Leaflet/blob/c50688a25f437f978561ce68194f73618727c0ba/src/map/Map.js#L358

OpenLayers `flyTo` example seems worse than our current solution with the `AnimatedMapController`:
https://openlayers.org/en/latest/examples/animation.html

PR #1273, opened 2022-06-11
[WIP] provide support for pitch

### LINKED ISSUE #1268: [FEATURE] tilt/pitch (map perspective)

greensopinion:
Provide a way to set the pitch of a map either via options or through the controller.
This enables maps to be displayed with a perspective applied so that it appears as if the viewer is looking at the map at an angle.

Marker layers can be rendered flat to the map (i.e. with pitch) or flat to the viewport (i.e. without pitch). This option can be used to create markers that either appear to be map features (such as arrows) or appear to stand up from the map surface, which is useful for pins or text bubbles.

This PR is a WIP (work in progress), there is more work to be done before this should be merged.

Fixes #1268
Makes obsolete #1269

### DIFF
--- a/example/ios/Flutter/AppFrameworkInfo.plist
+++ b/example/ios/Flutter/AppFrameworkInfo.plist
@@ -21,6 +21,6 @@
   <key>CFBundleVersion</key>
   <string>1.0</string>
   <key>MinimumOSVersion</key>
-  <string>8.0</string>
+  <string>9.0</string>
 </dict>
 </plist>
--- a/example/ios/Podfile
+++ b/example/ios/Podfile
@@ -1,0 +1,41 @@
+# Uncomment this line to define a global platform for your project
+# platform :ios, '9.0'
+
+# CocoaPods analytics sends network stats synchronously affecting flutter build latency.
+ENV['COCOAPODS_DISABLE_STATS'] = 'true'
+
+project 'Runner', {
+  'Debug' => :debug,
+  'Profile' => :release,
+  'Release' => :release,
+}
+
+def flutter_root
+  generated_xcode_build_settings_path = File.expand_path(File.join('..', 'Flutter', 'Generated.xcconfig'), __FILE__)
+  unless File.exist?(generated_xcode_build_settings_path)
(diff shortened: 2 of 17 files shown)


nimr77:
I was playing around with this issue, and something was not right, once I added it  didn't take the full screen, or the entire widget, yet when I scaled it to 2, It appeared more effective, 
```
  Widget _createLayerWithTransforms(
      LayerOptions options, List<MapPlugin> plugins) {
    var widget = _createLayer(options, plugins);
    if (options.applyPitch) {
      widget = _applyPitch(_applyRotation(widget));
    }
    return Transform.scale(scale: 2, child: widget);
  }
```
that was my change, but I think I can scale the whole flutter_map widget to resolve this issue.

greensopinion:
@nimr77 thanks for the feedback. There's a long ways to go before this is ready, and I've stopped working on it now. Feel free to improve it in any way that you want.

I haven't resolved the scaling or positioning problem. With perspective added (things getting shorter and thinner in the distance) both size and positioning are affected.

My plan was to figure out the translation and scaling needed to correct positioning and size using a matrix transform, but I haven't got around to it.

nimr77:
That's awesome, 
yes the size and position get affected, on scaling, I saw only the distance so the quality got affected too, I tried to do offset yet it didn't work for me the position became unstable, so calling worked for me, I'm trying to understand how to change the native zoom, so the renderer will be zoomed

JaffaKetchup:
As the author has stopped working on this PR, I'm going to close it. Thanks anyways for your contribution!

PR #1275, opened 2022-06-13
Callbacks for pointer gestures

pmjobin:
This allows registering callbacks on pointer-related gestures by adding corresponding properties to the public API. These callbacks are helpful to implement features such as, for instance, drawing a polygonal ROI over a map using the mouse.


ibrierley:
Thanks for this, will try and get chance to take a look over the next couple of days!

ibrierley (approved):
I think I'm fine with this change. Don't really see any major negatives, and can't spot any breaking changes..I don't think hover works over a marker, just to be aware, not sure there's any way around that, nor if it's desirable either.

(not directly related, but maybe we should think about bumping a minor version increase to pub.dev soon ?)

ibrierley:
Hah yes, I was actually thinking exactly the same before, but felt it a bit unfair on just one pull request :D. Gestures are getting a bit unweildy. I had wondered why we didn't just dump all events in one go, rather than separate types. The callbacks as options are maybe a bit trickier. If anyone has any clever ideas, I'm listening!

Thinking out loud...Just looking at https://api.flutter.dev/flutter/widgets/GestureDetector-class.html there's a lot of gestures, could somehow the gesturedetector be exposed, and let everyone do what they want ? Not even sure that makes sense without looking at it a bit more.

JaffaKetchup:
Yes, we could allow a gesture detector to be passed in, then override the child property (perhaps by implementing custom copy with if can't be done another way).

Issue #66, opened 2018-07-05
Anchor position top/bottom wrong way around


mjtalbot:
Firstly thank you, really enjoying the map!

The Anchor.top and Anchor.bottom look like they are the wrong way around, Anchor top actually pins the bottom of my marker to the correct spot. 

Should be easy to verify, just place a few markers and zoom in and out.

johnpryan:
Thanks for filing an issue! I think the original intention was that "top" meant "the top of the marker is pinned to the location." In Leaflet JS, the anchor is an X,Y position defining the "tip of the icon":

```
The coordinates of the "tip" of the icon (relative to its top left corner). The icon will be aligned so that this point is at the marker's geographical location. Centered by default if size is specified, also can be set in CSS with negative margins.
```

But it could be less confusing to define the anchor point as "below" and "above" the point on the map.  Either way, we need better terminology or documentation for this feature.

mjtalbot:
I could be totally misunderstanding something, but I think I agree with you, as in I also thought top would pin the top of the marker as well. However top pins the bottom of the marker to the geographical location.

Let me run through an example

Currently: 
```
var pixelPosY =
              (pos.y - (markerOpt.height - markerOpt._anchor.top)).toDouble();
```


lets say the geographical y position is calculated to be at 10, and we have a 5x5 marker. 

if the top anchor is selected, the top of the marker will be drawn at (10 - (5 - 0)) or 5. 
the bottom of the marker is at the geographical y position

if the bottom anchor is selected, the top of the marker will be drawn at (10 - (5 - 5)) or 10. 
the top of the marker is at the geographical y position.

I think it should be:
```
var pixelPosY =
              (pos.y - markerOpt._anchor.top).toDouble();
```
if the top anchor is selected, the top of the marker will be at (10 - 0), or 10. 
the top of the marker is at the geographical location. 

if the bottom anchor is selected, the top of the marker will be at (10-5) or 5. 
the bottom of the marker is at the geographical y position.

johnpryan:
There's been some changes to AnchorAlign since this issue, marking as stale

PR #1291, opened 2022-06-28
Proposed fix for unsymmetrical anchored markers disappearing

### LINKED ISSUE #918: [BUG] Markers without symmetrical anchors disappear before they are out of bounds
I believe the fix below handles the markers which have anchors in the center or edges.

https://github.com/fleaflet/flutter_map/pull/313

I have printed the marker width and height values and they are all as they are supposed to be.
If I increase the marker width artificially then the disappearing stops but the anchors are getting corrupted of course. So I think it might be a bug in the package. I might be wrong though.

https://user-images.githubusercontent.com/6442915/120219586-1e7ceb80-c244-11eb-92ad-954b0f8ebb60.mov

aytunch:
@ibrierley @JaffaKetchup please check and see if there is any hiding happening with your working projects.
Thank you very much for the discussion.

Fixes #918.

### DIFF
--- a/lib/src/layer/marker_layer.dart
+++ b/lib/src/layer/marker_layer.dart
@@ -275,8 +275,8 @@ class _MarkerLayerState extends State<MarkerLayer> {
               CustomPoint(pxPoint.x - rigthPortion, pxPoint.y + bottomPortion);
 
           if (!map.pixelBounds.containsPartialBounds(Bounds(sw, ne))) {
-            print("map.pixelBounds: " + map.pixelBounds);
-            print("Bounds(sw, ne):  " + Bounds(sw, ne));
+            print("map.pixelBounds: " + map.pixelBounds.toString());
+            print("Bounds(sw, ne):  " + Bounds(sw, ne).toString());
             continue;
           }
 
--- a/lib/src/layer/marker_layer.dart
+++ b/lib/src/layer/marker_layer.dart
@@ -275,6 +275,11 @@ class _MarkerLayerState extends State<MarkerLayer> {
               CustomPoint(pxPoint.x - rigthPortion, pxPoint.y + bottomPortion);
 
           if (!map.pixelBounds.containsPartialBounds(Bounds(sw, ne))) {
+            print("pxPoint:         " + pxPoint.toString());
+            print("leftPortion:     " + leftPortion.toString());
+            print("rigthPortion:    " + rigthPortion.toString());
+            print("topPortion:      " + topPortion.toString());
+            print("bottomPortion:   " + bottomPortion.toString());
             print("map.pixelBounds: " + map.pixelBounds.toString());
             print("Bounds(sw, ne):  " + Bounds(sw, ne).toString());
             continue;
--- a/lib/src/layer/marker_layer.dart
+++ b/lib/src/layer/marker_layer.dart
@@ -295,8 +295,8 @@ class _MarkerLayerState extends State<MarkerLayer> {
               key: marker.key,
               width: marker.width,
               height: marker.height,
-              left: pos.x - width,
-              top: pos.y - height,
+              left: pos.x - leftPortion,
+              top: pos.y - topPortion,
               child: markerWidget,
             ),
           );
--- a/lib/src/layer/marker_layer.dart
+++ b/lib/src/layer/marker_layer.dart
@@ -261,10 +261,18 @@ class _MarkerLayerState extends State<MarkerLayer> {
             _pxCache[i] = pxPoint;
           }
 
-          final width = marker.width - marker.anchor.left;
-          final height = marker.height - marker.anchor.top;
-          final sw = CustomPoint(pxPoint.x + width, pxPoint.y - height);
-          final ne = CustomPoint(pxPoint.x - width, pxPoint.y + height);
+          // See if any portion of the Marker resides in the map bounds
+          // If not, don't spend resources on build function.
+          // This calculation does not assume where the Anchor is or
+          // if it is symmetrical or not
+          final leftPortion = marker.width - marker.anchor.left;
+          final rigthPortion = marker.anchor.left;
+          final topPortion = marker.height - marker.anchor.top;
+          final bottomPortion = marker.anchor.top;
+
+          var sw = CustomPoint(pxPoint.x + leftPortion, pxPoint.y - topPortion);
+          var ne =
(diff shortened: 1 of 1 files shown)


aytunch:
I will enter one more commit. There is an error

aytunch:
Even after the commit, my markers are disappearing. I will investigate a little more...

aytunch:
Ok friends. I fixed the math and tested it.
In the video you can see that the logs inside of the `if (!map.pixelBounds.containsPartialBounds(Bounds(sw, ne)))` condition fires whenever a Marker is out of bounds. Even when the Anchor point is not symmetrical.
I would be very happy if you can test and merge the PR if you find fit.
Thanks
https://user-images.githubusercontent.com/6442915/176325434-1415b52a-3685-430f-8529-1c1686d21114.mov

ibrierley (commented):
Haven't had time to test, but I note there's a typo in "right vs rigth"

aytunch:
> Haven't had time to test, but I note there's a typo in "right vs rigth"

I always confuse them :) nice catch

aytunch:
For the namings, I agree with you. They can cause confusion.
I can swap right and left names and top and bottom naming.
But the calculations are true. You can see in the video as well.

Would this make more sense?

```
          final rightPortion = marker.width - marker.anchor.left;
          final leftPortion = marker.anchor.left;
          final bottomPortion = marker.height - marker.anchor.top;
          final topPortion = marker.anchor.top;

          final sw =
              CustomPoint(pxPoint.x + leftPortion, pxPoint.y - bottomPortion);
          final ne =
              CustomPoint(pxPoint.x - rightPortion, pxPoint.y + topPortion);

          if (!map.pixelBounds.containsPartialBounds(Bounds(sw, ne))) {
            continue;
@@ -287,8 +295,8 @@ class _MarkerLayerState extends State<MarkerLayer> {
              key: marker.key,
              width: marker.width,
              height: marker.height,
              left: pos.x - rightPortion,
              top: pos.y - bottomPortion,
              child: markerWidget,
            ),
```

I feel like, whatever we change, it will still not make too much sense because we are dealing with 2 different coordinate systems in the same code snippet. Map coordinate system having (0,0) at top-left and Anchor coordinate system having (0,0) at bottom-right

@JaffaKetchup I know you are busy with the exams. But can you at least confirm this is the case with the coordinate systems? And good luck with the grades ;)

ibrierley:
(btw I'm away from tomorrow for a few days, and as Jaffa may be busy, it might be a case of revisiting next week, I suspect I don' have any issue at all with the fix, just maybe it needs some extra docs or something)

JaffaKetchup:
@aytunch
I'm done with exams now, luckily for the recent high priority issue.
After I've dealt with that, I'll go through this. Should be quicker than next week, but I can't promise.

ibrierley:
Took a quick peak at leaflet out of interest...so I think _part_ of the confusion is that we don't behave the same with Anchors to them.


In my brain, I'm along with leaflet.

Lets say the Marker width/height is 300px/100px, the "point" of the Anchor would be at 75,100 (referenced top/left like they say in docs), however to have the same effect in flutter_map we need to set it to be 225,0.

Why we don't follow Leaflet in this case I don't understand.... :D.

JaffaKetchup:
Yeah I would agree. Using top left as (0, 0) definitely makes more sense. Whether we want to make as big of a breaking change as this is a different question.

ibrierley:
Interesting, just found this....

https://github.com/fleaflet/flutter_map/issues/66

ibrierley:
Actually, that may not be relevant thinking about it....

aytunch:
> Interesting, just found this....
> 
> #66

@ibrierley  Nice find from 2018 :)
As @JaffaKetchup said, then the decision is between introducing a breaking change or not. I don't think the devs would like playing with Anchor positions too much. Is there any non-breaking solution for having a (0,0) top-left Anchor coordinate system?

aytunch:
I think it is related. But the author only cared about the vertical convention not the horizontal

ibrierley:
Yeah, and don't worry, I won't be suggesting any breaking changes at all short term (one could conceivable consider your fix very slightly changing in behaviour if they've coded some hack workaround, so we probably need a minor release, but I doubt anyone ever has hacked around it).

My main concern with this type of stuff is making sure any "oddness" gets documented in the code (and main docs for usage), so there's as easy logic as possible (partly as people often reuse code later on), that way next time, any devs don't spend too long on it iyswim.

aytunch:
We should definitely state in documentation https://docs.fleaflet.dev/usage/layers/marker-layer#markers-markers that;

- Anchor.left corresponds to the logical distance between the Anchor point from the right side of the Marker
- Anchor.top corresponds to the logical distance between the Anchor point from the bottom side of the Marker

JaffaKetchup:
I'll try to have a look at this today if possible.

JaffaKetchup (approved):
Sorry for the slow reply @aytunch, totally my fault, keep saying I'll do it, then I never do.
By just looking at the code and the videos, I can tell it will work, and as it's a simple change, I don't feel the need to test it - we can always revert if absolutely necessary.

Note that it may take a while for the version to become available on pub.dev, unfortunately. This is because I need to write the CHANGELOG for the version, which needs approval and merging.

JaffaKetchup:
In the meantime, you may try using 'dependency_overrides'?

aytunch:
@JaffaKetchup there is no fault. We are very thankful to both of you for developing and maintaining this awesome package and your personal plugins as well.

As for the changelog, maybe something along these lines might be enough
_Logic to decide if a Marker is off-screen now takes the Anchor point in to account. (Previously it assumed symmetrical anchors)_

> which needs approval and merging.
Who does the approval and merging btw?

JaffaKetchup:
Thanks :)

CHANGELOG should be simple enough, but there are some other changes to put in there from previous commits.

If I submit the CHANGELOG PR - which is annoying at the moment, as I already have a fork for #1294, so I'll have to get involved with Git (~help me~) - then @ibrierley will have to approve it. Or vice versa. We can't merge our own PRs.

aytunch:
I see. Good luck with #1294 
If there is anything I can help with, please let me know guys.
On another note, my next milestone is integrating [flutter_map_tile_caching](https://github.com/JaffaKetchup/flutter_map_tile_caching) so expect some questions and issues there too :D
MapBox is very expensive and we don't have an investor yet. You plugin will come to the rescue ;)

PR #1293, opened 2022-06-29
[v2.1.0] Fix `LateInitializationError` with `MapController`

### LINKED ISSUE #1288: [BUG] Map fails to reinitialize while in ListView
When you put the `FlutterMap` inside a `ListView`, it can be hidden and then added anew. And it fails, because it sets a new state for a map controller, which is already initialized. Testing code is below: it adds 30 lines of text after the map. Scroll down and then back up again to see the error.

Mapybe mapController needs to be disposed properly?

Flutter_map 1.1.0, Flutter 3, Dart 2.17.

```dart
import 'package:latlong2/latlong.dart' show LatLng;

void main() {
  runApp(const MaterialApp(
    title: 'Flutter Demo',
    home: MyHomePage(),
  ));
}

class MyHomePage extends StatelessWidget {
  const MyHomePage({Key? key}) : super(key: key);

  @override
    return Scaffold(
      appBar: AppBar(
        title: const Text('Map in a ListView'),
      ),
      body: ListView(children: [
        SizedBox(
          height: 100.0,
          child: FlutterMap(
            options: MapOptions(
              center: LatLng(50, 30),
              zoom: 13,
              allowPanningOnScrollingParent: false,
            ),
            layers: [
              TileLayerOptions(
              ),
            ],
          ),
        ),
        for (int i = 1; i < 30; i++)
          Padding(
            padding: const EdgeInsets.all(20.0),
            child: Center(child: Text('Line $i', style: const TextStyle(fontSize: 20))),
          ),
      ]),
    );
  }
}
```

Zverik:
There is a setter for a final error, which calls for an exception. And it came! 

Fixes #1288.

### DIFF
--- a/lib/flutter_map.dart
+++ b/lib/flutter_map.dart
@@ -87,7 +87,7 @@ class FlutterMap extends StatefulWidget {
     this.children = const [],
     this.nonRotatedChildren = const [],
     this.mapController,
-  })  : super(key: key);
+  }) : super(key: key);
 
   @override
   FlutterMapState createState() => FlutterMapState();
--- a/lib/flutter_map.dart
+++ b/lib/flutter_map.dart
@@ -79,7 +79,7 @@ class FlutterMap extends StatefulWidget {
   /// A [MapController], used to control the map.
   final MapController? mapController;
 
-  FlutterMap({
+  const FlutterMap({
     Key? key,
     required this.options,
     this.layers = const [],
--- a/lib/flutter_map.dart
+++ b/lib/flutter_map.dart
@@ -77,7 +77,7 @@ class FlutterMap extends StatefulWidget {
   final MapOptions options;
 
   /// A [MapController], used to control the map.
-  final MapController mapController;
+  final MapController? mapController;
 
   FlutterMap({
     Key? key,
@@ -86,9 +86,8 @@ class FlutterMap extends StatefulWidget {
     this.nonRotatedLayers = const [],
     this.children = const [],
     this.nonRotatedChildren = const [],
-    MapController? mapController,
-  })  : mapController = mapController ?? MapController(),
-        super(key: key);
+    this.mapController,
+  })  : super(key: key);
 
   @override
   FlutterMapState createState() => FlutterMapState();
--- a/lib/src/map/flutter_map_state.dart
+++ b/lib/src/map/flutter_map_state.dart
@@ -33,8 +33,7 @@ class FlutterMapState extends MapGestureMixin {
   @override
   void initState() {
     super.initState();
-    if (widget.mapController == null)
-      _localController = MapControllerImpl();
+    if (widget.mapController == null) _localController = MapControllerImpl();
     mapState = MapState(options, (degree) {
       if (mounted) setState(() {});
     }, mapController.mapEventSink);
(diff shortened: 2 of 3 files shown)


JaffaKetchup:
Will try to review at some point soon, quite busy right now. Thanks!

JaffaKetchup (changes requested):
@Zverik

I seem to get another error now when I try to interact with the map:
image
...so not sure this is fixed.

I just used the code in the body of #1288 on the latest version.

Any ideas?

Zverik:
Yes, I did a test, and indeed `initState()` is called with the exactly same controller object (`Impl`) that was disposed in `dispose()`.

Zverik:
Okay, I fixed this. Could everybody please test that nothing (including examples) has broken? :)

LeoCreno:
Your branch worked for me !

JaffaKetchup:
@Zverik Will test this in a minute. Preparing the CHANGELOG for v2.0.0, so hoping to merge this in-time.

JaffaKetchup:
@Zverik Sorry, wasn't able to test in time for v2 :(. I'll try to test sooner rather than later for v2.1.0.

JaffaKetchup (changes requested):
See https://github.com/fleaflet/flutter_map/pull/1293#issuecomment-1187594169.

ibrierley:
Heya, I'm not getting a lot of time to look at things in depth just atm...

However, I'm a bit unclear where its at. Are we saying that with this PR, if the widget goes out of the widget tree and returns, it isn't at its last postion, but the initial one ? This doesn't fee right to me if so...(however, people can use the widget above I assume, or an Offstage widget to get around it, and we're in a better position that currently?).

JaffaKetchup:
@ibrierley Yeah, you're exactly right.

To save memory and increase performance, Flutter disposes of some widgets when they scroll out of view.

This PR fixes the issues arising from the map controller becoming 'detached' from the map, and then being unable to 'reattach'.

However, it does not prevent the widget getting rebuilt: not everyone wants or needs this, and it's fighting against Flutter to implement this unnecessarily. This is what the above comments were about. When the map gets rebuilt, it loses its state, and so returns to the initial position.

This can be easily fixed with the `KeepAlive` widget (or similar). If you would prefer another option from the 5 above, let me know.

ibrierley:
I don't think this is right as is (as a full solution). I think the PR fixes the main issue which is great, but I don't think the user would expect the map to get reset (and user shouldn't really care too much about underlying code), unless their own code would decide that. 

To me it feels like @JaffaKetchup's 2(ii) would make most sense still (I don't care if it's only there as an option, and whether the option enabled is default though...?), I don't think I quite understood the objection there, so may need a bit of clarity!

Would all of this work fine also in a navigator context, where someone navigates away and returns as well (as opposed to a listview) ?

JaffaKetchup:
I'm onboard with 2(ii). I'd leave the option disabled by default, but it can be implemented in under 10 additional lines. I'll make a PR.

I know that it would work with a `PageView` and tabbed setup, but I doubt it would work with navigating away from a page - although I have no real basis for this. I can test it easily.

JaffaKetchup:
@ibrierley Can confirm that the keep alive functionality has no effect on navigation.

aytunch:
would [RestorationMixin](https://api.flutter.dev/flutter/widgets/RestorationMixin-mixin.html) help keep the state of the `flutter_map` instances?

JaffaKetchup:
That looks way too confusing and high effort, I think the automatic keep alive mixin is good for now.

JaffaKetchup:
Thanks for your contributions!

PR #1294, opened 2022-06-29
'User-Agent' Header Generation & More

### LINKED ISSUE #1292: Include appropriate user-agent header when making map requests
I'm one of the admins of tile.openstreetmap.org. Currently flutter_map uses the default dart:io HTTP User-Agent (e.g. `Dart/2.17 (dart:io)`). This means that when looking at traffic, flutter traffic gets lumped in with other Dart traffic, and if **any** Dart users are causing usage problems, all Dart users including flutter could end up blocked. This happened recently, and because it was a generic user-agent, I was unable to figure out who the traffic was coming from.

The [tile usage policy](https://operations.osmfoundation.org/policies/tiles/) requires a Valid [HTTP User-Agent](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent) identifying application. For flutter, this could be something like `flutter_map/1.2.3`, where 1.2.3 is a version.

A better option for your users would be to automatically generate a user-agent from the name of their app, if possible. Frequently the name of the app is something like com.example.app, so a user-agent could be `com.example.app/456 flutter_map/1.2.3`. If an app-specific default is not set, then the instructions should be clear about the need to set an appropriate user-agent

JaffaKetchup:
**Fixes #1292. Includes breaking changes. Will require significant documentation changes.**

This PR's main aim is to add a `userAgentPackageName` parameter to `TileLayerOptions`, which can be automatically passed to `TileProvider`s for simplicity - developers do not have to specify tile providers if it is the default provider, just to add this parameter. This must then be combined with other headers correctly, resulting in one single correct 'User-Agent', and any other necessary headers.
There is also a 'lower level' headers implementation to `TileProvider`s, so custom headers can be passed if needed when `TileLayerOptions` is not in use, for whatever reason.

At the same time, this deprecates `NonCachingNetworkTileProvider` in favour of `NetworkNoRetryTileProvider`, as the old name was misleading. Also in this refactoring, `TileProvider`s lost their `const` ability. This will not affect developers that do not override the default provider, or did not include the `const` keyword when doing so - other developers will see breaking changes here.

In other news:

- Some of the `int`s have been changed to `Duration`s, as in the TODO list. This is a breaking change.
- The `_positionedForOverlay` method has had it's TODO solved as well, by removing unnecessary multiplication by one. Note that this overlaps with #1295, and will cause merge conflicts.
- Other TODOs have been removed, because they were not getting fixed or they were not worth fixing.
- Old deprecations (`placeholderImage` and `attributionBuilder`) have been completely removed.


pnorman:
A user-agent of `Dart/2.17 (dart:io), flutter_map (...)` is acceptable. If it's sending multiple of the same header, that would be a problem.

JaffaKetchup:
Well I'm not sure if there are multiple headers being sent or just one. It's showing as an array, which suggests to me that there are two, separate user-agents being sent.

Sorry for the pinging @pnorman, but can you provide some insight as to how these user-agents are checked?

Also note, on debug browser, the User-Agent appears to be the browser's and nothing else.

pnorman:
There is only one user-agent header, so I'm not sure which it'd send if it's an array. 

The user-agents are checked on the CDN

JaffaKetchup:
Waiting on https://github.com/dart-lang/sdk/issues/49382 now.

pnorman:
> I suspect it probably just concatenates all those in an Array

Well, it sends a string. If the string is a concatenation of all the array elements it would work

JaffaKetchup:
Interesting @pnorman. Would prefer to find a way to remove the old agent, but it's proving difficult.
Are you confirming that it's sending a string and concatenation, or not? I'm struggling to understand, sorry.

pnorman:
Headers are strings. I don't know what it's sending, but it's not sending an array, as there are no types in headers.

JaffaKetchup:
Yeah, I guessed/thought/knew as much. It's probably joining them, thinking about it - most servers/code would be sensible enough to not just throw away data.

I'll keep trying to see if I can get my idea to work, but if it takes too long, I'll give up and just go back to the two headers.

JosefWN:
If you want you can merge my additional fixes in my PR into yours, and I can close my PR. Would avoid the merge conflicts.

JaffaKetchup:
Thanks for the offer, but I'd prefer not to. Would make this PR just very big, and you'd lose some credit in the process. The conflicts shouldn't be too hard to iron out.

pubspec.yaml
```diff
@@ -21,6 +21,7 @@ dependencies:
   proj4dart: ^2.0.0
   tuple: ^2.0.0
   vector_math: ^2.1.0
+  universal_io: ^2.0.4
```
pablojimpas, line 24: same here

ibrierley:
Hopefully I should get a bit of time at the weekend to check this. Couple of questions in meantime...
There was a note about universal_io earlier and why it's needed...is this an issue for people ? 
Are there any other issues on NonCachingNetworkTileProvider being removed...are we fairly happy there's no missing functionality (I know we thought this provider didn't actually stop caching anyway...)

JaffaKetchup:
Wouldn't worry about checking it just yet, it still doesn't 100% work.
In terms of universal_io, I needed it to keep web support when using things from 'dart:_http', which can usually only be accessed through 'dart:io', which doesn't support web.
There is no lost functionality, only renaming. Everything that there was before can still be used, just with a different name and description.

JaffaKetchup:
Just wrote this: https://github.com/dart-lang/sdk/issues/49382#issuecomment-1179302726.

JaffaKetchup:
@ibrierley This is now ready for review.

I've reduced the reliance on 'universal_io', but it is still necessary to allow the library to compile for the web.

JaffaKetchup:
Ah, finally fixed it. Needed to remove the dependency on 'universal_io' in the end, it was not helping. Unfortunately, this has made the code less organised, as two 'tile_provider.dart' files are needed now.

ibrierley (approved):
This all tests fine for me. If anyone relies on things like user_agents in requests somehow I would test this (maybe give a couple of days for anyone to test and scream before merging :D?)

JaffaKetchup:
OK, I'm happy to wait for a couple of days, just in case anyone has any feedback.

I've prepared the documentation for v2 as well, so that's ready to go on release. Migration instructions have also been provided.

greensopinion:
> @greensopinion Looking through your plugin for vector tiles specifically, as this brought that to the forefront of my mind as I was editing the docs, you might need to make some changes to your custom tile providers to reflect these changes. 

@JaffaKetchup Thanks for the heads-up! Great to see more improvements to this library!

The `vector_map_tiles` plug-in doesn't use the `TileProvider` class of `flutter_map` because that one is specific to raster images. Instead, we have [VectorTileProvider](https://github.com/greensopinion/flutter-vector-map-tiles/blob/main/lib/src/vector_tile_provider.dart#L11) that is specific to vector tiles.

I've checked out this PR and run the example app. Everything appears to build and run without error. Not sure if you had something else in mind, but from what I can gather no changes are needed in the `vector_map_tiles` plug-in. Let me know if I'm missing something!

JaffaKetchup:
@greensopinion No problem!

In that case, I think the only thing to do is implement the 'User-Agent' header in your library.

PR #1297, opened 2022-07-06
Expose RetryClient in NetworkTileProvider initialisation

MichalTorma:
This pull request adds optional parameter `RetryClient` for `NetworkTileProvider`. I made this fork to scratch my own itch with some stubborn tile servers that are either too slow or don't follow the standard conventions. Some servers require longer cooldown between the requests therefore more than 3 retries are needed (as was the default in `NetworkImageWithRetry`).

### DIFF
--- a/lib/src/layer/tile_layer/tile_provider/network_image_with_retry.dart
+++ b/lib/src/layer/tile_layer/tile_provider/network_image_with_retry.dart
@@ -10,10 +10,24 @@ class NetworkImageWithRetry extends ImageProvider<NetworkImageWithRetry> {
   /// The scale to place in the [ImageInfo] object of the image.
   final double scale;
 
-  /// The http RetryClient that is used for the requests
-  final RetryClient retryClient = RetryClient(Client());
+  ///
+  final String? expectedFormat;
 
-  NetworkImageWithRetry(this.url, {this.scale = 1.0});
+  /// The http RetryClient that is used for the requests
+  late RetryClient retryClient;
+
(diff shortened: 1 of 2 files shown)


ibrierley (commented):
So, out of interest, what would be a good example code that someone would use as the optional RetryClient ?

MichalTorma:
> So, out of interest, what would be a good example code that someone would use as the optional RetryClient ?

Hi there,
Here is an example of layer definition that I needed for one open server that replies 200 every-time you ask for something, but response either contains png as one would expect, or content-type is different and error message (too many requests) is inside the response body. (don't as me why :D - It's Norwegian national cartography server...). I also needed to increase the number of retries because it is a bit sluggish...
```dart
        TileLayerOptions(
            tileProvider: NetworkTileProvider(
                retryClient: RetryClient(
              Client(),
              retries: 10,
              when: (p0) => p0.headers['content-type'] != 'image/png',
            )),
            wmsOptions: WMSTileLayerOptions(
                baseUrl: 'https://openwms.statkart.no/skwms1/wms.topo4?',
                layers: ['topo4_WMS'],
                version: '1.3.0',
                format: 'image/png',
                crs: const Epsg3857(),
                transparent: false)),
```

JaffaKetchup:
@MichalTorma Hi there,

I've had to make some unexpected changes to #1294, which includes the changes that PR made. You can see the commit which did this: https://github.com/fleaflet/flutter_map/pull/1294/commits/ad2adb651ef6374449f63e177a5fa169698784df.

As a result, I'm going to close this PR. Many thanks for your contribution, it's greatly appreciated!

PR #1319, opened 2022-07-25
[v2.2.0] Stricten Lints & Fix Tests Exceptions

lsaudon:
`strong-mode` is deprecated. It should be replaced by `language`. [Enabling stricter type checks](https://dart.dev/guides/language/analysis-options#enabling-additional-type-checks)

The exception `type 'Null' is not a subtype of type 'HttpHeaders'` does not appear in tests anymore.

Add sames rules for example

Replace dynamic by the right type when possible.


lib/src/layer/tile_layer/tile_provider/network_no_retry_image_provider.dart
```diff
@@ -29,7 +29,6 @@ class FMNetworkNoRetryImageProvider
     FMNetworkNoRetryImageProvider key,
     DecoderCallback decode,
   ) {
-    //ignore: close_sinks
```
JaffaKetchup, line 32: This line needs to be added back, as not having it is causing the checks to fail.

example/lib/pages/overlay_image.dart
```diff
@@ -15,7 +14,7 @@ class OverlayImagePage extends StatelessWidget {
     final bottomRightCorner = LatLng(53.475, 0.275);
     final bottomLeftCorner = LatLng(52.503, -1.868);
 
-    final overlayImages = [
+    final overlayImages = <OverlayImage>[
```
JaffaKetchup, line 17: This needs to be changed to support #1315.

```suggestion
    final overlayImages = <BaseOverlayImage>[
```

JaffaKetchup (changes requested):
Hi there, and thanks for your contribution. However, there's a few things that need to be changed.

lib/src/layer/tile_layer/tile_provider/network_no_retry_image_provider.dart
lsaudon: `close_sinks` is already ignore in [analysis_options.yaml](https://github.com/fleaflet/flutter_map/blob/a6fb00f5887b6417d9bcfee8c407377618487906/analysis_options.yaml).

JaffaKetchup: Ok, in that case, can you stop ignoring it overall, because that's a useful lint, and just ignore it on this occasion?

pablojimpas:
The strict-casts mode was introduced in Dart 2.16, right now we support all the way back to `2.12.0` https://github.com/fleaflet/flutter_map/blob/2d2a956337e41505fd5e44bbbd5e01ca2348ba72/pubspec.yaml#L8

I'm okay with applying the code changes to satisfy new linter rules, but we should either drop support for earlier versions of Dart (which I think it's not worth it) or left the `pubspec.yaml` untouched.

lib/src/geo/crs/crs.dart
```diff
@@ -350,21 +350,21 @@ abstract class Projection {
 
   LatLng unproject(CustomPoint point);
 
-  double _inclusive(Comparable start, Comparable end, double value) {
-    if (value.compareTo(start as num) < 0) return start as double;
-    if (value.compareTo(end as num) > 0) return end as double;
+  double _inclusive(double start, double end, double value) {
+    if (value.compareTo(start) < 0) return start;
```
pablojimpas, line 354: Do we really need to keep using the `compareTo` method now that they are all doubles?

lib/src/geo/crs/crs.dart
```diff
@@ -350,21 +350,21 @@ abstract class Projection {
...
+    if (value.compareTo(end) > 0) return end;
```
pablojimpas, line 355:
```suggestion
    if (value < start) return start;
    if (value > end) return end;
```

pablojimpas (changes requested):

JaffaKetchup:
I agree with @pablojimpas, I think backwards-compatibility is important, and we need to support Dart 2.12 for now at least.

JaffaKetchup (changes requested):
I agree with @pablojimpas, I think backwards-compatibility is important, and we need to support Dart 2.12 for now at least.

Also, please stop ignoring `close_sinks` for the whole library, and just ignore it where I previously asked. It's a useful lint to have, and in this case, it's a false positive.

lsaudon:
> I agree with @pablojimpas, I think backwards-compatibility is important, and we need to support Dart 2.12 for now at least.
> 
> Also, please stop ignoring `close_sinks` for the whole library, and just ignore it where I previously asked. It's a useful lint to have, and in this case, it's a false positive.

I did not add ignoring `close_sinks` for the lib.  The rule was already there.
I just put the same rules in the example.

lib/src/layer/tile_layer/tile_provider/network_no_retry_image_provider.dart
lsaudon: This lint is not in flutter_lints or custom add rules

analysis_options.yaml
```diff
@@ -1,9 +1,10 @@
...
 analyzer:
-  strong-mode:
-    implicit-casts: false
-    implicit-dynamic: false
+  language:
+    strict-casts: true
+    strict-inference: true
+    strict-raw-types: true
   errors:
     close_sinks: ignore
```
JaffaKetchup, line 9: Ok, I'm kind of seeing what you mean. I'm going to remove this line, then approve this PR so it will run the checks. If the checks fail, we'll need to add that 'ignore' line back, otherwise we should be OK.

```suggestion
```

JaffaKetchup (approved):
As above.

JaffaKetchup (changes requested):
Yeah the checks are still failing, so the 'ignore' line in the tile provider needs to be added back. Let's keep the analysis_options.yaml without the ignore statement though.

lib/src/layer/tile_layer/tile_provider/network_no_retry_image_provider.dart
JaffaKetchup: As below.

lsaudon: readded

lsaudon:
> Yeah the checks are still failing, so the 'ignore' line in the tile provider needs to be added back. Let's keep the analysis_options.yaml without the ignore statement though.

It's weird because I ran the same commands as in the CLI and this error doesn't appear.
image

JaffaKetchup:
Yeah, it's because the new package score analyser workflow I added uses a different lint set, so only that one fails: https://github.com/axel-op/dart-package-analyzer. Theoretically, this means that the pub.dev score for this package, should this have been pushed without this change, would have only been 120/130.

JaffaKetchup:
@pablojimpas Thanks for the heads-up, but I'm a bit confused. How should we support Dart 2.12? When you said 'leave the pubspec untouched', this PR hasn't changed it. Please can you clarify?

Once Dart 2.12 is supported, this should be good to go.

lsaudon:
> Yeah - I had this same situation as well. It's because the new package score analyser workflow I added uses a different lint set, so only that one fails: https://github.com/axel-op/dart-package-analyzer. Theoretically, this means that the pub.dev score for this package, should this have been pushed without this change, would have only been 120/130.

The package "dart-package-analyzer" does not use the same rules as the package "Pana" that gives the score pub.dev.

I think we should at least have the same rules as the package "dart-package-analyzer".

JaffaKetchup:
You might be right, in which case, we should probably just disable this check I added, because I'm not sure you can change the lint rules it uses.
I'll revert my PR or open a new one tomorrow.

lsaudon:
> You might be right, in which case, we should probably just disable this check I added, because I'm not sure you can change the lint rules it uses.
> I'll revert my PR or open a new one tomorrow.

No, don't cancel your PR, we can just add the rules and regulations if we have to.

JaffaKetchup:
I'm not sure how to do that, so feel free to do so here :)

lsaudon:
Ok, I found the rules that `pana` uses, I added the one that was not in `flutter_lints`. I sorted the rules.

The rules use by `pana` [analysis_options.dart](https://github.com/dart-lang/pana/blob/master/lib/src/analysis_options.dart)

- https://raw.githubusercontent.com/flutter/flutter/master/packages/flutter/lib/analysis_options_user.yaml
- https://raw.githubusercontent.com/dart-lang/lints/main/lib/core.yaml

JaffaKetchup:
OK, that seems good to me I think. @ibrierley Any opinions?

ibrierley:
No opinions from me!

JaffaKetchup (approved):
LGTM. Quite a large changeset, but I think it's all good. Many thanks for your contributions!

PR #1320, opened 2022-07-26
[v2.2.0] Improve GitHub Actions Workflow

lsaudon:
Upgrade actions.
Add `--no-pub` in `flutter test`, because `flutter pub` get is launch before.


JaffaKetchup:
Thanks for your contribution! These changes have been incorporated into #1321.

PR #1321, opened 2022-07-28
Improved GitHub Actions Workflows

JaffaKetchup:
I've been working on my plugin for flutter_map, and I have setup a multi-step workflow to increase productivity. I have carried over my changes to here, and you can see the things that have changed below. A run of these can be found at https://github.com/JaffaKetchup/flutter_map_tile_caching/actions/runs/2752638748.

In addition to testing the code contents of this library, there is now an automatically compiled package report reflecting what the [pub.dev score](https://pub.dev/packages/flutter_map/score) should be on publishing. If the expected score is less than the possible score, then this check fails. This can be changed if necessary.

If both of those checks (content and package score) pass, then there is an automatic example app builder, which builds an optimised APK for Android and ZIP archive for Windows. It's intelligent enough to only build the example app if its source has changed, as it's quite an expensive operation (up to 15 mins computing time). Note that that means that a change in the library to fix a bug will not update the example application, unless the example application has been changed. These are then outputted to a directory called 'prebuiltExampleApplications', which is committed.  
In the future, it is easily possible to set up automatic web building and deployment to Firebase Hosting (free), as I have done this in the past and still have the configuration for it. However, this will require @johnpryan's assistance as it requires adding repository secrets.

To reduce the number of workflow run logs (currently at 939), I've also included a simple workflow which will delete all old workflow runs (older than 7 days). This runs automatically every Monday.

---

**Things to consider:** 
- Should the example builder be run on every (content and package check passing) commit instead of just those that have changed the example? This will avoid issues where the example still contains bugs after they have been fixed because the example source hasn't changed.
- Do we want to allow a certain margin of error in the package analysis?
- Do we want to change the times at which the log cleaner runs?

### DIFF
--- a/.github/workflows/delete_old_runs.yml
+++ b/.github/workflows/delete_old_runs.yml
@@ -1,0 +1,16 @@
+name: Delete Old Workflow Runs
+on:
+  schedule:
+    - cron: "0 0 * * 1" # Run every Monday
+  workflow_dispatch:
+
+jobs:
+  del_runs:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: Mattraks/delete-workflow-runs@v2
+        with:
+          token: ${{ github.token }}
+          repository: ${{ github.repository }}
+          retain_days: 7
+          keep_minimum_runs: 0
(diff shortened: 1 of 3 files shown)


JaffaKetchup:
Removed auto-builder due to concerns discussed in Discord.

ibrierley (approved):
Thanks!

JaffaKetchup:
Now part of #1323.

PR #1318, opened 2022-07-25
Fixed Null-Safety Errors

JaffaKetchup:
_Replaces #1317 due to a Git messup_

This is essentially a blind fix, implementing https://docs.flutter.dev/development/tools/sdk/release-notes/release-notes-3.0.0#your-code. Help with testing is appreciated, as I can't see what I'm fixing!

Also upgraded package version to v2.1.1.

PR #1323, opened 2022-07-29
Addition To #1318: Null-Safety Bug Fix

JaffaKetchup:
#1318 appears not to have gotten all the possible bugs in the concerned line. This should get them all!

Also includes all changes from #1321.


ibrierley (approved):
Thanks!

PR #1325, opened 2022-07-29
[v2.2.0] Added `latLngToScreenPoint` & Refactored `pointToLatLng`

ibrierley:
Originally pointToLatLng was only available to mapController rather than State (for plugins), as well as rotatePoint (I'm unclear if anyone uses this, but probably should have been kept private originally, but may be handy for someone).

I've refactored this, and added latLngToScreenPoint in MapState, and added access to this in mapController as well.

Test page added as well, at bottom of examples.


example/lib/pages/latlng_to_screen_point.dart
```diff
@@ -0,0 +1,86 @@
...
+    if (mapEvent is! MapEventMove && mapEvent is! MapEventRotate) {
+      // do not flood console with move and rotate events
+      debugPrint(mapEvent.toString());
+    }
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    return Scaffold(
+        appBar: AppBar(title: const Text('Test out Interactive flags!')),
```
JaffaKetchup, line 51: This needs correcting:

```suggestion
        appBar: AppBar(title: const Text('LatLng To Screen Point')),
```

JaffaKetchup (changes requested):
This is the only issue, as far as I can see. Thanks for this!

JaffaKetchup (changes requested):
Just spotted this...

Maybe we should make the method names consistent?
At the moment, one specifies a 'ScreenPoint' whereas the other just uses 'Point', but they both mean the same thing. Personally, I'd prefer removing the 'Screen', as there are other methods that don't use it as well.

ibrierley:
I suppose the question is, which point ? layer point coordinate space, which may or may not be rotated, a widget coordinate space point, or screen coordinate space or something else ? Point doesn't really mean anything by itself does it ? So it's a fiddly question, be consistently ambiguous, or precisely inconsistent :D. I'm happy with whatever tbh (I think as long as the comments/documentation are clear about intention).

ibrierley:
It's also worth clarify it's NOT latLngToPoint here https://github.com/fleaflet/flutter_map/blob/e6a77940009bc8b646f8f1bf2b05bd24ab754f08/lib/src/geo/crs/crs.dart (line 26).

JaffaKetchup:
Ok, so maybe we need to be using 'latlngToScreenPointWithRotation' and all other possibilities of that? Seems sensible to me, although quite long. A priority (at least for me) is also to reduce breaking changes, so I'm not sure what's best for this.

ibrierley:
Yes, I know what you mean :D. I'd deliberated over that a bit. It's a kind of fiddly one, and depends a bit what may be likely to be provided in future.

For example, is there ever likely to be a ```latLngToScreenPointWithoutRotation``` ... not quite sure, but maybe needs some thought.

I'm not quite sure (yet) we need more than ```latLngToScreenPoint (or some other name of the same thing)```   and   ```latLngToPoint``` (now that already exists and is part of crs anyway). I tend to think of this as ```latLngTo<Map>Point``` the same as ```latLngTo<layer>Point``` ). Hmm maybe that's as clear as mud :D

What I typically try to do where applicable, when I write any methods is to add any new functionality as an optional parameter (like rotatePoint so as not to break it) . latLngToScreenPoint(latLng) can easily become latLngToScreenPoint(latLng, rotated = false) if there's a requirement without breaking it ?

JaffaKetchup:
Wow this is confusing :|

So, let me get this straight. Per conversion direction, there's two methods: one in CRS for the layer, and one for the map? Please correct me if I'm wrong.

I think the way you're thinking of it should be how it's named (ie. `map` and `layer`) is sensible to implement. Alternatively, if we don't want to do breaking changes (I'm assuming changing the existing method name is breaking, as I seem to remember it being publicly exposed - please correct me if I'm wrong), maybe we just document this somewhere. I need to write a CRS usage page anyways, so maybe this could go in there?

If there is no non-rotatable version yet, then just leaving it without any mention of rotation seems good to me. As you said, we can always add an optional argument in the future, if necessary.

ibrierley:
Ah sorry, I've probably confused you there, no, there's just one method,
crs project/unproject. Ignore what I said earlier, was just talking more
about how I think about what I sometimes call the same thing. Forget I ever
said it :D.

JaffaKetchup:
Ok! I'm happy for whatever you think is best, as long as it's consistent and hopefully non-breaking!

JaffaKetchup:
@ibrierley Can you resolve the conflicts please

ibrierley:
oh weird, where did they come from. Will take a look

ibrierley:
Bleh its a bit messed up, so will do a new one.

PR #1332, opened 2022-08-02
Add polygon label rotate

jetpeter:
When set in the Polygon, rotation will keep the polygon labels facing the top of the display similar to rotation of a marker.  Rotation is about the center of the text.


JaffaKetchup:
@jetpeter I'm giving priority to #1333 for now, as it's a breaking change. Then I'll review and merge also as part of v3.0.0. Thanks for your contributions so far, and sorry for the delay :)

JaffaKetchup (approved):
LGTM! Thanks for your contributions!

PR #826, opened 2021-02-21
Optimize markers calculations

TheLastGimbus:
This PR is attempting to optimize and shrink all calculations that are done when user drags map with **a lot of** markers

Hopefully, I will get it to state where we have full 60 fps with 5k markers, but we will se how it goes

Fixes #824 ~~I would really want to~~ Yeah, pretty much - I don't see we how could do much better than this

Issue #824, opened 2021-02-17
Map gets super laggy when there are a lot of Markers


TheLastGimbus:
Hi there!

I am currently making an app which will have *a lot of* markers - around 4000 spread across whole country. After some testing, it got obvious that all of this gets super laggy

When I use all 3.5k markers and zoom on area with *no markers at all*, and start dragging the map, I get ~12 fps, on *phisical Android tablet in `profile` mode*

### After some debugging, `print()`ing and `Stopwatch()`ing, I traced which parts use the most time:

In `lib/src/layer/marker_layer.dart` -> `MarkerLayer` -> `build()`, there is a `for-loop` that iterates over all given `Markerks`:
https://github.com/fleaflet/flutter_map/blob/f09b193505ad31eb496b99a379346d99ec944cc7/lib/src/layer/marker_layer.dart#L130

#### The whole thing was taking ~25ms, and with `Stopwatch()`, I traced which parts of loop use the most time:
```dart
// I did it like this:
int sumStep = 0;
for(var i in list){
  var w = Stopwatch()..start();
  i.someStep();
  sumStep += w.elapsedMicroseconds();
  w.reset();
  // Of course, some precision is lost, because some steps take 1.something microsecond
}
print('Step: $sumStep');  // Summed microseconds
```
This was taking around 7500 us:
https://github.com/fleaflet/flutter_map/blob/f09b193505ad31eb496b99a379346d99ec944cc7/lib/src/layer/marker_layer.dart#L131

This 3600:
https://github.com/fleaflet/flutter_map/blob/f09b193505ad31eb496b99a379346d99ec944cc7/lib/src/layer/marker_layer.dart#L132-L133

And the rest around 0:
https://github.com/fleaflet/flutter_map/blob/f09b193505ad31eb496b99a379346d99ec944cc7/lib/src/layer/marker_layer.dart#L140-L152

### So I started to dig into `.project()`:
https://github.com/fleaflet/flutter_map/blob/f09b193505ad31eb496b99a379346d99ec944cc7/lib/src/map/map.dart#L417-L420

https://github.com/fleaflet/flutter_map/blob/f09b193505ad31eb496b99a379346d99ec944cc7/lib/src/geo/crs/crs.dart#L26-L34

`this.scale()` and `transformation.transform()` don't seem to be expensive - they are only some basic calculations...

So, `projection.project`:
[...]

PR #1333, opened 2022-08-03
v3.0.0: Simplify External & Internal Layer Handling

### LINKED ISSUE #1308: [BUG] Cannot Pan Vertically Inside `ListView`
**Describe The Bug**
Map inside ListView example not works as expected in android device.

**Expected Behavior**
Map scroll does not scroll the list view

**Screenshots & Recordings**

https://user-images.githubusercontent.com/36578847/179428019-3f8f6ff5-68cf-4c16-bf5a-77ce40383954.mp4

---

**Doctors Report**

```shell
[✓] Flutter (Channel stable, 3.0.3, on macOS 12.4 21F79 darwin-arm, locale
    es-419)
    • Flutter version 3.0.3 at /Users/alejo/Documents/flutter
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision 676cefaaff (4 weeks ago), 2022-06-22 11:34:49 -0700
    • Engine revision ffe7b86a1e
    • Dart version 2.17.5
    • DevTools version 2.12.2

[!] Android toolchain - develop for Android devices (Android SDK version 33.0.0)
    • Android SDK at /Users/alejo/Library/Android/sdk
    ✗ cmdline-tools component is missing
      Run `path/to/sdkmanager --install "cmdline-tools;latest"`
      See https://developer.android.com/studio/command-line for more details.
    ✗ Android license status unknown.
      Run `flutter doctor --android-licenses` to accept the SDK licenses.
      See https://flutter.dev/docs/get-started/install/macos#android-setup for
      more details.

[✓] Xcode - develop for iOS and macOS (Xcode 13.4.1)
    • Xcode at /Applications/Xcode.app/Contents/Developer
    • CocoaPods version 1.11.3

[✓] Chrome - develop for the web
    • Chrome at /Applications/Google Chrome.app/Contents/MacOS/Google Chrome

[✓] Android Studio (version 2021.2)
    • Android Studio at /Applications/Android Studio.app/Contents
    • Flutter plugin can be installed from:
      🔨 https://plugins.jetbrains.com/plugin/9212-flutter
    • Dart plugin can be installed from:
      🔨 https://plugins.jetbrains.com/plugin/6351-dart
    • Java version OpenJDK Runtime Environment (build
      11.0.12+0-b1504.28-7817840)

[✓] VS Code (version 1.69.0)
    • VS Code at /Applications/Visual Studio Code.app/Contents
    • Flutter extension version 3.44.0

[✓] Connected device (3 available)
    • SM A305G (mobile) • R58M84Q30VV • android-arm64  • Android 11 (API 30)
    • macOS (desktop)   • macos       • darwin-arm64   • macOS 12.4 21F79
      darwin-arm
    • Chrome (web)      • chrome      • web-javascript • Google Chrome
      103.0.5060.114

[✓] HTTP Host Availability
    • All required HTTP hosts are available
```

**To Reproduce**
Run the example app

---

**Severity**
This will help us to label the issue quicker and decide what needs attention first. Only choose fatal if the entire app crashes, otherwise choose non-fatal.

- [x] Non-Fatal
- [ ] Fatal / App Crashes

**Frequency/Rarity**
This will help us to label the issue quicker and decide what needs attention first.

- [ ] Once
- [ ] Uncommon
- [ ] Common
- [x] Always

**Applicable Platforms**
Only select those that you've tested on - one or more. If possible, test on a variety of platforms.

- [x] Android
- [ ] iOS
- [ ] Web
- [ ] Windows
- [ ] Others (beta platforms)

### LINKED ISSUE #1322: [BUG] Changing CRS Does Not Rebuild Other Children
### What is the bug?

hi there

flutter_map: 2.0.0

in my case, i have 2 diffenert crs epsg4490 and epsg3857 , let's say we set it as 4490 in the first time and has a polyline on map. when i press a button to switch crs to 3857, the polyline disappear untill i move the map.

the polyline still there when change crs.

```dart
FlutterMap(
              options: MapOptions(
                  crs: _currentMapIndex == 0 ? epsg4490 : const Epsg3857(),
                  center: LatLng(23.388149672864074, 116.7156679026907),
                  interactiveFlags:
                      InteractiveFlag.all & ~InteractiveFlag.rotate,
                  zoom: 14,
                  minZoom: 0,
                  maxZoom: 22),
              children: [
                ..._buildBaseMap(),
                PolylineLayerWidget(
                    options: PolylineLayerOptions(polylines: [
                  Polyline(strokeWidth: 4, color: Colors.red, points: [
                    LatLng(23.3884495, 116.7191143),
                    LatLng(23.3806962, 116.7124767),
                    LatLng(23.3799523, 116.7246074),
                    LatLng(23.3716553, 116.7208881),
                  ])
                ])),
                Positioned(
                  bottom: 20,
                  left: 20,
                  child: TextButton(
                    onPressed: () {
                      _currentMapIndex = _currentMapIndex == 0 ? 1 : 0;
                      setState(() {});
                    },
                    child: const Text("切换"),
                  ),
                )
              ],
            )
```

_No response_

_No response_

### Platforms Affected

Android, iOS

### Severity

### Frequency

### Requirements

### LINKED ISSUE #1334: [BUG] Markers not displaying on first map display (map doesn't have a size on certain devices).
### What is the bug?

When you look at the example HomePage, there should be 3 markers. Only one displays until you move the map. 

(This works fine on the web, but not on some devices, tested on Android Redmi Note 10).

All 3 markers should appear.

```dart
Just run the example home page.
```

This is the likely issue.

When we run the marker_layer code, we do the following, to try and eliminate drawing markers out of bounds.
```
if (!map.pixelBounds.containsPartialBounds(Bounds(sw, ne))) {
        continue;
}
```

However, if we look at the bounds, they are ``` Bounds(CustomPoint (4093.0, 2724.0), CustomPoint (4093.0, 2724.0)) ``` , so one pixel wide. So the markers get rejected apart from the one placed in the middle coincidentally.

The reason for that, is that in map.dart we call ```getPixelBounds()```
```
Bounds getPixelBounds(double zoom) {
    final mapZoom = zoom;
    final scale = getZoomScale(mapZoom, zoom);
    final pixelCenter = project(center, zoom).floor();
    final halfSize = size / (scale * 2);
    return Bounds(pixelCenter - halfSize, pixelCenter + halfSize);
  }
```

And here, on first run it has a size of ```(0.0, 0.0)``` 

So basically any code that calls things like ```pixelBounds.containsPartialBounds()``` won't work if it hasn't had chance to get a size (I think this is because the LayoutBuilder in flutter_map_state.dart context/constraints also has no size then, and even things like ```WidgetsBinding.instance.window.physicalSize``` returns 0,0).

There's probably some other subtle bugs all around this on first display.

Not sure of the best solution on this, as we've been here before. Maybe we should force FlutterMap to run 2 passes first time around, but then will it draw something wrong on the first pass, then right on the 2nd. Not quite sure of an elegant solution to this one (but there is some possibly boiler plate code that could be removed if we did).

Or possibly something like place it in an Offstage widget that doesn't draw first pass, then move it into view (could actually be a feature to make FlutterMap visible or not ? Not sure, just thinking out loud).

_No response_

### Platforms Affected

Android

### Severity

### Frequency

### Requirements

MooNag:
## What is it?
This PR removes LayerOptions and Plugins entirely. This has multiple benefits.

LayerOptions were used as a way to pass down map state like this:
```dart
class CircleLayerOptions extends LayerOptions {

	final List<CircleMarker> circles;
	CircleLayerOptions({
		Key? key,
		this.circles = const [],
		Stream<void>? rebuild,
	}) : super(key: key, rebuild: rebuild);
}
```
```dart
class CircleLayer extends StatelessWidget {

	final CircleLayerOptions circleOpts;
	final MapState map;
	final Stream<void>? stream;

	CircleLayer(this.circleOpts, this.map, this.stream) : super(key: circleOpts.key);
```
Then inside of the build the class would need to use a StreamBuilder
```dart
Widget _build(BuildContext context, Size size) {
	return StreamBuilder<void>(
		stream: stream, // a Stream<void> or null
			builder: (BuildContext context, _) {
```
This is a lot of boilerplate code; confusing, and also requires a plugin registration to handle the layer options.

Removing LayerOptions entirely makes everything easier and more intuitive.
```dart
class CircleLayerWidget extends StatelessWidget {
	final List<CircleMarker> circles;

	const CircleLayerWidget({super.key, this.circles = const []});
```
Now, the stream is not required. You can easily get map state within the build method.
```dart
@override
	final map = MapState.maybeOf(context)!;
```

## Migration of built in layers (This should be the most common migration)
Anything that uses a plugin will require more effort, but I have already migrated some community plugins.
**BEFORE**
```dart
layers: [
	TileLayerOptions(
		urlTemplate: 'examplexyz',
		tileProvider: NetworkTileProvider(),
	),
	MarkerLayerOptions(markers: markers)
],
```
**AFTER**
```dart
children: [
	TileLayer(
		urlTemplate: 'examplexyz',
		tileProvider: NetworkTileProvider(),
	)),
	MarkerLayer(markers: markers)
],
```

## TODO:
- [ ] Docs still need to be updated.
- [ ] TileLayer still listens to events and calls setState on itself. It could be migrated to only update on build.

## Other changes
- MarkerLayer widget has been improved with about a 10% performance increase while zooming with 5000 markers.
- removed map onReady
- removed streamGroups (internal and not functional)
- Fixed a bunch of state inconsistencies
- MapState is now FlutterMapState
- Map now initializes during initState which means no need to "wait" for the first build
- https://docs.fleaflet.dev/faqs/performance-issues this entire page is now incorrect (minus the last two points)

closes #1334, closes #1308, closes #1322

### Plugin specific migration
- .getPixelOrigin() is now pixelOrigin

### I have also created a PR for some plugins migrating them preemptively:
Everything is breaking unless the plugin already used children and did not depend on another plugin
- [x] flutter_map_tile_caching JaffaKetchup/flutter_map_tile_caching#61
- [x] flutter_map_marker_cluster (requires flutter_map_marker_popup) lpongetti/flutter_map_marker_cluster#128
- [x] flutter-vector-map-tiles greensopinion/flutter-vector-map-tiles#82
- [x] flutter_map_line_editor (requires flutter_map_dragmarker) ibrierley/flutter_map_line_editor#28
- [X] flutter_map_location_marker tlserver/flutter_map_location_marker#39
- [x] flutter_map_marker_popup rorystephenson/flutter_map_marker_popup#51
- [ ] map_elevation
- [ ] flutter_map_floating_marker_titles
- [ ] flutter_map_tappable_polyline
- [x] flutter_map_dragmarker ibrierley/flutter_map_dragmarker#17
- [ ] flutter_map_move_marker
- [ ] lat_lon_grid_plugin
- [ ] line_animator

## Extra Notes
I noticed almost a 1.5x performance increase in debug mode, I did not see much or any improvement in profile mode.

State is now handled "by flutter" using setState; which rebuilds all of the map children. Event stream is now a callback for widgets above FlutterMap

Consider any other breaking changes to make at the same time?


JaffaKetchup (changes requested):
As discussed on the Discord server. Also, we need to support older than Dart 2.17, so we (unfortunately) cannot use the `super.` constructor properties.

ibrierley:
So far it looks good to me (this isn't related as happens with existing version, but when testing I wonder why on the home page all 3 markers don't display immediately..).

Can you explain the Dart 2.17 ```super``` thing a bit more. I'm using that version currently....

JaffaKetchup:
https://dart.dev/guides/language/language-tour#super-parameters
Notice in how all of the `LayerWidgets`, `Key` is now a super parameter.

ibrierley:
Sorry, I misread, you said older than 2.17 didn't you, not 2.17 itself. Yes, sure, good point that makes sense (unless there's some real useful reason to force it).

MooNag:
@TheLastGimbus on your issue #824 and PR #826 you were able to fix by adding caching. I am working on a re-write of the state system to improve performance overall and found in my testing with 5000 markers on the example app that performance is better without the cache now due to how state is now handled with panning and zooming. Would you be willing to check this on your project to confirm that performance is as good or better now with my PR? Migration should be fairly easy and I can help if you have any issue.

MooNag:
Made breaking changes to state management. Behavior is almost the same as before, but the name has changed and internally it is much different. This means I will need to migrate plugins again.

Largest changes:
- MapState is now FlutterMapState
- Streams are now gone; except for map events. Map events may move or get changed still. They are still bulky and make state management more difficult.
- onMoved stream is now a callback for widgets above flutter map, and children (map layers) are automatically updated of state through build context.


Example app should have identical behavior to before; as I have not made any intentional modifications to behavior, other than ensuring state always updates, and the map now properly gets initialized in setState which means anything inside of state that is non-nullable should be initialized after initState, except for size which gets initialized after build. I have not made it nullable however.

JaffaKetchup (changes requested):
I found this small mistake. Haven't had time to test or profile yet though.

TheLastGimbus:
Hi @MooNag - thanks for this effort! I didn't check up on my project for some time, but I can try to see this in a few days :eyes:

See, because we had that many markers, we created our own marker layer: https://github.com/KanarekApp/flutter_map_fast_markers

fast_markers are drawn on pure canvas - and currently, we are just drawing literally **two dots** as marker :neutral_face: - and when I benchmarked it at the time, the map had problems even if i *disabled* the drawing! That's why i had issues to marker *calculations* themselves - but maybe things changed from then...

But the question is, now that plugins are gone, how can i test my project/add my plugin/markers to the map?

JaffaKetchup:
@MooNag Does the above change anything?

JaffaKetchup (changes requested):
So I've done some testing on Chrome using the example. It looks mostly good, but there are a few issues.

- The 'Point To `LatLng`' page no longer works: I'm assuming the underlying method has been broken, or the state isn't propagating as expected. I can't seem to see any errors in the console.
- The Interactive Flags page no longer reports events: again, I'm assuming the state isn't working as expected.
- Wasn't able to test the Tile Loading Error Handler page, as I'm doing this on Chrome.

In terms of performance, I couldn't see any noticeable difference, but I haven't used anything other than debug mode yet (ie. I haven't profiled). I'll try to do that tomorrow.

In terms of the recent conversation, it would be great to make any performance improvements that can be made, as this is a great opportunity to simplify and make use of Flutter's power.

Thanks for all the work so far, it's really great!

MooNag:
@TheLastGimbus 

More or less all you need to to do migrate is remove the plugin wrapper class; then in your widget that is put inside the map children, do FlutterMapState.maybeOf(context) inside of build to get the latest map state! The map should (clearly there are still some bugs to iron out), automatically rebuild your plugin whenever the map state is changed. Just like standard widgets in flutter.

I can help you migrate your plugin to this PR, which branch of it should I migrate? I see there are multiple branches!

MooNag:
I re-implemented the eventStream in the mapcontroller, however I might revert that change later. I did it exclusively for the "map_controller.dart" example page due to how it handles state. But it can easily cause bad practice especially with plugins. Ideally the mapcontroller can provide a registerCallback function to register multiple callbacks; however I have not figured out how to implement that yet.

@JaffaKetchup  Point To LatLng hero issue is in 2.2.0 for me. I did not fix the issue with this commit but I made sure the behavior on the page was the same in both versions. The other issues are fixed.

I don't really see any significant performance changes with this PR other than maybe improved jank and more potential for flutter to optimize. I noticed less jank in debug mode but not any performance difference in release mode. I did optimize the MarkerLayer (by removing the caching process entirely) to improve the frame time while zooming by about 10% and the frame time in general by 1%. I haven't done anything with the other layers, and still need to look at the fast marker plugin. Waiting on info for which branch to migrate.

TheLastGimbus:
@MooNag thanks for asking, but I propably need to sit with my project to see how it is doing now - i don't know when i'll have the time :confounded: . I will let you know if i will need any help

But if it interests you and you want to try on your own, `canary` branch is our own internal "master"

JaffaKetchup (approved):
Ok, I think I'm happy with this. Because it is such a big update, I will not merge it until @ibrierley also approves after he's tested.

We also need to update the documentation for this. I'm happy to do it, or you can write something and I'll upload it.

JaffaKetchup (changes requested):
Retracted approval, as discussed in Discord server.

ibrierley (commented):
Thanks for your work on this MooNag, all the examples test fine for me so far. I'll try and have a rummage through the code over the next day or so, but all looking good so far for me.

JaffaKetchup:
Now available as a prerelease on pub.dev: see [v3.0.0-beta.1](https://pub.dev/packages/flutter_map/versions/3.0.0-beta.1).

ibrierley:
I'm fine with all of this, thanks again. 

I've created beta version/branches of my plugins as a bit of a test (line_editor and dragmarker and my own vector test stuff, line_animator doesn't have any flutter_map dependency). Whilst it's a bit of work (hadn't spotted your conversions ready!), relatively painless. 

I guess this is one we should give people a lot of notification and try hard to encourage people to do some testing. Things that are a bit more integrated (maybe vector tile plugin) I suspect may need a bit more work, but nothing crazy.

MooNag:
Alright I am happy with where this PR is and am happy to merge. I will work on migrating plugins over the next day or two. Still need to write some docs and check for any last breaking changes before publishing. I have gotten 3 people to test in their apps and I have it working on my app, it should be quite stable at this point and closes 4 distinct issues.

JaffaKetchup:
Ok, in that case, I'm happy to merge. Thanks for all your hard work!

This is my plan of action:

1. Release as v3.0.0-beta.2
2. Give notice on Discord
3. a. Wait for feedback and conduct more testing if necessary
b. Try to merge the other PRs that are waiting
4. Write CHANGELOG and publish new docs
5. Release as v3

JaffaKetchup (approved):
LGTM for @ibrierley and I.

PR #1348, opened 2022-08-28
Add Fallback URLs (for #1203)

### LINKED ISSUE #1203: [FEATURE] Fallback tile URLs in-case of HTTP error
**Describe The Problem**
I'm try to catch an error callback when the tile failed, so, the idea is when the first tile url map failed, called another, but this method errorTileCallback doesn't call, so, don't find in examples an example, maybe someone know somehow? Thanks

**Additional Information**
TileLayerOptions getMapBoxStyle(Function refresh, {Key key}) =>
    TileLayerOptions(
        key: key,
        urlTemplate: globalTileMap,
        errorTileCallback: (tile, error) {
          globalTileMap =
              'https://tiles.picap.app/styles/osm-bright/{z}/{x}/{y}.png';
          refresh.call();
        },
        backgroundColor: PiColors.backgroundTileMap);

TesteurManiak:
Fix #1203 

- Updated `NetworkTileProvider` and `NetworkNoRetryTileProvider` from `tile_provider_io.dart` & `tile_provider_web.dart` to support a `fallbackUrl`
    - Replaced references to `RetryClient` by a simple `Client` class as I think it is not needed to enforce a retry client, we should simply be able to use any `Client` class we want (+ allows to use `FMNetworkImageProvider` inside `tile_provider_web.NetworkNoRetryTileProvider.getImage`)
- Putted back `AssetTileProvider` which was missing for some unknown reasons and added support for `fallbackUrl` by implementing a private class `_FlutterMapAssetBundle` which will try to load an asset in the same way as the default `PlatformAssetBundle`, but will also try to load a fallback asset if the first one wasn't found.
- 2 examples have been added: `FallbackUrlNetworkPage` & `FallbackUrlOfflinePage`

Changes have been tested on Android and Chrome, no regressions were found.


JaffaKetchup:
Thanks @TesteurManiak! I'll have a look at this for v3.1.0 :)

TesteurManiak:
Hi, @JaffaKetchup I've merged the latest changes from master on the branch.
I kept the `AssetTileProvider` implementation inside it's own separated file with the `_FlutterMapAssetBundle`. (Seemed better for readability)

JaffaKetchup (changes requested):
See the changes I suggested below.
Can you also bring this up to date with v3 if needs be? For example, I know the `AssetTileProvider` was added back and this might cause problems.

TesteurManiak:
@JaffaKetchup everything should be ok now 👍

JaffaKetchup (approved):
LGTM! Thanks.

PR #1351, opened 2022-08-30
Use preferred tile.openstreetmap.org URL

Firefishy:
https://github.com/openstreetmap/operations/issues/737

Signed-off-by: Grant Slater <git@firefishy.com>


JaffaKetchup (approved):
LGTM. I will update the documentation with this info :)

Issue #1338, opened 2022-08-11
[BUG] Crossing W/E Extremity With `Polyline` Results In Unexpected Lines


WorldpixelSoftware:
### What is the bug?

I was trying to implement a great circle map between two points using the OSM Map.

All is good as long as you don't cross 180E/W. Then the line suddenly jumps to the other side of the map.

![flutter_map_bug](https://user-images.githubusercontent.com/77431217/184257980-c3f2403b-5b70-4dc5-861b-83928b593485.png)

### What is the expected behaviour?

Having a closed line between the two points. On low zoom levels (like OSM 0), there should be just two renders of any layer at a given coordinate.

### Do you have a potential solution?

Maybe mirroring the coordinates is possible somehow? So that a marker or a line is shown twice on low zoom levels.

### Platforms Affected

Android, iOS, Windows, MacOS

### Severity

Obtrusive: Prevents normal functioning but causes no errors in the console

### Frequency

Consistently: Always occurs at the same time and location

### Requirements

- [X] I agree to follow this project's [Code of Conduct](https://github.com/fleaflet/flutter_map/blob/master/CODE_OF_CONDUCT.md)
- [X] My Flutter/Dart installation is unaltered, and `flutter doctor` finds no relevant issues
- [X] I am using the [latest stable version](https://pub.dev/packages/flutter_map) of this package
- [X] I have checked the FAQs section on the documentation website
- [X] I have checked for similar issues which may be duplicates

JaffaKetchup:
Hi there, and thanks for the report! 
We've known about this issue for a while, but have never tried to resolve it as it seems to be a rare-ish occurrence. Can you try splitting your line into two parts, where one starts/ends at 180 and the other at -180?

KKStudios:
@JaffaKetchup Yeah splitting the lines was basically my workaround for this situation. I just think that looks weird though.

What would be really great is to have a map view, that has no physical boundaries left and right, basically like when you go on google maps and pan from one side to the other and the map just "rotates" in view

I've included two pictures of what I would consider expected behaviour for a map projection.

But thank you for your time and effort that went into this project!

![maps_ew_crossing](https://user-images.githubusercontent.com/77431217/185460713-d1e11bef-fce5-41da-9822-f744826d6781.png)
![gcmapper_ew_crossing](https://user-images.githubusercontent.com/77431217/185460737-0fdac27d-cba1-421e-8f4a-76283facfbe8.png)
.

[...]

Issue #1337, opened 2022-08-11
[FEATURE] Add Option To Preload Tiles Around Current Area


JaffaKetchup:
### What do you want implemented?

A property that can take a number of tiles in each direction outside of the current viewport to preload and render for smoother appearance.
Additionally (if possible), a method to preload and render a list of tile coordinates or URLs.

### Can you provide any other information?

Related to #1336.

### Platforms Affected

Android, iOS, Web, Windows, MacOS, Linux

### Severity

Obtrusive: No workarounds are available, and this is essential to me

### Requirements

- [X] I agree to follow this project's [Code of Conduct](https://github.com/fleaflet/flutter_map/blob/master/CODE_OF_CONDUCT.md)
- [X] I am using the [latest stable version](https://pub.dev/packages/flutter_map) of this package
- [X] I have checked for similar feature requests which may be duplicates

SalsabeelaHasan:
Hello ! any updates?

JaffaKetchup:
Not yet, we will wait on a greater rewrite first.

SalsabeelaHasan:
All the best.
if there is clear timeline, it would be great. thank you!

JaffaKetchup:
Unfortunately no timeline at the moment. We maintain this in our spare time, and at the moment, we are quite busy.

If your business/employer is waiting for this, it may help to offer a bounty to @MooNag (probably best to contact him via Discord), as he was the one originally suggesting the rewrite - but I'm not him so I can't say. Of course, absolutely 0 requirement/pressure to do this (we are open source after all), but it has worked in the past to fix a rather specific bug that someone needed.
At the moment, I'm far too busy to do any major work here - I've got exams approaching.

ibrierley:
it shouldn't be too fiddly I would have thought if someone needs it to have a bash and do a PR. Pass in a param for number of extra tiles in each direction. Then subtract this from the min extent of the tile range, and add to the max in both x,y directions. Eg around line 605 of [
](https://github.com/fleaflet/flutter_map/blob/master/lib/src/layer/tile_layer/tile_layer.dart)
```
// create a queue of coordinates to load tiles from
    for (var j = tileRange.min.y; j <= tileRange.max.y; j++) {
      for (var i = tileRange.min.x; i <= tileRange.max.x; i++) {
```
Add a test that the values aren't below 0 or above the max tile extent for that xoom.

JaffaKetchup:
[...]

PR #1356, opened 2022-09-05
Simplify tiles

JosefWN:
A suggestion to simplify the tile handling a bit:
- Move `AnimatedTile` into former `TileWidget` (no need for state, the tiles are managed by the `TileManager`), the `AnimationBuilder` does the heavy lifting for us
- Do not recreate `AnimationController` for every animation

Haven't gotten rid of the opacity widget yet, I think we'd need to crossfade between old and new tile on zoom


JaffaKetchup:
@JosefWN
Hi there, just getting round to looking at all of these.
Rather than dealing with our own animation controller and ticker etc., can we just use a [TweenAnimationBuilder](https://api.flutter.dev/flutter/widgets/TweenAnimationBuilder-class.html)?

JosefWN:
One reason I kept the animation controller is that we allow for this:

```dart
animationController?.forward(from: from);
```

So you can choose to play the animation from, say, 50% to 100%. I don't know why this is necessary, but we have exposed the option to the users as well (`TileLayer.tileFadeInStart`). I could have another look at if we can further simplify if we were to remove the option?

JaffaKetchup:
I see. I guess it depends on how much things can or cannot be simplified, and how often it's used. I'd say if it's not used in the example app, then we can probably get rid of it: although this will be a breaking change. @ibrierley @MooNag @TesteurManiak What do you guys think?

TesteurManiak:
IMO we can replace with a `TweenAnimationBuilder` the use of the `AnimationController` seems too specific.

ibrierley:
I'm a bit hesitant to do breaking changes atm (we've had quite a lot of late), but no problems with these type of changes implemented in a future version where there's several breaking changes if that makes sense.

JaffaKetchup:
Yeah, I would agree with that. I think leaving an at least 2-3 month gap between breaking releases is the minimum.

In this case, @JosefWN can you explore how much you can simplify things with this option - and any similar options you think fall under the same bracket - and maybe we merge this into a separate branch 'delayed-breaking', which we can merge into main and release at a later date.
Of course, if you don't have time, let us know and we'll just merge this for now.

Thanks :)

MooNag:
I know the tile layer stuff is all kinda spaghetti together and hard to keep separate, but we could look at making this the migration to a new tile layer system. Give it a new name, RasterTileLayer, WMSTileLayer, VectorTileLayer, etc.... And then deprecate the old stuff. This would allow for non-breaking changes and allow us to basically rewrite the entire tiling system. I don't think this change has to be breaking for end users.

One thing I would like to see be worked on, I don't have the time right now, but support wrapping on the map left and right; and an extension of that would be to extend the system to support rendering on a sphere.

JaffaKetchup:
I'm not sure, if you deprecate something in 1.2, and replace it in 1.3, is that OK with SemVer? Or would you need to move it 2.0? Are we even sticking with SemVer (we are atm, and it seems kind of sensible to remain doing so)?

But yes, I would like to see that rewrite and separation as well. This might even pave the way for #1337. But alas, I don't have the time (or much knowledge) to do so either at the moment.
I wonder how @greensopinion would feel about helping out, do you have time? You seem to be pretty knowledgeable about this stuff - and the naming/separation scheme could benefit your plugin, if you see what I mean.

Should we be making this aim more formal? Ie. use a Discussion, Discord channel, and/or Git branch.

---

> support wrapping on the map left and right

See #1338, #1201

> extend the system to support rendering on a sphere

I think someone had a crack at 3D rendering, but I think they ran out of time.

greensopinion:
> I wonder how @greensopinion would feel about helping out, do you have time?

Hey, thanks for the ping. I'm pretty busy with other things at the moment, unfortunately I don't have any spare cycles to chip in.

> I'm a bit hesitant to do breaking changes atm

Improvements are great, but any breaking change is a pain for users and for plug-in maintainers. Unless there's substantial benefit, we should try to avoid it as much as possible. I know that might sound like a drag, but it's a big deal. For example, going to 3.0.0 was a lot of work. For the VectorTileLayer I had to understand the nature of the change and then get it working. It cost me about a full day, and although the code is a little cleaner on the consumer side, it's an incremental benefit. I'm not sure that it's worth it to break the API for such a minor improvement.

I'm not sure what's planned for the new tile layer system. There is quite a lot of logic in [vector_map_tiles](https://github.com/greensopinion/flutter-vector-map-tiles) related to loading tiles of different sizes, depending on the data that's available and to optimize the frame rate. It would be great if we could avoid breaking that.

JaffaKetchup:
Thanks anyway @greensopinion :)

Yeah, understood with the breaking changes. For me, I don't mind them so much, but I can see & understand they are a pain for a lot of people, especially when the changes are as big as last time.
As I mentioned above, this PR will probably go into a delayed breaking changes branch, until we have accumulated them over maybe 2-3 months. This change should break much anyway - I can't see why anyone would be using it.

Those greater tile system changes probably won't be coming any time soon, so no need to worry about them for now.

JaffaKetchup:
@JosefWN I've converted to a draft just for now, so we can differentiate between needs-review/in-progress/ready. If you are ready to merge, let me know.

JaffaKetchup:
Hi @JosefWN, is this ready for review? If not, no pressure!

MooNag:
I propose we just work on new tile layer stuff from the ground up, give it a new widget (ideally plural with different behaviors) with new logic. That is realistically the best way to improve the tile handling at this point. The current system is not extensible, has too many edge cases, and way too many options.

JaffaKetchup:
@MooNag I agree with this, however I'm not sure when this would get finished?

JaffaKetchup:
Hey @MooNag, is there any chance of that rewrite happening any time soon? If not, we should probably get round to merging this.

MooNag:
not right now unfortunately unless someone else wants to take it on

JaffaKetchup:
Ok. @ibrierley When you're back, do you want to review this for merge? I'll do the same.

ibrierley (approved):
I'm a little out of touch with the tile handler recently, but I like the visible difference. My acid "feel" test is to open Animated MapController example in Web release, and repeatedly animate from London <-> Paris. It certainly feels better to me with this change (mainly less slight outer grey on animation). 

Regarding the code itself, I'm fairly easy and can't initially spot any issues, but it's hard with the tile handler as it's typically been a mess to follow, but maybe that's just me:D. I'm happy with this.

Thanks for the work on this!

JaffaKetchup:
@ibrierley I can't seem to find where this is a breaking change?

ibrierley:
Good question, I'm not actually sure if this is or not now...there was mention of the fadeInStart I think...

Issue #1346, opened 2022-08-27
[BUG] Invalid cast in getScaleZoom


JosefWN:
### What is the bug?

By calling the following:
```dart
final zoom = map.getScaleZoom(1.0, 10.0);
```

I'm experiencing an invalid cast in `getScaleZoom` (type 'int' is not a subtype of type 'double' in type cast):
```dart
double getScaleZoom(double scale, double? fromZoom) {
  final crs = options.crs;
  fromZoom = fromZoom ?? _zoom;
  return crs.zoom(scale * crs.scale(fromZoom)) as double; // <- here
}
```

It's a bit puzzling, it looks as if the abstract class is returning a `double`, but perhaps an implementing class is overriding the `zoom` method and returning an `int`. Not sure if other parts of the code are affected by this assumption as well.

### What is the expected behaviour?

The code should not break.

### How can we reproduce this issue?

This is on `3.0.0-beta3` (so I'm not running the latest stable, as I had to guarantee in this somewhat rigid template), running Flutter `3.0.5` (macOS, ARM).

### Do you have a potential solution?

By editing `flutter_map` in my pub cache I could get around the issue with a convert instead, but perhaps the return type of `crs.zoom` should not be `num` if we really expect it to be a `double`?

```dart
  double getScaleZoom(double scale, double? fromZoom) {
    final crs = options.crs;
    fromZoom = fromZoom ?? _zoom;
    return crs.zoom(scale * crs.scale(fromZoom)).toDouble();
  }
```

### Platforms Affected

MacOS

### Severity

Erroneous: Prevents normal functioning and causes errors in the console

### Frequency

Rarely: Occurs every so often

### Requirements

- [X] I agree to follow this project's [Code of Conduct](https://github.com/fleaflet/flutter_map/blob/master/CODE_OF_CONDUCT.md)
- [X] My Flutter/Dart installation is unaltered, and `flutter doctor` finds no relevant issues
- [X] I am using the [latest stable version](https://pub.dev/packages/flutter_map) of this package
- [X] I have checked the FAQs section on the documentation website
- [X] I have checked for similar issues which may be duplicates

JaffaKetchup:
Similar to https://github.com/fleaflet/flutter_map/issues/1296

PR #1357, opened 2022-09-05
Reduce unnecessary casts, use convert when unsafe

JosefWN:
- Fixes https://github.com/fleaflet/flutter_map/issues/1346
- `LatLngBounds.fromPoints` converted degrees to radians, then radians back to degrees... Also used num and nullables although we were only dealing with doubles.
- Removed some unnecessary casts coming from inferred types (where double can be inferred rather than num).

Not sure if we can guarantee that all `CustomPoint<num>` are in fact doubles, but left them for now since it would be a bigger change.

### DIFF
--- a/example/lib/pages/scale_layer_plugin_option.dart
+++ b/example/lib/pages/scale_layer_plugin_option.dart
@@ -101,18 +101,16 @@ class ScalePainter extends CustomPainter {
 
     const sizeForStartEnd = 4;
     final paddingLeft =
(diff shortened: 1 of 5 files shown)


TesteurManiak (approved):
Looks good to me, @JaffaKetchup do you want to double check?

JaffaKetchup:
I'll take your word for it, I'll merge when I get a minute.

JaffaKetchup (approved):
Submitting approval on behalf of @TesteurManiak

PR #1359, opened 2022-09-07
Upgrade dependencies, fix EPSG:3413 example

JosefWN:
Upgrades the project to the latest compatible dependencies.

Also adds `assets/` path prefix to a broken `Image.asset()` in the EPSG:3413 example.

Tested on all examples, seems to work except for https://github.com/fleaflet/flutter_map/issues/1360.

### DIFF
--- a/example/lib/pages/epsg3413_crs.dart
+++ b/example/lib/pages/epsg3413_crs.dart
@@ -154,7 +154,7 @@ class _EPSG3413PageState extends State<EPSG3413Page> {
                           LatLng(85.2802493, 79.794166),
                         ),
                         imageProvider: Image.asset(
-                          'map/epsg3413/amsr2.png',
+                          'assets/map/epsg3413/amsr2.png',
                         ).image,
                       )
                     ],
--- a/example/pubspec.yaml
+++ b/example/pubspec.yaml
@@ -4,7 +4,8 @@ publish_to: "none"
 version: 1.0.0
 
 environment:
-  sdk: ">=2.17.0 <3.0.0"
+  sdk: ">=2.18.0 <3.0.0"
+  flutter: ">=3.3.0"
 
 dependencies:
   flutter:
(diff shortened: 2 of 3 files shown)


JaffaKetchup:
Hi @JosefWN,
`flutter pub outdated` says that all dependencies are up to date without this PR:
image
image
However, I can see that your versions in this PR are the latest.
Do you have an idea why this might be?
Many thanks.

JosefWN:
Ah: https://dart.dev/tools/pub/dependencies

> For example, `^1.2.3` is equivalent to `>=1.2.3 <2.0.0`, and `^0.1.2` is equivalent to `>=0.1.2 <0.2.0`.

I assumed `^` would only allow patch version bumps regardless of major version, so I've effectively only made the dependency spec a little stricter.

Two things which might be good:
- Bumped to Dart 2.18 for `example/` just to align with the `flutter_map` spec.
- If I got it right, `flutter_lints: ">=1.0.4"` will allow any version greater than or equal to 1.0.4 (including future 3.0.0), so it might be safer to use the caret syntax there too like I did in my PR?

JaffaKetchup:
Good spot on the example project!

Still not sure why they would be showing differences on our systems, as they almost all use carat syntax anyway?

But I think your second point is an issue: I would prefer carat syntax.

JosefWN:
> Still not sure why they would be showing differences on our systems, as they almost all use carat syntax anyway?

I _think_ it will depend on when we last ran `flutter pub upgrade` or equivalent, so by bumping dependency versions in the spec we ensure that everyone using `flutter_map` are using these later versions.

JaffaKetchup:
Ok, I'm happy with this. Can someone else (@ibrierley, @TesteurManiak, @MooNag) give this a check to make sure it runs fine on their system? (Note the SDK version change)

TesteurManiak:
LGTM

JaffaKetchup (approved):
LGTM. Thanks for your changes!

PR #1365, opened 2022-09-19
Restrict tile layer with bound radius

nhtlquan:
If want restrict tile layer with bound radius. Add centerDraw and radiusMeter to TitleLayerWidget.

image

### DIFF
--- a/lib/src/layer/tile_layer/circle_clipper.dart
+++ b/lib/src/layer/tile_layer/circle_clipper.dart
@@ -1,0 +1,17 @@
+import 'dart:ui';
+
+import 'package:flutter/material.dart';
+class CircleClipper extends CustomClipper<Path> {
+  final double radius;
+  final Offset positionOffset;
+
+  CircleClipper(this.positionOffset, this.radius);
+
+  @override
+  Path getClip(Size size) {
+    return Path()..addOval(Rect.fromCircle(center: positionOffset, radius: radius));
+  }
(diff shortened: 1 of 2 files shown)


Robbendebiene:
I'm not a maintainer, but I feel like a generic solution that accepts a `CustomClipper` would be a better fit for a library. The circle only implementation feels too specific.

Maybe something like this:

```dart
abstract class TileLayerClipper<T> extends CustomClipper<T> {
  late final FlutterMapState  map; // needs to be initialized from the outside, there is probably a better way to do this
}
```
Any user can then create its own custom clipper shape like this:

```dart
class CircleClipper extends TileLayerClipper<Path> {
  @override
  Path getClip(Size size) {
    // do math and projection here based on the `map` variable.
    // calc and return shape
  }

  @override
  bool shouldReclip(CustomClipper<Path> oldClipper) => true;
}
```

The tile layer widget could then accept a property of type `TileLayerClipper<T>`.

JaffaKetchup:
Hi @nhtlquan, and thanks for your PR!

In this case, I think an approach similar to what @Robbendebiene suggested would be better.

Would it be possible for you to implement this as a plugin? If not, please try their suggestion here - if you don't have time, just let me know.

JaffaKetchup:
Hi @nhtlquan, have you had any time to consider the above comments? If not, no pressure!

PR #1367, opened 2022-09-19
Add parameter to fit to integer zoom levels

Robbendebiene:
Since flutter map is mostly used with pixel tiles, it is often desirable to use integer zoom levels in order to avoid any pixel tile upscaling. Currently this isn't possible using the `fitBounds` and `centerZoomFitBounds` methods.

This PR adds an additional parameter to the `FitBoundsOptions` called `forceIntegerZoomLevel` in order to force the calculation of applicable integer zoom levels of the aforementioned functions.

### DIFF
--- a/test/flutter_map_controller_test.dart
+++ b/test/flutter_map_controller_test.dart
@@ -1,0 +1,145 @@
+import 'package:flutter/material.dart';
+import 'package:flutter_map/flutter_map.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:latlong2/latlong.dart';
+
+void main() {
+  testWidgets('test fit bounds methods', (tester) async {
+    final controller = MapController();
+    final bounds = LatLngBounds(
+      LatLng(51, 0),
+      LatLng(52, 1),
+    );
+    final expectedCenter = LatLng(51.50274289405741, 0.49999999999999833);
+
+    await tester.pumpWidget(TestApp(controller: controller));
+
+    {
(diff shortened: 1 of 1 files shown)


TesteurManiak (approved):
LGTM, if everybody is okay with this feature it can be merged

TesteurManiak (approved):

ibrierley (commented):
Do you have any code to show the effectiveness of this at all ?

Robbendebiene:
@ibrierley I require this in a project I'm working on, which is also the place where I've tested it. What kind of example do you have in mind? From a degree of effectiveness it is similar to the ``inside`` parameter.
Normally one can simply round the zoom level outside of the flutter map code basis, but this doesn't work when fitting to bounds.

ibrierley:
I'm just thinking of a way to test it is all, so maybe future code changes can use to check its still working (don't really mean a full example page, but just something people can use to highlight its working or not.

Robbendebiene:
@ibrierley I've added a test that checks the fit bounds methods for some parameters. The change of the expected zoom values and the consistent center point give an impression of what is going on. At least this test can fail in the future if any internal implementation details change.

JaffaKetchup:
@ibrierley This good for you?

ibrierley (approved):
Can't see any obvious problems if others are happy with it!

JaffaKetchup:
Sure, you merge it if you're happy.

PR #1368, opened 2022-09-19
Deprecate obsolete fit parameter

Robbendebiene:
I always wondered what the use of the `zoom` parameter of the `FitBoundsOptions` class was.
When working on #1367 I finally had the chance to look into it. As it turns out this parameter is never used nor is it documented.

Since this is a somewhat breaking change I made a separate PR.

### DIFF
--- a/lib/flutter_map.dart
+++ b/lib/flutter_map.dart
@@ -333,13 +333,11 @@ class MapOptions {
 class FitBoundsOptions {
   final EdgeInsets padding;
   final double maxZoom;
-  final double? zoom;
   final bool inside;
 
   const FitBoundsOptions({
     this.padding = EdgeInsets.zero,
     this.maxZoom = 17.0,
-    this.zoom,
     this.inside = false,
   });
 }
--- a/lib/flutter_map.dart
+++ b/lib/flutter_map.dart
@@ -333,11 +333,16 @@ class MapOptions {
 class FitBoundsOptions {
   final EdgeInsets padding;
   final double maxZoom;
+  @Deprecated('This property is unused and will be removed in the next major release.')
+  /// TODO: remove this property in the next major release.
+  final double? zoom;
   final bool inside;
 
   const FitBoundsOptions({
     this.padding = EdgeInsets.zero,
     this.maxZoom = 17.0,
+    @Deprecated('This property is unused and will be removed in the next major release.')
+    this.zoom,
     this.inside = false,
   });
 }


JaffaKetchup:
(dammit not another one :D)

We'll get round to looking at this, please don't take the silence badly :)

Robbendebiene:
I knew that this PR wouldn't be welcomed with joy. 😅

Since this has no importance at all it would probably make sense to delay this PR till another more relevant breaking change kicks in. But maybe it also turns out that the removed property has an actual usage.

JaffaKetchup:
Nah, I was more annoyed that someone found yet another 'remnant of the old world' as I call them!
Where did it come from? Where does it go? Nobody knows :D

Thanks for the PR, they are always appreciated. As this literally did nothing, no one should be using it (or if they are, they shouldn't care that it's gone), and therefore I wouldn't call it a breaking change, just a minor change (v0.x.0).

I'm quite busy atm, so I'll see what I can get done when :)

TesteurManiak:
I have a suggestion, as this is a "breaking change" even if the parameter is not used, it could be considered to deprecate it and force to use `MapOptions.zoom` instead, also adding a comment as a reminder to remove this property for the next major release.
By doing so the PR can be merged  immediately and there won't be breaking change until further changes.

```dart
class FitBoundsOptions {
  final EdgeInsets padding;
  final double maxZoom;
  final bool inside;

  const FitBoundsOptions({
    this.padding = EdgeInsets.zero,
    this.maxZoom = 17.0,
    @Deprecated('This property is never used, use MapOptions.zoom instead.')
        double? zoom,
    this.inside = false,
  });
}
```

JaffaKetchup:
Sure, that seems sensible.

Robbendebiene:
Sounds good to me. The only question I have is, should we really point to the `MapOptions.zoom` property? As far as I know this is just setting the default/initial zoom value. Honestly I have no clue what this property was meant to do, but I doubt that it was forcing any zoom value since this would break the functionality of the fit bounds method.

TesteurManiak:
Then I guess you can just put: `This property is unused and will be removed in the next major release.` 🤔

Robbendebiene:
I've added two deprecation notices so we get:

grafik

TesteurManiak (approved):
LGTM

JaffaKetchup (approved):
Approving on behalf of @TesteurManiak.

Issue #1362, opened 2022-09-13
[BUG] Can't do any interaction on the map if I pass bounds param


tijee:
### What is the bug?

If I pass `bounds` to my FlutterMap, I cannot move or zoom anymore. If I remove the `bounds` (and pass a `center` for example) I can move, zoom, rotate, as normal.

### What is the expected behaviour?

I would expect to be able to move or zoom in the map even when passing an initial `bounds`.

### How can we reproduce this issue?

```dart
FlutterMap(
          options: MapOptions(
            bounds: LatLngBounds.fromPoints([......]),
            minZoom: 0,
            maxZoom: 18,
          ),
          children: [
            TileLayer(
              urlTemplate:
                  'https://api.maptiler.com/maps/basic-v2/{z}/{x}/{y}.png?key=......',
            ),
          ],
)
```

### Do you have a potential solution?

Pass a `center` instead of the `bounds` but I would also need to calculate the default `zoom` level matching the bounds I want.

### Platforms Affected

Android

### Severity

Obtrusive: Prevents normal functioning but causes no errors in the console

### Frequency

Consistently: Always occurs at the same time and location

### Requirements

- [X] I agree to follow this project's [Code of Conduct](https://github.com/fleaflet/flutter_map/blob/master/CODE_OF_CONDUCT.md)
- [X] My Flutter/Dart installation is unaltered, and `flutter doctor` finds no relevant issues
- [X] I am using the [latest stable version](https://pub.dev/packages/flutter_map) of this package
- [X] I have checked the FAQs section on the documentation website
- [X] I have checked for similar issues which may be duplicates

muhammadSaqib7860:
I have facing the same issue , any solution.

JaffaKetchup:
Thanks for the heads up! We'll look into this.

muhammadSaqib7860:
Kindly resolve it asap, I have a deadline so need to finish it asap.

ibrierley:
If you're desperate, revert to < 3.0.0 assuming thats ok

JaffaKetchup:
@muhammadSaqib7860

Many of us maintainers have jobs as well or things to do other than this, so please be patient!

In the meantime, please send a minimal reproducible example and a screen recording of it. This will allow us to verify this issue and assign it high priority.

muhammadSaqib7860:
Container(
              width: double.infinity,
              height: double.infinity,
              child: FlutterMap(
                mapController: mapController,
                options: MapOptions(
[...]

PR #1369, opened 2022-09-20
Fix `bounds` Issue (for #1362)

ibrierley:
Possible fix for the recent ```bounds``` issue. Needs proper testing before merging as I'm not quite familiar with bounds and recent changes. 

What this does is remove the fitBounds in the build method (I'm not sure this is needed there at all, as long as its run once?), and move it just to the initState (but it looks like it needs to run in the WidgetsBinding.instance.addPostFrameCallback for it to take effect...I'm not entirely sure if this is best or if its better before/after options.onMapReady?.call()).

Few thoughts...should ```bounds``` (I think we should think of it as initialFitBounds if my understanding of it is correct, and maybe rename one day) only run here...are there any other situations when we need to reset bounds ? Are there any possible race conditions where this is in the wrong place.

We can test the basics with something like this in one of the examples...
```
options: MapOptions(
                    bounds: LatLngBounds(
                      LatLng(56.6877, 11.5089),
                      LatLng(56.7378, 11.6644),
                    )
                ),
```

### DIFF
--- a/lib/src/map/flutter_map_state.dart
+++ b/lib/src/map/flutter_map_state.dart
@@ -42,11 +42,11 @@ class FlutterMapState extends MapGestureMixin
 
     move(options.center, zoom, source: MapEventSource.initialization);
 
-    // Funally, fit the map to restrictions
-    if (options.bounds != null) {
-      fitBounds(options.bounds!, options.boundsOptions);
-    }
     WidgetsBinding.instance.addPostFrameCallback((_) {
+      // Finally, fit the map to restrictions
+      if (options.bounds != null) {
+        fitBounds(options.bounds!, options.boundsOptions);
+      }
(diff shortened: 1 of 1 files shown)


JaffaKetchup:
I will try to have a look into this tomorrow when I get a moment, as this is high priority. Or if you're totally happy with it, let me know and I'll just approve and merge.

ibrierley:
Would prefer others to test really on this one, as there may be some edge cases I haven't thought of, especially with initialisation and timing. 
Would also be helpful if those who had the problem could give it a test to check it fixes it for them.

JaffaKetchup (approved):
LGTM. It's resolved the issue, and I cannot see why it would have side effects.

ibrierley:
Great, thanks.

Issue #1373, opened 2022-09-23
[BUG] when onPositionChanged is called because of controller.move the position is before the move


SecretUnicorn:
### What is the bug?

When moving via controller.move() the onPositionChanged position attribute does not hold the correct bounds. The southwest and northeast corner are the same. When slightly moving the map again after the move the bounds are correct.

### What is the expected behaviour?

The position attribute of the controller should hold the correct bounds

### How can we reproduce this issue?

```dart
FlutterMap(
      options: MapOptions(
        onPositionChanged: onMapChanged,
      ),

mapController.move(
                pointOfIntrest,
                mapController.zoom,
              );

 void onMapChanged(MapPosition position, bool gesture) {
    if (_mapChangeDebounce?.isActive ?? false) {
      _mapChangeDebounce!.cancel();
    }
    _mapChangeDebounce = Timer(
      const Duration(
        milliseconds: 500,
      ),
      () {
        Provider.of<GeoProvider>(context, listen: false)
            .fetchSearchesInBoundingBox(
          position.bounds!.southWest!,
          position.bounds!.northEast!,
        );
      },
    );
```

### Platforms Affected

Android

### Severity

Obtrusive: Prevents normal functioning but causes no errors in the console

### Frequency

Consistently: Always occurs at the same time and location

### Requirements

- [X] I agree to follow this project's [Code of Conduct](https://github.com/fleaflet/flutter_map/blob/master/CODE_OF_CONDUCT.md)
- [X] My Flutter/Dart installation is unaltered, and `flutter doctor` finds no relevant issues
- [X] I am using the [latest stable version](https://pub.dev/packages/flutter_map) of this package
- [X] I have checked the FAQs section on the documentation website
- [X] I have checked for similar issues which may be duplicates

PR #1374, opened 2022-09-23
fix for 1373, where bounds doesn't get updated after a mapController …

ibrierley:
For https://github.com/fleaflet/flutter_map/issues/1373

I think setState gets called, but the onPositionChanged gets called before _bounds is updated (inside build). I've also added an extra check for a valid bounds check to be called on a test new bounds (we don't want to update _bounds if the new position is erm out of bounds).

### DIFF
--- a/lib/src/map/flutter_map_state.dart
+++ b/lib/src/map/flutter_map_state.dart
@@ -418,7 +418,7 @@ class FlutterMapState extends MapGestureMixin
     newZoom = fitZoomToBounds(newZoom);
     final mapMoved = newCenter != _center || newZoom != _zoom;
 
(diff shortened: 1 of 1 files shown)


JaffaKetchup:
If you're happy with this, I'll just approve it. Or I can test it tomorrow (but I haven't been keeping up with the conversation).

ibrierley:
Give it a day anyway, in case someone spots a flaw.

JaffaKetchup:
@ibrierley You happy for this to be merged?

ibrierley:
Yes, lets update I guess, and fix if anyone screams.

JaffaKetchup (approved):
LGTM. Thanks!

PR #1361, opened 2022-09-13
update vector_map_tiles docs

greensopinion:
Remove reference to mixed mode which
is no longer supported. Minor wording
change to raster vs vector discussion.

PR #1289, opened 2022-06-27
the zoom speed using the mouse wheel can be adjusted

mboe:
The standard zoom speed using the mouse wheel does not fit all needs. It would be nice if it can be adjusted. This tiny code change allows to override the currently fixed value.

PR #1377, opened 2022-09-26
Minor documentation updates

danbits:
When setting up `flutter_map` with the `vector_map_tiles` plugin, I noticed a couple of things to update in the docs.
* The basic `flutter_map` usage example uses a named argument `controller`. I needed to update the argument to `mapController` in order to get the example working.
* The section on vector tiles mentions the "mixed" mode for `vector_map_tiles`, which has been removed. It looks like the paragraph was removed in #1361, but that change did not make its way into 1dfdc38193c2954bbc8c4471cc3364cb76b987e8. Sorry to bring this up again if that was the intention.

Thank you for all of the work on this package. It's been a joy to use!

### DIFF
--- a/usage/basics.md
+++ b/usage/basics.md
@@ -4,7 +4,7 @@
 
 ```dart
 FlutterMap(
-    controller: MapController(),
+    mapController: MapController(),
     options: MapOptions(),
     children: [],
 ),
@@ -18,7 +18,7 @@ This is the main widget for this library, and it takes three main properties. op
 *   ``[`children`](layers/) (required)
 
     Takes a list of `Widget`s that will be displayed on the map. These can be any widget, but are usually`Layer`s.
-*   ``[`controller`](controller.md) (optional)
+*   ``[`mapController`](controller.md) (optional)
 
     Takes a `MapController` that can be used to programmatically control the map, as well as listen to an events `Stream`.
 
--- a/getting-started/explanation/raster-vs-vector-tiles.md
+++ b/getting-started/explanation/raster-vs-vector-tiles.md
@@ -23,5 +23,3 @@ However it does add complexity to the rendering process as each element needs to
 ### Using Vector Tiles
 
 Due to the complications mentioned above, 'flutter\_map' does not natively support vector tiles. However, vector tiles can be used with a [community maintained plugin (`vector_map_tiles`)](https://github.com/greensopinion/flutter-vector-map-tiles) to do this.
-
-The plugin also supports 'mixed' mode to get the best of both worlds: using raster images during animations to improve performance, and vector rendering to provide sharp visuals and custom theming when idle.


JaffaKetchup:
Hi @danbits. Number 1 seems a bit odd to me, can you show some of your code please? And you are correct about number 2, thanks for pointing that out.

JaffaKetchup:
Ah. I'm not sure this is supposed to be like this? I can't remember it being that way in the v3 release? If it wasn't, that needs reverting I think.

TesteurManiak:
Just checked, the last occurence of `controller` was in https://github.com/fleaflet/flutter_map/pull/1289 then it became `mapController` in https://github.com/fleaflet/flutter_map/pull/1294

TesteurManiak:
Well, the thing is that apparently the name `mapController` is used since the version 2.1 of the package, so changing it back to `controller` will have an impact on all users who have migrated to 3.0.
I would recommend to deprecate `mapController` and add back the `controller` parameter to avoid any breaking changes 🤔

ibrierley:
Does it really matter, if it was part of a breaking change already ? I.e I would guess most of the old examples out in the wild are already incorrect with other changes anyway (eg layer changes to children), so it feels like simplest would just be to go with it and update the docs to match, or am I missing something (quite likely!) ?

TesteurManiak:
My main concern is that changing it back to `controller` in the next release would break project that are using version 3.0.
Usually, when updating code implies breaking change you want to make it a major version bump to avoid current users to encounter a regression in their codebase. The change is small but still, I'm not a big fan of breaking the code of current users of the package.
This might seem a bit overkill, but imo it's better to deprecate anything that can be deprecated before introducing breaking changes.

ibrierley:
Just checking, are we talking about the same thing here...hasn't it actually been mapController since the dawn of time...?

ibrierley:
Ah, maybe there's a bit of confusion with MapOptions vs FlutterMap mapController (not surprising if so!)

TesteurManiak:
You're right, I'm the one who inverted `controller` and `mapController` 🤣
So yeah there's no problem actually, just updating the documentation to be cohesive with the code

TesteurManiak:
So yeah, I guess it can be merged (sorry again for my mistake it has been a long day 😅 )

ibrierley:
Hah no worries, this gets most of us at some point, and one of the reasons we've tried to simplify stuff :D. 

Cue someone out there getting confused about bounds next :D

JaffaKetchup:
Ok, so if we're happy with this, I'll make the changes manually ASAP, as it's just a small changeset. Haven't been following closely (been busy), so please just confirm that's the case :)

ibrierley:
Fine with me

JaffaKetchup:
Great. Will add those changes ASAP. Thanks for the heads up!

usage/basics.md
```diff
@@ -4,7 +4,7 @@
 
 ```dart
 FlutterMap(
-    controller: MapController(),
+    mapController: MapController(),
```
danbits, line 7: @JaffaKetchup Should this line be included as well?

JaffaKetchup: Yes. Sorry, missed this one!

PR #1378, opened 2022-09-29
Add documentation on Stadia Maps tiles

ianthetechie:
This PR adds documentation for how to access tiles from Stadia Maps. As I mentioned on Discord, I'm one of the cofounders, but I've tried to keep the language as neutral/objective as possible ;)

Note that I haven't been able to test this out since it looks like the GitBook tooling for local dev is like 5+ years old and doesn't work out of the box on my new Mac with LTS node, so a sanity check on the styling would be appreciated.

### DIFF
--- a/SUMMARY.md
+++ b/SUMMARY.md
@@ -40,6 +40,7 @@
 ## Tile Servers
 
 * [Using Mapbox](tile-servers/using-mapbox.md)
+* [Using Stadia Maps](tile-servers/using-stadia-maps.md)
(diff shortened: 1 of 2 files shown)


JaffaKetchup:
Thanks @ianthetechie! I've applied these changes manually, as there was some other changes that this reminded me of. Please let me know if [docs.fleaflet.dev/tile-servers/using-stadia-maps](https://docs.fleaflet.dev/tile-servers/using-stadia-maps) is good for you!

ianthetechie:
Works for me. Thanks!

PR #1376, opened 2022-09-25
fix for flickering with fitbounds, moving calculations inside the lay…

ibrierley:
Fix for flickering with fitbounds, moving calculations inside the layoutBuilder…

PR #1380, opened 2022-10-07
Improve Attribution

### LINKED ISSUE #1379: [FEATURE] Improved attribution

I have a number of suggestions for how to improve attributions, and would be happy to contribute code.

* Allow multiple attribution sources. Right now it's just kinda one with a widget builder, but the helper functions could be a lot more interesting IMO.
* Make attribution collapse on small screens. This is a huge problem right now IMO. The next logical step once we have the above default builder is to allow it to collapse on small screens and pop up a list or something (need to do a bit of research the best way to do this in Flutter but I'm sure it's possible).
* Option to pass a URL rather than a tap handler. This is really what 99% of people want anyways. I'm not sure if you're OK with another dependency, but url_launcher is what I use. Just making this part of the builder helpers would make things SO much cleaner. As a bonus, attribution to flutter_map would be a hyperlink rather than just text ;)

Let me know if you have any thoughts/objections, but I have already implemented most of the first and last items locally.

Yeah, just doing a bunch of boilerplate myself for some of them. There actually isn't a proper way to attribute multiple sources at this point correctly since you only have one tap handler (ok sure pop up a selection box but that's SUPER clunky).

_No response_

### Platforms Affected

### Severity

### Requirements

JaffaKetchup:
The previous `AttributionWidget` was brought in fairly quickly to replace a deprecated and removed API. Whilst it was acceptable for simple maps with only a few attributions, it is not designed well to handle too many.

Designed to close #1379 by introducing breaking changes to improve the experience. Now, a `Map<Text, void Function()>` is passed to the default builder, which will be laid out in an expandable column to prevent clutter.

This is a breaking change and may not be to the liking of everyone. It is difficult to develop a feature like this - because everyone will want it to behave slightly different. Therefore, I would like a significant amount of feedback before merging.

### DIFF
--- a/example/lib/pages/home.dart
+++ b/example/lib/pages/home.dart
@@ -58,9 +58,14 @@ class HomePage extends StatelessWidget {
                   zoom: 5,
                 ),
                 nonRotatedChildren: [
-                  AttributionWidget.defaultWidget(
-                    source: 'OpenStreetMap contributors',
-                    onSourceTapped: () {},
+                  //AttributionLayer.defaultWidget(
+                  //  source: '© OpenStreetMap contributors',
+                  //  onSourceTapped: () {},
+                  //),
+                  AttributionLayer(
+                    attributions: {
+                      const Text('© OpenStreetMap contributors'): () {}
+                    },
                   ),
                 ],
                 children: [
--- a/lib/src/layer/attribution_layer.dart
+++ b/lib/src/layer/attribution_layer.dart
@@ -1,11 +1,104 @@
-import 'package:flutter/widgets.dart';
+import 'package:flutter/material.dart';
 
-/// Attribution widget layer, usually placed in `nonRotatedChildren`
+class AttributionLayer extends StatefulWidget {
+  const AttributionLayer.custom({
+    super.key,
+    required Widget Function(BuildContext context) customBuilder,
+  })  : _customBuilder = customBuilder,
+        attributions = null,
+        alignment = Alignment.bottomRight,
+        backgroundColor = const Color(0xCCFFFFFF),
+        animationDuration = const Duration(milliseconds: 250),
+        animationCurve = Curves.fastOutSlowIn;
+
+  const AttributionLayer({
+    super.key,
+    this.attributions,
+    this.alignment = Alignment.bottomRight,
+    this.backgroundColor = const Color(0xCCFFFFFF),
+    this.animationDuration = const Duration(milliseconds: 250),
+    this.animationCurve = Curves.fastOutSlowIn,
+  }) : _customBuilder = null;
+
+  final Widget Function(BuildContext context)? _customBuilder;
+
+  final Map<Text, void Function()>? attributions;
+  final Alignment alignment;
+  final Color backgroundColor;
+
+  final Duration animationDuration;
+  final Curve animationCurve;
+
+  @override
+  State<AttributionLayer> createState() => _AttributionLayerState();
+}
+
+class _AttributionLayerState extends State<AttributionLayer>
+    with TickerProviderStateMixin {
+  late final _animationController = AnimationController(
+    duration: widget.animationDuration,
+    vsync: this,
+  );
+  late final _animation = CurvedAnimation(
+    parent: _animationController,
+    curve: widget.animationCurve,
+  );
+
+  late final Map<Text, void Function()> _attributions = {
+    ...widget.attributions ?? {},
+    const Text('flutter_map & the authors'): () {},
+  };
+
+  @override
+  Widget build(BuildContext context) => widget._customBuilder != null
+      ? widget._customBuilder!(context)
+      : Align(
+          alignment: widget.alignment,
+          child: ColoredBox(
+            color: widget.backgroundColor,
+            child: Padding(
+              padding: const EdgeInsets.symmetric(vertical: 4, horizontal: 6),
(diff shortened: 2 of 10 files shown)


lib/src/map/flutter_map_state.dart
```diff
@@ -42,11 +42,11 @@ class FlutterMapState extends MapGestureMixin
...
     move(options.center, zoom, source: MapEventSource.initialization);
 
-    // Funally, fit the map to restrictions
-    if (options.bounds != null) {
-      fitBounds(options.bounds!, options.boundsOptions);
-    }
     WidgetsBinding.instance.addPostFrameCallback((_) {
+      // Finally, fit the map to restrictions
+      if (options.bounds != null) {
+        fitBounds(options.bounds!, options.boundsOptions);
```
MooNag, line 48: Was this change intentional? 

https://github.com/fleaflet/flutter_map/pull/1376/files#diff-0712fe61169af14b5ce2bed28ca88305d26d4d91630a99158d64f19c6cb7283dL46

MooNag:
> This is a breaking change and may not be to the liking of everyone
Is it possible to leave the old AttributionWidget; deprecate it. Then create this new one as AttributionLayer?

JaffaKetchup:
Yeah good call @MooNag, I'll do that. For some reason, this seems to have a messy commit history, so I'll need to sort that out.

example/lib/pages/home.dart
```diff
@@ -58,9 +58,14 @@ class HomePage extends StatelessWidget {
...
-                  AttributionWidget.defaultWidget(
-                    source: 'OpenStreetMap contributors',
-                    onSourceTapped: () {},
+                  //AttributionLayer.defaultWidget(
+                  //  source: '© OpenStreetMap contributors',
+                  //  onSourceTapped: () {},
+                  //),
+                  AttributionLayer(
+                    attributions: {
+                      const Text('© OpenStreetMap contributors'): () {}
```
ianthetechie, line 67: I think it would be worth refactoring this into an `AttributionSource`. It can have a detailed constructor like this with an onTapped action, but I would propose a convenience method as well to construct one with just a source name and a URL to launch on tap. This is (IMO) what 99.99% of them should be anyways. It would also save you from having to hunt around for the copyright glyph and would save a TON of boilerplate. Here's an example:

```dart
class AttributionSource {
  final String name;
  final void Function()? onSourceTapped;

  // TODO: Convenience constructor that accepts a URL rather than a function
}
```

The default onSourceTapped with said helper would look something like this:

```dart
() async {
  // Depends on package url_launcher
  if (!await launchUrl(uri)) {
    // Someone who actually understands dart logging correct this; if flutter_map doesn't already have a logger, maybe it should create one.
    log('Could not launch url', level: Level.WARNING.value);
  }
}
```

lib/src/layer/attribution_layer.dart
```diff
@@ -1,11 +1,104 @@
...
+                          child: Column(
+                            crossAxisAlignment: CrossAxisAlignment.end,
+                            children: [
+                              ..._attributions.keys.toList(),
+                              const Divider(),
+                            ],
+                          ),
+                        ),
+                      ),
+                      const Text('Show Map Attributions'),
```
ianthetechie, line 86: A few thoughts on this:

1. Extremely minor nitpick, but I think this can be shortened to just "Attributions" if it's going to be pure text; it should be obviously hyperlink styled, right?
2. Any objections to an info image? This is pretty standard practice and could look better if done right.
3. I don't know if this is easily doable, but the ideal solution IMO is to do something like Leaflet and collapse into in info button only if the text doesn't fit. If this is to difficult though, something like this would be great.

I may have some other suggestions about the UX after giving it a try later, but these are my thoughts off the top of my head.

example/lib/pages/home.dart
ianthetechie: Valid. I don't have a really strong opinion about this, but this is what like 99% of people probably want so it seems logical enough to add.

Does Dart have a package-level feature construct which conditionally includes code and deps based on feature flags (ex: cargo features in Rust)? You could solve any objection to the added weight by adding an opt-out this way.

TesteurManiak: Unfortunately not, a package has to be referenced in the `pubspec.yaml`, even if it is not used in the project in the end, the code might not be included when building an application with tree shaking but the package dependency will remain.

JaffaKetchup: We can't depend on url_launcher because it requires extra setup for platform specifics: I would prefer to avoid this confusion.

lib/src/layer/attribution_layer.dart
JaffaKetchup: Yeah I'm still not sure on the best way to do this. At the moment, the attributions expand upwards, with each one on a new line. This doesn't look great.
If you want it to look a specific way (which I have no problem with, open to all ideas) can you just post a very rough sketch from something like Paint, just so we can review that before spending time implementing it.

Maybe a 50% opacity flutter_map logo (the same one on the website, and I still have the original design files if it needs modifying) and 50% opacity info icon will look ok. When tapped, it can open a floating box of some sort? Additionally this would add appropriate space for an image attribution like Mapbox requires. What do you think of this?

JaffaKetchup:
Closed, due to reasons discussed on the Discord server.

PR #1381, opened 2022-10-08
Improve PolygonLayer rendering performance

MooNag:
Changes:
- Better Offset generation code. Uses fixed-sized lists now.
- Added batching code. Every sequential polygon that generates the same renderHashCode will have optimized draw calls to the canvas. In many cases this can more than double the performance. With 15 batches and 1.5k polygons I profiled a framerate change from 13 fps -> 38 fps. 
- Batching code overhead is minimal. I profiled <1ms on a high end phone (z fold 4), and <2ms on a budget phone (pixel 4a) with 1.5k polygons. I don't think there would be any situation where performance would become worse, combined with the other optimizations. (this needs to be tested though)
- Added batched polygons to example app.
- Polygons with holes now render labels (not sure if this is intentional to be the case or not).
- Refactored the renderer to de-duplicate the code a reasonable amount.

Possible side effects:
- Since polygons are batched there may potentially be slight differences between this renderer and the old one.
- Polygons with holes didn't get refactored because I have limited knowledge of how the hole renderer works (TODO),


mohammedX6:
I tested it out, looking good I noticed some really good performances uplift.

mohammedX6:
But I have a question, why are you not saving the offsets into the polygon offsets variable? you are doing the calculation of the offsets twice.

MooNag:
I did this since the offsets are computed on every build anyways, and the compute function has low overhead. Potentially this could be optimized. Previously, the offsets were being stored in a growable list, which wasted multiple milliseconds of time just resizing lists, whereas now using a single ListGenerator the whole compute can be done in less time.

The offsets being calculated twice is only for if you are using a label. Since those are computed after the main polygons and I don't want to commit on having a cache variable for only a **tiny** amount of performance. Especially given that at this point the limiting factor actually seems to be the GPU rather than the CPU on most phones where many polygons are involved. Let me know if you have any questions! I can also post a profiler if you have any questions about the compute time percentages for each kind of thing.

mohammedX6:
i learned new thing toady from you today!.
I already implemented your modification and i saw a huge amount of performance increase in my app especially on older tablets, you did a great job!

MooNag:
this isn't stale bot Just going really slowly. I know its hard to tell the difference

JaffaKetchup:
Hey @MooNag. Did you ever fix the issue with the hole points not working? I think I reported to you on Discord. That needs to be fixed before we can merge.

lib/src/layer/polygon_layer.dart
```diff
@@ -202,77 +271,12 @@ class PolygonPainter extends CustomPainter {
...
-    }
-  }
-
   @override
   bool shouldRepaint(PolygonPainter oldDelegate) => false;
 
-  double _dist(Offset v, Offset w) {
-    return sqrt(_dist2(v, w));
-  }
+  double _dist(Offset v, Offset w) => sqrt(_dist2(v, w));
```
JosefWN, line 277: I don't think we need these helpers? They are just overriding the standard library, most of which we have already imported in this file as well, which is less intuitive to devs.

Don't know about the standard lib but this excessive nesting will also just make longer call chains/slower code... if even just a tiny bit.

`_dist(v, w)` -> `(v - w).distance`
`_dist2(v, w)` -> `(v - w).distanceSquared`
`_sqr(x)` -> If need be we could just use `math.pow(x, 2)` or `x * x`

ignatz:
I also ran into performance issues with polygon/polylines. Instead of batching canvas paints, I experimented with minimizing repaints: https://github.com/fleaflet/flutter_map/pull/1423 (currently only for polylines) but the same approach should work for polygons, too. I'm mostly raising this here because I don't think the approaches would be compatible. Concretely: batching and culling require you to redraw on every pan (unless you also merge bounding boxes).  It might be worth thinking a little bit about with approach might be more promising going forward, i.e. minimizing number of canvases vs minimizing number of draws.

lib/src/layer/polygon_layer.dart
MooNag: Yeah these things are true, longer chains will cause more cache misses. Duplicating standard libraries is pointless here. I will fix that.

MooNag:
Closing in favor of #1442

PR #1385, opened 2022-10-15
Modify circle layer to represent more type of curves

GT020:


lib/src/layer/arc_layer.dart
```diff
@@ -0,0 +1,112 @@
...
+        final map = FlutterMapState.maybeOf(context)!;
+        final arcWidgets = <Widget>[];
+        for (final arc in arcs) {
+          arc.offset = map.getOffsetFromOrigin(arc.centre);
+
+          if (arc.useRadiusInMeter) {
+            final r = const Distance().offset(arc.centre, arc.radius, 180);
+            final delta = arc.offset - map.getOffsetFromOrigin(r);
+            arc.realRadius = delta.distance;
+          }
```
MooNag, line 55: Maybe move this code into your custom paint so you can make the ArcMarker have only final fields. You would just pass in your MapState into your CustomPaint.

lib/src/layer/cubic_curve_layer.dart
```diff
@@ -0,0 +1,135 @@
...
+        handlePointTwo.hashCode ^
+        color.hashCode ^
+        borderStrokeWidth.hashCode ^
+        borderColor.hashCode ^
+        start.hashCode ^
+        end.hashCode ^
+        handleOne.hashCode ^
+        handleTwo.hashCode;
+  }
+}
```
MooNag, line 60: Please use Object.hash here. I believe this is the recommended hash function for flutter at the moment.
```dart
int get hashCode => Object.hash(endPoint, handlePointOne, handlePointTwo, color, .... etc);
```

lib/src/layer/cubic_curve_layer.dart
```diff
@@ -0,0 +1,135 @@
...
+    return LayoutBuilder(
+      builder: (BuildContext context, BoxConstraints bc) {
+        final size = Size(bc.maxWidth, bc.maxHeight);
+
+        final curveWidgets = <Widget>[];
+        for (final curve in curves) {
+          curve.start = map.getOffsetFromOrigin(curve.startPoint);
+          curve.end = map.getOffsetFromOrigin(curve.endPoint);
+          curve.handleOne = map.getOffsetFromOrigin(curve.handlePointOne);
+          curve.handleTwo = map.getOffsetFromOrigin(curve.handlePointTwo);
```
MooNag, line 81: These can also be done inside of the custom painter. This would allow for CubicCurveMarker to have only final values.

lib/src/layer/quadractic_curve_layer.dart
```diff
@@ -0,0 +1,131 @@
...
+        endPoint.hashCode ^
+        handlePoint.hashCode ^
+        color.hashCode ^
+        borderStrokeWidth.hashCode ^
+        borderColor.hashCode ^
+        start.hashCode ^
+        end.hashCode ^
+        handle.hashCode;
+  }
+}
```
MooNag, line 56: Use Object.hash here too

lib/src/layer/quadractic_curve_layer.dart
```diff
@@ -0,0 +1,131 @@
...
+  Widget build(BuildContext context) {
+    return LayoutBuilder(
+      builder: (BuildContext context, BoxConstraints bc) {
+        final size = Size(bc.maxWidth, bc.maxHeight);
+        final map = FlutterMapState.maybeOf(context)!;
+        final curveWidgets = <Widget>[];
+        for (final curve in curves) {
+          curve.start = map.getOffsetFromOrigin(curve.startPoint);
+          curve.end = map.getOffsetFromOrigin(curve.endPoint);
+          curve.handle = map.getOffsetFromOrigin(curve.handlePoint);
```
MooNag, line 75: These can also be computed in custom paint to allow for only final variables in QuadraticCurveMarker

MooNag (changes requested):
I like the idea of having more built in layers. I don't know if we **need** these layers, but they might be nice to have and worth discussing. I provided some feedback on the code, I have not tested these layers in a project myself for functionality though.

lib/src/layer/arc_layer.dart
GT020: On it

GT020:
Let It stay here for while

PR #1390, opened 2022-10-24
Improve Attributions

ianthetechie:
See #1379.

This PR adds a new attribution control, `RichAttributionWidget` as an alternative to the existing `AttributionWidget` (which is still valid if you want a very minimal style) and has a more expressive API built around `SourceAttribution` and its concrete subclasses.

* Toggle full display of non-logo attributions via the info button
* Concrete implementations for text and logo attributions
* Ability to do "hyperlink" style URI launching for all attribution types, while still enabling a completely custom onTap handler for the few that want that.

Possible discussion points:
* Logo attributions are always visible; I have yet to see a case where this isn't required in my map work but the API could be extended if desired
* I opted NOT to change the opacity of the logos as in a prototype passed around on Discord because my understanding is that this could be a big performance impact.
* In my understanding, the current attribution widget implementation is implemented in a very strange way. The default static helper doesn't even construct an attribution widget. I fixed this. I additionally think it should actually just be implemented as a named constructor rather than a static method, but open to opinions.

TODOs:
- [ ] Decide if we want any more logo attribution constructors (I just implemented the obviously necessary ones that I cared about)
- [ ] Tests (probably?)
- [ ] This project always gets a logo attribution if the control is added. There should probably be a text option instead (discussion welcome)? I also don't know where the official logo is hosted so I just grabbed it from an HTML inspection of the docs. This *probably* needs to change.
- [ ] Probably a version bump (not sure project policy on this)

Screenshot gallery:

image
image


JaffaKetchup:
Superseded/continued/replaced by #1487.

PR #1393, opened 2022-10-28
new attributionwidget constructor

### LINKED ISSUE #1392: [FEATURE] new AttributionWidget without default text

I would like a new constructor for `AttributionWidget` where I can control all of the text. Adding 'flutter_map | © ' makes 1.) the attribution box quite big and 2.) the different styles of text don't match nicely.
Implementation: new named constructor like `AttributionWidget.customText` that doesn't add 'flutter_map | © '

provide an `attributionBuilder` to `AttributionWidget`s normal constructor building the widget yourself

_No response_

### Platforms Affected

### Severity

Minimum: Not required for my use

### Requirements

Friendly-Banana:
fixes #1392
might become obsolete by #1390 but that is a draft as of now

### DIFF
--- a/lib/src/layer/attribution_layer.dart
+++ b/lib/src/layer/attribution_layer.dart
@@ -65,4 +65,39 @@ class AttributionWidget extends StatelessWidget {
           ),
         ),
       );
+
+
+  /// Quick constructor for a more classic styled attibution box
+  ///
+  /// Displayed as a padded translucent white box with the text from [source].
+  ///
+  /// Provide [onSourceTapped] to carry out a function when the box is tapped. If that isn't null, the source text will have [sourceTextStyle] styling - which defaults to a link styling.
+  static Widget customText({
+    required String source,
+    void Function()? onSourceTapped,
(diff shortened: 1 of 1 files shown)


JaffaKetchup:
Hi @Friendly-Banana,

Thanks for this PR, however, I am going to close it for three reasons:
- There is already a way to construct a custom `AttributionWidget` - it doesn't create any of the UI for you, but it's not too hard to copy from the current implementation if necessary
- This method of implementation creates a lot of code duplication - if it were to be accepted, it would be better to just have an argument to disable/remove the default text
- #1390 should be on the way _relativley_ soon - we are still in discussion on the Discord server (note that it may or may not become a plugin: we are still undecided, and it will depend on the extent to which the PR goes). Whilst it won't replace this current method of attribution, it should resolve the issues you mentioned in your original issue (#1392).

Please don't be dissuaded from contributing in the future: we are always happy to have new ideas and PRs! Once again, many thanks for your PR!

Friendly-Banana:
H @JaffaKetchup,
Ok, I understand. Looking forward to #1390.

PR #1403, opened 2022-11-06
Add key to Polyline, CircleMarker + Test utilities

TesteurManiak:
I've added a key property to `Polyline` and `CircleMarker` classes. Moreover I've written some basic tests and made reusable test utilities.

I've replaced the `dev_dependency` to `mockito` by `mocktail` as it's easier to use and doesn't rely on code generation which might be cumbersome when testing.
The API is almost the same and doesn't change the current usage in `MockHttpClientResponse`, `MockHttpHeaders`, etc.


JaffaKetchup (approved):
Merging as discussed in Discord.

PR #1404, opened 2022-11-09
Polyline stroke width in meters

LeonTenorio:
Implementation: 
- useStrokeWidthInMeter boolean parameter. When that value is true the PolylinePainter use the map to get the stroke width in meters and render using that.

### DIFF
--- a/lib/src/layer/polyline_layer.dart
+++ b/lib/src/layer/polyline_layer.dart
@@ -70,6 +70,7 @@ class PolylineLayer extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
     final map = FlutterMapState.maybeOf(context)!;
+
     return LayoutBuilder(
       builder: (BuildContext context, BoxConstraints bc) {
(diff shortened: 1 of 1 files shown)


JaffaKetchup:
@ibrierley @MooNag @TesteurManiak What do we think about this being in core? It makes sense to me, and I think using `Distance()` is accurate - at least it has been accurate enough when I've done this same thing before.

lib/src/layer/polyline_layer.dart
```diff
@@ -10,7 +10,7 @@ class Polyline {
   final Key? key;
   final List<LatLng> points;
   final List<Offset> offsets = [];
-  final double strokeWidth;
+  double strokeWidth;
```
JaffaKetchup, line 13: This doesn't need to be mutable.

```suggestion
  final double strokeWidth;
```

JaffaKetchup (approved):
LGTM, I think this is a very useful feature :)

PR #1408, opened 2022-11-20
Simplify FlutterMapState by:

ignatz:
1. Explicitly separating core state form derived quantities. This let's us handle state updates and invalidation more consistently.
 2. Preserve the current memoization behavior for derived quantities by introducing an explicit caching abstraction.
 3. Reduce the extend of the render tree being rebuild by avoiding `setState`, therefore avoid invalidation of the entire stateful widget.

(Re (1): to be clear I'm not aware of any consistency issues today. I'm only stating that the updates and invalidation are done very selectively in a way that requires intimate knowledge of how derived quantities functionally depend on core state.

Arguably this is a heavy handed change with some aspects more valuable than others. Take it as an RFC we can work through, break up, remold and even throw away :).

Ultimately, this was the result of me trying to squeeze out a few ms from my frame times. I was struggling to reason, which quantities are updated where and when? How do state changes relate to render tree rebuilding? I do hope that this refactor makes it a little clearer. The good or bad news: I didn't find any low hanging fruit for improving performance. While this change most certainly has some performance implications, it seemed pretty much unchanged with my local testing. Do you have any benchmarks that you run? Let me know what you think - thanks!

### DIFF
--- a/lib/src/map/flutter_map_state.dart
+++ b/lib/src/map/flutter_map_state.dart
@@ -17,8 +17,6 @@ class FlutterMapState extends MapGestureMixin
 
   final MapController _localController = MapControllerImpl();
 
-  bool _hasFitInitialBounds = false;
-
   @override
   MapOptions get options => widget.options;
 
@@ -34,17 +32,23 @@ class FlutterMapState extends MapGestureMixin
 
     mapController.state = this;
 
-    // Initialize all variables here, if they need to be updated after the map changes
-    // like center, or bounds they also need to be updated in build.
-    _rotation = options.rotation;
-    _center = options.center;
-    _zoom = options.zoom;
-    _pixelBounds = getPixelBounds(zoom);
-    _bounds = _calculateBounds();
+    LatLng center = options.center;
+    double zoom = options.zoom;
+    if (options.bounds != null) {
+      final target =
+          getBoundsCenterZoom(options.bounds!, options.boundsOptions);
+      center = target.center;
+      zoom = target.zoom;
+    }
 
-    WidgetsBinding.instance.addPostFrameCallback((_) {
-      options.onMapReady?.call();
-    });
+    // Initialize core state here. State that needs to be updated after a map
+    // changes like: move center, or zoom in/out.
+    _state = ValueNotifier(_State(options.crs, center, zoom, options.rotation));
+
+    final onMapReady = options.onMapReady;
+    if (onMapReady != null) {
+      WidgetsBinding.instance.addPostFrameCallback((_) => onMapReady());
+    }
   }
 
   //This may not be required.
@@ -157,43 +161,36 @@ class FlutterMapState extends MapGestureMixin
     );
 
     return LayoutBuilder(
-        builder: (BuildContext context, BoxConstraints constraints) {
-      //Update on layout change
-      setSize(constraints.maxWidth, constraints.maxHeight);
+      builder: (BuildContext context, BoxConstraints constraints) {
+        //Update on layout change
+        _nonrotatedSize =
+            CustomPoint<double>(constraints.maxWidth, constraints.maxHeight);
+        _state.value = _State(options.crs, center, zoom, rotation);
 
-      if (options.bounds != null &&
-          !_hasFitInitialBounds &&
-          constraints.maxWidth != 0.0) {
-        final target =
-            getBoundsCenterZoom(options.bounds!, options.boundsOptions);
-        _zoom = target.zoom;
-        _center = target.center;
-        _hasFitInitialBounds = true;
-      }
-
-      _pixelBounds = getPixelBounds(zoom);
-      _bounds = _calculateBounds();
-      _pixelOrigin = getNewPixelOrigin(_center);
-
-      return MapStateInheritedWidget(
-        mapState: this,
-        child: Listener(
-          onPointerDown: onPointerDown,
-          onPointerUp: onPointerUp,
-          onPointerCancel: onPointerCancel,
-          onPointerHover: onPointerHover,
-          onPointerSignal: onPointerSignal,
-          child: PositionedTapDetector2(
-            controller: _positionedTapController,
-            onTap: handleTap,
-            onLongPress: handleLongPress,
-            onDoubleTap: handleDoubleTap,
-            child:
-                RawGestureDetector(gestures: gestures, child: _buildMap(size)),
-          ),
-        ),
-      );
-    });
+        return ValueListenableBuilder<_State>(
+            valueListenable: _state,
+            builder: (BuildContext ctx, _State settings, Widget? __) {
+              return MapStateInheritedWidget(
+                mapState: this,
+                child: Listener(
+                  onPointerDown: onPointerDown,
+                  onPointerUp: onPointerUp,
+                  onPointerCancel: onPointerCancel,
+                  onPointerHover: onPointerHover,
+                  onPointerSignal: onPointerSignal,
+                  child: PositionedTapDetector2(
+                    controller: _positionedTapController,
+                    onTap: handleTap,
+                    onLongPress: handleLongPress,
+                    onDoubleTap: handleDoubleTap,
+                    child: RawGestureDetector(
+                        gestures: gestures, child: _buildMap(size)),
+                  ),
+                ),
+              );
+            });
+      },
+    );
   }
 
   Widget _buildMap(CustomPoint<double> size) {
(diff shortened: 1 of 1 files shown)


ibrierley:
Thanks for your work on this. I'm sure we'll all have a rummage though and give some thoughts before too long.

If you are also getting other performance issues unrelated to his PR, it's also worth just posting a separate git issue with your specific problem. It both helps devs realise what issues users are hitting, and sometimes the have workarounds or how it could be improved in future also.

lib/src/map/flutter_map_state.dart
```diff
@@ -34,17 +32,24 @@ class FlutterMapState extends MapGestureMixin
...
+    // Initialize core state here. State that needs to be updated after a map
+    // changes like: move center, or zoom in/out.
+    _state = ValueNotifier(_State(options.crs, center, zoom, options.rotation));
 
-    WidgetsBinding.instance.addPostFrameCallback((_) {
-      options.onMapReady?.call();
-    });
+    if (options.onMapReady != null) {
+      WidgetsBinding.instance.addPostFrameCallback((_) {
+        options.onMapReady!.call();
```
TesteurManiak, line 50: I would use a variable to beneficiate from type shadowing
```dart
final onMapReady = options.onMapReady;
if (onMapReady != null) {
  WidgetsBinding.instance.addPostFrameCallback((_) {
    options.onMapReady.call();
  }
}
```

ignatz: Hi [Guillaume](https://github.com/TesteurManiak), nice to meet you. I adapted the PR to your proposal - thanks.

Quickly taking a step back to be mindful of everyone's time: do you think this PR overall is going in the right direction? Just want to avoid a late surprise, e.g. us finding the change too large/invasive. Either way, happy to work though any issues with you. Cheers.

TesteurManiak: Hi, @ignatz I need myself to take a deeper look into your code which I haven't time to do for now. Your contribution is very much welcome, especially if code responsibility is better handled, but it might need some testing to ensure it doesn't cause any regression.

ignatz: Sounds good. Keep me posted. Thanks

ignatz:
Any thoughts on whether there's anything worth salvaging in this change?

JaffaKetchup:
Hi @ignatz, and sorry for the long delay!
I'll try to get a look at this tomorrow. If I can't, please do keep this open. Contributions are always appreciated, even if we sometimes are slow to act.
At the moment, there is some internal discussion on how to proceed with FM and its tile layers (potentially a complete rewrite of the tile layer system or larger), but we're limited on time and not sure what option is best. Maybe this is a good step toward a rewrite (looking at the number of line changes).
I've removed the stale tag, don't worry about that :)

ibrierley:
Have been having a bit of a ponder on this one. I'm just thinking out loud here for discussion, and I may be missing some elements..

I like the idea of a separate state class. I had even previously wondered about having a state class that could be created and passed direct to mapController/MapOptions, with only allowing certain methods to mutate the state (the idea being to separate it out, and be easier to test certain bits of code and be consistent...eg if we want to test what happens if we try and get a pixelOrigin from a Map that has no size initialised yet, it can become quite fiddly). I'm a little unsure on what the latest thinking of that kind of thing would be, good or a no-no. But should state (i.e FlutterMapState have things like Gestures and Layout builders within it?)

The Caching side of PR, I'm a little more uncertain of, and whether its just another element that kind of pollutes the space a bit with potential bugs (either by it not being used correctly, not understood or something else). I'm not against caching at all, I'm just not quite convinced it's needed. If there are elements that are recalculated more than every frame, or movement change, I'd be interested as I lose track, and may be missing some that happens.

I do get the point with the render tree, but am not probably expert enough to give a valid opinion on this atm, but I thought Flutter was pretty good at eliminating updates where there haven't been changes, naturally I may be wrong. I'd definitely be interested to see some example or evidence of where there are unnecessary updates/workload.

ignatz:
Hey @ibrierley , thanks for your thoughts. It's hardly going to be a discussion, since I agree with everything :). Let me try to shed some light on how the change adheres to your thinking:

> I like the idea of a separate state class. I had even previously wondered about having a state class that could be created and passed direct to mapController/MapOptions, with only allowing certain methods to mutate the state (the idea being to separate it out, and be easier to test certain bits of code and be consistent...eg if we want to test what happens if we try and get a pixelOrigin from a Map that has no size initialised yet, it can become quite fiddly). I'm a little unsure on what the latest thinking of that kind of thing would be, good or a no-no. But should state (i.e FlutterMapState have things like Gestures and Layout builders within it?)

I'm guessing (and please don't take this the wrong way) that the split of responsibilities has grown somewhat organically over time. Some aspects might be relics of the pre v3 state (though, I didn't really use earlier versions) when there was a separate API from plain widgets for map plugins. In the current state of Flutter map, I'd like to think about the different responsibilities as:

1.  MapOptions: kitchen sink for all things initialization
2.  FlutterMap (Widget): responsible for building the UI, i.e. widget-tree
3.  State: everything you need to know to restore a new FlutterMap from scratch.
4.  MapController: the things that let's query and alter map state. Ideally, FlutterMap users and plugins should only change state through the controller.

Today,FlutterMapState is 2 and 3 and 4. It's a widget, it has the state, and it used to be the MapController API for plugins. More than once did I look for functionality on the MapController and only found it on FlutterMapState :).

Moreoever, looking at FlutterMapState it wasn't clear to me what is "core" state, i.e. sufficient to restore a map from scratch, and what's derived, cached state that functionally depends on core state. It was also not clear to me *when* derived state should be updated *consistently*.

That's why I tried to separate out what is "core" state and separate it from derived (cached) state that needs to be invalidated whenever the core state changes to avoid inconsistencies.

> The Caching side of PR, I'm a little more uncertain of, and whether its just another element that kind of pollutes the space a bit with potential bugs (either by it not being used correctly, not understood or something else). I'm not against caching at all, I'm just not quite convinced it's needed. If there are elements that are recalculated more than every frame, or movement change, I'd be interested as I lose track, and may be missing some that happens.

To be transparent, I'm also not sure how needed the caching is. The benefit will greatly depend on how often users lookup "derived" state.
To be clear, *I didn't add any caching*. I preserved the caching that was already in the FlutterMapState, collected the derived state in a single place (`Cache`), and tried to simplify the lifecycle by not have "derived state updates" in various places. For example, try to reason on all the places `_updateSizeByOriginalSizeAndRotation()` need to be invoked consistently to avoid having `this.size` inconsistencies. 

> I do get the point with the render tree, but am not probably expert enough to give a valid opinion on this atm, but I thought Flutter was pretty good at eliminating updates where there haven't been changes, naturally I may be wrong. I'd definitely be interested to see some example or evidence of where there are unnecessary updates/workload.

Yes and no. Flutter is as good as it can be. Calling `setState()` on FlutterMapState invalidates it, thus it and all its children need to be rebuild. For example, I'm resizing the map window, then the LayoutBuilder will re-execute its builder function for its children to relayout, repaint, ... . How expensive that is in practice depends :woman_shrugging:

ibrierley:
> I'm guessing (and please don't take this the wrong way) that the split of responsibilities has grown somewhat organically over time. Some aspects might be relics of the pre v3 state (though, I didn't really use earlier versions) when there was a separate API from plain widgets for map plugins. In the current state of Flutter map, I'd like to think about the different responsibilities as:
> 
> 1. MapOptions: kitchen sink for all things initialization
> 2. FlutterMap (Widget): responsible for building the UI, i.e. widget-tree
> 3. State: everything you need to know to restore a new FlutterMap from scratch.
> 4. MapController: the things that let's query and alter map state. Ideally, FlutterMap users and plugins should only change state through the controller.
> 
> Today,FlutterMapState is 2 and 3 and 4. It's a widget, it has the state, and it used to be the MapController API for plugins. More than once did I look for functionality on the MapController and only found it on FlutterMapState :).
> 
> Moreoever, looking at FlutterMapState it wasn't clear to me what is "core" state, i.e. sufficient to restore a map from scratch, and what's derived, cached state that functionally depends on core state. It was also not clear to me _when_ derived state should be updated _consistently_.
> 
> That's why I tried to separate out what is "core" state and separate it from derived (cached) state that needs to be invalidated whenever the core state changes to avoid inconsistencies.
> 


Yes, I think what happens is everyone who comes along including myself, typically just puts things in a "near" place visually to where the same action is (oherwise also the changes become quite large every time), as no one necessarily knows the whole thing and reasonings for original implementation and if still relevant, so it's not a bad thing to visit this type of issue.


> To be transparent, I'm also not sure how needed the caching is. The benefit will greatly depend on how often users lookup "derived" state. To be clear, _I didn't add any caching_. I preserved the caching that was already in the FlutterMapState, collected the derived state in a single place (`Cache`), and tried to simplify the lifecycle by not have "derived state updates" in various places. For example, try to reason on all the places `_updateSizeByOriginalSizeAndRotation()` need to be invoked consistently to avoid having `this.size` inconsistencies.

Maybe the caching is just a bit of a redherring with the name...maybe if it was Calculated/Derived/Computed or something, I feel like it wouldn't have quite the same implications (I tend to think of things like expiries etc with caches). However, maybe it would still be useful to imply data could be out of date (although it shouldn't be).


> Yes and no. Flutter is as good as it can be. Calling `setState()` on FlutterMapState invalidates it, thus it and all its children need to be rebuild. For example, I'm resizing the map window, then the LayoutBuilder will re-execute its builder function for its children to relayout, repaint, ... . How expensive that is in practice depends 

I don't think I have a problem with this if tested by quite a few people and performance checked and any other issues.

There's also a few other subtle changes that could lead to issues like nonRotatedSize being potentially zero (rather than null) and things like that which I haven't tested (and can be hard to with different speed devices). I'm wondering if this PR should be broken down a bit, although I appreciate some aspects rely on others being implemented first, so that's potentially a bit fiddly.

ignatz:
> Yes, I think what happens is everyone who comes along including myself, typically just puts things in a "near" place visually to where the same action is (oherwise also the changes become quite large every time), as no one necessarily knows the whole thing and reasonings for original implementation and if still relevant, so it's not a bad thing to visit this type of issue.

No need to justify, this is totally normal. Otherwise, FlutterMap would be the first software project in existence w/o entropy. That's why we're here, trying to leave the place a little cleaner than we found it :+1:. In the worst case we test our understanding, in the best case we make it a little easier for the next person :smile:.

> Maybe the caching is just a bit of a redherring with the name...maybe if it was Calculated/Derived/Computed or something, I feel like it wouldn't have quite the same implications (I tend to think of things like expiries etc with caches). However, maybe it would still be useful to imply data could be out of date (although it shouldn't be).

I'm open to other names. While there's not TTL, there is "cache" invalidation. If nothing else, preserving the "caching" for now let's us move forward w/o worrying about regressing performance for certain users and use-cases. Going forward with everything in one place we can re-evaluate how useful certain cached state is to then possibly remove them (i.e. evaluate (access-rate x cost)).

> I don't think I have a problem with this if tested by quite a few people and performance checked and any other issues.

How do we build the necessary confidence. While I've been running these changes for a few month in a non-trivial app, I cannot claim that I'm using every single last feature. Would you suggest to submit to a branch? Do you have an RFC process to get the input of others?
Same for performance, I'm not aware of any process I could follow. I did some best-effort comparison and profiling, however didn't notice any difference. FWIW, this is more the material for micro-benchmarks, since for any real application GPU canvas rendering (in the order of tens of milliseconds) will easily dominate these simple computations, which rather take in the order of micro-seconds.

> There's also a few other subtle changes that could lead to issues like nonRotatedSize being potentially zero (rather than null) 

Let me have another look to see where this could be an issue.

> and things like that which I haven't tested (and can be hard to with different speed devices).

Fortunately, I'm only running pretty old devices + web.

> I'm wondering if this PR should be broken down a bit, although I appreciate some aspects rely on others being implemented first, so that's potentially a bit fiddly.

I feel the same way. I'm very open to trying to break it up further. For example, I could see myself pulling out the widget tree changes. I didn't do that already because I didn't know you folks and I wanted to gauge interest first. Let me know what you think.

Best!

JaffaKetchup:
@ibrierley @ignatz I'm just preparing and reviewing https://github.com/fleaflet/flutter_map/pull/1442 for either v3.2.0 or v4.0.0. It would be great if this could be merged, but I'm a bit out of the loop here!

ignatz:
After a longer hiatus, I took my changes for another spin (the PR should reflect my latest changes). The good news: I managed to fix the issue I previously had with map resizes and also reduce the size of the change. However,  there might be a slight performance regression. It's hard to say, w/o dedicated performance tests.

Here's what I saw using the example application while rotating the map:

before: 
Screenshot_2023-03-05-22-20-23-84_35700e11be2ac49d4a90b3b128d28763

after:
Screenshot_2023-03-05-22-23-12-49_35700e11be2ac49d4a90b3b128d28763

relevant is only the UI thread here, which sees the average build time jump from 1.9ms to 2.5ms. While my confidence is limited, with the changes it could totally be that we, either:
 * memoize something that was memoized before
 * or eagerly compute something that was previously lazily calculated and isn't always needed.

Unfortunately,  I don't expect to have much time in the near future to look into this more. I'll close the pull-request for now to reduce the noise on your end. Feel free to re-open if you think it's worth keeping this in the ice-box and or further pursuing the change. Thanks!

ibrierley:
Thanks for the update, and it's great you tested it for performance, and it's still useful to know whatever. Apologies, I'm pretty snowed under with other stuff atm, and never really had chance to give it a run.

I do still think it's an interesting idea though.

PR #1427, opened 2023-01-02
Delete unused and broken LatLngBounds.pad() method.

ignatz:
The method fails to deal correctly with LatLng wrap arounds. A more correct version would probably look more like:

    void pad(double bufferRatio) {
      final heightBuffer = (_sw!.latitude - _ne!.latitude).abs() * bufferRatio;
      final widthBuffer = (_sw!.longitude - _ne!.longitude).abs() * bufferRatio;

      final point1 = LatLng((90 + _sw!.latitude - heightBuffer) % 180 - 90,
          (180 + _sw!.longitude - widthBuffer) % 360 - 180);
      final point2 = LatLng((90 + _ne!.latitude + heightBuffer) % 180 - 90,
          (180 + _ne!.longitude + widthBuffer) % 360 - 180);

      final newBounds = LatLngBounds(point1, point2);
      _sw = newBounds.southWest;
      _ne = newBounds.northEast;
    }

On a tangent, it probably should be a factory rather than a mutable method and the `bufferRatio` parameter is hard to use correctly. For example, bufferRatio = 0.5 would increase the bounds to 4 times its original size.

### DIFF
--- a/lib/src/geo/latlng_bounds.dart
+++ b/lib/src/geo/latlng_bounds.dart
@@ -172,15 +172,6 @@ class LatLngBounds {
     return true;
   }
 
-  /// Expands bounds by decimal degrees unlike [extend] or [extendBounds]
-  void pad(double bufferRatio) {
-    final heightBuffer = (_sw!.latitude - _ne!.latitude).abs() * bufferRatio;
-    final widthBuffer = (_sw!.longitude - _ne!.longitude).abs() * bufferRatio;
-
-    _sw = LatLng(_sw!.latitude - heightBuffer, _sw!.longitude - widthBuffer);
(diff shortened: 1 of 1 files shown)


ignatz:
Any thoughts on deleting already broken code?

JaffaKetchup:
Hey @ignatz, sorry for the massive delay. I'll take a look tommorrow.

JaffaKetchup (approved):

JaffaKetchup:
Thanks for the contribution :)

S-ecki:
Thanks for the contribution! 
I am maintaining a codebase where `.pad()` was used and want to upgrade the dependency now. How would you go about changing the code using `.pad()`?

ignatz:
Naively, grab above version and provide it as a function or extension on latlngbounds

PR #1431, opened 2023-01-13
Migrate LatLngBounds to nullsafety

TesteurManiak:
**BREAKING CHANGES**

Updated `LatLngBounds` constructor and its methods to rely on non-nullable values.

## Migration guide

**BAD:**
```dart
final bounds = LatLngBounds();
bounds.extend(dublin);
bounds.extend(paris);
bounds.extend(london);
```

**GOOD:**
```dart
final bounds = LatLngBounds.fromPoints([dublin, paris, london]);
```

And all method related to the `LatLngBounds` now relies on non-nullable parameters and returns non-nullable values:

* `LatLngBounds`: takes 2 `LatLng` parameters instead of `LatLng?`
* `extend()`: takes a `LatLng` instead of `LatLng?`
* `southWest`: returns a `LatLng` instead of `LatLng?`
* `northEast`: returns a `LatLng` instead of `LatLng?`
* `isValid`: became deprecated as it'll always be true
* `contains()`: takes a `LatLng` instead of `LatLng?`
* `isOverlapping()`: takes a `LatLngBounds` instead of `LatLngBounds?`

I've added some unit tests to ensure that the values are still relevant even after the migration.


JaffaKetchup (approved):
LGTM. Thanks :)

PR #1440, opened 2023-01-29
renaming "add pins" in drawer to match file

augustweinbren:
Tiny change to the drawer naming of `tap_to_add.dart` to match the file's App Bar. I believe this makes it easier to find the file within the example's `lib` directory after browsing through a built version of the Example.


JaffaKetchup (approved):
Thanks for your contribution!

PR #1442, opened 2023-02-04
Batch canvas draw operations and minimize redraws on map pans for the polygon and polyline layers to significantly improve performance.

ignatz:
The batching is based on @MooNag 's proposal.

Based on a makeshift benchmark and my phone, I'm seeing:

* On the raster thread:
  * max: 2521ms => 8ms  (314x improvement)
  * avg: 1219ms => 6ms  (202x improvement)
* On the UI thread:
  * max: 49.6ms => 3.6ms (12.7x improvement)
  * avg: 19.2ms => 2.0ms (8.6x improvement)

* Before: 
image
* After: 
image

### DIFF
--- a/lib/src/layer/polygon_layer.dart
+++ b/lib/src/layer/polygon_layer.dart
@@ -6,6 +6,7 @@ import 'package:flutter_map/flutter_map.dart';
 import 'package:flutter_map/src/layer/label.dart';
 import 'package:flutter_map/src/map/flutter_map_state.dart';
 import 'package:latlong2/latlong.dart' hide Path; // conflict with Path from UI
+import 'package:flutter/foundation.dart' show kIsWeb;
 
 enum PolygonLabelPlacement {
   centroid,
@@ -79,23 +80,27 @@ class PolygonLayer extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
-    return LayoutBuilder(
-      builder: (BuildContext context, BoxConstraints bc) {
-        final size = Size(bc.maxWidth, bc.maxHeight);
-        final map = FlutterMapState.maybeOf(context)!;
+    final map = FlutterMapState.maybeOf(context)!;
+    final size = Size(map.size.x, map.size.y);
+    final origin = map.pixelOrigin;
+    final offset = Offset(origin.x.toDouble(), origin.y.toDouble());
 
-        final Iterable<Polygon> pgons = polygonCulling
-            ? polygons.where((p) {
-                return p.boundingBox.isOverlapping(map.bounds);
-              })
-            : polygons;
+    final Iterable<Polygon> pgons = polygonCulling
+        ? polygons.where((p) {
+            return p.boundingBox.isOverlapping(map.bounds);
+          })
+        : polygons;
 
-        return CustomPaint(
-          key: UniqueKey(),
-          painter: PolygonPainter(pgons, map),
-          size: size,
-        );
-      },
+    final paint = CustomPaint(
+      painter: PolygonPainter(pgons, map),
+      size: size,
+      willChange: false,
+      isComplex: true,
+    );
+    return Positioned(
+      left: -offset.dx,
+      top: -offset.dy,
+      child: kIsWeb ? paint : RepaintBoundary(child: paint),
     );
   }
 }
@@ -103,13 +108,31 @@ class PolygonLayer extends StatelessWidget {
 class PolygonPainter extends CustomPainter {
   final Iterable<Polygon> polygons;
   final FlutterMapState map;
+  final double zoom;
+  final double rotation;
 
-  const PolygonPainter(this.polygons, this.map);
+  PolygonPainter(this.polygons, this.map)
+      : zoom = map.zoom,
+        rotation = map.rotation;
+
+  int get hash {
+    _hash ??= Object.hashAll(polygons);
+    return _hash!;
+  }
+
+  int? _hash;
+
+  List<Offset> getOffsets(List<LatLng> points) {
+    return points.map((pos) => getOffset(pos)).toList();
+  }
+
+  Offset getOffset(LatLng point) {
+    final delta = map.project(point);
+    return Offset(delta.x.toDouble(), delta.y.toDouble());
+  }
 
   @override
   void paint(Canvas canvas, Size size) {
-    final rect = Offset.zero & size;
-
     var path = ui.Path();
     var paint = Paint();
     var borderPath = ui.Path();
@@ -129,7 +152,7 @@ class PolygonPainter extends CustomPainter {
     }
 
     for (final polygon in polygons) {
-      final offsets = _getOffsets(polygon.points, map);
+      final offsets = getOffsets(polygon.points);
       if (offsets.isEmpty) {
         continue;
       }
@@ -142,7 +165,7 @@ class PolygonPainter extends CustomPainter {
 
       final holeOffsetsList = List<List<Offset>>.generate(
           polygon.holePointsList?.length ?? 0,
-          (i) => _getOffsets(polygon.holePointsList![i], map));
+          (i) => getOffsets(polygon.holePointsList![i]));
 
       if (holeOffsetsList.isEmpty) {
         if (polygon.isFilled) {
@@ -195,7 +218,6 @@ class PolygonPainter extends CustomPainter {
     }
 
     drawPaths();
-    canvas.clipRect(rect);
   }
 
   Paint _getBorderPaint(Polygon polygon) {
@@ -262,11 +284,12 @@ class PolygonPainter extends CustomPainter {
   }
 
   @override
-  bool shouldRepaint(PolygonPainter oldDelegate) => true;
-}
-
-List<Offset> _getOffsets(List<LatLng> points, FlutterMapState map) {
-  return points.map((pos) => map.getOffsetFromOrigin(pos)).toList();
+  bool shouldRepaint(PolygonPainter oldDelegate) {
+    return kIsWeb ||
+        oldDelegate.zoom != zoom ||
+        oldDelegate.rotation != rotation ||
+        oldDelegate.hash != hash;
+  }
 }
 
 double _dist(Offset v, Offset w) {
--- a/lib/src/layer/polygon_layer.dart
+++ b/lib/src/layer/polygon_layer.dart
@@ -1,4 +1,5 @@
 import 'dart:math';
+import 'dart:ui' as ui;
 
 import 'package:flutter/widgets.dart';
 import 'package:flutter_map/flutter_map.dart';
@@ -12,11 +13,9 @@ enum PolygonLabelPlacement {
 }
 
 class Polygon {
-  final Key? key;
   final List<LatLng> points;
-  final List<Offset> offsets = [];
   final List<List<LatLng>>? holePointsList;
-  final List<List<Offset>>? holeOffsetsList;
+
   final Color color;
   final double borderStrokeWidth;
   final Color borderColor;
@@ -25,15 +24,19 @@ class Polygon {
   final bool isFilled;
   final StrokeCap strokeCap;
   final StrokeJoin strokeJoin;
-  late final LatLngBounds boundingBox;
   final String? label;
   final TextStyle labelStyle;
   final PolygonLabelPlacement labelPlacement;
   final bool rotateLabel;
 
+  LatLngBounds? _boundingBox;
+  LatLngBounds get boundingBox {
+    _boundingBox ??= LatLngBounds.fromPoints(points);
+    return _boundingBox!;
+  }
+
   Polygon({
     required this.points,
-    this.key,
     this.holePointsList,
     this.color = const Color(0xFF00FF00),
     this.borderStrokeWidth = 0.0,
@@ -47,9 +50,19 @@ class Polygon {
     this.labelStyle = const TextStyle(),
     this.labelPlacement = PolygonLabelPlacement.centroid,
     this.rotateLabel = false,
-  }) : holeOffsetsList = null == holePointsList || holePointsList.isEmpty
-            ? null
-            : List.generate(holePointsList.length, (_) => []);
+  });
+
+  /// Used to batch draw calls to the canvas.
+  int get renderHashCode => Object.hash(
+      holePointsList?.length ?? 0,
+      color,
+      borderStrokeWidth,
+      borderColor,
+      isDotted,
+      isFilled,
+      strokeCap,
+      strokeJoin,
+      labelStyle);
 }
 
 class PolygonLayer extends StatelessWidget {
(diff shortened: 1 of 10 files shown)


MooNag:
Reviewed this and tested the polygon layer in a a second app. I have made some improvements for performance and code quality. Ill try and commit them as a suggestion or a PR to your branch (not sure the exact right way to do that yet so i might cause some chaos. I might need to just do it as a review).

There is a chance that this will have an issue with culling enabled and caching the layer render, but i was not able to get it to consistently occur without toggling off isComplex in the CustomPaint.

Performance test in my real world app:
FM 3.1.0      - 15 FPS
This PR        - 42 FPS
My old PR   - 41 FPS

Id be happy to submit an approving review after taking a look at my recommendations.

lib/src/layer/polygon_layer.dart
```diff
@@ -47,9 +51,19 @@ class Polygon {
...
     this.labelPlacement = PolygonLabelPlacement.centroid,
     this.rotateLabel = false,
-  }) : holeOffsetsList = null == holePointsList || holePointsList.isEmpty
-            ? null
-            : List.generate(holePointsList.length, (_) => []);
+  });
+
+  /// Used to batch draw calls to the canvas.
+  int get renderHashCode => Object.hash(
+      holePointsList?.length ?? 0,
```
MooNag, line 58:
```suggestion
      holePointsList,
```

Since 2 lists with the same length would hash to the same value otherwise!

lib/src/layer/polygon_layer.dart
```diff
@@ -1,22 +1,22 @@
 import 'dart:math';
```
MooNag, line 1:
```suggestion
```
Unused import

lib/src/layer/polygon_layer.dart
```diff
@@ -58,133 +72,190 @@ class PolygonLayer extends StatelessWidget {
...
+    final Iterable<Polygon> pgons = polygonCulling
+        ? polygons.where((p) {
+            return p.boundingBox.isOverlapping(map.bounds);
+          })
+        : polygons;
+
+    final paint = CustomPaint(
+      painter: PolygonPainter(pgons, map),
+      size: size,
+      willChange: false,
```
MooNag, line 97:
```suggestion
```
Remove this since it is the default value already

lib/src/layer/polygon_layer.dart
```diff
@@ -58,133 +72,190 @@ class PolygonLayer extends StatelessWidget {
...
+  int get hash {
+    _hash ??= Object.hashAll(polygons);
+    return _hash!;
+  }
+
+  int? _hash;
 
-  PolygonPainter(this.polygonOpt, this.rotationRad);
+  List<Offset> getOffsets(List<LatLng> points) {
+    return points.map((pos) => getOffset(pos)).toList();
```
MooNag, line 126:
```suggestion
    return List.generate(points.length, (index) {
      final delta = map.project(points[index]);
      return Offset(delta.x.toDouble(), delta.y.toDouble());
    }, growable: false);
```

This is significantly more performant since it generates the list directly. I saw 8ms -> 0.4ms in profiling 500 polygons

lib/src/layer/polygon_layer.dart
```diff
@@ -58,133 +72,190 @@ class PolygonLayer extends StatelessWidget {
...
+  }
+
+  Offset getOffset(LatLng point) {
+    final delta = map.project(point);
+    return Offset(delta.x.toDouble(), delta.y.toDouble());
+  }
```
MooNag, line 132:
```suggestion
```

Not needed in the polygon layer since it was moved directly into getOffsets

lib/src/layer/polygon_layer.dart
```diff
@@ -58,133 +72,190 @@ class PolygonLayer extends StatelessWidget {
...
+
+      final hash = polygon.renderHashCode;
+      if (lastHash != null && lastHash != hash) {
+        drawPaths();
+      }
+      lastHash = hash;
+
+      final holeOffsetsList = List<List<Offset>>.generate(
+          polygon.holePointsList?.length ?? 0,
+          (i) => getOffsets(polygon.holePointsList![i]));
```
MooNag, line 168:
```suggestion
          (i) => getOffsets(polygon.holePointsList![i]), growable: false);
```

Make it not growable for some potential performance gains.

lib/src/layer/polygon_layer.dart
```diff
@@ -195,87 +266,36 @@ class PolygonPainter extends CustomPainter {
...
+        oldDelegate.hash != hash;
   }
+}
 
-  double _dist2(Offset v, Offset w) {
-    return _sqr(v.dx - w.dx) + _sqr(v.dy - w.dy);
-  }
+double _dist(Offset v, Offset w) {
+  return sqrt(_sqr(v.dx - w.dx) + _sqr(v.dy - w.dy));
+}
```
MooNag, line 297:
```suggestion
```

Not needed if the built in function is used

lib/src/layer/polygon_layer.dart
```diff
@@ -195,87 +266,36 @@ class PolygonPainter extends CustomPainter {
...
 
-  double _sqr(double x) {
-    return x * x;
-  }
+double _sqr(double x) {
+  return x * x;
```
MooNag, line 300:
```suggestion
```

Also not needed if the built in function is used.

lib/src/layer/polygon_layer.dart
```diff
@@ -58,133 +72,190 @@ class PolygonLayer extends StatelessWidget {
...
+        }
+      }
+    }
+  }
+
+  void _paintDottedLine(
+      ui.Path path, List<Offset> offsets, double radius, double stepLength) {
     var startDistance = 0.0;
     for (var i = 0; i < offsets.length; i++) {
       final o0 = offsets[i % offsets.length];
```
MooNag, line 261: **ACTUALLY FOR LINE 263:**

\-      final totalDistance = _dist(o0, o1);
\+      final totalDistance = (o0 - o1).distance;

There is a built in method that will calculate the distance, this should be more performant and also have cleaner code (since we can remove the helper functions)

lib/src/layer/polyline_layer.dart
```diff
@@ -1,15 +1,14 @@
 import 'dart:math';
```
MooNag, line 1:
```suggestion
```

Unused

lib/src/layer/polyline_layer.dart
```diff
@@ -233,46 +275,46 @@ class PolylinePainter extends CustomPainter {
...
-  bool shouldRepaint(PolylinePainter oldDelegate) => false;
+  bool shouldRepaint(PolylinePainter oldDelegate) {
+    return kIsWeb ||
+        oldDelegate.zoom != zoom ||
+        oldDelegate.rotation != rotation ||
+        oldDelegate.hash != hash;
+  }
 }
 
 double _dist(Offset v, Offset w) {
```
MooNag, line 316: Also remove these if we use the built in function!

lib/src/layer/polyline_layer.dart
```diff
@@ -127,95 +115,149 @@ class PolylinePainter extends CustomPainter {
...
+
+  int get hash {
+    _hash ??= Object.hashAll(polylines);
+    return _hash!;
+  }
+
+  int? _hash;
+
+  List<Offset> getOffsets(List<LatLng> points) {
+    return points.map((pos) => getOffset(pos)).toList();
```
MooNag, line 133:
```suggestion
    return List.generate(points.length, (index) {
      return getOffset(points[index]);
    }, growable: false);
```

Use a list generator again for performance reasons. However we will not embed getOffset since it is used in other parts of the code here

lib/src/layer/polyline_layer.dart
```diff
@@ -127,95 +115,149 @@ class PolylinePainter extends CustomPainter {
...
   }
 
-  void _paintDottedLine(Canvas canvas, List<Offset> offsets, double radius,
-      double stepLength, Paint paint) {
-    final path = ui.Path();
+  void _paintDottedLine(
+      ui.Path path, List<Offset> offsets, double radius, double stepLength) {
     var startDistance = 0.0;
     for (var i = 0; i < offsets.length - 1; i++) {
       final o0 = offsets[i];
```
MooNag, line 263: **FOR LINE 265**

For the same performance and code quality reasons as before.

\-      final totalDistance = _dist(o0, o1);
\+      final totalDistance = (o0 - o1).distance;

Using the built in function will let us remove _dist(v, w) and _sqr (x)
Though I cannot suggest removing them in review you will have to create another commit to do that.

MooNag (changes requested):
Looks good so far! Just a few things.

Is there a specific reason we are using zoom and rotation in the shouldRepaint functions vs just using map.hashCode? We dont have to change it, it just seems like it would be cleaner code with the same results and less susceptible to breaking in the future. 

Let me know if you have any questions on the feedback!

ignatz:
> Is there a specific reason we are using zoom and rotation in the shouldRepaint functions vs just using map.hashCode? We dont have to change it, it just seems like it would be cleaner code with the same results and less susceptible to breaking in the future.

I'm not sure I understand the suggestion of using map.hashCode correctly. shouldRepaint tells the repaint boundary on when to invalidate the cash, i.e. when two Painter objects should be considered visually different. If by map.hashCode you mean the FlutterMapState's dart-Object hashCode (which doesn't seem to be overridden), i'm not sure how that would achieve this.

lib/src/layer/polygon_layer.dart
ignatz: Makes sense

lib/src/layer/polygon_layer.dart
ignatz: Nice!

lib/src/layer/polygon_layer.dart
ignatz: Happy to rely on library functionality. The performance argument seems a bit far fetched. it's implemented as double get distance => math.sqrt(dx * dx + dy * dy) in (sky_engine/lib/ui/geometry.dart).

lib/src/layer/polygon_layer.dart
ignatz: Like me some less mutability but the performance argument seems far fetched.

lib/src/layer/polyline_layer.dart
ignatz: Done

lib/src/layer/polyline_layer.dart
ignatz: Done

lib/src/layer/polygon_layer.dart
MooNag: When I profiled it using a growable list generator (growable: true) it only went from 8ms to about 3ms in the main getOffsets function, rather than 0.4ms. It could be a non-issue but generally non-mutable lists have significantly higher performance, or can have better performance in certain situations.

lib/src/layer/polygon_layer.dart
MooNag: I haven't profiled it here but generally code runs faster when it is reused (especially in low level libraries), because the CPU cache does not miss code that is run frequently as much. Also generally lower level libraries can have extra compiler optimizations (like using a CPU specific instruction for SQRT for instance) that just "work" correctly where you might only see the performance in some situations. Lastly, sky_engine is going to have more people auditing the performance of the distance function than in flutter_map so any odd tricks or issues will be more likely to be fixed there; let me know if that makes sense or if you have questions!

MooNag (approved):
Looks good to me!

lib/src/layer/polygon_layer.dart
ignatz: Using standard functionality always makes sense to me. I'm just not buying in the performance argument.

The rest of the argument is more of theoretical nature but who doesn't like a good geeky argument? :) You're arguments are:
 * improved instruction cache hit rate: It's gonna basically be the same
 * improved instructions: any optimization you're thinking of would happen on the math.sqrt level and not on the sky_engine level. Arguable SQRT is slow that's were fast square-root algorithms would come in (quake for the win).
 * better maintainance for sky_engine: sure but in practice there's really only so many ways to implement pythagoras :)

Look i'm happy to use the distance function. I'm just saying that the performance will be the same

ignatz:
> Just going to have 1 more person check since its a fairly large rendering change and we should get it merged in soon. Please send me a ping if we don't get it merged in by next Friday (Feb 17th)!

That makes sense. I appreciate all the eyes we can get. Given the testing situation, I'm certainly not convinced that we're not introducing any new edge cases. I'd not be surprised if something bubbles up and we'll have to iterate. The less we need to, the better. Thanks

lib/src/layer/polygon_layer.dart
ignatz: I appreciate that you profiled that beats certainly anything that I can argue. That said, if your statement: 

>  but generally non-mutable lists have significantly higher performance

was true, that would generally strike me as surprising. Growable or not, for any given size it's just a hunk of continuous memory and on the instruction level you generally don't have a notion of immutable memory (you can make pages read-only but that's different).
Similar implementations in other languages sometimes pull off a so-called small-vector optimization but that's also generally unrelated to mutability.

I could certainly imagine the underlying implementations to be different but in that case I would consider any slowness to be a bug. Mind sharing your benchmark?

JaffaKetchup (approved):
As far as I can see, everything is OK. As you said, there might be issues further down the road, but I think everything is pretty good.

In terms of the actual performance gain on my device, the UI thread is better, but the raster thread is the same, if not a little worse (@MooNag).

(master vs fast_poly in that order):

https://user-images.githubusercontent.com/58115698/219422860-10fdd16c-a547-4d0c-a256-0d11dbd3edb7.mp4

ignatz:
> As far as I can see, everything is OK. As you said, there might be issues further down the road, but I think everything is pretty good.
> 
> In terms of the actual performance gain on my device, the UI thread is better, but the raster thread is the same, if not a little worse (@MooNag).
> 
> (master vs fast_poly in that order):

Thank you for taking the time to actually validate. Unfortunately for you, your observation is very much expected. The performance should be the same. The two optimizations kick in when you have:
 * polylines with the same paint property (bigger optimization)
 * and independently when you pan (smaller optimization)

In your case you're having few polylines with different properties + you're rotating/zooming. If you're still up for further validation, consider drawing some thousands polylines with the same color. Sorry, I should have been clearer..

JaffaKetchup:
Ah no worries. As far as I'm concerened, two people saying performance has improved is enough, and this shows that normal performance has remained the ~same.
Apologies for taking so long to get this through! @MooNag's recently been invited to the repo with write perms, so he's going to merge when he's ready.

MooNag:
Just checked the video, the performance regression seems insignificant enough. We will merge and watch for any issues!

JaffaKetchup:
Thanks @ignatz :)

PR #1448, opened 2023-02-17
Add secondary tap to MapOptions

### LINKED ISSUE #1444: Add onSecondaryTap  (right-click for windows) callback

Please add the onSecondaryTap click callback so windows users can capture a right-click event when clicking on the map. 

Currently, having to use the onLongPresss callback but this is not intuitive too windows users:

        onLongPress: (tapPosition, point) {

            showDialog(
                context: context,
                builder: (BuildContext context) {
                  return MapContextMenuDialog(tapPosition: LatLng(point.latitude, point.longitude));
                });
          },

See https://api.flutter.dev/flutter/widgets/GestureDetector/onSecondaryTap.html

Heres a package that is capturing a right-click event to display a context menu:  https://pub.dev/packages/native_context_menu.

### Platforms Affected

Windows, MacOS, Linux

### Severity

### Requirements

TesteurManiak:
Fix #1444 

I've added a property `onSecondaryTap` to `MapOptions` which is usually associated to the right-click action. To implement this behavior I had to remove the dependency to the package [positioned_tap_detector_2](https://pub.dev/packages/positioned_tap_detector_2) which I've re-implemented directly inside flutter_map as it seems like this package is not maintained anymore.

You can try the gesture in the example app. I've tested it on MacOS but didn't try on windows or linux but I trust that Flutter is able to recognize the right-click action on those platforms.

P.S: I had to re-generate the platform sources for the project to compile hence why there is changes for the android, ios, web and macos folders in `example`;


JaffaKetchup (approved):
LGTM AFAIK. Thanks :)

PR #1453, opened 2023-03-01
Prevent double scrolling

rorystephenson:
Before this change the `PointerScrollEvent` handler was not registered with `PointerSignalResolver` which meant that if scrolling was enabled for `FlutterMap` and it was within a scrollable widget or had a scrollable child widget then they would both be scrolled simultaneously. This is only noticeable on web since on mobile a drag gesture is used to scroll/pan instead of mouse scrolling.

### DIFF
--- a/example/ios/Podfile
+++ b/example/ios/Podfile
@@ -1,0 +1,41 @@
+# Uncomment this line to define a global platform for your project
+# platform :ios, '11.0'
+
+# CocoaPods analytics sends network stats synchronously affecting flutter build latency.
+ENV['COCOAPODS_DISABLE_STATS'] = 'true'
+
+project 'Runner', {
+  'Debug' => :debug,
+  'Profile' => :release,
+  'Release' => :release,
+}
+
+def flutter_root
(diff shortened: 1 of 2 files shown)


JaffaKetchup:
Thanks @rorystephenson, we'll try to have a look at this soon!

TesteurManiak:
The code looks good to me, do you think you can add an example page that would showcase your changes and help us to test them ?

rorystephenson:
@TesteurManiak You can test this with the "Map inside ListView" example **in web** by scrolling with the mouse over the map inside the ListView. Without these changes the map and the ListView will both react to the scrolling. With these changes only the map will react (by zooming).

TesteurManiak (approved):
I've been able to test the example you've mentioned @rorystephenson thanks 👍 
I wasn't able to find any regression in the other scrolling behavior so everything is good for me (you can double check if you want @JaffaKetchup)

jbournonville:
Hello, do you know when this will be released ?

JaffaKetchup:
Hi @jbournonville,
We're working towards v4 in #1475.

Issue #1454, opened 2023-03-01
[FEATURE] Remove `absorbPanEventsOnScrollables` flag.


rorystephenson:
### What do you want implemented?

Remove the `absorbPanEventsOnScrollables` flag. If you have a FlutterMap inside a scrollable and you want dragging the map to cause the scrollable to scroll then you can just disable the `drag` interactive flag. It seems to me that [this has the same effect](https://github.com/fleaflet/flutter_map/blob/3e8368d85d45e1431ed78488f0df102cef81b925/lib/src/map/flutter_map_state.dart#L117-L118) and I can't see why you would need dragging enabled but for the map to not absorb pan events (which makes the dragging not work anyway)?

If you agree I am happy to open a PR including removing unnecessary commented code in `flutter_map_state.dart`.

### Platforms Affected

Android, iOS, Web, Windows, MacOS, Linux

### Severity

Minimum: Not required for my use

### Requirements

- [X] I agree to follow this project's [Code of Conduct](https://github.com/fleaflet/flutter_map/blob/master/CODE_OF_CONDUCT.md)
- [X] I am using the [latest stable version](https://pub.dev/packages/flutter_map) of this package
- [X] I have checked for similar feature requests which may be duplicates

ibrierley:
Not quite sure this is quite the same thing (but I lose track fast!)....For example with the dragmarker (& line_editor) plugin, it allows panning, but needs absorbPanEventsOnScrollables disabled to be able to drag icons iirc ?

rorystephenson:
@ibrierley thanks for pointing that out! I did some further digging and realised that you are right that this would break some plugins. The good news is the fix in #1453 resolves this issues. I've decided to include removal of the flag in that PR (see the second part of the description in that PR for more info).

JaffaKetchup:
Closed by #1455 AFAIK.

PR #1455, opened 2023-03-02
Remove absorbPanEventsOnScrollables option

rorystephenson:
The only reason to disable absorbing of pan events is because some plugins which use drag gestures rely on it. It turns out if those plugins detect vertical/horizontal drag instead of pan this flag is not necessary.

Note that affected plugins need to be updated and this should be released in a major version bump as it is a breaking change.

See #1454 discussion leading to this PR.

## Before releasing

- [ ] Notify affected plugin authors of the required change.
- [ ] Create new major version with changelog entry to describe how other plugins must migrate.

## Understanding more

- When `absorbPanEventsOnScrollables` is true a horizontal & vertical drag gesture detector are added to FlutterMap and if they are triggered FlutterMap's scale gesture logic is triggered since the vertical/horizontal drag gesture and scale gesture are in the same `GestureArenaTeam` and the scale gesture is the captain of that team. Horizontal/vertical drag gestures are recognised faster than scale/pan gestures and therefore when there are competing gestures they will prevail.
- [Scrollable Flutter widgets detect horizontal/vertical drag gestures](https://github.com/flutter/flutter/blob/12cb4eb7a009f52b347b62ade7cb4854b926af72/packages/flutter/lib/src/widgets/scrollable.dart#L622-L662) which is why FlutterMap needs to detect them and then defer to its scale gesture if it wants to be pannable inside of a ListView.
- `flutter_map_dragmarker` and likely other plugins detect pan gestures instead of horizontal/vertical drag gestures. This means that when `absorbPanEventsOnScrollables` is enabled those plugins don't work because FlutterMap's horizontal/vertical drag gestures have precedence. This is a simple fix, affected code should listen to `onVerticalDrag(Start/Update/End)` and `onHorizontalDrag(Start/Update/End)` instead of `onPan(Start/Update/End)`.

## Demo

With `absabsorbPanEventsOnScrollables` removed such that horizontal/vertical gestures are always detected in FlutterMap (see this PR's code changes) and a modified `flutter_map_dragmarker` (detecting horizontal/vertical drag instead of pan) is used it is possible to have a pan-able FlutterMap inside of a ListView containing drag markers which can be dragged:

https://user-images.githubusercontent.com/3683599/222518631-4d350d44-eb50-4368-83a3-7a0a698312d3.mp4

### DIFF
--- a/example/lib/pages/map_inside_listview.dart
+++ b/example/lib/pages/map_inside_listview.dart
@@ -23,7 +23,6 @@ class MapInsideListViewPage extends StatelessWidget {
               height: 300,
               child: FlutterMap(
                 options: MapOptions(
-                  absorbPanEventsOnScrollables: true,
                   center: LatLng(51.5, -0.09),
                   zoom: 5,
                 ),
--- a/lib/flutter_map.dart
+++ b/lib/flutter_map.dart
@@ -251,8 +251,6 @@ class MapOptions {
   /// see [InteractiveFlag] for custom settings
   final int interactiveFlags;
 
-  final bool absorbPanEventsOnScrollables;
-
   final TapCallback? onTap;
   final TapCallback? onSecondaryTap;
   final LongPressCallback? onLongPress;
@@ -293,7 +291,6 @@ class MapOptions {
   final bool keepAlive;
 
   MapOptions({
-    this.absorbPanEventsOnScrollables = true,
     this.crs = const Epsg3857(),
     LatLng? center,
     this.bounds,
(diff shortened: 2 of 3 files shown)


JaffaKetchup:
Thanks @rorystephenson, we'll try to have a look at this soon!

TesteurManiak (approved):
LGTM 👍

JaffaKetchup:
@rorystephenson When you get a moment from working on #1475, can you resolve the conflicts here for merge? (Also note that #1475 needs the same conflicts resolved).

JaffaKetchup:
@rorystephenson You happy for this to be merged now? Not sure if it will cause more conflicts.

rorystephenson:
> @rorystephenson You happy for this to be merged now? Not sure if it will cause more conflicts.

Yep I think it's good to go.

PR #1462, opened 2023-03-10
Add a threshold for rasterization to avoid excessive fixed overhead c…

ignatz:
…ost for cheap redraws.

### DIFF
--- a/lib/src/layer/polygon_layer.dart
+++ b/lib/src/layer/polygon_layer.dart
@@ -71,10 +71,20 @@ class PolygonLayer extends StatelessWidget {
   /// screen space culling of polygons based on bounding box
   final bool polygonCulling;
 
+  /// The layer allows to rasterize and translate the underlying canvas on
+  /// panning the map (does not apply for rotations and zooming). This can
+  /// greatly improve performance if the set polygons in the map's viewport
+  /// doesn't change. However, rasterization comes at a fixed overhead, which
+  /// amortizes when drawing the canvas itself is costly and the use-case makes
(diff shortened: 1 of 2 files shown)


JaffaKetchup:
Thanks @ignatz, we'll try to have a look at this soon!

lib/src/layer/polygon_layer.dart
```diff
@@ -84,27 +94,29 @@ class PolygonLayer extends StatelessWidget {
     final origin = map.pixelOrigin;
     final offset = Offset(origin.x.toDouble(), origin.y.toDouble());
 
-    final Iterable<Polygon> pgons = polygonCulling
+    final List<Polygon> pgons = polygonCulling
```
TesteurManiak, line 97: Is there a reason why you converted the `Iterable<Polygon>` to a `List<Polygon>` I don't really see the benefits here 🤔 
Otherwise everything else looks good to me.

(Do you want to double check @JaffaKetchup?)

ignatz: Since I recently changed it from an List to an Iterable I felt like going full circle :). Jokes aside, I added the length comparison as quick true-positive check. Since both, calculating the length and the hashing, require a full traversal (e.g. when the raster cache is being hit), I felt the benefit of delaying the list conversion was harder to justify. Hope that makes sense. In other words, I can change it back and it might be marginally letss work in cases where the length actually differs and marginally more work in all others.

TesteurManiak (approved):

PR #1470, opened 2023-03-23
adjustCenterIfOutsideMaxBounds calculates correctly the center when size.x > width  or size.y > height

dumabg:
Sometimes size.x or size.y were greater than the total with or height, and, in this case, leftOkCenter is  > rightOkCenter or botOkCenter is < topOkCenter. When this occurs, the code  in the bottom,
```

 if (newCx < leftOkCenter ||
        newCx > rightOkCenter ||
        newCy < topOkCenter ||
        newCy > botOkCenter) {
      return null;
```
satisfies the conditions and return null.

The consequence is that when using the component, it can't return for a zoom.

This is the sample that I'm using:

```dart
FlutterMap(
      mapController: controller.mapController,
      options: MapOptions(
        center: controller.center,
        zoom: 19,
        enableScrollWheel: false,
        onPositionChanged: controller.onPositionChanged,
        onMapEvent: controller.onMapEvent,
        maxBounds: LatLngBounds(
          LatLng(-15.94720, -47.98902),
          LatLng(-15.94872, -47.98749),
        ),
        interactiveFlags: InteractiveFlag.all,
      ),
      children: [
        TileLayer(
          subdomains: const ['a', 'b', 'c'],
          maxZoom: 22,
          minZoom: 19,
          maxNativeZoom: 19,
        ),
      ],
    );
```
controller.center has the value lat: -15.94774,  lon: -47.98830.
The most important is the maxBounds values.

If zoom in, it works, but if zoom out to return to previous zoom, the zoom doesn't work. The reason is explained above.

I have changed the halfSize calculation, taking the min between the real size and the expected size and now it works correctly.

I only have tested with this sample. I don't know if this change could cause another problems.


JaffaKetchup:
@dumabg Thanks for the contribution! I'll try to review soon.

JaffaKetchup:
@dumabg Can you add an MRE for me to test with? If possible, can you also test with #1475 to see if that may have fixed the issue?

JaffaKetchup:
This PR is now included in the CHANGELOG edit in #1482/ 040c6f1, so it must be included in v4.

JaffaKetchup (changes requested):
Thanks for the contribution @dumabg, however this has an unexpected side effect.

If `maxBounds` is set smaller than the current viewport size, and the initial configuration means that the map is zoomed outside of those bounds, it should(?) prevent any map interaction at all, as all states are outside the bounds.
However, this PR seems to mean the map can be panned in at least one direction (vertically downwards) in my case.

We should check whether the current behaviour is what is wanted, but as this changes the current behaviour unintentionally, we are unable to merge this for v4.

If you still want to work on this, please create an MRE in order for us to more clearly see your issue.

dumabg:
I've tested the new version 4 and occurs the same.

JaffaKetchup:
@dumabg Not sure what you mean by that?
Can you provide an MRE to show the issue this PR resolves?

ibrierley:
Yes, I'm a bit unclear on this one. Typically this issue is when there is a zoom level and maxBounds set which can't be resolved (and it's also worth double checking rotated maps). I can't quite work from the initial description what the issue is that's happening to know if its correct or not.

dumabg:
The MRE:

[pre.zip](https://github.com/fleaflet/flutter_map/files/11491958/pre.zip)

ibrierley:
I'm a bit confused here though, have tested, and the PR has the same behaviour doesn't it ?

Maybe you could do a video with the with/without your PR, as I suspect I'm at cross purposes somewhat ?

JaffaKetchup:
Closing due to inactivity.

Issue #1394, opened 2022-10-31
[FEATURE] 


gpsim:
### What do you want implemented?

Hello, I tried to achieve Baidu layer tiles, but always render out of order.

http://online1.map.bdimg.com/onlinelabel/?qt=tile&x={x}&y={y}&z={z}&styles=pl&scaler=1&p=1

### What other alternatives are available?

I tried to modify the proj4 property with the following code.

import 'package:flutter/material.dart';
import 'package:flutter_map/plugin_api.dart';
import 'package:flutter_map_example/widgets/drawer.dart';
import 'package:latlong2/latlong.dart';
import 'package:proj4dart/proj4dart.dart' as proj4;

class EPSG3413Page extends StatefulWidget {
  static const String route = 'EPSG3413 Page';

  const EPSG3413Page({Key? key}) : super(key: key);

  @override
  _EPSG3413PageState createState() => _EPSG3413PageState();
}

class _EPSG3413PageState extends State<EPSG3413Page> {
  late final Proj4Crs epsg3413CRS;

  double? maxZoom;

  @override
  void initState() {
    super.initState();

    // 9 example zoom level resolutions
    final resolutions = <double>[
      262144,
      131072,
      65536,
      32768,
      16384,
      8192,
      4096,
      2048,
      1024,
      512,
      256,
      128,
      64,
      32,
      16,
      8,
      4,
      2,
      1
    ];

    final epsg3413Bounds = Bounds<double>(
      const CustomPoint<double>(0.0, 20037508.342789244),
      const CustomPoint<double>(20037508.342789244, 0.0),
    );

    maxZoom = (resolutions.length - 1).toDouble();

    // EPSG:3413 is a user-defined projection from a valid Proj4 definition string
    // From: http://epsg.io/3413, proj definition: http://epsg.io/3413.proj4
    // Find Projection by name or define it if not exists
    final proj4.Projection epsg3413 = proj4.Projection.add('EPSG:3857',
        '+proj=merc +a=6378206 +b=6356584.314245179 +lat_ts=0.0 +lon_0=0.0 +x_0=0 +y_0=0 +k=1.0 +units=m +nadgrids=@null +wktext  +no_defs');

    epsg3413CRS = Proj4Crs.fromFactory(
      code: 'EPSG:3857',
      proj4Projection: epsg3413,
      resolutions: resolutions,
      //bounds: epsg3413Bounds,
      origins: [const CustomPoint(0, 0)],
      scales: null,
[...]

Issue #1263, opened 2022-06-09
[FEATURE] Add function animatedMove() to MapController


LostInDarkMath:
**Describe The Problem**
I really like the `_animatedMapMove()` function from one of your example projects: https://github.com/fleaflet/flutter_map/blob/master/example/lib/pages/animated_map_controller.dart#L42
I find myself copying those in all of my projects taht uses your library.

**Describe Your Solution**
I think it would be nice if this function would became part of your API, maybe of the `MapController`. But I'm not sure if and how this could be implemented because if depends on the `TickerProviderStateMixin`.

**Applicable Platforms**
Select platforms that this request applies to.

- [x] All/any

LostInDarkMath:
`staleLabel.remove()`

ibrierley:
My bugbear with this one, is that actually it doesn't work properly. 

It looks better than nothing though (especially at similar zoom levels), but I'd probably wince at merging something that I know is broken :D. (Although we do have the same code in the double tap controller...so maybe we could just expose that and then fix it later, so it's at least in once place...or let it be overridden)

If you want to see what I mean, set the  AnimationController duration to 5000ms rather than 500ms, click on Paris and then London and vice versa (key being, they have different zooms). You will notice that London actually goes away and off the screen before coming back. 

You can't really tween in a simple way a scale and a translation, as the scale affects the translation. I did do a flyTo conversion which corrected that, but I think it was kinda felt maybe it polluted the code a bit, but maybe we should get a plugin working for it (or as I said, expose some method that you can pass your own thing to as well), as it is a good feature.

LostInDarkMath:
`staleLabel.remove()`

LostInDarkMath:
```dart
void onStaleLabelAdded(StaleLabel staleLabel){
  staleLabel.remove();
}
```

JosefWN:
I have forked @ibrierley's `flyTo` and included rotation which I needed. It's relatively straight-forward to extend the `MapControllerImpl` creating an `AnimatedMapController`, although it requires https://github.com/fleaflet/flutter_map/pull/1357 to function.

Perhaps we could change the _Animated MapController_ example to utilize this? It's better than what we have and wouldn't require any changes to the `flutter_map` core. We might need to test the curve a bit though, I have the impression (just visually) that the animation is fairly linear, or perhaps I'm just so used to `easeInOut`.

```dart
import 'dart:math' as math;

import 'package:flutter/material.dart';
[...]

PR #1475, opened 2023-03-29
`TileLayer` Reimplementation (& v4 Release Preparation)

rorystephenson:
This PR started with wanting to fix the broken `updateInterval` option before I got sucked in to the TileLayer rabbit hole. I hope I've significantly cleaned up the code without breaking too much. The old code was recalculating map boundaries and loading tiles on every build and in general I found it very hard to follow. Tile loading/pruning is done out of the build method now and it should be much more efficient. I've done my best to comment the code but please mention anywhere where that's missing! I am sure there are some mistakes since the changes are extensive. I'm also very happy for any suggestions for improvement.

**Breaking changes**

- Opacity is no longer an option, you can just wrap the layer with it if needed.
- Tile renamed to TileImage
- Coords renamed to TileCoordinate and the individual coordinates are not `int`s.
- minNativeZoom/maxNativeZoom now `int?` Instead of `double?` although they already were interpreted as whole numbers
- `updateInterval` removed since it was doing nothing. The same behaviour is now possible using the `tileUpdateTransformer` option.
- `tileFadeInDuration`/`tileFadeInStart`/`tileFadeInStartWhenOverride` now moved to a single `TileFadeIn` option.
- `ErrorTileCallback` signature now has Object instead of dynamic for the error, it also now contains  a nullable StackTrace

**Features**

- TileUpdateTransformer allows filtering/modifying tile updates triggered by movement. See AnimatedMapController example where this is used to preload tiles at the target location and disable loading/pruning whilst moving which causes a LOT of tile downloads for big distances.
- First tile layer tests added (not covering everything but its a start).
- Updates to individual tiles are now done only to that tile rather than calling setState(){}. This is possible because TileImage (previously Tile) is now a ChangeNotifier and Tile (previously AnimatedTile) listens for changes and updates as appropriate. In fact there are no calls to setState((){}) in TileLayer now.


example/lib/pages/animated_map_controller.dart
```diff
@@ -16,15 +16,9 @@ class AnimatedMapControllerPage extends StatefulWidget {
...
     with TickerProviderStateMixin {
-  // Note the addition of the TickerProviderStateMixin here. If you are getting an error like
-  // 'The class 'TickerProviderStateMixin' can't be used as a mixin because it extends a class other than Object.'
-  // in your IDE, you can probably fix it by adding an analysis_options.yaml file to your project
-  // with the following content:
-  //  analyzer:
-  //    language:
-  //      enableSuperMixins: true
-  // See https://github.com/flutter/flutter/issues/14317#issuecomment-361085869
-  // This project didn't require that change, so YMMV.
```
rorystephenson, line 27: Removed this note as it's no longer necessary since flutter 2.1:

https://github.com/flutter/flutter/issues/14317#issuecomment-435401569

example/lib/pages/animated_map_controller.dart
```diff
@@ -55,10 +49,25 @@ class AnimatedMapControllerPageState extends State<AnimatedMapControllerPage>
     final Animation<double> animation =
         CurvedAnimation(parent: controller, curve: Curves.fastOutSlowIn);
 
+    // Note this method of encoding the target destination is a workaround.
+    // When proper animated movement is supported (see #1263) we should be able
+    // to detect an appropriate animated movement event which contains the
+    // target zoom/center.
```
rorystephenson, line 55: See this comment for explanation of the hacky method.

lib/flutter_map.dart
```diff
@@ -30,12 +29,12 @@ export 'package:flutter_map/src/layer/marker_layer.dart';
 export 'package:flutter_map/src/layer/overlay_image_layer.dart';
 export 'package:flutter_map/src/layer/polygon_layer.dart';
 export 'package:flutter_map/src/layer/polyline_layer.dart';
-export 'package:flutter_map/src/layer/tile_layer/coords.dart';
-export 'package:flutter_map/src/layer/tile_layer/tile.dart';
 export 'package:flutter_map/src/layer/tile_layer/tile_builder.dart';
+export 'package:flutter_map/src/layer/tile_layer/tile_coordinate.dart';
```
rorystephenson: Tile was renamed to TileImage.

This is the only addition, the rest are just re-ordered because my IDE does it automatically and the order had been broken. Given that only a small amount of exports were out of order I think it makes sense to re-order them.

lib/src/layer/tile_layer/tile.dart
```diff
@@ -1,125 +1,74 @@
 import 'package:flutter/widgets.dart';
-import 'package:flutter_map/flutter_map.dart';
+import 'package:flutter_map/src/core/point.dart';
+import 'package:flutter_map/src/layer/tile_layer/tile_builder.dart';
+import 'package:flutter_map/src/layer/tile_layer/tile_image.dart';
+
+class Tile extends StatefulWidget {
```
rorystephenson, line 6: The diff is confusing because of renaming. Tile was renamed to TileImage and AnimatedTile to Tile.

lib/src/layer/tile_layer/tile_provider/base_tile_provider.dart
```diff
@@ -17,23 +17,24 @@ abstract class TileProvider {
...
+      String urlTemplate, TileCoordinate coords, TileLayer options) {
     final z = _getZoomForUrl(coords, options);
 
     final data = <String, String>{
-      'x': coords.x.round().toString(),
-      'y': coords.y.round().toString(),
-      'z': z.round().toString(),
+      'x': coords.x.toString(),
+      'y': coords.y.toString(),
+      'z': z.toString(),
```
rorystephenson, line 32: No rounding needed as they are now ints. The existence of rounding shows that tile coordinates can/should be ints instead of doubles as previously.

JaffaKetchup:
Hey @rorystephenson,
A special thanks for taking this on :)
We'll get round to looking at it soon I'm sure!

rorystephenson:
> Hey @rorystephenson, A special thanks for taking this on :) We'll get round to looking at it soon I'm sure!

It has been a BIG mission! Its been close to a year since I last dived in and whilst last time I fixed lag issues and tidied things up a fair bit I'm quite stoked with the improvements In this PR. I am in the process of writing up the changes so it will be easier to get a hang of the diff once I've done that.

JaffaKetchup:
> It has been a BIG mission! Its been close to a year since I last dived in and whilst last time I fixed lag issues and tidied things up a fair bit I'm quite stoked with the improvements In this PR. I am in the process of writing up the changes so it will be easier to get a hang of the diff once I've done that.

I can imagine, we've been putting it off for ages just because none of us have the time to do things like this unfortunately.
Looking forward to reviewing and getting this released! I'll add a special thanks to you on the Credits page of the docs for the next version.

ibrierley:
Yes, thanks for looking at this. Apologies, I'm away for over 2 weeks, so won't get chance short term to do a review (happy if anyone else wants to confirm all ok tho). Just wondering about the removal of Level stuff. It's a while since I looked at that, but I was thinking some of it was being uses as a form of caching to avoid recalculations, but I may be wrong. Is that the case, just wanted to make sure there wouldn't be any performance issues from the changes (altho tbh I doubt it would be that much anyway to calculate each zoom level update). I may be on the wrong track though, just can't really check from here, so apologies if so!

rorystephenson:
@TesteurManiak I've just encountered a race condition scenario between the TileLayer build and the map events triggering tile updating. It's possible for the TileLayer to build before a MapEvent causes new tiles to be added to the tile manager. In this scenario tiles may be missing until another build occurs. Surprisingly I've only seen this in a very particular scenario but I need to think about how it should be handled properly and whether this means I need to change how tile management works. So for no I'd suggest holding off on reviewing.

mootw:
No worries, let us know when you're ready again or if you need any assistance. I am happy to help troubleshoot if needed!

rorystephenson:
Thanks @TesteurManiak .

In this PR I have moved tile loading/pruning out of the TileLayer build(), it is now triggered by listening to the stream of  MapEvents. There are two reasons for doing this:

1. Make the build faster.
2. Allow tile loading/pruning to be debounced/filtered/modified.

When I started testing the app in profile mode a race condition occurred due to a quirk in FlutterMapState initialization where the FlutterMap's size is first null, then CustomPoint(0,0) and finally the actual size (a quirk caused by how LayoutBuilder works). This was causing only one tile to be loaded on the first load of the map. As I mentioned this is a race condition and it doesn't always happen, sometimes LayoutBuilder sets the actual size on the first go. I have just pushed a commit with some `debugPrint`s to clarify what is happening when the race condition does/doesn't occur.

**Race condition NOT occuring**

```code
I/flutter (20296): FlutterMapState LayoutBuilder set size: (395.42857142857144, 739.4285714285714)
I/flutter (20296): Size is non-zero, triggering tile update/prune.
I/flutter (20296): Updating tiles, map size: CustomPoint (395.42857142857144, 739.4285714285714)
I/flutter (20296): TileLayer build (map size CustomPoint (395.42857142857144, 739.4285714285714))
I/flutter (20296): FlutterMapState post frame callback (calling onMapReady)
```

**Race condition occurring**

```code
I/flutter (20426): FlutterMapState LayoutBuilder set size: (0.0, 0.0)
I/flutter (20426): Size is zero, scheduling tile update/prune in next frame. (TileLayer didChangeDependencies)
I/flutter (20426): TileLayer build (map size CustomPoint (0.0, 0.0))
I/flutter (20426): FlutterMapState post frame callback (calling onMapReady)
I/flutter (20426): TileLayer Post-frame callback update, size is: CustomPoint (0.0, 0.0)
I/flutter (20426): Updating tiles, map size: CustomPoint (0.0, 0.0)
I/flutter (20426): FlutterMapState LayoutBuilder set size: (395.42857142857144, 739.4285714285714)
I/flutter (20426): TileLayer build (map size CustomPoint (395.42857142857144, 739.4285714285714))
```

So this actually highlights an issue in FlutterMapState where onMapReady is called before the map has set its size. This is worth looking in to separately and I can add a workaround for TileLayer but it got me thinking: can this happen when the map moves?

When the map is moved it triggers both a rebuild of the TileLayer and a MapEvent which triggers tile update/prune. If the build happens before the update/prune then the TileLayer build may be missing tiles and they won't be shown until another build is triggered. We could add some setState() calls but that feels like a bandage and will cause extra builds which I want to avoid.

If it *is* possible for map movement to trigger the build before the emitted MapEvent triggers tile update/prune it wouldn't be very noticeable with normal movement since the build/tile update gets triggered many times a second but when the map is moved with MapController the tiles may not load.

I have just run some quick tests locally by modifying the MapController example page to zoom in a bit further so that the three locations do not share any tiles and added some debugPrints to see when the TileLayer build/tile update get fired with respect to each other when MapController triggers movement. I tapped t between London/Paris/Dublin around 50 times and every time the tiles loaded and the tile update/prune was triggered before the build but I don't know enough about flutter/dart to say for sure that this will always occur or if it's possible that the build happens first and the tiles do not appear.

If we can be sure that the update/prune will always happen before the build then I will just add a small workaround for the issue mentioned above and all is well. If not then I need to reconsider the design (I have an idea but it will require digging in to pruning logic to be sure it works).

JaffaKetchup:
(Little bit lost, I must admit, but the idea of being able to control tile loading (debounce etc) seems like a nice feature, although I can't really think of many use cases.)

rorystephenson:
> (Little bit lost, I must admit, but the idea of being able to control tile loading (debounce etc) seems like a nice feature, although I can't really think of many use cases.)
>
> EDIT: Re-read a little later, and makes more sense now.

So this should replace the broken `updateInterval` feature as you can debounce updates and it also allows for smarter tile loading/pruning based on the event that triggered movement. See the `AnimatedMapController` example for a demonstration of that.

---

@TesteurManiak / @JaffaKetchup / @mootw I have resolved the aforementioned issue by pre-creating (but not loading) missing TileImages in the build method. I am sure there are things to fix and improve but for now I think an initial review is possible and I've updated the PR description to outline the changes. Please let me know if you run in to issues or have suggestions for improvements!

JaffaKetchup:
@rorystephenson Great :) I'll try to look over this this weekend.

JaffaKetchup:
@rorystephenson `TileCoordinate` doesn't appear to be exported any more. Also, I would personally prefer to keep it called `Coords`, for compatibilities' sake, although I'm open to reasons why this would be a bad idea.

JaffaKetchup (changes requested):
I've spotted one bug that wasn't there previously in the example app. Toggling the dark mode in the tile builder page does nothing now.

Dark mode failure

Additionally, the non-exposing of `TileCoordinate`/`Coords` is also a bug - plugins rely on it quite heavily in some cases.  
I would also prefer to keep `Coords` named `Coords` instead of `TileCoordinate`, just for compatibility reasons and keeping things short. Open to other ideas on that!

---

I've also spotted an issue in the current 'master' and this branch, to do with the `InteractiveFlags`.
If dragging and flinging is enabled, then flinging is disabled mid-fling, the map gets a `setState() called during build` error.
If flinging is then re-enabled, the map splits into two, separately controllable halves.
Moving to another page crashes the entire app frame.

Fling1

Fling2

Fling3

rorystephenson:
> I've spotted one bug that wasn't there previously in the example app. Toggling the dark mode in the tile builder page does nothing now.
> 
> Dark mode failure
> 
> Additionally, the non-exposing of `TileCoordinate`/`Coords` is also a bug - plugins rely on it quite heavily in some cases. I would also prefer to keep `Coords` named `Coords` instead of `TileCoordinate`, just for compatibility reasons and keeping things short. Open to other ideas on that!
> 
> I've also spotted an issue in the current 'master' and this branch, to do with the `InteractiveFlags`. If dragging and flinging is enabled, then flinging is disabled mid-fling, the map gets a `setState() called during build` error. If flinging is then re-enabled, the map splits into two, separately controllable halves. Moving to another page crashes the entire app frame.
> 
> Fling1
> 
> Fling2
> 
> Fling3

Nice spotting I'll have a look at the dark mode bug and try add a test for it as well.

---

Maybe it's worth tackling the other issue in a separate pr if it's present in master as well in case it takes a while to merge this pr.

rorystephenson:
The dark mode issue was caused by me removing the tileContainerBuilder functionality (I'd forgotten to remove the option but it was doing nothing). The same behaviour can be achieved by wrapping the TileLayer with the desired widget and I have changed the example to do so. If there is a use case where the wrapping widget needs access to the list of tiles, which TileContainerBuilder provided, then removing it is a mistake and I will reinstate it. Are you aware of any?

I did the same thing when I removed the opacity option. I was trying to simplify the TileLayer without removing functionality as there are a lot of options however if you prefer that we keep some of these options or deprecate them before removing them I'm fine with that.

---

As for TileCoordinate I have re-exported it :+1:. I changed it from simply Coords because at first it wasn't clear to me that it was only used by the TileLayer and I was hesitant to modify it. It is slightly specific to TileLayer in that the distanceTo implementation ignores the z coordinate. That said I'm fine with changing it back if you think it's not worthwhile.

The conflict should be resolved now.

JaffaKetchup:
Thanks @rorystephenson.

I'm not aware of any reason why `tileContainerBuilder` is needed anymore, so I'm fine with that being removed. As it is a relatively niche option and this is a major update anyway, I'm happy with skipping the deprecation step.

However, it might be worth keeping `opacity`: if we can get it to the `Image` widget, we can use the [more efficient `opacity` variable there](https://api.flutter.dev/flutter/widgets/Image/opacity.html) (with an `AlwaysStoppedAnimation`). Not sure if that's possible though?

In terms of `TileCoordinate`, I think I'm coming around to it. Looking at the implementation differences (removal of generic types and changing of the constructor), I think it might make sense to renew the name. Although it definitely should be plural `TileCoordinates̳`.

rorystephenson:
> I'm not aware of any reason why `tileContainerBuilder` is needed anymore, so I'm fine with that being removed. As it is a relatively niche option and this is a major update anyway, I'm happy with skipping the deprecation step.

OK I've left it out 👍 .

> However, it might be worth keeping `opacity`: if we can get it to the `Image` widget, we can use the [more efficient `opacity` variable there](https://api.flutter.dev/flutter/widgets/Image/opacity.html) (with an `AlwaysStoppedAnimation`). Not sure if that's possible though?

That's definitely worth considering and it's not too difficulty to implement (we just need to make sure images are reloaded if the opacity changes). The behaviour will be different since overlapping tiles (from different zooms) will both be visible in this case so it would only make sense if fastReplace is enabled so that there are no overlapping tiles. I'm not sure what the typical use cases are for transparent tiles but if you think it's worth adding I can do so.

> In terms of `TileCoordinate`, I think I'm coming around to it. Looking at the implementation differences (removal of generic types and changing of the constructor), I think it might make sense to renew the name. Although it definitely should be plural `TileCoordinates̳`.

I've changed TileCoordinate to the plural 👍 .

rorystephenson:
I've added a `tileOpacity` option, let me know what you think.

rorystephenson:
I've also just pushed up a predefined TileUpdateTransformer which disables tile updates/prunes for tap/press events since they shouldn't cause the map to move and therefore shouldn't require loading/pruning. What do you think about making that the default?

On a similar note I realise I didn't share a video demonstrating the improvement in animated map movement when using a TileUpdateTransformer to preload the destination tiles and disable pruning whilst moving, have a look at the videos below:

*Current master*


https://user-images.githubusercontent.com/3683599/229725099-5053b516-4cf0-4d52-8051-a47aeca82bea.mp4



*This PR*


https://user-images.githubusercontent.com/3683599/229725141-6fdf8ce8-5726-4482-a8b7-3a32786bdc25.mp4

---

It seems to work quite well. Unfortunately the implementation of the TileUpdateTransformer is a little hacky because it needs to mark the move events with custom ids to distinguish animated start/inProgress/stop and to encode the destination location in the start event id so that the transformer can pick it up. In the comments I mention that it is a workaround in lieu of integrated animated movement support in flutter map but since I wrote that #1263 was closed as @TesteurManiak has created an animated map movement plugin (nice one!). If we want a cleaner implementation I think we have a couple of options:

1. Make it possible to encode more information when using mapController. This could be done by allowing custom map events or with optional target zoom/orientation/latLng arguments.
2. Consider whether it's worth integrating the map animation in to flutter map with dedicated events.

There are pros and cons of both. I think it's very nice having animation as a separate plugin in order to keep flutter_map as simple as possible. At the same time it requires more work from users to set up and it is complicated for other package maintainers to integrate with. For example in flutter_map_supercluster I have a separate implementation of animated movement (which is more limited than @TesteurManiak's implementation, e.g. it doesn't animate rotation). If I want to integrate with @TesteurManiak's implementation I need to make it a dependency since it would get called directly by my package's code and this means updating my package as well when it changes.

I don't think this is something that should be solved in this PR and it's maybe worth a fresh discussion so I will remove the reference to the closed map animation issue in the code. Maybe we can open a new issue to find out what is needed (both to make the TileUpdateTransformer less hacky and to make it easy for other packages to integrate with map movement animation) and I can update the comment to reference that issue instead.

JaffaKetchup:
> I've added a `tileOpacity` option, let me know what you think.

Implementation looks good, but we should probably keep it called `opacity` as it was before. We also need to update the examples.

> I've also just pushed up a predefined `TileUpdateTransformer` which disables tile updates/prunes for tap/press events since they shouldn't cause the map to move and therefore shouldn't require loading/pruning. What do you think about making that the default?

Nice! I think this should be forced rather than default - there is no reason to disable this behaviour, so it should be used in addition to any that the user specifies.
This feature definitely needs some documentation and explanation, but I've got that.

I definitely think that this needs some method of external control through a plugin API with more events: we definitely want to encourage plugins instead of adding these features to FM - we won't be adding animation into core.
I would like to get this out before v4, purely because of the implementation being a little 'hacky' as you called it, but I realise that this PR may not be the place to do it. It's relatively low priority.

I'll get a pre-release of this out today.

JaffaKetchup:
@rorystephenson In 8e50027 I've renamed `tileOpacity` back to `opacity` to reduce the number of breaking changes. I've also removed leftover deprecations and updated the changelog and versioning.

---

This PR is now also for v4 release preparation, and I've released prerelease [v4.0.0-dev.1](https://pub.dev/packages/flutter_map/versions/4.0.0-dev.1) based on this.
CHANGELOG and pubspec.yaml are ready, but no documentation or migration instructions are prepared yet.

rorystephenson:
Nice one @JaffaKetchup. I've just pushed up a commit which avoids reloading the images when opacity changes, it just modifies the opacity. In doing so I added a class to encapsulate the fade/fastReplace in to a single class since they have an impact on each other and it makes it clear that if fast replace is enabled fading is disabled.

Making these changes made me realise that there are essentially three different modes:

1. fastReplace: Replaces and prunes immediately.
2. No fade: Does not fade, waits 50ms then prunes.
3. Fade: Fades in tiles, waits 50ms + fade duration then prunes.

After my last commit number 2, no fade, is not possible since there is a default fade that gets applied. Before I re-enable it I wanted to check if you know why we would want to have a no fade option which is **not** fastReplace? As far as I understand if tiles are opaque then flickering should not be visible with fastReplace?

example/lib/pages/animated_map_controller.dart
```diff
@@ -198,3 +212,36 @@ class AnimatedMapControllerPageState extends State<AnimatedMapControllerPage>
...
   }
 }
+
+/// Causes tiles to be prefetched at the target location and disables pruning
+/// whilst animating movement. When proper animated movement is added (see
+/// #1263) we should just detect the appropriate AnimatedMove events and
+/// use their target zoom/center.
+final _animatedMoveTileUpdateTransformer =
+    TileUpdateTransformer.fromHandlers(handleData: (event, sink) {
+  final id = event is MapEventMove ? event.id : null;
```
TesteurManiak, line 222: Shouldn't you also check for `MapEventRotate` ?

JaffaKetchup:
@rorystephenson I'm a little confused why we can't use `opacity < 1` if `fastReplace` `false`: doesn't the current implementation support that?
But no, there's not much point having 2 - why wait for 50ms before replacing and pruning the tile.

rorystephenson:
> @rorystephenson I'm a little confused why we can't use `opacity < 1` if `fastReplace` `false`: doesn't the current implementation support that?

The old implementation was just wrapping the TileLayer with an Opacity widget, so the user can have the old functionality by wrapping TileLayer in an Opacity widget and enabling fastReplace.

With the new `opacity` behaviour, disabling fastReplace causes overlapping tile images (not yet pruned ones) to be simultaneously visible which is visually very messy. In [this comment](https://github.com/fleaflet/flutter_map/pull/1475#issuecomment-1494071999) you said:

>We can force `fastReplace` `true` if `opacity < 1`, we just need to document that behaviour.

This is what I ended up going with as well as documenting that if a user wants an interactive map with transparency they should just wrap the whole layer with an opacity widget.

>But no, there's not much point having 2 - why wait for 50ms before replacing and pruning the tile.

OK this is what I thought as well. I don't see why the old implementation waited 50ms, unless some flickering can happen, but that doesn't seem to happen as far as I can see.

---

I've just pushed a commit which consolidates opacity/tileFadeIn/fastReplace in to a single option `tileDisplay`. I think it does a good job of preventing inconsistent use of options (e.g. opacity 0.5 and fastReplace off) as well as simplifying the code. I've tried to document it well to describe the tradeoffs of the opacity option. Let me know what you think, we can always change it back.

JaffaKetchup:
OK, that seems like a good idea to me, although we should make these classes private if they are now accessed through `TileDisplay`.

https://github.com/fleaflet/flutter_map/blob/f8624b78723b0c5cd2b5e4b52e67e434b080dcf1/lib/src/layer/tile_layer/tile_display.dart#L74

https://github.com/fleaflet/flutter_map/blob/f8624b78723b0c5cd2b5e4b52e67e434b080dcf1/lib/src/layer/tile_layer/tile_display.dart#L91

And these lines of documentation seem to be wrong (should be to "1.0" not "0.1"):

https://github.com/fleaflet/flutter_map/blob/f8624b78723b0c5cd2b5e4b52e67e434b080dcf1/lib/src/layer/tile_layer/tile_display.dart#L34-L41

---

I've now realised the difference between the old `opacity` and new `opacity`, and I realise why you called it `tileOpacity`.
If you're happy to change it back, please do, but I understand if you're fed up with opacity discussion now :D (apologies). If it is changed back, we should consider adding `layerOpacity` and wrap the layer in a `Opacity` widget as before: I know this can be done by the end user, but I wish people RTFM! I can see people asking why it doesn't work as they think it should, so let's get this right now. Again, apologies for changing my mind so much!

---

Just noticed `overrideTilesWhenUrlChanges` hanging about in the new documentation and I've noticed you changed its implementation slightly. I'm not sure of a situation where users wouldn't want it `true`: as far as I'm concerned, it should at least default to `true`, if it isn't removed altogether. It has been a big cause of annoyance in the past!

JaffaKetchup:
@rorystephenson fbf782 fixes the documentation and makes the `TileDisplay` extenders' constructors private.

rorystephenson:
Those changes look good @JaffaKetchup. I've just pushed up two more:

- https://github.com/fleaflet/flutter_map/pull/1475/commits/e88d6004b8be88bad98c1daf6dd5da6f258b4b54 prevents taps from triggering load/prune by default. It also adds a predefined transformer to allow all events to trigger load/prune.
- https://github.com/fleaflet/flutter_map/pull/1475/commits/64d20070b8c2001e7c26860e4fc9af4503b75048 adds a throttling transformer.

JaffaKetchup:
Sounds good. What about this?

> Just noticed overrideTilesWhenUrlChanges hanging about in the new documentation and I've noticed you changed its implementation slightly. I'm not sure of a situation where users wouldn't want it true: as far as I'm concerned, it should at least default to true, if it isn't removed altogether. It has been a big cause of annoyance in the past!

JaffaKetchup:
As well as ^, also just had this come through on Discord.

rorystephenson:
> Sounds good. What about this?
> 
> > Just noticed `overrideTilesWhenUrlChanges` hanging about in the new documentation and I've noticed you changed its implementation slightly. I'm not sure of a situation where users wouldn't want it true: as far as I'm concerned, it should at least default to true, if it isn't removed altogether. It has been a big cause of annoyance in the past!

I'm all for dropping it, I don't see why you wouldn't want the existing tiles to hang around until the new ones are available.

Here's how it looks with that option removed:


https://user-images.githubusercontent.com/3683599/229994927-8a3cf82e-a6f0-400e-8137-f6eefde5fa25.mp4

rorystephenson:
@JaffaKetchup I just had a look at the fastReplace issue and when I enabled fastReplace on 3.1.0 the tiles remained but when I moved the map they did not load anymore. Since fastReplace has become part of the new TileDisplay.instantaneous() I gave that a double check and it is working fine.

JaffaKetchup:
> I'm all for dropping it, I don't see why you wouldn't want the existing tiles to hang around until the new ones are available.

Ah interesting: I was thinking the opposite. Showing two different tile styles might be confusing. I guess we're keeping it then!

> I just had a look at the fastReplace issue and when I enabled fastReplace on 3.1.0 the tiles remained but when I moved the map they did not load anymore. Since fastReplace has become part of the new TileDisplay.instantaneous() I gave that a double check and it is working fine.

I'll let him know.

> On an unrelated note. I'm not sure if now is the time but another tidy up that would be worth doing at some point is to stop passing the entire TileLayer to TileProvider's getImage as it makes it harder to reason about what options can be changed/removed since a provider can rely on any option. Ideally we'd pass a subset of the options, those which are necessary for fetching tiles. That said I haven't played with custom tile providers much so I don't have a great idea of what options they need.

That's not urgent: you're probably right that not everything needs to be passed, but some custom providers might be way beyond what we can consider. Let's leave that for now.

rorystephenson:
@JaffaKetchup I've noticed that when resizing the map new tiles are not loaded. This is not surprising because no events are triggered when resizing and loading is only triggered by map events. What do you think about adding a MapSizeChange event? This would also remove the need for [this workaround](https://github.com/fleaflet/flutter_map/pull/1475/files#diff-e590e19051c91489320fd6ad887adeb886111410002194672680e09099e757c3R667).

JaffaKetchup:
@rorystephenson I'm happy with that idea, seems sensible.
I'll do some more in depth testing and review this weekend.

(Started a rewrite of the gesture system yesterday, but I might abandon that: fixing that mess will take a lot longer than to v4.)

JaffaKetchup:
Just had an issue report on the Discord channel. Seems to be an issue with #1462 or #1442 (cc @ignatz @mootw), because it's reproducible both with and without this PR.

This issue is a blocker to the v4 release if it can be verified and reproduced. Unfortunately, they were not able to provide an MRE.

---

> With the V4 Release I'm getting realy "glitchy" Polygons and Lines that alter their shape when I Zoom in or move the screen. Is this a known issue? The polygons are Valid and have no self intersection.

> when i zoom in/out parts of the polyline disappear. Furthermore the line is getting thicker and thinnner and is not a line anymore like in previous Versions.

<details>
<summary>Working Polygons</summary>

https://user-images.githubusercontent.com/58115698/230732371-b483b1b7-1f97-4a2d-8fa7-1269c78e586e.mov

</details>

<details>
<summary>Working Polylines</summary>

https://user-images.githubusercontent.com/58115698/230732375-fd070647-fd1e-47d8-907f-cf1884163660.mov

</details>

<details>
<summary>Broken Polygons</summary>

https://user-images.githubusercontent.com/58115698/230732378-97666336-64c1-46bc-8a6a-e24f6e5ca8ef.mov

</details>

<details>
<summary>Broken Polylines</summary>

https://user-images.githubusercontent.com/58115698/230732376-ecbaad56-8c62-49af-8558-578d0cbf768a.mov

</details>

ignatz:
I'd try to disable the repaint boundary, eg. Use the threshold, sounds like the symptom you'd expect if not every frame gets a paint.

JaffaKetchup:
@ignatz Look at the polyline example, it's quite deformed. Could this be a cause?
Can you also elaborate please? I'd like to send it back to the reporter to see if your advice fixes their issue.

dkuemper:
Hi thid is KuDos81 from the discord. Raising the Threshholds to 16192 and then 16192000 did not solve the issue. 
As an MRE, you can take the example App and replace the following in the polygon.dart of the Example-App.
Although (due to only a few polygon points) it doesn't show the full dancing behaviour you can always se some gaps of the polygon inside and some strange danicing patterns in the dotted layer.

```
    final filledPoints = <LatLng>[
      LatLng(51.50626085219557, -0.075738326893169),
      LatLng(51.505686351002744, -0.07347482931832179),
      LatLng(51.50483157651988, -0.07446190453551053),
      LatLng(51.50537917661692, -0.07652181574310216)
    ];
    final filledDotedPoints = <LatLng>[
      LatLng(51.50483157651988, -0.07446190453551053),
      LatLng(51.50537917661692, -0.07652181574310216),
      LatLng(51.50469133895954, -0.07656471617782601),
      LatLng(51.50433072865138, -0.07521291370816201)
    ];
```

image

ignatz:
Thanks for the example. I was able to reproduce your issue. As it turns out the issue wasn't due to missing repaints but due to how the points were projected and limited precision. https://github.com/fleaflet/flutter_map/pull/1485 reverts all the changes related to the repaint boundary optimization and fixes the visual artifacting you're observing. Sorry for the inconvenience.

JaffaKetchup:
@ignatz Don't worry about it, no inconvenience at all: this stuff happens! Thanks for fixing it, I'll check it out today.

JaffaKetchup:
@rorystephenson Any luck with this yet? I know it's been Easter so no worries if not!

> I've noticed that when resizing the map new tiles are not loaded. This is not surprising because no events are triggered when resizing and loading is only triggered by map events. What do you think about adding a MapSizeChange event? This would also remove the need for [this workaround](https://github.com/rorystephenson/flutter_map/blob/6fbd8f70146a8b55f828f986355b0ae1891e28b0/lib/src/layer/tile_layer/tile_layer.dart#L662-L683).

JaffaKetchup (dismissed):
LGTM (once https://github.com/fleaflet/flutter_map/pull/1475#issuecomment-1503746061 is addressed) AFAIK!
Docs and migration info are not ready yet, I'll write those before release, but that shouldn't block this.
#1482 may get conflicts after this is merged (cc @tlserver), but it is already included in the updated CHANGELOG.

Would like at least one other maintainer's detailed approval (cc @mootw @TesteurManiak) before merging.

JosefWN:
Awesome! Sorry I can't be of much help, been bogged down in other projects lately, but something that came to mind without reading through the entire PR: Will this refactor also pave way for TMS support, for example? When I last dove into this pre-refactor the architecture turned out to be quite hairy with regards to extensibility: https://github.com/fleaflet/flutter_map/issues/1394

JaffaKetchup:
@JosefWN I wouldn't think it does atm, as that is managed by the tile providers instead of the tile layer, however we could probably include a fix for TMS in v4, if it's simple.

Looking back at https://github.com/fleaflet/flutter_map/issues/1394#issuecomment-1324075231, https://github.com/fleaflet/flutter_map/issues/1394#issuecomment-1324084429, and https://github.com/fleaflet/flutter_map/issues/1394#issuecomment-1340966959, it appears you suggested our `invertY` or `_getTileUrl` method is wrong, and correcting that should fix the issue? Am I understanding that right? Can you just clarify a bit, then I can form a fix.

In addition, the retina tiles is a little unclear, but I tried to clean it up with the web docs: https://docs.fleaflet.dev/usage/layers/tile-layer/recommended-options#retina-mode-retinamode. However, it's still not very clear: the "{r}" placeholder is redundant (prefer "@2x") in normal URLs, but it's still needed in WMS, or something like that AFAIK?

JosefWN:
IIRC the challenge I faced was the lack of access to stuff, if you look at Leaflet: https://github.com/Leaflet/Leaflet/blob/main/src/layer/tile/TileLayer.js#L178-L195

Where `getTileUrl` depends on the `_map` and `_globalTileRange` etc. They support TMS with both `tms: true` or supplying `-y` for TMS and potentially other tiling systems that support it.

The retina mode approach I just got from there:
```dart
r: Browser.retina ? '@2x' : '',
```

mootw:
I have not done any code review yet but i have run into a few practical issues in testing.

Migrating was fairly easy to do, only a couple of changes.

Example app looks good to me on web!

Testing a more complex app (on android) i am running into a few issues:

Biggest issue is this crash when zooming (maybe panning?) while replacing tile layers quickly (for animation):
```
ConcurrentModificationError (Concurrent modification during iteration: _Map len:20.)

_CompactIterator.moveNext (dart:collection-patch/compact_hash.dart:710)
TileImageManager.reloadImages ([REMOVED]/flutter_map/lib/src/layer/tile_layer/tile_image_manager.dart:122)
_TileLayerState.didUpdateWidget ([REMOVED]/flutter_map/lib/src/layer/tile_layer/tile_layer.dart:434)
StatefulElement.update ([REMOVED]/flutter/packages/flutter/lib/src/widgets/framework.dart:5260)
Element.updateChild ([REMOVED]/flutter/packages/flutter/lib/src/widgets/framework.dart:3686)
```


I am subjectively feeling a lot of jank when zooming and scrubbing the timeline, loading many tiles at once (mid hundreds of tiles); though i will need to profile to actually see if this is actually a problem compared to the old version, and if it is what is causing it. This behavior also basically guarantees the crash above.

This is just a preliminary review/testing. I will attempt to fix the crash or provide a mvp for reproducing the issue once I do a proper code review which will take a bit longer as there are many changes. Overall it seems good to me so far.

I had 1 issue with migration which was fixed by reading the comments (i will comment on that in my code review) which was fixed. Otherwise the only issue i have seen is the crash and potentially some jank.

JaffaKetchup:
@mootw (cc @rorystephenson) I can't understand why we're getting that `ConcurrentModificationError`. The lines mentioned don't appear to modify the map it's iterating through: unless it counts modifying the value (the `TileImage` object is mutable). Not sure if this could be the issue?

rorystephenson:
Just dropping in to say I was away the last couple of weeks and hope to be able to update this soon.

JaffaKetchup:
No problem :)
The v4 docs are ready on the docs site (the version chooser on the top left lets you choose it).

JaffaKetchup:
Published a new prerelease: https://pub.dev/packages/flutter_map/versions/4.0.0-dev.2.

ibrierley:
Apologies for not looking at this sooner, have been away several weeks, and not getting as much time as I'd like on this. 

One thing I think I notice, is that the tiles don't reload as optimally when zooming out at least, see if I can include comparison videos. It "feels" a bit like there is a longer delay when loading in previously seen tiles, so there is grey space for a longer time before they all kind of load. Interestingly the animated mapController example looks better though...

Original:

https://user-images.githubusercontent.com/3901173/233782826-f650863e-2173-4203-8cbb-c8d9cce4159d.mp4

V4

https://user-images.githubusercontent.com/3901173/233782840-311d426b-8a60-42e7-b32e-226c67478f79.mp4

It's quite subtle, and it may need someone else to try and check I'm not going mad...

JaffaKetchup:
@ibrierley From your video, I can definitely see what you mean, but I can't seem to reproduce on my device - haven't directly compared to old version, but from memory tile loading time seems about the same and acceptable.

In terms of the animated movement being much better, that's probably because of the introduction of the `TileUpdateTransformer` that allows the user to choose when they want to load/prune new tiles.
The implementation to do this is a little finicky, as below, but I think @TesteurManiak's animation plugin supports it (https://github.com/TesteurManiak/flutter_map_animations/pull/3)?
https://github.com/fleaflet/flutter_map/blob/8324c0c835b73b69f2a985e1569ed4d3ec3b849b/example/lib/pages/animated_map_controller.dart#L220-L250

ibrierley:
I've just uploaded this PR web version onto my test site, so it can be seen a bit easier the problem I was trying to describe.
If you try on a PC for example an old version eg

https://fluttermap.dabbles.info/v1.1.1/#/  and mousewheel scroll zoom in/out a lot

And then try 

https://fluttermap.dabbles.info/v4/#/

You will (possibly) see that the new version has more grey screen, and kind of flashes more. As though it has to wait for all tiles to download before displaying, rather than allowing partial sets of tiles to display and making it look smoother ?

JaffaKetchup:
I think that possibly the issue is that the tile loading waits for the gesture to complete before loading new tiles? Although its hard to be sure.

JosefWN:
> You will (possibly) see that the new version has more grey screen, and kind of flashes more.

IIRC the tiles in the old/current implementation are cross faded. Say I zoom in, the old tile is scaled until replaced by a higher resolution tile. When this happens, the old tile fades out and the new tile fades in simultaneously, which makes for a very smooth transition.

JaffaKetchup:
Interesting. For me, the issue was more noticeable when using a mouse scroll, so maybe it's just differences in mouse configuration? 
I've got a lot on my plate as well, but maybe I can get some time to do some profiling for jankiness.

rorystephenson:
I've just had a chance to have a quick look at why tiles seem to load slower when zooming out and at least part of the issue is that when zooming out fast tiles may be pruned before they finish loading.

Imagine a pinch-zoom from 10 to 8. When 9 is reached the tiles at that zoom begin loading and then when 8 is reached very shortly after (maybe 200ms) the loading tiles at zoom 9 are pruned because they are not `active` since they haven't finished loading yet. Therefore we don't see any new tiles until the tiles at zoom 10 finish loading. If the tiles load fast enough  then they will become visible before they are pruned however when zooming fast that often isn't the case.

What is confusing is why this does not affect the old version so badly. There are a couple of possibilities here:
1. The pruning logic has changed causing us to prune when previously we wouldn't.
2. The tile takes longer to become `active` in the new version and thus is pruned before it is loaded.

I didn't intentionally change the pruning logic and I can't see any differences after a quick review. This leaves option 2, tiles taking longer to become active. Similarly I didn't deliberately change the logic here and after another quick review the code has been heavily refactored but the logic appears to be the same. In both the old and the new version tile activation works as follows:

1. The tile's `load()` method is called.
2. We call `getImage()` on the TileProvider.
3. Once the image is loaded it sets the tile to active if there is no animation, otherwise it starts the animation and sets the tile to `active` once (if) the animation finishes.

The animation time is the same across the old and the new version which really just leaves the TileProvider itself. Again the implementations don't seem to have changed but strangely when I added some logging to see how long the tile provider was taking when pinch-zooming out there was a fairly consistent difference: around 300ms for the new version and 100ms for the old version. I'm a little lost as to why there is such a difference and I wonder if someone else can reproduce this. Testing info:

- Running on a real android device in profile mode.
- In the new version you can log the difference between loadStarted and loadFinishedAt. In the old version you need to add in a loadFinishedAt since it didn't exist yet.
- Old version is a908a898f5fd84664f3e7cd5aaca0e009504e86d (3.1)
- New version is 654db1474648a16293fe5c0f37b272ed41c8ad51 (latest commit on this branch at time of testing).
- First I zoom in quite far, pan to the side and then finally I make a large fast pinch-zoom to zoom out.

This could be caused by a variety of things (e.g. image eviction) but before digging further it would be nice to have someone else confirm the difference.

---

Aside from determining if/why the TileProvider is slower this also raises a question as to how we could better handle pruning for tiles which are currently being loaded.

ibrierley:
Heya, I think it is, but its a long time since I uploaded. Is there a specific version you want me to upload which will help isolate ?

JaffaKetchup:
(Also still need to resolve https://github.com/fleaflet/flutter_map/pull/1475#issuecomment-1500060259)
I'll do some profiling tonight just to see if there's any obvious & quick improvements that can be made.

mootw:
@rorystephenson are you on a high performance network? When I tested on web I barely noticed a difference if any at all. However, I was able to notice (not profile as of yet), a noticeable amount of jank when zooming on my phone using a test app which loads a significant number of layers and tiles (mid hundreds at once while zooming). compared to wherever I was testing from last. Unfortunately I did not write down the HEAD of that. This was also how I am able to consistently get the crashing issue. I discussed earlier. I apologize that I have not had the time to look at either of these yet.

JaffaKetchup:
Just done a little bit of testing on the example app (Windows). We're doing excellently on UI/Build Frame time, but Raster Time is often quite high:
image

I guess it's high because we've got loads of images, but I can't understand this:
image

The entire screen, except the map widget, seems to repaint every time the map repaints - but takes longer despite having barely anything within it! `RepaintBoundary`s make no difference.

rorystephenson:
> (Also still need to resolve [#1475 (comment)](https://github.com/fleaflet/flutter_map/pull/1475#issuecomment-1500060259)) I'll do some profiling tonight just to see if there's any obvious & quick improvements that can be made.

True I forgot about that 👍 .

> @rorystephenson are you on a high performance network? When I tested on web I barely noticed a difference if any at all. However, I was able to notice (not profile as of yet), a noticeable amount of jank when zooming on my phone using a test app which loads a significant number of layers and tiles (mid hundreds at once while zooming). compared to wherever I was testing from last. Unfortunately I did not write down the HEAD of that. This was also how I am able to consistently get the crashing issue. I discussed earlier. I apologize that I have not had the time to look at either of these yet.

I'm on decent but not extremely fast wifi. I don't see much of a difference on my computer when testing the web version but there is quite a noticeable difference when I test the web version with my phone (my phone is a few years old, maybe with a newer phone it is less noticeable).

I forgot about [the concurrency issue](https://github.com/fleaflet/flutter_map/pull/1475#issuecomment-1512454036) as well, that's definitely worth resolving if we can reproduce it. Does your test app change the tile image url? I ask because `reloadImages` is only called when the tile url changes.

rorystephenson:
Here's an update on the concurrent modification error that @mootw spotted:

I just had a quick dive in to see if I could understand how the concurrent modification can occur and I think I've found it. The `reloadImages` method [calls `load()`](https://github.com/rorystephenson/flutter_map/blob/a6d5f96e1a403fb7a045ca5ab8c768bb2af969b6/lib/src/layer/tile_layer/tile_image_manager.dart#L127) on the TileImage which in turn [listens for the image loading](https://github.com/rorystephenson/flutter_map/blob/a6d5f96e1a403fb7a045ca5ab8c768bb2af969b6/lib/src/layer/tile_layer/tile_image.dart#L142C1-L144). This looks like the callback will be called asynchronously but [as the documentation mentions](https://api.flutter.dev/flutter/painting/ImageStream/addListener.html) it can actually be called synchronously and when this happens [onLoadComplete is called](https://github.com/rorystephenson/flutter_map/blob/a6d5f96e1a403fb7a045ca5ab8c768bb2af969b6/lib/src/layer/tile_layer/tile_image.dart#L158) which can [trigger a prune](https://github.com/rorystephenson/flutter_map/blob/a6d5f96e1a403fb7a045ca5ab8c768bb2af969b6/lib/src/layer/tile_layer/tile_layer.dart#L647) that causes tiles to be removed whilst we are iterating them.

A similar issue could also theoretically occur [if a synchronous error is thrown when](https://github.com/rorystephenson/flutter_map/blob/a6d5f96e1a403fb7a045ca5ab8c768bb2af969b6/lib/src/layer/tile_layer/tile_image.dart#L148) when loading the tile image since it also calls onLoadComplete.

I can see a few different ways to solve this:

1. Change how `reloadImages` works. First make a copy of the list of tiles to reload (`List.from(_tiles.values)`), and then iterate it calling load().
2. Add a `newTile` parameter to the `onLoadComplete()` ([_activate already has logic to determine if a TileImage is new](https://github.com/rorystephenson/flutter_map/blob/a6d5f96e1a403fb7a045ca5ab8c768bb2af969b6/lib/src/layer/tile_layer/tile_image.dart#L173-L174). Then if `onLoadComplete()` is called with `newTile: false` we can avoid pruning since we wouldn't expect pruning to be necessary when reloading an existing tile, only when adding a new TileImage as that implies that the visual bounds have changed. On the other hand this may cause problems if all tiles are reloaded whilst the map is moving since [we don't trigger the prune in the onLoadComplete callback when there are remaining tiles to load](https://github.com/rorystephenson/flutter_map/blob/a6d5f96e1a403fb7a045ca5ab8c768bb2af969b6/lib/src/layer/tile_layer/tile_layer.dart#L642) and the last tile may be a reloaded tile.
3. Delay the calls to onLoadComplete  (`Future.delayed(Duration.zero, () => onLoadComplete(coordinates))`). This will work but it also feels very hacky and we'd have to check that it doesn't introduce any slow-down.

Any better ideas?

I am fairly certain that this issue can only happen when calling `reloadTiles` since it calls `load()` whilst iterating the TileImage collection. The only other place `load()` is called is [when adding new tiles](https://github.com/rorystephenson/flutter_map/blob/a6d5f96e1a403fb7a045ca5ab8c768bb2af969b6/lib/src/layer/tile_layer/tile_layer.dart#L615) but the collection being iterated is a List which is a filtered/copied list of TileImages.

JaffaKetchup:
Nice work! (It's 11pm here for me now, so I'll be back tomorrow and can check it out then.)

rorystephenson:
I've gone ahead and added a commit with my preferred fix for the concurrent modification issue (it now iterates a copy of the tile collection instead of iterating the tile collection directly).

I've also revisited the v1.1.1 example on mobile web and the performance differences are not consistent for me. If someone is experiencing noticeable and consistent performance differences between v1.1.1 and this version I'm interested in having a look. If so please specify the following:

1. Device used.
2. Target platform (if web please mention whether you overriding the web renderer with canvaskit or html).

---

So I think we're mostly just waiting on the addition of a MapEvent for map size change.

JaffaKetchup:
Thanks @rorystephenson, sorry it dropped off my radar. Your implementation looks good.

Are you going to be able to add that last point? Happy to try to add it if you can't - although this area (sizing) appears to be particularly vulnerable to bugs.

Still trying to figure out the long double raster times - if you have any insight into that, that would be helpful. But I don't think that's much different to any previous version, it's just something that's never been paid much attention.

rorystephenson:
@JaffaKetchup I've added a couple of commits with the implementation of the map size change event. I made some other changes while I was there as I noticed some things which could be simplified/improved. See the commit messages for more info.

JaffaKetchup:
Thanks @rorystephenson, I'll try to review in the coming day(s). I'll also update the docs with those last few breaking changes.

I've added a special credit for you to the [v4 docs](https://docs.fleaflet.dev/v/v4-beta/credits#:~:text=Notable%20contributors%3A-,%40rorystepenson,-%40ignatz).

mootw (approved):
Tested on complex app on android and web. The crash has been fixed, and performance seems acceptable (though I do note some potential optimizations and issues we have discussed above). I did not profile it against previous versions yet though. Interestingly loading tiles on the web seems like the image decode is faster (for cached tiles) though i did not measure it. I think the opacity comment for `TileDisplay.instantaneous` should be more prominent in some way (maybe as part of the migration docs?) to encourage using an opacity widget over the tile layer to prevent overlapping tiles with weird behavior when the opacity is not 1.

In general though, I like most of these changes and they should be incredibly helpful for the future of flutter_map. Thank you for your contributions! 🎉

(Though before we merge we should get another review or two because I probably missed something and there are a lot of changes)

JaffaKetchup (approved):
LGTM! Would like an additional review, but not strictly necessary, cc. @ibrierley @TesteurManiak.

TesteurManiak (approved):
The code looks good to me 👍 (Not a big fan of the use of the `!` operator, I prefer to use a transitive variable to benefit from the type shadowing but this can be part of a future refactor)

ibrierley (approved):
I'm fine with all of this! I can't see the performance issues now that I was seeing before, so happy for this to go ahead. Thanks for all of the work.

JaffaKetchup:
🎉 Thanks for the contribution.
Please join the Discord server to keep informed about updates.

lib/src/layer/tile_layer/tile_image_manager.dart
```diff
@@ -0,0 +1,263 @@
...
+
+    final toRemove = <String>[];
+    for (final entry in _tiles.entries) {
+      if (!entry.value.retain) toRemove.add(entry.key);
+    }
+
+    for (final key in toRemove) {
+      _removeWithDefaultEviction(key, evictStrategy);
+    }
+  }
```
josxha, line 211: Hi @rorystephenson, I currently working on some code improvements and wanted to ask if there is a reason to have the logic splitted into three loops here?

rorystephenson: Hi @josxha, it's great that you're digging in looking for a change to improve the code.

It looks like this code originates from [this commit](https://github.com/fleaflet/flutter_map/commit/d4aefaecc44d0c0b431375fd7bc97b2ef1a214e7).

I don't think we can reduce the for loops in the current form (more on that at the end). Tiles have an existing `retain` value (from a previous prune run) which might be set to true or false. Imagine if a tile, tile X, previously had a `retain` value of true but it is no longer current. In the current implementation we first set tile X's `retain` to the value of `current`, false, in the first for loop. Then we iterate through the tiles and, if we need to retain tile X because it is the ancestor/child of a current tile that is not loaded yet we will set its retain to `true`. 

If we combine those two for loops we might first encounter a non-loaded child of tile X when iterating and set its `retain` value to true before iterating over tile X and setting its `retain` back to false because it is not current.

We cannot combine the 2nd/3rd for loops because we need to iterate all tiles before we know that we do not need to retain a child (as mentioned before, it may be the child/parent of a non-loaded current tile).

Finally we can't combine the 3rd and 4th for loops because this would cause a concurrent modification exception. Calling _removeWithDefaultEviction removes an element from _tiles so we must first build a new list of removals from _tiles before iterating over the new list and performing the removals.

---

That said the fact that we have four loops to perform removals is a code smell and you're right to point it out. I have been experimenting with refactoring tile pruning to make the logic easier to follow and you have made me realise that I can probably remove an iteration in my [new implementation](https://github.com/rorystephenson/flutter_map/blob/306891d6194ed9a9a3625565724f38fa700c9eb1/lib/src/layer/tile_layer/tile_removal_state.dart#L52-L68). I need to finish my work on that before opening a PR though.

ibrierley: Just r.e the removals. I never really liked the whole prune methodology, it felt kind of wrong, and I did have a dabble at a change, but it wasn't quite as smooth.

The basic premise was to not think about pruning tiles, but starting from scratch, "what tiles do we need to display immediately", if a tile is outstanding, see if there's a parent (or child) tile zoom that covers it (similar in concept to the retainParent/Child iirc), and use that.

It's a subtle difference mentally, if you wanted a peak [
](https://github.com/ibrierley/flutter_map_vector_tiles/blob/master/lib/VectorTileWidget.dart) (basically line 244-384). As mentioned, it didn't work as well so I wouldn't just copy it, but it may just give a slightly different angle on it..I wonder if my problem was that it wouldn't necessarily try and use a tile from the last frame, it would pick any as a backup). Don't want to distract the work you're doing tho, so don't get bogged down on that this. Maybe it should be a separate change.

josxha: Thank you very much for this great explaination @rorystephenson!

PR #1485, opened 2023-04-11
Revert RepaintBoundary optimization due to visual artifacting.

ignatz:
In order to use the RepaintBoundary the each vector is broken up into two: a support vector (Positioned) and a local vector (map.project), so that the local projection can be repositioned w/o always requiring a full redraw. However, as it turns out, the implementation has not enough precision to correctly position vectors very close to each other (it's likely due to map projection and/or integer precision). This is particularly noticeable when e.g. drawing a dotted line (which is just a bunch of close by polygons) or a polygon with a boundary (which is just a polygon and a polyline precisely layered atop each other).


TesteurManiak (approved):

JaffaKetchup:
Thanks @ignatz for the quick fix :)
(Thanks @TesteurManiak)

PR #1487, opened 2023-04-12
Improved Attribution

JaffaKetchup:
Continuation and replacement of #1390, which aims to improve the attribution situation within 'flutter_map'.
Contains breaking changes, and addition of Windows support for the example app.


lib/src/layer/attribution_layer.dart
```diff
@@ -66,3 +85,382 @@ class AttributionWidget extends StatelessWidget {
...
+  State<StatefulWidget> createState() => RichAttributionWidgetState();
+}
+
+class RichAttributionWidgetState extends State<RichAttributionWidget> {
+  StreamSubscription<MapEvent>? mapEventSubscription;
+
+  final persistentAttributionKey = GlobalKey();
+  Size? persistentAttributionSize;
+
+  late bool popupExpanded = widget.popupInitialDisplayDuration != Duration.zero;
```
TesteurManiak, line 290: Maybe a getter would be better to avoid code duplication on `widget.popupInitialDisplayDuration != Duration.zero`:
```dart
bool get hasDuration => widget.popupInitialDisplayDuration != Duration.zero;
```

lib/src/layer/attribution_layer.dart
```diff
@@ -66,3 +85,382 @@ class AttributionWidget extends StatelessWidget {
...
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    final persistentAttributionItems = [
+      ...widget.attributions
+          .whereType<LogoSourceAttribution>()
+          .map((e) => [e, const SizedBox(width: 16)])
+          .expand((e) => e)
+          .toList(),
```
TesteurManiak, line 329: `toList()` is unneeded the spread operator works on basic `Iterable` objects

lib/src/layer/attribution_layer.dart
```diff
@@ -66,3 +85,382 @@ class AttributionWidget extends StatelessWidget {
...
+    mapEventSubscription?.cancel();
+    super.dispose();
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    final persistentAttributionItems = [
+      ...widget.attributions
+          .whereType<LogoSourceAttribution>()
+          .map((e) => [e, const SizedBox(width: 16)])
```
TesteurManiak, line 327: May I suggest to use a method like this:
```dart
Iterable<Widget> buildSeparatedWidgets(Iterable<Widget> children) sync* {
  for (int i = 0; i < children.length; i++) {
    yield children[i];
    if (i < children.length - 1) {
      yield const SizedBox(width: 16);
    }
  }
}
```

This will allow to avoid having an unnecessary `SizedBox` widget at the end.

lib/src/layer/attribution_layer.dart
```diff
@@ -66,3 +85,382 @@ class AttributionWidget extends StatelessWidget {
...
+/// Base class for attributions that render themselves as widgets
+///
+/// Extended/implemented by [TextSourceAttribution] & [LogoSourceAttribution].
+@internal
+abstract class SourceAttribution extends StatelessWidget {
+  final void Function()? _onTap;
+
+  const SourceAttribution._({
+    super.key,
+    void Function()? onTap,
```
TesteurManiak, line 98: You can use `VoidCallback?` instead of `void Function()?`

TesteurManiak (changes requested):
Here's my first feedbacks, I'll try to do a second pass on this when I have a bit more time.

pubspec.yaml
```diff
@@ -21,6 +21,7 @@ dependencies:
   polylabel: ^1.0.1
   proj4dart: ^2.1.0
   tuple: ^2.0.0
+  url_launcher: ^6.1.6
```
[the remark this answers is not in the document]
JaffaKetchup: That's a good point, forgot we talked about it and considered removing it myself. I'll remove it.

lib/src/layer/attribution_layer.dart
```diff
@@ -66,3 +85,382 @@ class AttributionWidget extends StatelessWidget {
...
+
+    if (widget.popupInitialDisplayDuration != Duration.zero) {
+      Future.delayed(
+        widget.popupInitialDisplayDuration,
+        () => setState(() => popupExpanded = false),
+      );
+    }
+
+    WidgetsBinding.instance.addPostFrameCallback(
+      (_) => WidgetsBinding.instance.addPostFrameCallback(
```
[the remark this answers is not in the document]
JaffaKetchup: I'm not entirely sure, but delaying one frame doesn't get the right size. I'm assuming it's because FM takes an additional frame in the first place, but I'm not sure.

lib/src/layer/attribution_layer.dart
JaffaKetchup: We should keep this space AFAIK: it is between the last user-defined logo, and the flutter_map logo.

TesteurManiak: Ok but you might still want to use a similar method to avoid using `expand` and iterating again on the widget list 🤔

lib/src/layer/attribution_layer.dart
JaffaKetchup: Fixed in db944a1.

lib/src/layer/attribution_layer.dart
JaffaKetchup: Fixed in db944a1.

lib/src/layer/attribution_layer.dart
JaffaKetchup: Fixed in db944a1.

lib/src/layer/attribution_layer.dart
```diff
@@ -66,3 +85,382 @@ class AttributionWidget extends StatelessWidget {
...
+                    ),
+                    child: Padding(
+                      padding: const EdgeInsets.all(8),
+                      child: Column(
+                        mainAxisSize: MainAxisSize.min,
+                        crossAxisAlignment: CrossAxisAlignment.start,
+                        children: [
+                          ...widget.attributions
+                              .whereType<TextSourceAttribution>()
+                              .toList(),
```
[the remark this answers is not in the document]
JaffaKetchup: Fixed in db944a1.

pubspec.yaml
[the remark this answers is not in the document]
JaffaKetchup: Fixed in db944a1.

lib/src/layer/attribution_layer.dart
JaffaKetchup: Fixed in 1670453.

lib/flutter_map.dart
```diff
@@ -25,7 +25,8 @@ export 'package:flutter_map/src/geo/latlng_bounds.dart';
 export 'package:flutter_map/src/gestures/interactive_flag.dart';
 export 'package:flutter_map/src/gestures/map_events.dart';
 export 'package:flutter_map/src/gestures/multi_finger_gesture.dart';
-export 'package:flutter_map/src/layer/attribution_layer.dart';
+export 'package:flutter_map/src/layer/attribution_layer.dart'
+    hide SourceAttribution;
```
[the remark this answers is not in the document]
JaffaKetchup: Fixed in a3706de.

It has been replaced with a warning stating that:
> unknown subtypes will not be displayed

TesteurManiak (approved):
LGTM

PR #1495, opened 2023-04-21
Refactored `FlutterMapState`'s `maybeOf` method

JaffaKetchup:
Refactored `FlutterMapState`'s `maybeOf` method into itself and `of` to meet good practise guidelines and avoid unnecessary nullable return types.
Also changed the error that `of` throws from `FlutterError` to `StateError`.

### DIFF
--- a/lib/src/map/flutter_map_state.dart
+++ b/lib/src/map/flutter_map_state.dart
@@ -761,9 +761,7 @@ class FlutterMapState extends MapGestureMixin
       ?.mapState;
 
   static FlutterMapState of(BuildContext context) =>
-      context
-          .dependOnInheritedWidgetOfExactType<MapStateInheritedWidget>()
-          ?.mapState ??
+      maybeOf(context) ??
       (throw StateError(
           '`FlutterMapState.of()` should not be called outside a `FlutterMap` and its children'));
 }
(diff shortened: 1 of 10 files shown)


JaffaKetchup:
(Build will fail because of formatting issues fixed in #1475, so it's safe to ignore)

lib/src/map/flutter_map_state.dart
```diff
@@ -756,15 +756,16 @@ class FlutterMapState extends MapGestureMixin
...
+      .dependOnInheritedWidgetOfExactType<MapStateInheritedWidget>()
+      ?.mapState;
+
+  static FlutterMapState of(BuildContext context) =>
+      context
+          .dependOnInheritedWidgetOfExactType<MapStateInheritedWidget>()
+          ?.mapState ??
+      (throw StateError(
+          '`FlutterMapState.of()` should not be called outside a `FlutterMap` and its children'));
 }
```
TesteurManiak, line 769: You can directly call `maybeOf(context)` to avoid code duplication (I also find it clearer to use a `if` to throw an error and beneficiate of type shadowing but it's a personal preference):

```dart
static FlutterMapState of(BuildContext context) {
  final state = maybeOf(context);
  
  if (state == null) {
    throw StateError('...');
  }

  return state;
}
```

Otherwise the rest looks good to me 👍

TesteurManiak (changes requested):

lib/src/map/flutter_map_state.dart
JaffaKetchup: Good catch, I was intending to use `maybeOf` in `of`, but I got sidetracked.
I'll keep using `??`, as it's just shorter and similar in clarity (if you're familiar with Dart).

Fixed in fc56357.

TesteurManiak (approved):

PR #1499, opened 2023-04-24
Add layer that directly takes widgets als children

Robbendebiene:
Hi, recently I wrote a map layer for another project that basically works like the Stack widget by directly taking widgets instead of custom Marker objects. Just wanted to ask if there is any interest in such kind of feature for the library.

Note for now every widget is positioned at its center (no configurable anchor point) and there is no counter rotation when rotating the map. However both of these features should be fairly easy to implement.

Additionally this layer optionally allows some basic sizing of its children in meters (so they scale on zoom in/out) but I could also easily remove this feature.

Usage:

```dart
WidgetLayer(
   children: [
       WidgetLayerPositioned(
          position: LatLng(0,0),
          child: Container(
            color: Colors.red,
            width: 30,
            height: 100,
          ),
       ),
       WidgetLayerPositioned(
          position: LatLng(0,0),
          child: Container(
            color: Colors.blue,
            width: 20,
            height: 20,
          ),
       ),
   ],
);
```

Main benefit in my eyes is that it's more flutter like and widgets with intrinsic/dynamic sizes can be used.
Also going down to elements and render objects allows for more performance optimizations.
This can probably be even more optimized when implemented using the [Sliver](https://api.flutter.dev/flutter/widgets/SliverMultiBoxAdaptorWidget-class.html) way.

Other ideas:
- Widgets placed inside the `WidgetLayer` that are not of type `WidgetLayerPositioned` could be rendered/positioned as an overlay.
- Another constructor or layer that exposes a `builder` parameter that lazily builds it's children given the current bounding box.


JaffaKetchup:
Thanks @Robbendebiene.
However, I'm not clear why this is necessary? You can use a `LayoutBuilder`/`Align` combination in `children` to position anything as you'd like. Any widget in the `children` list can also use `FlutterMapState.of()` to access the map state: this can be used to create counter rotation as `Marker`s do - but you can also use `nonRotatedChildren`.
Maybe I'm missing something? Sorry if I am!

Robbendebiene:
> However, I'm not clear why this is necessary? You can use a LayoutBuilder/Align combination in children to position anything as you'd like. Any widget in the children list can also use FlutterMapState.of() to access the map state: this can be used to create counter rotation as Markers do - but you can also use nonRotatedChildren.
> Maybe I'm missing something? Sorry if I am!

The main purpose of this is to replace the marker layer or basically make it obsolete. Not in it's current form, but it could get there pretty easily.

By using a `LayoutBuilder` you mean you would create a Widget that contains a `LayoutBuilder` containing a `Stack` which contains a custom widget (which gets `LatLng`) which itself uses a `Positioned` widget to position itself accordingly in the `Stack`? I guess in this setup the `LayoutBuilder` isn't even necessary. I'm confused, can you please explain how the `LayoutBuilder` targets the MarkerLayer?

> Is it mainly the calculation in meters ? What use case have you had for a resizing widget based on size in meters out of interest ?

As I wrote before the `WidgetLayer` allows placing widgets with dynamic/intrinsic sizes on the map. This is the main benefit from a users point of view. It doesn't require a fixed size as the Markers do. The meters stuff is just something I required personally for showing an animated circle/area on the map, something like this: https://i.stack.imgur.com/olcrj.gif

PR #1514, opened 2023-05-11
Fix jitter of polygon/polyline after reverting from 2-step positioning

ignatz:
Previously, the canvas was positioned in two steps using a Positioned widget, which was removed due to its own precision issues. Now the canvas needs to be redrawn whenever the viewport changes, not just rotations and zooming but also panning.

### DIFF
--- a/lib/src/layer/polygon_layer.dart
+++ b/lib/src/layer/polygon_layer.dart
@@ -98,12 +98,9 @@ class PolygonLayer extends StatelessWidget {
 class PolygonPainter extends CustomPainter {
   final List<Polygon> polygons;
   final FlutterMapState map;
-  final double zoom;
-  final double rotation;
+  final LatLngBounds bounds;
 
-  PolygonPainter(this.polygons, this.map)
-      : zoom = map.zoom,
-        rotation = map.rotation;
+  PolygonPainter(this.polygons, this.map) : bounds = map.bounds;
 
   int get hash {
     _hash ??= Object.hashAll(polygons);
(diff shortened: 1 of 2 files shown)


JaffaKetchup:
Hey @ignatz.
Please also note #1513: we think the batching changes may have caused some issues when rendering polylines with borders - not sure how we didn't catch this before!
Fixes for the broken status checks (support for Flutter 3.10) available in #1512.

ignatz:
> Hey @ignatz. Please also note #1513: we think the batching changes may have caused some issues when rendering polylines with borders - not sure how we didn't catch this before!

Why the "saveLayers" issue wasn't caught is probably easy to answer: there are no tests, no examples, and the docstring seems to discourage its use.

As to this fix, this is unrelated.

> Fixes for the broken status checks (support for Flutter 3.10) available in #1512.

Unrelated to this fix but FYI, I was able to convert my App to Flutter 3.10 w/o many issues. I guess that's due to the use of my own custom tile provider :shrug:

JaffaKetchup:
> Unrelated to this fix but FYI, I was able to convert my App to Flutter 3.10 w/o many issues. I guess that's due to the use of my own custom tile provider 🤷
Yeah, everything is working, but the checks aren't - there have been some deprecations and stuff.

JaffaKetchup:
(Please pull from master. Only Flutter 3.10 is now supported.)

JaffaKetchup (approved):
LGTM. Thanks :)

PR #1515, opened 2023-05-11
Apply stronger typing of FlutterMap4 more broadly and opportunistic cleanup/fix

ignatz:
FlutterMap4 changes project()'s return type from CustomPoint<num> to CustomPoint<double>, henceforth push the stricter typing further up the stack and do some opportunistic cleanups/fixes.

### DIFF
--- a/lib/src/map/flutter_map_state.dart
+++ b/lib/src/map/flutter_map_state.dart
@@ -41,7 +41,7 @@ class FlutterMapState extends MapGestureMixin
     _rotation = options.rotation;
     _center = options.center;
     _zoom = options.zoom;
-    _pixelBounds = getPixelBounds(zoom);
+    _pixelBounds = getPixelBounds();
     _bounds = _calculateBounds();
 
     WidgetsBinding.instance.addPostFrameCallback((_) {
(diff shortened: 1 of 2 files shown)


lib/src/map/flutter_map_state.dart
```diff
@@ -516,19 +515,21 @@ class FlutterMapState extends MapGestureMixin
...
 
   CustomPoint<int> getNewPixelOrigin(LatLng center, [double? zoom]) {
-    final viewHalf = size / 2.0;
-    return (project(center, zoom) - viewHalf).round();
+    final halfSize = size / 2.0;
+    return (project(center, zoom) - halfSize).round();
   }
 
-  Bounds<double> getPixelBounds(double zoom) {
-    final mapZoom = zoom;
```
ignatz, line 528: I believe this to be a bug. It should have been:

    final mapZoom = this.zoom;

Otherwise scale will always be `1.0` :shrug:

JaffaKetchup: Interesting spot! Will have to test this just to make sure there's no side effects.

JaffaKetchup:
(Please pull from master. Only Flutter 3.10 is now supported.)

JaffaKetchup (approved):
LGTM. Thanks :)

PR #1518, opened 2023-05-12
Attempt to fix actions

JaffaKetchup:


JaffaKetchup:
As we can see from #1520, this still doesn't work. Unfortunately, I think we need to wait for some permissions for the original owner of this project to allow secrets to be passed to forks: https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#enabling-workflows-for-forks-of-private-repositories

PR #1523, opened 2023-05-12
Fine-grained MediaQuery dependencies to rebuild only when needed

ignatz:
Previously, the broad dependency triggered rebuilds when the keyboard came up even when rendered in a Scaffold() with resizeToAvoidBottomInset set to false. This caused a lot of jank for me, i.e. triggering full-map-rebuilds at animation pace.


TesteurManiak (approved):
LGTM and thanks a lot for the added test case 👍

JaffaKetchup:
(Untested by myself.)
Thanks @ignatz!

PR #1530, opened 2023-05-23
Upgrade 'latlong2' & 'http' dependencies

josxha:
- migrate LatLngBounds to use the now immutable LatLng 
- run `flutter fix` and `dart format` to add const for the LatLng constructor


JaffaKetchup (approved):
Thanks @josxha!

PR #1532, opened 2023-05-25
Multiple Improvements & v5 Release Preparation

JaffaKetchup:
Includes a major refactoring and reorganization with the aim of reducing the scope of responsibility for source files.

For changes to functionality and API, see the [CHANGELOG](https://github.com/fleaflet/flutter_map/blob/jaffa/CHANGELOG.md). Note that this PR includes breaking changes.

Example application improvements:
- Reorganized drawer menu
- Added welcome dialog to example app when running at 'demo.fleaflet.dev' only
- Removed unnecessary/bloat demo pages
- Overhauled Marker Layer page
- Improved other demo pages

Other meta changes:
- Updated CHANGELOG and pubspec.yaml
- Fixed GitHub not being able to interpret license
- Add automated pub.dev publishing action on tag creation
- Cut ties with leaflet.js in README
_I think we're starting to progress in our own direction, and I want to [reduce the expectations for feature parity with it](https://github.com/fleaflet/flutter_map/issues/1537#issuecomment-1569082219). Of course, I'm happy to look at it for inspiration as we sometimes do._
  > _It is true that the concepts of this library are similar, but our aim is not necessarily feature parity - we want to implement core features that there's high demand for, and don't really suit plugin architectures._
  > _This is partially because FM is on a much smaller scale than leaflet.js, and there's only 4 maintainers, so we rely on the community for a lot of feature additions - hence, we prefer plugins, as it also reduces the maintenance cost for us._


rorystephenson:
Hey @JaffaKetchup nice work!

Great to see the demo app too.

FYI I just found a bug I introduced in the v4 changes which is worth including in this: #1534 .

TBH it's probably worth cherry-picking and releasing as a hotfix on v4 as well because it breaks tileupdatetransformers that rely on movement ids (like the one for animated movement in the example app).

JaffaKetchup:
Hi @rorystephenson,
Great spot! I've copied your fix into this manually in commit 434ce32 to avoid conflicts, and added you as a co-author. I'll also merge #1534 into 'master' to publish a patch version.

JaffaKetchup:
This appears to have broken tile loading on movement altogether.

lib/src/layer/polyline_layer.dart
```diff
@@ -55,33 +55,35 @@ class Polyline {
...
   const PolylineLayer({
     super.key,
     this.polylines = const [],
     this.polylineCulling = false,
-    @Deprecated('No longer needed and will be removed.')
+    @Deprecated(
+      'No longer has an effect, and no alternative is available. '
+      'This option overcomplicated the situation, and is now decided automatically internally. '
+      'This feature is removed (and this option deprecated) since v5.',
+    )
```
mootw, line 69: I think this should just be removed as it has had a default value, and has no effect now anyways. It is already "broken". However I decided to look at the internal behavior of polyline culling and i don't see any culling actually happening internally.

mootw (changes requested):

lib/src/layer/polyline_layer.dart
JaffaKetchup: Fixed in 59d3cc1.

mootw (approved):

example/lib/pages/fallback_url_offline_page.dart
ibrierley, line 1: There's a few pages removed, like this and maxBounds which are quite fiddly and repetative to test without an example, I feel like some of these are useful, so may be worth pondering their removal more (unless included as part of another example, but maybe we should make that clear if so).

lib/plugin_api.dart
ibrierley, line 1: Any reason we'll be using "misc" as a directory ? Feels a bit unclear what would be in there, but I'm probably easy on it, just trying to check the reasoning.

ibrierley (commented):
I'm pretty ok with all of this, just a note, my mobile is bust atm, so just testing on web, so some bits like rotation I haven't tested, but don't see any obvious reason why that should be changed.

Main ponderings. 

I think there's a few examples that have been removed, but I feel like some of them are handy, that keep cropping up like anchors and maxBounds, so wondering if they should be kept. Not sure the removal outweighs the "bloat".

Any reason for moving some core stuff to misc ? I don't really care tbh, just sometimes in general I find misc folders a kind of slack dumping ground :D.

Couple of bits that are making me scratch my chin, but I don't think are related to this PR (but maybe V4?), but wondering if we should double check before releasing...

The Reset Tile Layer example, doesn't seem to display the tiles after a reset. I don't think they did in very old releases, but then did in recent ones unless I'm mistaken...
I note in the example at https://demo.fleaflet.dev/#/ on my max two finger zoom doesn't work. I know there's been a few odd mac related things that probably aren't flutter_map specific, but may be worth double checking as well (it may just be me).

JaffaKetchup:
@ibrierley Thanks.

> I think there's a few examples that have been removed, but I feel like some of them are handy, that keep cropping up like anchors and maxBounds, so wondering if they should be kept. Not sure the removal outweighs the "bloat".

I removed the following example pages:

- Fallback URL for `AssetTileProvider`  
I originally removed this because I considered it to be similar to the same functionality for the `NetworkTileProvider` - happy to add it back if you think it would be better
- Live Location demo & Live Location feature in Map Controller page  
This didn't work anyway, and there's much better solutions with plugins, and we don't have to deal with permissions
- Network Tile Provider page  
With the changes made to the providers previously, this page is redundant and is the same as the Home page
- Widgets  
There's nothing special anymore - since (I think?) v2, widgets can just be inserted directly into `children`, and I don't think this needs much demonstration

I consolidated the following pages into one new Markers page:
- Tap To Add
- Marker Rotation
- Marker Anchor

> Any reason for moving some core stuff to misc ? I don't really care tbh, just sometimes in general I find misc folders a kind of slack dumping ground :D.

That's precisely what it is! I would say the most 'core' part of this library is the widget and widget state. This was never in "core". 'core' suggests utmost importance to me, and things that are coupled everywhere, whereas 'misc' or 'utils' implies shared code but that is independent. All files in "misc" are independent (not closely related to FM state, etc.).

> The Reset Tile Layer example, doesn't seem to display the tiles after a reset. I don't think they did in very old releases, but then did in recent ones unless I'm mistaken...

As far as I can see, the example is broken. I don't know when it broke, but I'm assuming it broke with the Tile Layer rewrites. But is resetting even useful anymore? Not sure myself.

> I note in the example at https://demo.fleaflet.dev/#/ on my max two finger zoom doesn't work. I know there's been a few odd mac related things that probably aren't flutter_map specific, but may be worth double checking as well (it may just be me).

Mac gestures have been buggy for ages. I unfortunately can't check (don't have a touch screen Mac). I'm thinking that the next update should be a gestures cleanup and addition of rotation support with mice/cursor, but that's just a pipedream at the moment.

lib/plugin_api.dart
JaffaKetchup: See https://github.com/fleaflet/flutter_map/pull/1532#issuecomment-1574992027.

example/lib/pages/fallback_url_offline_page.dart
JaffaKetchup: See https://github.com/fleaflet/flutter_map/pull/1532#issuecomment-1574992027.

example/lib/main.dart
```diff
@@ -31,47 +25,44 @@ import 'package:flutter_map_example/pages/reset_tile_layer.dart';
...
-import 'package:flutter_map_example/pages/tap_to_add.dart';
 import 'package:flutter_map_example/pages/tile_builder_example.dart';
 import 'package:flutter_map_example/pages/tile_loading_error_handle.dart';
-import 'package:flutter_map_example/pages/widgets.dart';
 import 'package:flutter_map_example/pages/wms_tile_layer.dart';
+import 'package:url_strategy/url_strategy.dart';
 
-void main() => runApp(const MyApp());
+void main() {
+  setPathUrlStrategy();
```
TesteurManiak, line 34: Can't we use `usePathUrlStrategy` instead of relying on the url_strategy dependency? Here's the related [Flutter documentation](https://docs.flutter.dev/ui/navigation/url-strategies#configuring-the-url-strategy)

TesteurManiak (approved):

example/lib/main.dart
JaffaKetchup: I used the dependency as it no-ops for non-web platforms. Either way, a dependency is needed (either on this or flutter_web_plugins (which I will admit is from SDK, but still needs to be installed)), and this is just less setup. If you think it's a bigger deal, I'm happy to change it.

TesteurManiak: Well, it's not that bad, merely a suggestion (hence why I've still approved the PR), I'm okay if it stays like this for now 👍 
But as the package hasn't been updated in 2 years it might be wise to separate from it in the future before it becomes an issue.

JaffaKetchup: Sure. Although, I can't see it being changed soon - it's implementation is incredibly simple and literally boils down to calling from flutter_web_plugins and a platform dependent import.

Thanks for reviewing!

example/lib/pages/max_bounds.dart
ibrierley, line 1: I think I'd quite like this to reappear. It's often problematic and hard to find decent examples.

JaffaKetchup: Ah sorry, forgot to say, but this is now applied to the home page.

ibrierley (changes requested):
I think I'm fine with all of it, except I'd quite like maxBounds example to be retained (or included as part of another one).

ibrierley (approved):

rorystephenson (commented):
Hey @JaffaKetchup one little comment/question. Sorry I didn't have time to review this and raise it before merging.

lib/src/layer/marker_layer.dart
```diff
@@ -1,75 +1,107 @@
...
   final double width;
 
   /// Bounding box height of the marker
   final double height;
-  final Anchor anchor;
 
-  /// If true marker will be counter rotated to the map rotation
+  /// Positioning of the [builder] widget relative to the center of its bounding
+  /// box defined by its [height] & [width]
+  final AnchorPos? anchorPos;
```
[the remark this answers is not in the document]
JaffaKetchup: Hi,

I changed it because the constructor used `AnchorPos` anyway, and because I added the ability for `MarkerLayer` to override the anchor position of its descendants. There was also a few other changes that meant this was the best way of doing things.
I didn't think about the consequences of this causing a new calculation every rebuild though. Does this cause a significant performance disadvantage?

It's now nullable because `MarkerLayer` is now able to override it. If you leave it null, it will default to anchoring at the center.

(Apologies for the screenshot instead of line refs, I'm on mobile)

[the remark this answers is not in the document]
rorystephenson: >Does this cause a significant performance disadvantage?

Honestly I don't feel like I have a reliable way to measure this, so far my attempts at performance comparisons with flutter_map feel quite random.

>Can you post your code snippet that you're having trouble migrating?

https://github.com/rorystephenson/flutter_map_marker_popup/blob/8139b1874141885a04bd95d3e2b089a64eea1fd7/lib/src/marker_layer.dart#L80-L83

Here you can see where I used the calculated anchor value, I am able to migrate but I must now calculate the anchor on each build and this requires assuming a default anchorPos if none is defined.

---

We can support defining the anchorPos on MarkerLayer and still store the anchor on Markers if the user defines it by just storing the `Anchor?` on Marker instead of `AnchorPos` and only setting it when the user provides an `anchorPos` on the Marker. However I feel this hides the performance implications (it is more efficient to define anchor per-Marker than at the layer level) from the users. I definitely see the appeal of being able to define such options at the MarkerLayer level (like rotation etc) however for parameters which we can pre-calculate, like Anchor, I'm not sure it's a worthwhile tradeoff.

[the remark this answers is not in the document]
rorystephenson: I've opened an issue to discuss this further: https://github.com/fleaflet/flutter_map/issues/1552

PR #1534, opened 2023-05-30
Reinstate missing id in MapController's move method

rorystephenson:
When refactoring for the TileLayer changes I seem to have accidentally removed the id for move events via MapController. This, amongst other things, breaks TileUpdateTransformers which rely on movement event ids.


JaffaKetchup:
This fix has been implemented in #1532 (434ce32), and will not be cherry-picked or used for a patch release (~v4.0.1~): see https://github.com/fleaflet/flutter_map/pull/1532#issuecomment-1568149553.
Thanks @rorystephenson!

PR #1540, opened 2023-05-31
Emphasis on isFilled = true

altilunium:
Polygon class "isFilled" parameter is set to "false" by default. New users might be confused since they can't easily see the polygon after following the code example.


JaffaKetchup:
Thanks @altilunium, I've added this manually into the 'v5 (beta)' docs.

Issue #1354, opened 2022-09-01
[BUG] MacOS: Multitouch introduction causes issues (Flutter 3.3.0)


JosefWN:
### What is the bug?

With Flutter's new multitouch support [extending to desktop](https://github.com/flutter/flutter/issues/23604), the default behavior in `flutter_map` on Flutter 3.3.0 is two finger pan/rotate and pinch to zoom. This is completely in line with the Apple Maps app on Mac for example.

A couple of issues:
- [x] When I rotate clockwise with my fingers, the map rotates counter-clockwise. EDIT: Fixed in Flutter engine in a coming release.
- [x] ~The panning threshold is a bit tight, so you need some speed when you two-finger swipe to get it to continue the panning as an animation. With slight movements there is no animation and the pan stops abruptly, which is not the case in the native Maps application on macOS for example.~
- [ ] Pinch to zoom does not center zoom on mouse pointer.
- [x] Pinch to rotate does not center rotation on mouse pointer.

It seems like https://github.com/fleaflet/flutter_map/issues/1358 is also related to this umbrella issue (where a `resolutions` list is provided, there seems to be bound checks for other gestures like scroll to zoom, but not for pinch to zoom). Tracked in a separate ticket.
 
Combined rotation and zoom in a single gesture (which works just fine on iPad) doesn't seem to work on Mac at all, neither with `flutter_map` nor Apple Maps, unless it's some setting I need to change on my end. Tried disabling some more complex multitouch gestures in the macOS settings to see if they could be interfering but that was not the case.

As things stand right now I almost liked the old behavior (with scroll to zoom and click & drag to pan) a bit more, although that prevented me from rotating the map intuitively.

### What is the expected behaviour?

Rotating clockwise with a touch gesture in Apple Maps or rotating images in Preview (macOS) rotates the canvas clockwise. The same goes for `flutter_map` built for iPad. It would make sense if it worked the same way on macOS.

### How can we reproduce this issue?

Run on macOS with a touchpad, try to rotate the canvas with a touch gesture.

### Can you provide any other information?

Running `flutter_map` 3.0.0-beta3 on M1 Macbook Air.

### Platforms Affected

MacOS

### Severity

Minimum: Allows normal functioning

### Frequency

Consistently: Always occurs at the same time and location

### Requirements

- [X] I agree to follow this project's [Code of Conduct](https://github.com/fleaflet/flutter_map/blob/master/CODE_OF_CONDUCT.md)
[...]

PR #1543, opened 2023-06-04
Center zoom on pinch zoom gesture (not just on move gesture)

JosefWN:
This should fix the last issue tracked in https://github.com/fleaflet/flutter_map/issues/1354. https://github.com/fleaflet/flutter_map/pull/1081 seems to omit the pinch to zoom case. I tried adding it and it seems to be working on MacOS now, but I haven't tested extensively to see if this change breaks anything else.

### DIFF
--- a/lib/src/gestures/gestures.dart
+++ b/lib/src/gestures/gestures.dart
@@ -435,7 +435,7 @@ abstract class MapGestureMixin extends State<FlutterMap>
                 }
               }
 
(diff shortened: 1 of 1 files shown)


JaffaKetchup (approved):
LGTM, thanks @JosefWN. I can't test this myself, but if this same technique is used elsewhere, I don't see why it shouldn't be used here.

Issue #1544, opened 2023-06-04
[FEATURE] Re-expose `MapControllerImpl` after v5.0.0


JosefWN:
### What is the bug?

Since  `v5.0.0`, the `FlutterMapControllerImpl` does not export its constructors. This makes it hard to use a custom map controller, which I'm doing, and which also @TesteurManiak is doing: https://github.com/TesteurManiak/flutter_map_animations/blob/main/lib/src/animated_map_controller.dart#L21-L25

```
The member 'MapControllerImpl' can only be used within its package.dart[invalid_use_of_internal_member](https://dart.dev/diagnostics/invalid_use_of_internal_member)
```

FYI, also ran into this issue with the tests when debugging, but it seems to be something on flutters end: https://github.com/flutter/flutter/issues/127571

### How can we reproduce it?

Try to implement a custom map controller.

### Do you have a potential solution?

Export the constructor.

### Platforms

All

### Severity

Fatal: Causes the application to crash

TesteurManiak:
To me it's not that big of a deal, I'm currently updating flutter_map_animations so it can take an instance of `MapController` (or instantiate it internally) but it will cause a few breaking changes for my package.

JosefWN:
Okay, my map controller works a bit differently so it might be tricker for me 🥹

JaffaKetchup:
Hi @JosefWN,
Not sure why the constructor needs to be exposed? `MapController`'s constructor redirects to it, and it doesn't take any arguments itself. But I'm probably misunderstanding something, so an example would be helpful.

TesteurManiak:
@JaffaKetchup to give you an example here's my implementation of `AnimatedMapController`:

## Before flutter_map v5

```dart
class AnimatedMapController extends MapControllerImpl {
  // ...
}
```
Which allowed to completely replace `MapController`:

```dart
late final mapController = AnimatedMapController(vsync: this);

@override
void dispose() {
  mapController.dispose();
  super.dispose();
}

@override
Widget build(BuildContext context) {
  return FlutterMap(
    mapController: mapController,
  );
}
```

## After flutter_map v5

Now it acts more like a wrapping of `MapController`:

```dart
class AnimatedMapController {
  AnimatedMapController({
    required this.vsync,
    MapController? mapController,
   // ...
  })  : mapController = mapController ?? MapController(),
        // used to know if the controller should be disposed internally
        _internal = mapController == null;

  // ...
}
```

```dart
[...]

PR #1548, opened 2023-06-06
Re-expose `MapControllerImpl`

JosefWN:
Simple proposal for https://github.com/fleaflet/flutter_map/issues/1544.

In line with `v4.0.0` behavior. Tried and it works with my custom map controller.

### DIFF
--- a/lib/src/map/controller.dart
+++ b/lib/src/map/controller.dart
@@ -4,7 +4,6 @@ import 'package:flutter/material.dart';
 import 'package:flutter_map/flutter_map.dart';
 import 'package:flutter_map/src/map/state.dart';
 import 'package:latlong2/latlong.dart';
-import 'package:meta/meta.dart';
 
 /// Controller to programmatically interact with [FlutterMap], such as
 /// controlling it and accessing some of its properties.
@@ -20,7 +19,7 @@ abstract class MapController {
   /// instance.
   ///
   /// Factory constructor redirects to underlying implementation's constructor.
-  factory MapController() = MapControllerImpl._;
+  factory MapController() = MapControllerImpl;
 
   /// Moves and zooms the map to a [center] and [zoom] level
   ///
@@ -172,9 +171,8 @@ abstract class MapController {
   void dispose();
 }
 
-@internal
 class MapControllerImpl implements MapController {
-  MapControllerImpl._();
+  MapControllerImpl();
 
   @override
   bool move(


TesteurManiak (approved):
LGTM! @JaffaKetchup was there any specific reason to make this constructor private in v5?

JaffaKetchup:
@TesteurManiak No reason other than cleanup of external APIs.

Btw, doesn't it seem redundant to even have this separation anyway (between impl and standard), if they're always exposed together? If they're not merged, we need to copy the docs across to ensure better docs coverage.

TesteurManiak:
Yeah the separation between `MapController` and `MapControllerImpl` is redundant to me too. I don't see why they couldn't be merged into one `MapController` class, that would probably make more sens.

JaffaKetchup (changes requested):
I think that should be done then.

JaffaKetchup:
@JosefWN I'm not 100% sure what you mean by that I'm afraid!

JaffaKetchup:
> I suppose it boils down to whether we think that `MapControllerImpl` will grow larger over time.

Given that #1549 was only recently opened to add new functionality, I think we have to assume it might change.

> because merging them would impose needlessly strict and potentially "clunky" constraints on any extension of MapControllerImpl vs an implementation of MapController.

However, I'm not sure it's needless. Anyway, if we move the implementation methods into the current `MapController`, and make it a `base` (`abstract`?) class, extenders still don't have to implement all methods - they can just override ones important to them, or add new ones? They can also override all getters and variables.

Perhaps I'm missing something. But I do understand that extension/implementation is different.

I would also like @mootw and @ibrierley s' inputs to this!

ibrierley:
I think I'm kind of indifferent on this one, as I never quite understood why it was designed in the way it was originally, so don't feel that aware to spot any issues.

rorystephenson:
My 2 cents (sorry it's late here and I don't have time to read this all now):

I prefer to keep the API size as small as possible by only exposing what we want to which in turn makes refactoring much easier. For that reason I am for keeping MapControllerImpl and not exporting it to hides methods/variables we don't want to expose (like the stream sink or the setter for the map state for example). If there is another way to achieve the same thing I'm all for it.

Again I haven't had time to read this or even check the PR changes and it's laaate here so I this may already be addressed.

JaffaKetchup:
Converting to a draft for now. Let's see how #1551 plays out first!

JaffaKetchup:
@rorystephenson @JosefWN With #1551 now merged, what needs to happen here (if anything)?

rorystephenson:
My vote is to close this and leave things as-is and close this. If we can find a more elegant way to allow extending MapController without exposing MapControllerImpl that would be great but for now it doesn't look like it.

JaffaKetchup:
Maybe we can export impl through the plugins API import? Doesn't make much of a difference, but at least it would be something.

If not, I'm happy to close this.

rorystephenson:
We don't want the MapControllerImpl to be available even to plugins. It gives access to methods which should only be used by flutter_map internals.

I've upgraded the latest version of flutter_map_animations which wraps the MapController and I think it's an elegant solution. IMO we can close this PR.

JaffaKetchup:
Ok. We'll close for now, but if anyone has any objections, just ask for this to be reopened.

Issue #1342, opened 2022-08-16
fitBounds does not work as expected when a Map is rotated


SheenaJacob:
### What is the bug?

The aim of this example is to be able to fit the map to contain the four specified bounds. When the Fit Bounds button is clicked with no Map Rotation then it works as expected, but when a Map is first rotated and then the FitBounds button is pressed this is no longer the case.

### What is the expected behaviour?

When a map is not rotated fitBounds works as expected

![expected](https://user-images.githubusercontent.com/60810797/184839568-84490ba0-d9e4-4f14-b498-9292e064e79f.png)

When a map is rotated then the map no longer shows all the bounds,  but centers it correctly.
![After_rotating](https://user-images.githubusercontent.com/60810797/184839728-d11e420d-449e-4756-abb5-6ba34174b36b.png)

### How can we reproduce this issue?

```dart
import 'package:flutter/material.dart';
import 'package:flutter_map/flutter_map.dart';
import 'package:latlong2/latlong.dart';

void main() {
  runApp(FitBounds());
}

class FitBounds extends StatelessWidget {
  FitBounds({Key? key}) : super(key: key);

  static LatLng topRightCorner = LatLng(4.214943, 33.925781);
  static LatLng topLeftCorner = LatLng(3.480523, 30.844116);
  static LatLng bottomLeftCorner = LatLng(-1.362176, 29.575195);
  static LatLng bottomRightCorner = LatLng(-0.999705, 33.925781);
  final MapController mapController = MapController();

  @override
  Widget build(BuildContext context) {
    final markers = <Marker>[
      Marker(
        width: 100,
        height: 100,
        point: topLeftCorner,
        builder: (ctx) => const Icon(Icons.circle, color: Colors.green),
      ),
      Marker(
        width: 100,
        height: 100,
        point: topRightCorner,
        builder: (ctx) => const Icon(Icons.circle, color: Colors.blue),
      ),
      Marker(
        width: 100,
        height: 100,
        point: bottomLeftCorner,
        builder: (ctx) => const Icon(Icons.circle, color: Colors.red),
      ),
      Marker(
        width: 100,
        height: 100,
        point: bottomRightCorner,
        builder: (ctx) => const Icon(Icons.circle, color: Colors.purple),
      ),
    ];

    return MaterialApp(
      home: Scaffold(
        appBar: AppBar(title: const Text('Animated MapController')),
        body: Padding(
          padding: const EdgeInsets.all(8),
          child: Column(
            children: [
[...]

PR #1549, opened 2023-06-08
Added `MapController.fitCoordinates` method

jjoelson:
Hi all! With this PR I'm proposing to add a new map controller method called `fitCoordinates`. 

When we have a list of lat/lng coordinates and want to fit the map viewport to those coordinates, the current best practice is to first calculate a `LatLngBounds`, and then use `fitBounds`:
```
final coordinates = <LatLng>[...];
final bounds = LatLngBounds.fromPoints(coordinates);
mapController.fitBounds(bounds);
```

With this PR, we can instead write:
```
mapController.fitCoordinates(coordinates);
```

The origin of this PR is that I was trying to implement a fix for https://github.com/fleaflet/flutter_map/issues/1342 to make `fitBounds` work correctly when the map is rotated. I implemented [this fix](https://github.com/fleaflet/flutter_map/issues/1342#issuecomment-1298082024) which was proposed by @JosefWN but the result was a bit surprising to me:

fitBounds with rotation

I was puzzled by the additional padding to the left of the green marker, until I remembered `fitBounds` is fitting the `LatLngBounds`, _not_ the 4 coordinates themselves. Given `LatLngBounds` can only represent a rectangle bounds in a north-up orientation, then what is happening here is correct; that extra margin on the left is required to fit the northwestern most point of the bounds. If we draw a black rectangle polyline that represents the `LatLngBounds` of the four points, then we can see it's working as it should:

fitBounds with rotation and bounding box

So I realized `fitBounds`, even if corrected to work with rotation, is not what I actually need, and is probably not what most users of this package need when trying to fit a list of coordinates in the viewport when the map is rotated. Rather, I needed a method like `fitCoordinates` that fits specific coordinate points, regardless of rotation. With `fitCoordinates`, the result looks like this:

fitCoordinates with rotation

Rather than fitting a north-up bounding box of the coordinates, `fitCoordinates` fits the coordinates themselves.

I plan to add tests and make some minor tweaks to the code to avoid some duplicate computations, but I wanted to post this PR before putting in too much more work so I can get feedback from the experienced maintainers of this package about this idea. Am I correct that `fitCoordinates` does something useful and different than `fitBounds` when the map is rotated?

### DIFF
--- a/lib/src/gestures/map_events.dart
+++ b/lib/src/gestures/map_events.dart
@@ -20,6 +20,7 @@ enum MapEventSource {
   doubleTapZoomAnimationController,
   interactiveFlagsChanged,
   fitBounds,
+  fitCoordinates,
   custom,
   scrollWheel,
   nonRotatedSizeChange,
--- a/lib/src/map/controller.dart
+++ b/lib/src/map/controller.dart
@@ -126,6 +126,22 @@ abstract class MapController {
     FitBoundsOptions? options,
   });
 
+  /// Move and zoom the map to perfectly fit [coordinates], with additional
+  /// configurable [options]
+  ///
+  /// For information about return value meaning and emitted events, see [move]'s
+  /// documentation.
+  bool fitCoordinates(List<LatLng> coordinates, {FitBoundsOptions? options});
+
+  /// Calculates the appropriate center and zoom level for the map to perfectly
+  /// fit [coordinates], with additional configurable [options]
+  ///
+  /// Does not move/zoom the map: see [fitCoordinates].
+  CenterZoom centerZoomFitCoordinates(
+    List<LatLng> coordinates, {
+    FitBoundsOptions? options,
+  });
+
   /// Convert a screen point (x/y) to its corresponding map coordinate (lat/lng),
   /// based on the map's current properties
   LatLng pointToLatLng(CustomPoint screenPoint);
@@ -240,6 +256,22 @@ class MapControllerImpl implements MapController {
         const FitBoundsOptions(padding: EdgeInsets.all(12)),
   }) =>
       _state.centerZoomFitBounds(bounds, options!);
+
+  @override
+  bool fitCoordinates(
+    List<LatLng> coordinates, {
+    FitBoundsOptions? options =
+        const FitBoundsOptions(padding: EdgeInsets.all(12)),
+  }) =>
+      _state.fitCoordinates(coordinates, options!);
+
+  @override
+  CenterZoom centerZoomFitCoordinates(
+    List<LatLng> coordinates, {
+    FitBoundsOptions? options =
+        const FitBoundsOptions(padding: EdgeInsets.all(12)),
+  }) =>
+      _state.getCoordinatesCenterZoom(coordinates, options!);
 
   @override
   LatLng pointToLatLng(CustomPoint localPoint) =>
--- a/lib/src/map/state.dart
+++ b/lib/src/map/state.dart
@@ -468,6 +468,25 @@ class FlutterMapState extends MapGestureMixin
     return zoom;
   }
 
+  bool fitCoordinates(
+    List<LatLng> coordinates,
+    FitBoundsOptions options, {
+    Offset offset = Offset.zero,
+  }) {
+    final target = getCoordinatesCenterZoom(coordinates, options);
+    return move(
+      target.center,
+      target.zoom,
+      offset: offset,
+      source: MapEventSource.fitCoordinates,
+    );
+  }
+
+  CenterZoom centerZoomFitCoordinates(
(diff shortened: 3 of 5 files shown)


JosefWN:
Hi @jjoelson!

Yes, you are spot-on. I only found the bug in `fitBounds`, but just like you, I realized that `fitBounds` might not be all that good for coordinate/vector-based layers. I have since been pretty busy in other projects, but I'm glad you have come up with a solution. I'm off to bed now, but will have a look at your PR when I wake up!

jjoelson:
> Hi @jjoelson!
> 
> Yes, you are spot-on. I only found the bug in `fitBounds`, but just like you, I realized that `fitBounds` might not be all that good for coordinate/vector-based layers. I have since been pretty busy in other projects, but I'm glad you have come up with a solution. I'm off to bed now, but will have a look at your PR when I wake up!

Great, thank you!

While writing tests I realized that the calculations to handle padding are not quite correct, so I'm going to push an update at some point today or tomorrow to address that.

JosefWN:
Okay, I will have a look a bit later then!

Oh, and if you have the fixed `fitBounds`-method, feel free to make a PR with that too (or I can do it, but my fork looks a bit different so I would need to port my fix). The `fitBounds`-method is still relevant for raster layers etc.

jjoelson:
OK, I pushed an update with the fixed padding computation and tests. I have two questions:

- Currently `fitCoordinates` uses `FitBoundsOptions`, as none of the properties in `FitBoundsOptions` seem to be specific to the bounds use-case. But would it make sense to create a separate `FitCoordinatesOptions` class in case there's a need to add use-case specific properties in the future?
- Are there any other things that need to be updated as part of this PR e.g. documentation, examples, changelog, etc.?

> Oh, and if you have the fixed fitBounds-method, feel free to make a PR with that too

I should be able to put up a separate PR for that some time next week. 👍

JaffaKetchup:
Hey @jjoelson,

Thanks for the contributions so far! It looks like #1342 may have been 'complaining' about a similar thing, so definitely open to fixes.

Just to let you know, the checks appear to be failing atm due to an unnecessary import.

> Are there any other things that need to be updated as part of this PR e.g. documentation, examples, changelog, etc.?

Don't worry about documentation or changelog, I do that before a release. If you can add an example, please do!

> Currently `fitCoordinates` uses `FitBoundsOptions`, as none of the properties in `FitBoundsOptions` seem to be specific to the bounds use-case. But would it make sense to create a separate `FitCoordinatesOptions` class in case there's a need to add use-case specific properties in the future?

If they can share now, they should share now. However, the class does need some improvements in terms of documentation, clarity, and ensuring consistency.

jjoelson:
I put up a second PR to update `fitBounds`: https://github.com/fleaflet/flutter_map/pull/1550

rorystephenson:
Hey @jjoelson, nice work on this PR and the fitBounds fix!

I was looking at bounds related code in #1551 before I noticed these two PRs. We have three different ways of expressing map bounds right now (four with your coordinate bounds) and I was hoping we could clean that up by creating a MapBounds base class which the different bounds implementations can subclass. I mentioned it [in this comment](https://github.com/fleaflet/flutter_map/pull/1551#issuecomment-1586901114) but disregard the suggestions for the specific methods as I'm starting to think they don't make sense. I think we have the following variations of map bounds (I've given them names for convenience, the names can change):

- `VisibleCenterBounds` - Map bounds that limit the center point of the map.
- `VisibleEdgeBounds` - Map bounds that limit the visible edges of the map, i.e. prevents coordinates from outside of a defined LatLng bounds from being viewed. This is useful for preventing grey borders when zooming out or stopping missing tiles from being loaded when using asset tiles.
- `FitBounds` - bounds that should contain LatLng bounds, with optional padding etc.
- `CoordinateBounds` - Map bounds that should contain a list of coordinates (your addition).

These would all subclass `MapBounds` and expose relevant methods to detmine if we are inside the bounds, CenterZoom, etc. This would allow us to use a single method in the controller for fitting the map to any of these different bounds. MapBounds would also be useful in MapOptions:

- `bounds` (which is initial bounds) could be a MapBounds instead of just a LatLngBounds.
- maxBounds/adaptive bounds/swPanBoundary/nePanBoundary would all become a single option (perhaps `maBounds`).

If you think this is doable and a good idea, maybe you could help me determine what methods MapBounds would need to expose?

JaffaKetchup:
(I've converted this to a draft due to it's conflicting interaction with #1551. @rorystephenson Any commits that integrate this change, if this PR is no longer needed, please add @jjoelson as co-author and mention this PR ID in commit message to give appropriate credit.)

jjoelson:
I will resolve the conflicts and push shortly.

JaffaKetchup:
Replaced by #1551. Thanks @jjoelson for your contribution!

Issue #1547, opened 2023-06-06
[FEATURE] `FlutterMapState` over-notifies dependent widgets via `updateShouldNotify`


rorystephenson:
### What is the bug?

FlutterMapState over-notifies dependent widgets. `InheritedWidget`s should only notify their dependents when it is necessary and right now FlutterMapState will always notify dependents. This was causing me confusion when debugging one of the plugins I maintain and leads to unnecessary builds.

### How can we reproduce it?

See the diff in this branch: https://github.com/fleaflet/flutter_map/compare/master...rorystephenson:flutter_map:did-change-dependencies-bug?expand=1

If you run hot reload you will see that didChangeDependencies is called even though nothing changed. I haven't tested it but I presume didChangeDependencies will also get triggered when changing parent widgets of FlutterMap even if those changes don't affect FlutterMap itself.

### Do you have a potential solution?

I'm sharing this here because it's probably a symptom of a bigger issue.

As far as I can see we would need to create a real implementation for `_MapStateInheritedWidget.updateShouldNotify`. I don't see how we can do so without making FlutterMapState immutable and emitting a new one whenever it changes. We can't currently compare with the oldWidget's mapState because they are the same object which is mutated when changes occur.

So the proposal would be for FlutterMapState to just become an immutable object that only contains final variables that describe the current map state (center, size, zoom...) and for it to implement `==` so that we can determine if the map state has changed and give a proper implementation for `_MapStateInheritedWidget.updateShouldNotify`. The rest of the logic in FlutterMapState would move to appropriate classes.

Obviously this is a huuuge change so I want to kick off a discussion about it and see if anyone has different ideas.

### Platforms

All

### Severity

Obtrusive: Prevents normal functioning but causes no errors in the console

JaffaKetchup:
Just to note, this was the comment before the recent reorganization:
![image](https://github.com/fleaflet/flutter_map/assets/58115698/f5e5392e-6004-469a-89f5-5991bfa0c6cc)

ibrierley:
Reminds me a little of https://github.com/fleaflet/flutter_map/pull/1408 but not sure if that's a redherring, but may be worth a quick scan.

My own take I think is that it makes sense. I've also pondering before about the idea of a state that could even be serialized (for debug at least), where you could effectively record a state, replay it etc, but that's probably a bit of a distraction as well.

[...]

PR #1550, opened 2023-06-09
Updated fitBounds to handle rotation

jjoelson:
Per the discussion in https://github.com/fleaflet/flutter_map/pull/1549, here's a separate PR to address https://github.com/fleaflet/flutter_map/issues/1342.

I've added tests no padding, symmetric padding, and asymmetric padding for every 10 degree intervals between -360 and 360. Here are visuals of the test cases:

**No padding:**

https://github.com/fleaflet/flutter_map/assets/824653/9a60f89a-142d-4dbc-bc37-241f1ab11315

**Symmetric padding:**

https://github.com/fleaflet/flutter_map/assets/824653/ce191d52-01d1-4e3b-9a41-593a70fa6398

**Asymmetric padding:**

https://github.com/fleaflet/flutter_map/assets/824653/c2a376ad-58b1-4f83-8f0b-087c5a1069cc

PR #1551, opened 2023-06-09
Major State Refactoring

rorystephenson:
This PR can be considered exploratory, prompted by #1547. I think overall the code quality is significantly improved but there are some major breaking changes so I want to see what other contributors think about these changes.

It's late where I am so I will try to update this summary better tomorrow or in the next couple of days but for now some highlights are:

- Immutable FlutterMapState: This will make testing far easier.
- Dropped children/nonrotatedChildren in favour of a builder. This resolves the issue of not being able to interlace rotated and non rotated layers. That said it's obviously a big breaking change although it's quite easy to migrate (see changes to examples). Could investigate re-adding children/nonrotatedChildren as deprecated options if we feel that's necessary.
- All changes to the map state are now managed by FlutterMapStateController.
- General tidy-ups.
- Added MapController.of(context) which plugins can use.
- Many map events now pass the FlutterMapState before & after the event instead of just the zoom/latLng.

Some more potential improvements which I haven't tackled yet, some of which are probably for another PR:

- [x] More tidying up of the gesture code (splitting monster methods in to smaller more understandable ones).
- [x] Replace methods in MapController that proxy to FlutterMapState with a getter for FlutterMapState.
- [x] Make the InheritedWidget an InheritedModel with aspects (FlutterMapState/MapController)
- [x] Consider whether MapOptions should be removed from FlutterMapState. This would allow us to expose it as a separate aspect if the previous point about InheritedWidget is implemented. It would also reduce builds caused by options changes which don't end up affecting the state.
- [ ] ~~Consider merging PositionedTapDetector2 code in to the FlutterMapInteractiveViewer. This would mean we just have one gesture detector (raw gesture detector).~~


JaffaKetchup:
Thanks @rorystephenson! I'll have look over everything, but just one thing so far:

> Dropped children/nonrotatedChildren in favour of a builder. This resolves the issue of not being able to interlace rotated and non rotated layers. That said it's obviously a big breaking change although it's quite easy to migrate (see changes to examples). Could investigate re-adding children/nonrotatedChildren as deprecated options if we feel that's necessary.

I unfortunately don't like this. FM is supposed to be simple to use: having the user handle creating `Stack`s and handle moving `mapState` around is far too complex.
Is there something we can do with a wrapper widget for layers? For example, a layer could be wrapped in `LayerZIndex` (that name probably isn't ideal), which would explicitly state the z-index of the layer.

rorystephenson:
> Thanks @rorystephenson! I'll have look over everything, but just one thing so far:
> 
> > Dropped children/nonrotatedChildren in favour of a builder. This resolves the issue of not being able to interlace rotated and non rotated layers. That said it's obviously a big breaking change although it's quite easy to migrate (see changes to examples). Could investigate re-adding children/nonrotatedChildren as deprecated options if we feel that's necessary.
> 
> I unfortunately don't like this. FM is supposed to be simple to use: having the user handle creating `Stack`s and handle moving `mapState` around is far too complex. Is there something we can do with an optional wrapper widget for layers? For example, a layer could be wrapped in `LayerZIndex` (that name probably isn't ideal), which would explicitly state the z-index of the layer. Alternatively we create a public/exposed `mixin` that adds a `zIndex` property to any widget. Then apply this mixin to all existing layers.

Yeah I get where you're coming from. Personally I prefer the builder because the children/nonrotatedChildren feels broken in that you can't interleave them but this is just preference.

That has given me an idea, we can support both and I think it can be fairly clean.

The existing method which we can keep:

```dart
FlutterMap(children: [...], nonrotatedChildren: [...])
```

Alternative builder method:

```dart
FlutterMap.builder(builder: (context, mapState) {...})
```

The mapState parameter in the builder could be removed since the FlutterMapTranslation.rotate will just look for it in the BuildContext if you don't provide it (which is maybe slightly less performant than providing it if you already have it stored in a variable but I guess it's probably negligible).

JaffaKetchup:
I'm still not convinced by the builder I'm afraid.

What if we create a new layer `NonRotatingLayer`? Here's an implementation that works. It keeps `nonRotatedChildren` available, and doesn't need any migration (no breaking changes) to use.

https://github.com/fleaflet/flutter_map/blob/79b54e4008b7d77ac54a55e9407839a4cd9854bc/lib/src/map/state.dart#L164-L178

```dart
child: Stack(
  children: [
    ...widget.children.map(
      (child) => child is NonRotatedLayer
        ? child
        : OverflowBox(
            minWidth: size.x,
            maxWidth: size.x,
            minHeight: size.y,
            maxHeight: size.y,
            child: Transform.rotate(
              angle: rotationRad,
              child: child,
            ),
          ),
    ),
    ...widget.nonRotatedChildren,
  ],
),
```

```dart
class NonRotatedLayer extends StatelessWidget {
  const NonRotatedLayer({
    super.key,
    required this.child,
  });

  final Widget child;

  @override
  Widget build(BuildContext context) => child;
}
```

rorystephenson:
I started wrote a response before deciding that it's not worthwhile including the builder change in this PR since I don't think we're going to reach a consensus easily and the builder changes are actually a small part of the changes. The main changes are immutable FlutterMapState and gesture refactoring. I am going to remove the builder change so that we can focus on the main parts of the PR.

For future reference when revisiting the removal of nonrotatedChildren my response is below.

---

>I'm still not convinced by the builder I'm afraid.

Can you elaborate on why?

>What if we create a new layer NonRotatingLayer? Here's an implementation that works. It keeps nonRotatedChildren available, and doesn't need any migration (no breaking changes) to use.

At first sight I didn't like this idea because it required wrapping children in a widget just to serve as a type signature, however the same thing can be achieved with a mixin e.g.

```dart
mixin FlutterMapNoRotateMixin on Widget {}

class NonRotatedLayer extends StatelessWidget with FlutterMapNoRotateMixin {
  final Widget child;

  /// Wrapping a child of the 
  const NonRotatedLayer({
    super.key,
    required this.child,
  });

  @override
  Widget build(BuildContext context) => child;
}
```

Then for example the attribution widget can just mix in the mixin.

Unfortunately though there is another issue which, for me, makes this a no-go. If you move a class which uses this mixin down one level in the widget tree it no longer works. E.g. if you wrapped a NonRotatedLayer in a Container it will not work anymore. This would be very confusing for users.

JaffaKetchup:
Sure, I think that's probably a good idea! I've responded in #1553, and we can continue the discussion there in the meantime.

JaffaKetchup:
On another note, is this ready for a preliminary review, or did you have more plans?

lib/src/gestures/flutter_map_state_controller.dart
```diff
@@ -0,0 +1,440 @@
...
+  // the FlutterMapInteractiveViewerState is initialized.
+  set interactiveViewerState(
+    FlutterMapInteractiveViewerState interactiveViewerState,
+  ) =>
+      _interactiveViewerState = interactiveViewerState;
+
+  void linkMapController(MapControllerImpl mapControllerImpl) {
+    _mapControllerImpl = mapControllerImpl;
+    _mapControllerImpl.stateController = this;
+  }
```
JaffaKetchup, line 29: Should this be a setter?

lib/src/gestures/interactive_flag.dart
```diff
@@ -39,4 +42,25 @@ class InteractiveFlag {
   static bool hasFlag(int leftFlags, int rightFlags) {
     return leftFlags & rightFlags != 0;
   }
```
JaffaKetchup, line 44: If we now have individual checkers, should `hasFlag` even be public anymore?

lib/src/map/flutter_map_state.dart
```diff
@@ -0,0 +1,443 @@
...
+  }
+
+  FlutterMapState withOptions(MapOptions options) => FlutterMapState(
+        options: options,
+        center: center,
+        zoom: zoom,
+        rotation: rotation,
+        nonRotatedSize: nonRotatedSize,
+        size: _calculateSize(rotation, nonRotatedSize),
+      );
```
JaffaKetchup, line 109: Unless I'm missing something, this should be part of `copyWith`.

lib/src/map/map_controller.dart
```diff
@@ -20,7 +21,16 @@ abstract class MapController {
...
+  factory MapController() = MapControllerImpl;
+
+  static MapController? maybeOf(BuildContext context) => context
+      .dependOnInheritedWidgetOfExactType<MapStateInheritedWidget>()
+      ?.mapController;
+
+  static MapController of(BuildContext context) =>
+      maybeOf(context) ??
+      (throw StateError(
+          '`MapController.of()` should not be called outside a `FlutterMap` and its children'));
```
JaffaKetchup, line 33: Just a question. Does this mean that the controller can now be access by the end user in descendants, without having to specify a controller manually?

lib/src/map/map_controller.dart
```diff
@@ -117,179 +127,12 @@ abstract class MapController {
...
-  }) =>
-      _state.centerZoomFitBounds(bounds, options!);
-
-  @override
-  LatLng pointToLatLng(CustomPoint localPoint) =>
-      _state.pointToLatLng(localPoint);
-
-  @override
-  CustomPoint<double> latLngToScreenPoint(LatLng latLng) =>
-      _state.latLngToScreenPoint(latLng);
```
JaffaKetchup, line 250: These need to be restored.

JaffaKetchup (commented):
Just a few little nitpicks and questions, haven't gone through in detail yet. Feel free to argue if you think otherwise, I've probably misunderstood something.

lib/src/gestures/flutter_map_state_controller.dart
rorystephenson: Given that it's not just setting the mapControllerImpl, it's also setting the MapControllerImpl's FlutterMapStateController, I think it's clearer to keep it a method with 'link' in the name.

lib/src/gestures/interactive_flag.dart
rorystephenson: `hasFlag` lets you check multiple flags so I think we should keep it. Maybe we should rename it to hasFlags to make that clearer (with a redirecting method and deprecation).

lib/src/map/map_controller.dart
rorystephenson: This was previously possible with FlutterMapState.of(context).mapController. Now it is given a dedicated method to allow only looking up the MapController if that's all you need. This will also increase efficiency when the inherited widget is an inherited model since updating the state won't trigger an update of code which just wants the controller.

lib/src/gestures/flutter_map_state_controller.dart
JaffaKetchup: Sure, guess that makes sense.

lib/src/gestures/interactive_flag.dart
JaffaKetchup: I wouldn't bother then :D

lib/src/map/map_controller.dart
rorystephenson: It's now available via the `mapController.mapState.centerZoomFitBounds` etc. If we want to avoid breaking changes I can add those methods back with deprecations?

JaffaKetchup: I'm not sure why they need to be moved in the first place? I guess because it doesn't actually change any state?

lib/src/map/flutter_map_state.dart
rorystephenson: The withRotation method returns the current FlutterMapState if the rotation doesn't change. This avoids recalculating the size as well as retaining lazy calculated fields (pixelBounds/bounds/pixelOrigin).

rorystephenson:
> On another note, is this ready for a preliminary review, or did you have more short-term plans?

Actually I think I'm going to work on the InhertiedModel change so I would say wait for now.

I do, however, have a question which as arised from refactoring.

In `MapOptions` what is the difference between `bounds`, `swPanBoundary`/`nePanBoundary`and `maxBounds`.

My rough understanding right now is that `swPanBoundary`/`nePanBoundary` are only for use when the `adaptiveBoundaries` is enabled. However the difference between `bounds`/`maxBounds` is confusing me.

ibrierley:
I think...

bounds is basically fitBounds (i.e on initialisation), i.e make the map fit as best possible to the bounds provided, but doesn't restrict during future pan/zoom iirc. 

maxbounds is basically, don't show outside these bounds at any point at all (to eliminate grey areas outside a map area for example on init and during pan/zooms). 

I think you're right on the pan ones, but it's a while since I've tried, so may need to double check the code for the adaptive boundaries thing (unless anyone else can remember specifically).

rorystephenson:
> bounds is basically fitBounds (i.e on initialisation), i.e make the map fit as best possible to the bounds provided, but doesn't restrict during future pan/zoom iirc.
Ah yep looks like you're right, it would be worth renaming this/center/zoom to initialBounds/initialCenter/initialZoom (with deprecations) as it's quite confusing.

> maxbounds is basically, don't show outside these bounds at any point at all (to eliminate grey areas outside a map area for example on init and during pan/zooms).

Hmm this is where it gets confusing for me because that's what I though adaptive bounds is supposed to do?

JaffaKetchup:
It's explained in the docs, and I'm not really a fan of making breaking changes just for naming, but that will affect all users, at the moment. 

Tbh, I can't remember what the point of adaptive boundaries was. I think it was based on screen size instead of coordinates, or something like that, but I have no idea.

ibrierley:
I think when people get confused, it's always worth revisiting what the concepts are, just in case there are tweaks in method names or documentation etc. After all if people ask, there's a reason, I certainly forget :D.

I "think" there's separate things here though, i.e a boundary of the center you can drag to, vs a boundary of the edges displayed.

rorystephenson:
>It's explained in the docs, and I'm not really a fan of making breaking changes just for naming, but that will affect all users, at the moment.

I tried setting a bound limit on a personal project of mine not long ago but didn't bother in the end because `bounds` wasn't working and I couldn't be bothered jumping in to the code to figure out why. I also remember when I first used FlutterMap a few years ago I was confused why changing center/zoom did not update the map. I think if the changes are made with appropriate deprecations it's a trivial change for users.

---

As far as adaptive versus max bounds I've never used either so I'm also a bit lost.

This is the relevant part of FlutterMapState's `move()` method which uses these values:

https://github.com/fleaflet/flutter_map/blob/79b54e4008b7d77ac54a55e9407839a4cd9854bc/lib/src/map/state.dart#L406-L420

It uses them as follows:
1. It uses `isOutOfBounds()` to determine if we are out of bounds. `isOutBounds` relies on the adaptive bounds or, if they are not defined, ne/sw pan boundaries to determine if we are out of bounds. If outside of bounds and...
    **a.** If `slidingBoundaries` is disabled the move is aborted.
    **b.** Otherwise it uses `containPoint()` to move the center so that we are inside the bounds. `containPoint` uses the adaptive bounds, if they are defined, or ne/sw pan boundaries to determine the new center.
2. If `maxBounds` is defined `adjustCenterIfOutsideMaxBounds()` will try to find a center position which places us inside the maxBounds. If it fails to find one (maybe because we are zoomed too far out) the move is aborted.

So we seem to have two different ways of trying to stay within the bounds. The next thing to figure out is why there is two and if we can combine them in to one.

rorystephenson:
@ibrierley looks like you're right from an old comment of yours: https://github.com/fleaflet/flutter_map/issues/177#issuecomment-1279154556

So I will integrate the various bounds options in to a single bounds option to make this clear.

ibrierley:
As always, I forget what I wrote in the past :D

rorystephenson:
So adaptive bounds come from [this PR](https://github.com/fleaflet/flutter_map/pull/430) and they were added to stop the visible bounds from going outside of the defined LatLng bounds. This was initially intended for an asset tile provider to prevent loading tiles for which there is no tile asset (if only within a certain range were downloaded).

`maxBounds` was added a couple of years later in [this PR](https://github.com/fleaflet/flutter_map/pull/1211). It was added to get rid of the grey borders when the map is zoomed out.

These options were added for different reasons and, whilst I'm not 100% certain, they _seem_ to essentially achieve the same thing: prevent viewing of coordinates outside of the defined LatLng boundary. The algorithms are different although they may be different ways of achieving the same thing, some testing will be needed to determine that and choose a preferred one.

I did notice that `maxBounds` doesn't prevent seeing the gray borders when rotating. This may also be the case with adaptive boundaries but i'll need to modify the sliding map example to determine that because I don't see missing tiles but I'm not sure if the asset tiles finish where the provided adaptive bounds finish. In both cases some digging will be necessary to determine if there is a problem with the algorithm or something else (e.g. bounds checking isn't done whilst rotating, only when moving).

P.S.
I had a laugh when I noticed that all three of us have been confused by `bounds` before, see the comments from [this comment](https://github.com/fleaflet/flutter_map/pull/1211#issuecomment-1094925762) onwards.

ibrierley:
Hah yes, it's always been a bit confusing, but in some ways I'm not surprised, as it's not even always clear when talking about it :D.
So nothing like actually testing it to see...and I was kinda surprised actually haha. You're kind of right, but the implementation has different effects, so they aren't quite the same afaics.

So firstly, maxBounds aim was to avoid grey areas and pretty much match leaflet.js implementation of maxBounds (hence why it's called that, and something not a bit more clear), i.e https://leafletjs.com/reference.html search for maxBounds "When this option is set, the map restricts the view to the given geographical bounds, bouncing the user back if the user tries to pan outside the view."

swPanBoundary (with adaptive off), seems to keep the center constrained to the boundary. Indeed adaptiveBoundaries true then appears to have the aim of keeping the visible area constrained, and it works similar, but if you zoom out for example, you will be able to see outside of the pan area (but can't pan). So it is more of a literal pan boundary, rather than a pure visible boundary. 

I also note in the original PR for adaptiveBoundaries it says "Caching a few tiles outside the boundaries is recommended for cases where the user is allowed to zoom out so much that the entire area fits on the screen." so the implication is that it won't limit in the same way.

Here's an adjusted sliding_map example with markers draw at the boundary, just pop this in place of the existing example one in examples folder, and comment/uncomment the relevant lines around 32 & 39 to swap between them.

I think I did test originally maxBounds with rotation, but annoyingly my mobile is bust to be able to tell properly whether this is the case (kinda hard to tell with just a manual force), so it would be good for someone to test that at least and see if it does work or not, and whether I'm imagining it :D.

It's slightly annoying that there are two quite similar features, I suspect both are used, but may be wrong.

```
import 'package:flutter_map_example/widgets/drawer.dart';

class SlidingMapPage extends StatelessWidget {
  static const String route = '/sliding_map';

  const SlidingMapPage({Key? key}) : super(key: key);

  @override
    return Scaffold(
      appBar: AppBar(title: const Text('Sliding Map')),
      drawer: buildDrawer(context, route),
      body: Padding(
        padding: const EdgeInsets.all(8),
        child: Column(
          children: [
            const Padding(
              padding: EdgeInsets.only(top: 8, bottom: 8),
              child: Text(
                  'This is a map that can be panned smoothly when the boundaries are reached.'),
            ),
            Flexible(
              child: FlutterMap(
                options: MapOptions(
                  center: const LatLng(56.704173, 11.543808),
                  //minZoom: 12,
                  //maxZoom: 14,
                  zoom: 15,
                  //comment out these 4 lines
                  swPanBoundary: const LatLng(56.6877, 11.5089),
                  nePanBoundary: const LatLng(56.7378, 11.6644),
                  adaptiveBoundaries: true,
                  slideOnBoundaries: true,


                  //comment out this line
                  //maxBounds: LatLngBounds(const LatLng(56.6877, 11.5089),
                  //    const LatLng(56.7378, 11.6644)),

                  screenSize: MediaQuery.of(context).size,
                ),
                children: [
                  TileLayer(
                    urlTemplate:
                    'https://tile.openstreetmap.org/{z}/{x}/{y}.png',
                  ),
                  MarkerLayer( markers: [
                    Marker(
                      point: const LatLng(56.6877, 11.5089),
                      width: 64,
                      height: 64,
                      anchorPos: AnchorPos.align(AnchorAlign.center),
                      builder: (context) => const ColoredBox(
                        color: Colors.lightBlue,
                        child: Align(
                          alignment: Alignment.centerRight,
                          child: Text('-->'),
                        ),
                      ),
                    ),
                    Marker(
                      point: const LatLng(56.7378, 11.6644),
                      width: 64,
                      height: 64,
                      anchorPos: AnchorPos.align(AnchorAlign.center),
                      builder: (context) => const ColoredBox(
                        color: Colors.lightBlue,
                        child: Align(
                          alignment: Alignment.centerRight,
                          child: Text('-->'),
                        ),
                      ),
                    ),
                  ],
                  ),
                ],
              ),
            ),
          ],
        ),
      ),
    );
  }
}
```

rorystephenson:
Thanks for having a look @ibrierley ! So I can confirm that on mobile web using the demo when I zoom all the way out and rotate the grey border is partially visible in the corners. I'd need to have a further look to determine if that's due to the algorithm or because the bounds limiting is not called in all the places is should be (e.g. when calling `rotate`).

So right now I'm looking to create a MapBoundary base class with the following subclasses:

- `VisibleCenterBoundary`: This would match the old sw/ne pan boundary implementation so it prevents the visible center from going outside of a given `LatLngBounds`.
- `VisibleEdgeBoundary`: This would replace adaptive and max bounds. It would prevent points from outside of a given `LatLngBounds` from being shown on the map.

I'm thinking that `MapBoundary` would define methods something like this:

- `contains(nonRotatedSize, center, rotation, zoom)`: Determine if a given center/rotation/zoom is within the boundary.
- `minimumZoom(nonRotatedSize, center, rotation)`: Determine the minimum possible zoom for a given center & rotation. This can be used to determine if a given center/rotation/zoom can be moved to remain within the bounds without zooming (if the desired zoom is >= the minimum).
- `clamp(center, rotation, zoom)`: Determine the nearest center value which would place the visible boundary within this boundary.

I wonder if `FitBoundsOptions` can also become a subclass of `MapBoundary`. In fact it seems like it could be combined with `VisibleEdgeBoundary` by adding the `padding`, `maxZoom`, `inside` and `forceIntegerZoomLevel` options.

I'm not attached to the names, if we can come up with names which better describe the behaviour I'm all for it.

rorystephenson:
I was diving in to the various bounds implementations when I noticed #1549 and #1550 which add a new fit bounds method and add a fix for fit bounds, respectively. I'm hopeful that we can create a MapBounds abstraction which can encapsulate all of this. So then rather than adding a new fitCoordinates method to the controller we can have a MapCoordinateBounds subclass with its own implementation for determining if we are within the coordinate bounds and the center zoom required to be inside the coordiante bounds if we are not already.

JaffaKetchup:
> `minimumZoom(nonRotatedSize, center, rotation)`
> `clamp(nonRotatedSize, center, rotation, zoom)`

What about `clampZoom` & `clampCenter`? Makes them kind of related in a way, as they are.

I've merged #1550, feel free to refactor it however.

jjoelson:
> I was diving in to the various bounds implementations when I noticed #1549 and #1550 which add a new fit bounds method and add a fix for fit bounds, respectively. I'm hopeful that we can create a MapBounds abstraction which can encapsulate all of this. So then rather than adding a new fitCoordinates method to the controller we can have a MapCoordinateBounds subclass with its own implementation for determining if we are within the coordinate bounds and the center zoom required to be inside the coordiante bounds if we are not already.

Hi @rorystephenson, I really like the idea of unifying the various types of boundaries and performing boundary checks for rotation!

As far as unifying this with `fitCoordinates`, I think it's conceptually tricky. The idea behind `fitCoordinates` is not to define a "bounds" per se, but rather to set the center and zoom so that each of the individual coordinates is visible. In fact, as I'm thinking about this, `fitCoordinates`'s parameter should probably be `Set<LatLng>` rather than `List<LatLng>`, because the order of the coordinates or duplicates coordinates have no effect on the result.

Using a collection of coordinates as a map boundary feels like a different and broader problem, and it could mean a few different things:
- It could be the bounding rectangle of the coordinates
- It could be the [convex hull](https://en.wikipedia.org/wiki/Convex_hull) of the coordinates
- It could be an [alpha shape](https://gis.stackexchange.com/a/1231) of the coordinates
- It could be an arbitrary polygon defined by the _order_ of the coordinates 

Fitting coordinates doesn't have this type of ambiguity; given a set of coordinates, there's really only one way to fit them on the map.

So if we wanted to try to implement specific coordinate-based boundaries such as `ConvexHullBoundary` or `BoundingBoxBoundary` then there might be some opportunity to share logic internally with `fitCoordinates`. But I'm not sure it makes sense to create a single type like `MapCoordinateBounds` to try to unify the public API because fitting a set of coordinates and defining a map boundary based on a collection of coordinates are quite different things in my estimation.

JaffaKetchup:
(
> `fitCoordinates`'s parameter should probably be `Set<LatLng>` rather than `List<LatLng>`, because the order of the coordinates or duplicate coordinates have no effect on the result.

Whilst that would make logical sense, it's probably fine as it is. It's less work to import from an API or export to a storage medium. Open to argument though.

> Using a collection of coordinates as a map boundary feels like a different and broader problem, and it could mean a few different things

Of those options, I feel like bounding box and convex hull are enough for most users - and implementing arbitrary (potentially concave) polygon boundaries feels like it might be a lot of work (not saying a convex hull algorithm wouldn't be).
)

rorystephenson:
>I'm not sure it makes sense to create a single type like `MapCoordinateBounds` to try to unify the public API because fitting a set of coordinates and defining a map boundary based on a collection of coordinates are quite different things in my estimation.

Thanks for the reply @jjoelson ! After reading your comments I've settled on two separate abstractions, obviously still up for discussion, FrameFit and FrameConstraint (I took inspiration for the naming [from ArcGIS](https://pro.arcgis.com/en/pro-app/latest/help/layouts/customizing-your-map-extent.htm)).:

- `FrameFit` Places the map frame (the visible portion of the map) at a given location. This can be used to set the initial map position or to move the map to a new position. It replaces the old fit bounds behaviour.
- `FrameConstraint`: Does nothing (returns the provided map state) if the frame is already within the constraints, otherwise will either return null if it finds no suitable constraint or a new map state. This replaces maxBounds, adaptive bounds and ne/sw pan boundaries.

Let me know if you like the new abstractions.

---

Aside from that:
 - The FrameConstraint.contain(), which replaces the old maxBounds option, is fixed for rotation (@ibrierley). I updated the sliding map example to place markers in the corners so it is clear that we don't go outside of their bounds.
 - I've reinstated the removed MapController methods with deprecations (@JaffaKetchup).
 - MapOptions is now no longer part of FlutterMapState. I have added a MapOptions.of() class to look them up separately if needed. When the InheritedWidget is an InheritedModel this will allow rebuilding only when the options change if desired (or avoiding rebuilding when the options change but the FlutterMapState does not).

I still need to work on the InheritedModel change so this PR is not finished but I wanted to give an update on the coordinate fit changes. I won't work on merging PositionedTapDetector2 in to FlutterMapInteractiveViewer in this PR as there are enough changes anyway and it's something that can be done without breaking changes in a future PR.

jjoelson:
Bravo @rorystephenson, the `FrameFit` abstraction is great!

lib/src/misc/frame_fit.dart
```diff
@@ -0,0 +1,207 @@
...
+
+    return math.max(min, math.min(max, boundsZoom));
+  }
+}
+
+class FitCoordinates extends FrameFit {
+  final List<LatLng> coordinates;
+  final EdgeInsets padding;
+  final double maxZoom;
+  final bool inside;
```
jjoelson, line 127: I included this `inside` parameter in my `fitCoordinates` PR for parity with `fitBounds`, but upon further reflection I think that may have been a mistake on my part.

For `fitBounds`, so far as I can tell `inside: true` basically means "fit the map frame inside the bounds instead of fitting the bounds inside the map frame" (aside: it's possible my fix for rotation in https://github.com/fleaflet/flutter_map/pull/1550 didn't properly handle `inside: true`, so I will look into that further and maybe open a separate issue to discuss that).

For `fitCoordinates`, I migrated that implementation fairly directly but I now suspect the result is actually incoherent; as discussed above, a list of coordinates does not define a "bounds" in any unambiguous sense, so the idea of fitting the map frame _inside_ the coordinates doesn't really make sense. So I think I would advocate removing this parameter from `FitCoordinates`.

My apologies for causing confusion by including it in the first place!

rorystephenson: No problem! I've remove `inside` from FitCoordinates. If you figure out the rotation fix for FitBounds I'm happy to update it in this PR.

lib/src/map/map_controller.dart
rorystephenson: I have restored them as deprecated methods, what do you think? In my opinion accessing the state via a getter in a controller is an anti-pattern as it makes the control flow bi-directional (MapController <-> Map state/frame). It would be preferable to either:

-  If in a child of FlutterMap use FlutterMapState.of(context) (now MapFrame.of(context)) which will ensure the widget gets rebuilt when the frame changes.
- If not in a child widget, use a callback (we would have to add onFrameChanged) which the user can use with a Stream that they manage themselves. i.e. (onFrameChanged: (oldFrame, newFrame) => _frameStream.add(newFrame)).

That way the control flow is unidirectional which is easier to reason about. (MapController -> FlutterMap -> MapFrame -> children/listeners). At the same time there are examples of Flutter widgets that have bi-directional flow with their controller e.g. TextEditingController 🤷.  Anway I don't want to make this change (it's a huge breaking change for users) but I did want to reduce the number of methods which expose the state from the controller so that the MapFrame itself becomes the main API for accessing the state of the map.

If the general opinion is that it's better to leave the old methods without deprecations I'm not too worried about changing it back.

rorystephenson:
OK I've now marked this as ready for review. I've tested it on mobile web comparing it with the demo app (both running canvaskit) and I don't notice any performance/behaviour regressions.

In general I have tried to reduce our API as much as possible without removing useful methods. It's clear that many, if not most, of the methods in FlutterMapState were for internal use but became part of our API in v3, I think, due to refactoring. FlutterMapState no longer exists. It is replaced by an immutable MapFrame which only holds the minimal information necessary for displaying the map.

Let me know what your thoughts are. I've only added a couple of tests so far and I will add some more once there is a consensus on the changes.

example/lib/pages/interactive_test_page.dart
```diff
@@ -166,4 +164,49 @@ class _InteractiveTestPageState extends State<InteractiveTestPage> {
...
+      case MapEventRotateEnd():
+        return 'MapEventRotateEnd';
+      case MapEventNonRotatedSizeChange():
+        return 'MapEventNonRotatedSizeChange';
+      case null:
+        return 'null';
+      default:
+        return 'Unknown';
+    }
+  }
```
JaffaKetchup, line 211: Aware that this is within the example, but would it be better to have each `MapEvent` class define it's friendly name within itself, rather than this massive switch statement?

lib/src/misc/frame_constraint.dart
```diff
@@ -0,0 +1,102 @@
...
+
+import 'package:flutter_map/src/geo/latlng_bounds.dart';
+import 'package:flutter_map/src/map/flutter_map_frame.dart';
+import 'package:flutter_map/src/misc/point.dart';
+import 'package:latlong2/latlong.dart';
+
+abstract class FrameConstraint {
+  const FrameConstraint();
+
+  const factory FrameConstraint.unconstrained() = UnconstrainedFrame._;
```
JaffaKetchup, line 11: Would a user ever manually specify this?

JaffaKetchup (changes requested):
Maybe this is just me, but `Frame` seems to be confused in here.
It's replaced `FlutterMapState`, which would therefore suggest it is the map state, but is also the 'viewport'/'camera', which is only a part of the state.
If it is the state, why bother having `MapController`, and doesn't "`State`" more clearly show that it is state? Wasn't the point of the `MapController` to hide state from the user before - and now we are exposing it to all users? I guess I'm missing something here, but I feel like the number of things the user has to memorize and access just to programmatically handle the map has grown significantly.
As an example, if `Frame` refers to viewport/camera, `mapController.mapFrame.center` should return the screen point of the center of the camera, not the geographical coords, just as an example.
This does not apply to something like `fitFrame`, where I would consider it to essentially mean `fitFrameTo`.

Can you write an explanation of the new differentiated `MapController`/`Frame`(s), and anything similar, and why you would use one instead of the other?

---

Instead of `Frame`, maybe we should consider `Camera` - that seems to be what other libraries are using. Unless `Frame` means `State`.

---

I'm not a fan of the renaming of `center`/`zoom`/`rotation` to `initial_`. Whilst it may be more intuitive, it is now clearly stated within the documentation, and I feel like the migration cost for existing users would outweigh the clarity benefits for new users.

image

---

This needs more deprecations where possible. I understand they may now be difficult to retrofit, but for things like `interactiveFlags` and the other interactive flags, I think a grace period to migrate to `InteractiveOptions` would be nice, as well as all the changed map controller getters.

---

TLDR; I think the ideas are good and mostly well implemented, however, I'm really unclear on the whole new `Frame`/`Controller`/state interaction, and I'm not a fan of the many many breaking changes to **all** users that are included in this.

lib/src/gestures/flutter_map_interactive_viewer.dart
```diff
@@ -0,0 +1,856 @@
...
+  late double _scaleCorrector;
+  late double _lastRotation;
+  late double _lastScale;
+  late Offset _lastFocalLocal;
+  late LatLng _mapCenterStart;
+  late double _mapZoomStart;
+  late Offset _focalStartLocal;
+  late LatLng _focalStartLatLng;
+
+  late final AnimationController _flingController;
```
TesteurManiak, line 65: Prefer assigning the controller directly on the declaration line to make it more readable (listeners can be kept inside the `initState`):
```dart
late final AnimationController _flingController =  AnimationController(vsync: this);
```

lib/src/misc/frame_fit.dart
```diff
@@ -0,0 +1,204 @@
+import 'dart:math' as math;
+
+import 'package:flutter/widgets.dart';
+import 'package:flutter_map/src/geo/latlng_bounds.dart';
+import 'package:flutter_map/src/map/flutter_map_frame.dart';
+import 'package:flutter_map/src/misc/point.dart';
+import 'package:flutter_map/src/misc/private/bounds.dart';
+import 'package:latlong2/latlong.dart';
+
+abstract class FrameFit {
```
TesteurManiak, line 10: Can't this be converted to a `sealed` class ?

TesteurManiak (commented):
I'll have to give a second read to this as there are a lot of changes, but here's my first observations:
* Some minor code style recommendations on the declaration of `late` variables (note that I'm also not a big fan of the use of `var`, I prefer to explicitly indicate the variable type but it really is a matter of preference).
* @JaffaKetchup following what we talked about on discord I agree that changing some parameters such as `center` to `initialCenter` or `zoom` to `initialZoom` can be seen as unneeded but as long as the old name remains with the `Deprecated` annotation it seems ok to me 🤔

example/lib/pages/interactive_test_page.dart
rorystephenson: Yeah I hesitated on that too, I ended up putting it in the example because I couldn't imagine why a user would want to use friendly names for events. If they need to match on an event type they can use a case statement. Do you think there is a use for them?

lib/src/misc/frame_constraint.dart
rorystephenson: Not that I can think of. I created it as the default value to avoid needing to write special-case code that handles nulls.

lib/src/misc/frame_fit.dart
rorystephenson: Yeah you're probably right. I mean a user could implement their own FrameFit if they wanted to but is that really likely?

JaffaKetchup (changes requested):
I think a lot of what you said in response was just what I (imagined I? 😀) said in the first place.

> I wouldn't expect a map frame/camera to return me a screen point if I asked it for center (especially because unless the map moves or changes size on the screen the center value would never change). Maybe that's just me 🤷 .

I agree with you, ~which is why I would've thought `mapController.mapFrame.center` should've been something like `mapController.map.center` instead. You're not getting the center of the viewport, that would be useless, as you said, you're getting the center of the viewport projected onto the map. But anyway, I think it's alright and I don't think `map.center` is much better anyway, because that would always be (0, 0).~ So just ignore me!

That explanation helps somewhat. I'm ~not~ looking forward to writing the docs and migration info for this 😂.  
I also hadn't realised you'd added some deprecations before I posted that review.

----

I can't see any regressions in the example app, which is good news.

So here's _my_ remaining list of things to do I think. Happy to help with this, especially the last point (as that's just brute force really), but can't until you're sure you're not going to make any incompatible changes whilst I'm working on it (eg. renaming `Frame` to `Camera`). Also, please try to avoid so many force pushes, it makes it difficult to track commits and collaborate!

* Rename `Frame` to `Camera`

* Rename filenames under 'map/' to not include 'flutter_map_'  
I think it just makes them too long and it's kinda redundant.

* Add more internal documentation  
Every new class should have documentation, and every major member should be documented.

* Ensure deprecations follow 'standard' format  
I think the deprecations you've added are for the right methods, but they should be in a slightly different format (being very picky here), which is very similar to [what Flutter uses](https://github.com/flutter/flutter/wiki/Tree-hygiene#deprecations):  
    ```dart
    @Deprecated(
      'Prefer `<replacement>` instead. <?additional short migration info>'
      'This <member type, eg. method/getter> has been changed because <brief reasoning info> . '
      'This feature is deprecated since v6.',
    )
    ``` 

---

At a minimum, it's looking like this will be waiting for another week, as @ibrierley can't review until next weekend at the minimum. Thanks for spending so much time on this recently!

I'm going to construct a project board or some sort of tracking for v6. If we're including a change as big as this, we might as well include some other big stuff.

jjoelson:
Hi @JaffaKetchup and @rorystephenson! I just want to provide an update regarding my comment [here](https://github.com/fleaflet/flutter_map/pull/1551#discussion_r1229746904):
> it's possible my fix for rotation in https://github.com/fleaflet/flutter_map/pull/1550 didn't properly handle inside: true, so I will look into that further and maybe open a separate issue to discuss that

I have the basic fix working but I still need a bit of time to finalize the handling of padding and adding tests. 

It turns out that once we take rotation into account, the fit algorithms for `inside: false` and `inside: true` are quite distinct. This got me thinking that from the perspective a library _user_, the two fits are also quite distinct in their behavior. I don't know the history, but I have a feeling that `inside` only ended up as a parameter on `fitBounds` because the implementations were nearly identical so long as rotation was ignored. 

So since we're introducing this `FrameFit` abstraction with this refactor, I think we should consider removing `inside` from `FitBounds` and instead have a separate `FitInsideBounds` frame fit. What do you two think of that approach?

lib/src/misc/frame_fit.dart
JaffaKetchup: Continued at https://github.com/fleaflet/flutter_map/pull/1551#issuecomment-1597631016 and after...

JaffaKetchup:
Hey @jjoelson, thanks for popping back in!

Agreed. From your explanation, setting `inside` causes a completely different behaviour when rotating, so should be represented by a completely separate object.

We've decided on `Camera` instead of `Frame`. So maybe more appropriate class names could be `CameraFitAround` & `CameraFitWithin`, both of which implement `CameraFit`. Then the user can select either one using constructors on `CameraFit` (`.around`, `.within`) instead of referencing the class names themselves.

rorystephenson:
OK @jjoelson no problem, when it's ready let me know if you want a hand integrating the fixes in to this PR.

@JaffaKetchup thanks for the feedback, I think I've addressed everything!

JaffaKetchup:
Thanks @rorystephenson, I'll have another look through in the next few days, and experiment with migrating my plugin, and we'll see what comes out of that. I don't think there'll be anything else major though!

JaffaKetchup (commented):
Just noticed that https://github.com/fleaflet/flutter_map/pull/1551#issuecomment-1597695109 hasn't yet been added. Not sure if you realized?

rorystephenson:
I was waiting for @jjoelson's changes for the inside behaviour. Once that's ready I can incorporate that in to these changes or, if you prefer @jjoelson , you can make a PR on this branch. Either works for me!

I agree that separating inside from normal fit bounds is a good idea.

jjoelson:
Hi @JaffaKetchup and @rorystephenson! I put up a draft PR to this branch with my progress so far: https://github.com/rorystephenson/flutter_map/pull/1

As I'm working on this, I have a question about `CameraFit` that I hope we might discuss: should options like `padding`, `maxZoom`, and `forceIntegerZoomLevel` be properties on the individual fit classes rather than the parent class? By having them on the parent class, we are saying they are universal options that will apply to all fit algorithms. Given the breadth of possible fitting algorithms, how certain are we that they are truly universal?

The particular scenario that I'm thinking about right now is that `maxZoom` doesn't seem to make sense with respect to `FitInsideBounds`; if anything, it would make sense for this fit algorithm to have a `minZoom` option to be analogous to how `FitBounds` uses `maxZoom`.

We probably have to continue to support `maxZoom` on `FitInsideBounds` for backwards compatibility in this case, but this type of thing could arise in the future with other fit types.

Robbendebiene:
@jjoelson I'm not closely following your PR, so maybe I'm missing something, my appologize for that.

> The particular scenario that I'm thinking about right now is that maxZoom doesn't seem to make sense with respect to FitInsideBounds; if anything, it would make sense for this fit algorithm to have a minZoom option to be analogous to how FitBounds uses maxZoom.

Why do you think that `maxZoom` doesn't make sense in this case? To me it translates to "Please try to fit the camera into this box (inside or outside depending on the definition) BUT if the calculated zoom value that is required to fit it properly exceeds `maxZoom` or undercuts `minZoom` (see my PR https://github.com/fleaflet/flutter_map/pull/1562) then use the respective zoom limit instead to calculate the center point.

JaffaKetchup:
@jjoelson Just to note, I moved the properties into the parent class in dabf734fe7442396762894967fb9409fe7dbfd0e.

rorystephenson:
Any options that don't conceptually make sense for any imaginable CameraFit should stay in the subclasses. This will avoid future refactoring/breaking changes when introducing new CameraFits and will make it easier to change individual fits without needing to update the other ones. Same with CameraConstraints (and subclasses in general).

JaffaKetchup:
Feel free to rearrange!

JaffaKetchup (changes requested):
- `options.maxBounds` needs a deprecation to `CameraConstraint.contain`
- `MapCamera` should be exposed through the standard import library, not the plugin API
  - or, the documentation on `MapController.camera` needs updating to remove the suggestion of this method of accessing the camera as being an anti-pattern

ibrierley:
Just had a pre-review of this, think I'm fine with it all, feels coherent and all tests well for me.

jjoelson:
> @jjoelson I'm not closely following your PR, so maybe I'm missing something, my apologies for that.
> 
> > The particular scenario that I'm thinking about right now is that maxZoom doesn't seem to make sense with respect to FitInsideBounds; if anything, it would make sense for this fit algorithm to have a minZoom option to be analogous to how FitBounds uses maxZoom.
> 
> Why do you think that `maxZoom` doesn't make sense in this case? To me it translates to "Please try to fit the camera into this box (inside or outside depending on the definition) BUT if the calculated zoom value that is required to fit it properly exceeds `maxZoom` or undercuts `minZoom` (see my PR #1562) then use the respective zoom limit instead to calculate the center point.

Hmm, you might be right after thinking about it some more. My reasoning was that it's a bit of a self-defeating parameter. If you're fitting the camera inside the bounds then `maxZoom` can never come into play unless it's _preventing_ the camera from fitting inside the bounds. And similarly, if you're fitting bounds inside the camera then `minZoom` can never come into play unless it's _preventing_ the bounds from fitting inside the camera.

But you're correct that some people might actually have a need for that behavior, and in fact it's already the case that the camera's own min and max zoom can prevent the fit from succeeding. So I think you're correct that `minZoom` and `maxZoom` could apply to all fits.

Robbendebiene:
@jjoelson Okay I see where you are coming from 🙂 At the end it is all about making a decision of what should win in such a scenario, the fitting or the defined limits.

> That said, the default maxZoom value of 18 may make for confusing behavior when trying to fit inside bounds if it prevents the fit from succeeding. Maybe we can keep the properties on the parent class but have each subclass determine default values?

Sounds good, even though raising the default maxZoom in the base class would also work for me. It probably would even be possible to set it to `double.infinity` or `double.maxFinite` (and `minZoom` to 0) if one wants to completely disable any limits.

rorystephenson:
> * `options.maxBounds` needs a deprecation to `CameraConstraint.contain`
> * `MapCamera` should be exposed through the standard import library, not the plugin API

Done 👍 

---

I've also merged in @jjoelson's changes to this PR!

Finally I've tidied up some exports/imports and added a lint to keep them alphabetically ordered (99% were already).

jjoelson:
I think where we landed is that it's OK for `maxZoom` to be an option for all fits, but since applying a `maxZoom` can cause certain fits (like `FitInsideBounds`) to fail, then its default value should be  a "no-op" value like `double.infinity` so people don't experience confusing default behavior. Please correct me if I'm wrong @Robbendebiene.

@JaffaKetchup Given that the old `fitBounds` and `FitBoundsOptions` continue to exist with their existing default values for backwards compatibility, would it be OK if all parameters for the new `CameraFit` abstraction have "no-op" values by default? I.e.:
```
padding = EdgeInsets.zero,
maxZoom = double.infinity,
forceIntegerZoomLevel = false,
```
I think that would be the best way to avoid unexpected behavior going forward, especially if we imagine that new fit types will be added in the future.

JaffaKetchup:
@jjoelson Sorry for the slow response, I think having those defaults make sense.

rorystephenson:
Your proposed defaults sound good @jjoelson , do you want to open a PR or shall I make the changes?

rorystephenson:
@JaffaKetchup I've just been upgrading some packages to use this branch:

 - flutter_map_marker_popup
 - flutter_map_supercluster
 - flutter_map_animations

Those deprecation changes were a result of that. Otherwise no other changes were needed for these plugins to work.

So as far as I can see we are just missing the changed CameraFit defaults.

rorystephenson:
@jjoelson and @JaffaKetchup I've just added a commit which updates the maxZoom default (the rest were already as suggested). I opted for null rather than double.infinity for the default value as I think that the intent is clearer; no maximum. It also makes it more explicit in the calculations.

Let me know what you think. If you're both happy with I think we're ready to review and release a dev version for more testing.

lib/src/misc/frame_fit.dart
rorystephenson: @TesteurManiak On further consideration I don't think this should be a sealed class. The interface of the class is safe in the sense that it can only calculate a non-nullable MapCamera result. If the user wishes to implement their own CameraFit it shouldn't cause any issues.

I will resolve this for now but if you disagree let me know!

TesteurManiak: No problem, I'm okay with it staying an `abstract class` 👍

JaffaKetchup (approved):
LGTM for a preview release, although if bugs are found, a new PR will be needed, @rorystephenson if you're up for that.

Will try to merge the other remaining PRs before publishing the preview, and whilst those are being updated, I'm going to make a couple of changes in a PR as well. Just for reference, here's the v6 plan: https://github.com/fleaflet/flutter_map/projects/3.

Robbendebiene:
Not sure if this is intentional or if I'm doing something wrong, but it seems like I can no longer access/import `FlutterMapState`?

PR #1556, opened 2023-06-16
Suppress Warning

EliaTolin:
The 'children' argument should be last in widget constructor invocations. Try moving the argument to the end of the argument  Warning : dart sort_child_properties_last

Happy clean coding! 💯


JaffaKetchup:
Thanks @EliaTolin, I've made this change manually :)

Issue #1552, opened 2023-06-10
[FEATURE] Re-instate Anchor on Marker


rorystephenson:
### What do you want implemented?

I'm opening this issue to continue a discussion I started here: https://github.com/fleaflet/flutter_map/pull/1532#pullrequestreview-1466723876

In flutter_map 5.0.0 Marker's Anchor was replaced with AnchorPos. This was done for consistency with the newly added option to define AnchorPos at the MarkerLayer level. There are, however, two drawbacks to this approach:

1. The Anchor is calculated every time the Marker is built, previously it was pre-calculated when creating the Marker instance.
2. The default AnchorPos for a Marker is ambiguous for other packages that use Marker. The AnchorPos is nullable and null by default which makes sense to allow for the fallback to the MarkerLayer anchorPos. Unfortunately this means that other packages that use Marker need to check the source of MarkerLayer to determine what the default AnchorPos is and if that ever changes there will be inconsistencies.

### What other alternatives are available?

1. Re-instate the old behaviour. The pros are less calculations when building markers and the default AnchorPos is clear. The con is that it's no longer possible to define a default anchorPos at the MarkerLayer level which is clearly convenient given that for apps that don't have a huge number of Markers the performance impact of doing so will be zero.
2. Change Marker's AnchorPos? to Anchor? (previously it was non-nullable) and define a AnchorPos.default (which just redirects to the default value). The pros are that users who are concerned about the performance impact can define the anchor at the Marker level, and the default AnchorPos is not as clear as a default parameter but it's clear enough. The cons are that Marker.anchor will be inconsistent with MarkerLayer.anchorPos and it may not be obvious that defining at the MarkerLayer level can increase calculations. I think that both of these can be mitigated by good documentation. 

### Can you provide any other information?

@JaffaKetchup I've elaborated on a couple of options. Let me know what you think.

### Severity

Annoying: Currently have to use workarounds

JaffaKetchup:
I think the best option out of those is the first one. The second one is too complex for users to understand.

I don't think the re-calculation will be much of an issue, but I can imagine your second point about  ambiguity being potentially annoying.

However, I also wonder whether something like this would be OK? Unless I'm completely missing something, this wouldn't require a new calculation for every marker build, and it solves the second point. 

```dart
// In constructor
    AnchorPos? anchor,
}) : _anchor = whatever(anchor);

// In class
final Anchor? _anchor;
Anchor get anchor => _anchor ?? layerAnchor;
```

[...]

PR #1558, opened 2023-06-19
Improve `Marker.anchor` performance

rorystephenson:
This addresses #1552, it re-instates the precalculated Anchor on a Marker whilst leaving the argument as an anchorPos which makes it consistent with the MarkerLayer options.


JaffaKetchup (approved):
LGTM. I think this behaves correctly.

PR #1560, opened 2023-06-19
Added `hasGesture: true` to `onPointerSignal`'s `move` call

### LINKED ISSUE #1465: [BUG] `hasGesture` is `false` after double tap zoom in `onPositionChanged` callback
### What is the bug?

The onPositionChanged callback delivers _false_ for `hasGesture` when zooming, especially double tapping to zoom. Normally, when the user zooms with two fingers, he also moves the map slightly, so hasGesture will eventually be true. But if you zoom with mouse and keyboard or zoom via double tap, `hasGesture` is just false. 

This makes it hard to disguish between zooms made by code and zooms made by human. 

hasGesture == true, when onPositionChanged callback is called from a zoom. 

```dart
FlutterMap(
 options: MapOptions(
 onPositionChanged: (
    MapPosition posChange,
    bool hasGesture,
 ) {
   print(hasGesture);
}
```

_No response_

I was just able to test it on Android, but I expect it to happen on all platforms.

### Platforms Affected

Android

### Severity

### Frequency

### Requirements

JaffaKetchup:
Should fix #1465?

### DIFF
--- a/lib/src/gestures/gestures.dart
+++ b/lib/src/gestures/gestures.dart
@@ -72,8 +72,12 @@ abstract class MapGestureMixin extends State<FlutterMap>
             _offsetToPoint(pointerSignal.localPosition), newZoom);
 
         // Move to new center and zoom level
-        mapState.move(newCenterZoom[0] as LatLng, newCenterZoom[1] as double,
-            source: MapEventSource.scrollWheel);
+        mapState.move(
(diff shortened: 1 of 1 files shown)


JaffaKetchup:
Actually, I'm not sure this will fix it. This line is concerned with the scroll wheel, not a double tap gesture.

PR #1584, opened 2023-07-13
Add test to ensure gestures always work

rorystephenson:
Recently a change was made to move the TileLayer backgroundColor option to become an option of FlutterMap. This accidentally fixed an issue where gestures would stop working when going beyond a TileLayer's max zoom because FlutterMap's gesture detectors' HitTestBehaviour is the default deferToChild and when above a TileLayer's maxZoom potentially there would be no hit-testable children of FlutterMap.

Moving the color option to FlutterMap means that there is always a ColoredBox child widget which is hit testable. To avoid accidentally breaking hit testing again in the future this commit adds a test which ensures it always works.

In passing I made TileLayer only wrap its tiles in a ColoredBox widget if the user defines a backgroundColor (which is deprecated).


JaffaKetchup (approved):
> > accidentally fixed an issue
>
> 😂 I think you mean 'intentionally' 😉

LGTM, thanks for the contribution. More tests are always appriciated!

Issue #1522, opened 2023-05-12
[BUG] CustomPoint has potential cast exception at runtime


josxha:
### What is the bug?

While working on https://github.com/greensopinion/flutter-vector-map-tiles/pull/149 i discovered that the current CustomPoint implementation lead to a cast exception at runtime.

This behaviour is documented in the parent class math.Point but isn't mentioned in CustomPoint:
>   /// **Important Note**: This function accepts a `num` as its argument only so
  /// that you can scale `Point<double>` objects by an `int` factor. Because the
  /// `*` operator always returns the same type of `Point` as it is called on,
  /// passing in a double [factor] on a `Point<int>` _causes_ _a_
  /// _runtime_ _error_.

### How can we reproduce it?

```dart
final pDouble = CustomPoint<double>(1, 2);
final pInt = CustomPoint<int>(3,4);
var point;

point = pInt + pDouble; // runtime exception
point = pInt + pDouble; // runtime exception
point = pInt * pDouble; // runtime exception
point = pInt / pDouble; // runtime exception
```

Potentially this would throw an exception too:
```dart
point = CustomPoint<int, int>(5, 5) / CustomPoint<int, int>(2, 2); // runtime exception
```

### Do you have a potential solution?

The issue is that `other` may be a num or double and turns the result into a double while the function casts it to an int.

For example here:
```dart
  /// Create new [CustomPoint] where [x] and [y] values are added to [other]
  /// point [x] and [y] values
  @override
  CustomPoint<T> operator +(math.Point other) {
    return CustomPoint<T>((x + other.x) as T, (y + other.y) as T);
  }
```

Possible solutions:
- Add documentation to the dangerous functions like it's done in math.Point. This could lead to potential errors for users but might not be that problematic since CustomPoint is mostly used internally or by plugin developers.
- force `other` to be the same type as the CustomPoint instance. This requires transforming every working double/int compination to be converted to a double/double.
- Use named functions like: 
```dart
CustomPoint<T> multiplyWithInt(CustomPoint<int> other) {} // int/int, double/int
CustomPoint<double> multiplyWithDouble(CustomPoint<double> other) {} // double/double, int/double
```

### Platforms

tested on android but should 

### Severity

Erroneous: Prevents normal functioning and causes errors in the console

josxha:
The flutter_map implementation had this issue at some point too and used a workaround:
```dart
final tileZoomDouble = tileZoom.toDouble();
[...]

PR #1585, opened 2023-07-14
Replace CustomPoint with extension methods on Point

rorystephenson:
The CustomPoint class predates dart extension methods. The name is confusing (custom in what sense, and why) and in looking at the implementation it became clear that it was essentially just adding extension methods.

Changing to a Point will avoid conversions in plugins which deal with libraries that use Point.

Code will work as-is without changes with the exception of the subtraction operator. Previously CustomPoint<int> - CustomPoint<double> did not trigger analyzer warnings but would cause a runtime exception. This is now replaced by subtract() extension methods which allow subtracting num from a Point<double> and int from Point<int>.

---

Issue #1522 prompted me to look at this and I believe it resolves that issue (we no longer add methods which are not runtime-safe to Point and Point already documents the potential runtime error with the * operator).

Note that the vast majority of changes are just renaming CustomPoint to Point and importing it.


lib/src/geo/crs.dart
```diff
@@ -1,6 +1,6 @@
 import 'dart:math' as math;
+import 'dart:math';
```
JaffaKetchup, line 2: Use only one import for the same library.

lib/src/layer/tile_layer/tile_layer.dart
```diff
@@ -1,5 +1,6 @@
 import 'dart:async';
 import 'dart:math' as math;
+import 'dart:math';
```
JaffaKetchup, line 3: Use only one import for the same library.

lib/src/layer/tile_layer/tile_range.dart
```diff
@@ -1,7 +1,8 @@
 import 'dart:math' as math;
+import 'dart:math';
```
JaffaKetchup, line 2: Use only one import for the same library.

lib/src/map/camera/camera.dart
```diff
@@ -1,11 +1,12 @@
 import 'dart:math' as math;
+import 'dart:math';
```
JaffaKetchup, line 2: Use only one import for the same library.

lib/src/map/camera/camera_constraint.dart
```diff
@@ -1,9 +1,10 @@
 import 'dart:math' as math;
+import 'dart:math';
```
JaffaKetchup, line 2: Use only one import for the same library.

lib/src/map/camera/camera_fit.dart
```diff
@@ -1,10 +1,11 @@
 import 'dart:math' as math;
+import 'dart:math';
```
JaffaKetchup, line 2: Use only one import for the same library.

lib/src/misc/point_extensions.dart
```diff
@@ -0,0 +1,128 @@
...
+  /// values in [other].
+  Point<int> scaleBy(Point<int> other) {
+    return Point<int>(x * other.x, y * other.y);
+  }
+}
+
+extension OffsetToPointExtension on Offset {
+  @Deprecated(
+    'Prefer `toPoint()`. '
+    'This method has been renamed as a result of CustomPoint\'s removal. '
```
JaffaKetchup, line 121: Just for consistency, prefer double quotes with single inner quotes over single quotes with escaped inner quotes.

```suggestion
    "This method has been renamed as a result of CustomPoint's removal. "
```

lib/src/misc/private/bounds.dart
```diff
@@ -1,22 +1,21 @@
 import 'dart:math' as math;
-
-import 'package:flutter_map/src/misc/point.dart';
+import 'dart:math';
```
JaffaKetchup, line 2: Use only one import for the same library.

JaffaKetchup (changes requested):
LGTM, apart from these few nitpicks for consistency and to keep things concise.

lib/src/geo/crs.dart
rorystephenson: Nice catch 👍

rorystephenson:
Nice catch on the redundant imports @JaffaKetchup !

First I started by removing the `import 'dart:math' as math;` imports but that didn't work well with some classes which had `min`/`max` getters in classes in some of the tile related code. So then I started removing the new `import 'dart:math';` imports but this meant that `Point` was always prefixed with `math` which makes quite a mess. I think importing math with a prefix makes sense for methods like `min`/`max` and getters like `pi` but in the case of `Point` it is pretty unambiguous.

For that reason I've updated the imports to:

```dart
import 'dart:math' as math hide Point;
import 'dart:math' show Point;
```

WDYT?

JaffaKetchup:
It's less than ideal, I'm sure you'll agree, but OK for now!

mootw (approved):
lgtm

PR #1590, opened 2023-07-17
add CustomPoint.toOffset

tlserver:
Just adding `CustomPoint.toOffset()` for consistency with OffsetToCustomPointExtension. With PR #1585, this change need to be moved to point_extension.dart


JaffaKetchup:
Hey @tlserver, thanks for the contribution. Just to avoid breaking changes - @rorystephenson can you add this change manually to #1585 and add @tlserver as co-author for attribution?

rorystephenson:
This is now incorporated in #1585 :+1:.

PR #1591, opened 2023-07-17
Ignore gestures on closed `RichAnimationWidget` when using `FadeRAWA`

### LINKED ISSUE #1589: [BUG] `FadeRAWA` allows invisible widgets to receive gesture events
### What is the bug?

On iOS (only tested on iOS) - Using the following code the "OpenStreetMap contributors" text is still tappable even through it does not show up on the screen. If a user is to tab in the general location of where the text would be the link opens

### How can we reproduce it?

Using the following on a FlutterMap
```
                nonRotatedChildren: [
                  RichAttributionWidget(
                    attributions: [
                      TextSourceAttribution(
                        'OpenStreetMap contributors',
                        onTap: () => launchUrl(
                          Uri.parse('https://openstreetmap.org/copyright'),
                        ),
                      ),
                    ],
                  ),
                ],
```

Tap the "info" button to see where the text would be, close the widget, and tab in the region where the text would be if the widget was open.. This will cause the link to be launched

Haven't looked deep into how the widget is shown/hidden but maybe it's done through opacity then the items would still interactive even though they are not seen

### Platforms

iOS 16

### Severity

JaffaKetchup:
Fixes #1589.
Added `IgnorePointer` to `FadeRAWA.popupAnimationBuilder` to ignore gestures when closed. `ScaleRAWA` does not suffer from the same issue as the child becomes 0x0 px.

### DIFF
--- a/lib/src/layer/attribution_layer/animation.dart
+++ b/lib/src/layer/attribution_layer/animation.dart
@@ -179,6 +179,9 @@ class FadeRAWA implements RichAttributionWidgetAnimation {
         opacity: isExpanded ? 1 : 0,
         curve: isExpanded ? popupCurveOut : popupCurveIn,
         duration: buttonDuration,
-        child: child,
+        child: IgnorePointer(
+          ignoring: !isExpanded,
+          child: child,
+        ),
       );
 }


TesteurManiak (approved):
LGTM! 👍

PR #1592, opened 2023-07-18
Support desktop (click and drag) rotation

### LINKED ISSUE #1568: [FEATURE] Implement rotation functionality on mouse/cursor based platforms

Have you considered the function of rotating the map by right-clicking and long-pressing the map?

Have you considered the function of rotating the map by right-clicking and long-pressing the map?

Have you considered the function of rotating the map by right-clicking and long-pressing the map?

### Severity

JaffaKetchup:
Resolves #1568.

Thanks to @TesteurManiak for the infrastructure implementation and POC and @ibrierley for the rotation algorithm itself. Also thanks to https://stackoverflow.com/questions/48916517/javascript-click-and-drag-to-rotate and https://github.com/openlayers/openlayers/blob/7099f14bd2348491c1550f7b0be929e5da00d83d/src/ol/interaction/DragRotate.js#L65.

Still to do before this can be release ready:

- [ ] Ensure this does not break other existing gestures
- [x] Allow for other keyboard triggers
- ~Add `InteractiveFlag.cursorRotate`~ (redundant with 'Allow for other keyboard triggers')
- ~Consider renaming `InteractiveFlag.rotate` for clarity~ (`InteractiveFlag.cursorRotate` will not be added)
- ~Allow for customizing rotation degree scale (1 by default)~ (does not work correctly and can cause rotation jumps)

Also moved `InteractionOptions.enableScrollWheel` to `InteractiveFlags.scrollWheelZoom` with deprecation.


lib/src/gestures/flutter_map_interactive_viewer.dart
```diff
@@ -366,6 +414,14 @@ class FlutterMapInteractiveViewerState
     }
   }
 
+  // Thanks to https://stackoverflow.com/questions/48916517/javascript-click-and-drag-to-rotate
+  double getCursorRotationDegrees(Offset offset, BuildContext context) =>
+      round((math.atan2(offset.dx - MediaQuery.sizeOf(context).width / 2,
+                  offset.dy - MediaQuery.sizeOf(context).height / 2) *
+              (180 / pi) *
+              -1) +
+          100);
```
TesteurManiak, line 423: Maybe this would be more readable:
```dart
double getCursorRotationDegrees(Offset offset, BuildContext context) {
  final size = MediaQuery.sizeOf(context);
  return round((math.atan2(offset.dx - size.width / 2, offset.dy - size.height / 2) * (180 / math.pi) * -1) + 100);
}
```
Also I cannot seem to find `round(/* ... */)` method, usually you use the [`.round()`](https://api.dart.dev/stable/dart-core/double/round.html) and `(180 / pi)` shouldn't work as the import to `dart:math` is made with `as math` does this code compile for you? 🤔

TesteurManiak (changes requested):

lib/src/gestures/flutter_map_interactive_viewer.dart
JaffaKetchup: Here's a revised function:

```dart
// Thanks to https://stackoverflow.com/questions/48916517/javascript-click-and-drag-to-rotate
double getCursorRotationDegrees(Offset offset) {
  final size = MediaQuery.sizeOf(context);
  return -math.atan2(
          offset.dx - size.width / 2, offset.dy - size.height / 2) *
      (180 / math.pi);
}
```

It looks like `round` and `pi` came from 'latlong2'. I've switched to using `math.pi`.

Asking for the `context` as an argument turned out to be unnecessary.

I've also removed the rounding function, as I can't see a visible effect on the result. @ibrierley Can you confirm if you included the rounding for a reason?

I've also removed the `+ 100`, as that seems to be a correction term of sorts.
 - +180 or -180: map North when cursor top
 - +100: map North when cursor right
 - -100: map North when cursor left
 - 0: map North when cursor bottom

Not sure why the values are quite like that, or which has the best feeling? I guess +/- 180 kind of feels OK, as the cursor aligns with the North of the map. Open to your opinions!

ibrierley:
I only included the round function as that was in the original solution I was looking at. I never had chance to do any detailed digging into whether it causes any issues without it. I did ponder it at the time and couldn't really figure why it was necessary.

lib/src/gestures/flutter_map_interactive_viewer.dart
JaffaKetchup: See 0a52a93 for resolution. I've added a local scope `correctionTerm` constant for easy identification, and set North = cursor (180).

TesteurManiak (changes requested):
Just got some very minor remarks, otherwise LGTM! 👍

lib/src/map/options.dart
```diff
@@ -398,12 +399,25 @@ final class InteractionOptions {
...
 
+  /// Whether to allow rotation by moving the cursor dependent on the currently
+  /// pressed keyboard [LogicalKeyboardKey]
+  ///
+  /// Fix to returning `false` to disable cursor/keyboard rotation.
+  ///
+  /// Defaults to allowing rotation by cursor if any of the Control keys are
+  /// pressed.
+  final bool Function(LogicalKeyboardKey key)?
+      isCursorRotationKeyboardKeyTrigger;
```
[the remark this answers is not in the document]
JaffaKetchup: Resolved in 70bffdb.

androidseb:
Thank you very much for adding in this feature! I've tested it just now on the demo page here:
https://demo.fleaflet.dev/

For me it works mostly fine, but I noticed the rotation jumps around a little bit in unexpected ways (maybe I don't have the correct expectations?)...

Here are some reproduction steps of the issue in case this helps:

* Load the demo page here in the latest version of chrome (Version 116.0.5845.140 (Official Build) (64-bit)): https://demo.fleaflet.dev/
* Hold the "Ctrl" keyboard key
* With the left mouse button, drag the map from the top left corner down by just a few pixels
* UX issue:
    * EXPECTED: the map rotates by just a few degrees
    * ACTUAL: the map is rotated instantly by 45 degrees, then by a few additional degrees

It appears wherever the mouse is dragged to relative to the center of the map view becomes the new north of the map. @JaffaKetchup Is this how it's expected to work? I would have expected the map rotation to allow us to "Drag" the map for rotation by a few degrees at once, not set the new north.

JaffaKetchup:
Hi @androidseb,
Indeed, this is the way it's supposed to function at the moment, but I do agree it might be better if we could just apply an offset to the current rotation, instead of setting the rotation outright. I'll see what I can do, but if I can remember correctly, trying to implement it like this caused issues.

PR #1598, opened 2023-07-25
Fix some polygon edge cases, avoid unecessary work, and tidy up the code

ignatz:
Fix some polygon edge cases, e.g. double drawing on labels or not correctly setting the border color.

I also feel that the code re-organization really helped the separation of concerns and made it much clearer what each part is doing. Specifically:

* Fill and border are clearly separated.
* The loop is only responsible for batching paths. The Paint is now handled once during drawing.

image

### DIFF
--- a/example/lib/pages/polygon.dart
+++ b/example/lib/pages/polygon.dart
@@ -98,7 +98,7 @@ class PolygonPage extends StatelessWidget {
                       points: filledPoints,
                       isFilled: true,
                       color: Colors.purple,
-                      borderColor: Colors.purple,
+                      borderColor: Colors.yellow,
                       borderStrokeWidth: 4,
                     ),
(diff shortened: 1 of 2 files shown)


example/lib/pages/polygon.dart
```diff
@@ -132,12 +134,28 @@ class PolygonPage extends StatelessWidget {
                     ),
                     Polygon(
                       points: holeOuterPoints,
-                      //holePointsList: [],
+                      isFilled: true,
```
ignatz, line 137: FYI, this is one of the edge cases, where previously it would fill even though isFilled is false by default. Arguably it should have not been filled before.

JaffaKetchup (approved):
LGTM, thanks for the improvements!

PR #1599, opened 2023-07-25
Fix polygon draw batching

ignatz:
This PR addresses a pre-existing condition and is independent from the cleanup. The cleanup is more a result of my debugging. The commits could be split.

### DIFF
--- a/lib/src/layer/polygon_layer.dart
+++ b/lib/src/layer/polygon_layer.dart
@@ -11,6 +11,17 @@ enum PolygonLabelPlacement {
   polylabel,
 }
 
+bool isClockwise(List<LatLng> points) {
+  double sum = 0;
+  for (int i = 0; i < points.length; ++i) {
+    final a = points[i];
+    final b = points[(i + 1) % points.length];
+
+    sum += (b.longitude - a.longitude) * (b.latitude + a.latitude);
+  }
+  return sum >= 0;
+}
+
 class Polygon {
   final List<LatLng> points;
   final List<List<LatLng>>? holePointsList;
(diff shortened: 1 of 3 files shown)


lib/src/layer/polygon_layer.dart
```diff
@@ -27,6 +38,10 @@ class Polygon {
   final TextStyle labelStyle;
   final PolygonLabelPlacement labelPlacement;
   final bool rotateLabel;
+  // Designates whether the given polygon points follow a clock or anti-clockwise direction.
+  // This is respected during draw call batching for filled polygons. Otherwise, batched polygons
+  // of opposing clock-directions cut holes into each other leading to a leaky optimization.
+  final bool filledAndClockwise;
```
JaffaKetchup, line 44: This is extra responsibility for the user to get right. Can you give a little more detail, or perhaps think of a way around needing this?

TesteurManiak: I'm not sure that it adds responsibility to the user considering that the value is set by the [constructor](https://github.com/fleaflet/flutter_map/pull/1599/files#diff-963bb1a45c292ea8c2c6558adae2bf75441edeb0dec681bf4235dac4b2bcf026R67) but maybe it should be in a private variable 🤔

JaffaKetchup: Ah, sorry missed that. Agreed, in that case it should be private.

ignatz: Making it private sounds good... done. Note that this change is explicitly intended to require less responsibility from users, i.e. you don't have to get your normals right to not be screwed by the batching details

JaffaKetchup (approved):
LGTM, thanks for the improvements!

PR #1606, opened 2023-08-04
Add a polygon-layer-level switch to toggle labels and micro-optimize

ignatz:
Polygon labels are expensive. Add a layer-level switch to disable lab…els to allow reusing the same polygons while toggling the labels, e.g. depending on zoom level.

Also a bunch of small optimizations to minimize canvas operations (save/restore) and reduce strain on garbage collection (fewer ephemeral allocations)


ignatz:
Close in favor of strict super-set https://github.com/fleaflet/flutter_map/pull/1607 as suggested by @JaffaKetchup

PR #1607, opened 2023-08-04
Increase scope of polygon draw batching for polygons with labels.

ignatz:
This change contains all previous micro-optimizations plus a larger optimization extending the scope and applicability of polygon draw batching.

Previously, we'd eagerly flush a batch whenever we'd encounter a polygon with a label. However, depending on the rendered polygon and label size, it would never be drawn. In this case we don't have to and should flush.

From a composition and visual point of view, this change is a noop.

### DIFF
--- a/lib/src/layer/label.dart
+++ b/lib/src/layer/label.dart
@@ -5,14 +5,14 @@ import 'package:flutter/material.dart';
 import 'package:flutter_map/plugin_api.dart';
 import 'package:polylabel/polylabel.dart';
 
-void paintLabelText(
-  Canvas canvas, {
+void Function(Canvas canvas)? buildLabelTextPainter({
   required String labelText,
   required List<Offset> points,
   required double rotationRad,
   bool rotate = false,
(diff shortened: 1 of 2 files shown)


JaffaKetchup:
@ignatz Is it OK to close #1606? I'll try to have a look at this today.

ignatz:
> @ignatz Is it OK to close #1606? I'll try to have a look at this today.

Of course. Closed it already. I was thinking/hoping that breaking my changes up would aid the adoption. Happy to go with whichever approach you prefer.

ignatz:
> I was just thinking that if this contains all of the other, and this one includes better optimizations, why not prefer this one instead? Have I missed something?

No, missing nothing :). This is a clear superset. Visually both changes are noops (minus the "padding").

JaffaKetchup (approved):
LGTM. I haven't compared performance/benchmarks, but the code looks good and the example has not changed behaviour.
Thanks for the contribution!

Issue #1564, opened 2023-06-20
[FEATURE] Implement better method to organize the position of layers


JaffaKetchup:
<div type='discussions-op-text'>

<sup>**Originally discussed in #1553,** converted to issue to allow tracking in [v6 Release Planning](https://github.com/fleaflet/flutter_map/projects/3).
</sup>
In flutter_map, all `nonRotatedChildren` always appear on top of all `children`. This can cause issues for some users, who need to be able to control exactly where each layer is positioned within the layer stack, and also need to control whether the layer rotates.  

As such, it may be desirable in the long term to find a solution to this issue. Three proposals discussed somewhat in #1551 (https://github.com/fleaflet/flutter_map/pull/1551#issuecomment-1585563458, https://github.com/fleaflet/flutter_map/pull/1551#issuecomment-1585621572, https://github.com/fleaflet/flutter_map/pull/1551#issuecomment-1585637588, https://github.com/fleaflet/flutter_map/pull/1551#issuecomment-1586121816) are listed below, but all FM users are welcome to add their ideas or opinions!

## The `builder` proposal

An initial suggestion by @rorystephenson, started in #1551, was to introduce (or potentially replace the current `children`/`nonRotatedChildren` arguments with) a `builder` method. The user would create a `Stack` widget themselves, using a new widget `FlutterMapTransform.rotate` to enable rotation on specific layers.

I (@JaffaKetchup) think that having the user handle the stack and opt-in to rotation on layers:
 * increases the time required to implement the map first time round: even though only by maybe half a minute, converting users from other libraries needs to be as quick and easy as possible
   * means the user requires knowledge about the difference between a rotated and non-rotated layer before they can implement their first layer
 *  is perhaps a little 'backwards' considering most layers are rotatable - surely the non-rotatable layers are the odd-ones out, and so should explicitly state that they don't support rotation
 * causes friction when adding new/multiple rotatable layers, as another sub-`Stack` is required, or another `FlutterMapTransform` wrapper is required
 * makes code less readable as it's hard to see what the map widget is doing immediately, and somewhat breaks the general convention of interactive maps being formed from layers: layers are now descendants of the user's custom widget tree, not `FlutterMap` or a related widget
 * ~requires migration across all users~ (a later proposal made the builder optional via an alternative constructor)

## The wrapper widget and mixin proposal 

[...]

PR #1615, opened 2023-08-15
Reworked layers mechanism

JaffaKetchup:
There are four issues with the current way of defining layers:

1. One has already been extensively discussed in #1564, where 'non rotated layers' must always come on top
2. One has been spotted as a point at which people often get stuck (eg. #1616), where they put a layer over an interactive `MarkerLayer`, meaning it cannot receive gesture events
3. The name 'non rotated layer' is not ideal, the layers also don't move - something like 'anchored layer' might be preferable
4. For extremely simple maps, the concept of layers might be a bit too much

I've decided to make the following changes:
* (1, 3, 4): Reworked the way layers and non-rotated layers are defined and used
  * Deprecated `FlutterMap.nonRotatedChildren`
  * Transferred responsibility for whether to be 'mobile' or 'static' to the layer itself, via the new  `MobileLayerTransformer` widget which can be included in a layer's `build` method
* (2): Added `TranslucentPointer` widgets around all layers (optionally, on by default)
* (4): Added a new `FlutterMap.simple` constructor that just supports one very basic `TileLayer` and an attribution layer

The changes I've made are extremely intrusive, so open to suggestion! Doubtless someone'll spot a better way to do something.

---

Also includes some refactoring and cleanup that I couldn't resist doing. More to come in another PR!


mootw:
i am not a huge fan of these specific changes on first reading. i feel uneasy about the layers stack changes specifically.



i really like the concept of removing non-rotated children. the end user should not be able to place a map layer somewhere where it can go, but will function incorrectly and might not be obvious (as in they should not be concerned about whether a layer rotates; the layer should decide if it wants to rotate and how).

i think the implementation of the layer stack is sub-optimal though. it is quite convoluted and the concept of an Anchored Layer is introduced. as a developer i would expect these behaviors from fm
1. i can place any widget inside of the flutter map children and they will display as a stack in the map. (like i can place an Align bottom right with child text widget to create an attribution for example).
2. i can place a tile layer in (or create one as a FM developer), and Flutter Map will serve 2 primary goals. Create a view port that my widget renders in, provide bounds and projection helpers and the camera position, zoom, and rotation. Secondly, FM will capture touch/input events that handle navigation around the map. (i think it is ideal if FM avoids messing with gestures as much as possible to avoid creating wrappers around the input system). 

I  think this philosophy also applies to the layers_stack implementation. I don't think it is necessary to extend Widget to create an AnchoredLayer widget to isolate layers that need rotating and then picking those out of the stack, its convoluted in my opinion and gets further away from the everything is a widget philosophy (and might even cause flutter optimization issues).

My immediate thoughts are this. Treat all layers as non-rotated children, then have a FM provided widget that is literally just rotates the map layer just like how all other layers are already (basically returning the old behavior); we can wrap all of our internal layers that want that rotation in it, and any external developers who are creating layers can easily just wrap their widgets in our "old behavior" widget to return the old behavior. this makes migration trivial.

I think the benefits are obvious with this implementation:
- mixing and matching rotated and non rotated children is trivial
- we remove non-rotated children (App developers should not need to figure out where to put the layers, they should all just go in a stack top to bottom; layer widget developers decide how the widget behaves)
- the less we add "on top" of flutter behavior the less risk of performance issues or things we need to maintain.

**these are my immediate thoughts**, i haven't looked deeply at the translucent_pointer, but i have a feeling both of these things can get mitigated by a slightly easier design. we shouldn't need to be injecting touch behavior, if someone wants to modify a layers touch properties they should do it by wrapping the widget themselves (in my opinion; we should avoid modifying "default" touch/input capturing behavior when possible).

i can hopefully take a closer look at this project soon, ive been incredibly busy. but i think this is worth looking at because we should solve the gesture and layer rotation problem. it is quite strange that we are just rotating all of the widgets in my opinion when they can easily handle it internally with a single widget. i think the "flutter way" is that it is okay to have a widget handle behavior change. not everything needs to be a parameter, especially not things related to gesture handling, we can provide the widgets to handle these changes, just like flutter already has IgnorePointer and others.

JaffaKetchup:
@mootw

> My immediate thoughts are this. Treat all layers as non-rotated children, then have a FM provided widget that is literally just rotates the map layer just like how all other layers are already (basically returning the old behavior); we can wrap all of our internal layers that want that rotation in it, and any external developers who are creating layers can easily just wrap their widgets in our "old behavior" widget to return the old behavior. this makes migration trivial.

Ok, I think what you're saying is the inverse of what it currently is. However, I think there are still problems with this approach:

1. As we want to still allow any type of `Widget` to be used as a layer, some sort of detection mechanism would be required to ensure/check that the 'rotator widget' is being used at the top level, in order to allow FM to apply the necessary transformation effects to it. This is exactly what this PR implements, but with the opposite instead.

2. This is potentially less performant that opting-out of rotation, as is current. In most maps, there will be more rotated widgets than non-rotated widgets - therefore, the detection mechanism will be run more times. The detection mechanism is not cheap, as it requires an ancestor lookup.

3. It is not necessarily performant to wrap the layers with a 'rotator widget'. The detection mechanism will have to lookup more ancestors (2+): `_DetectorAncestor` <- `TileLayer` <- `RotatorWidget`. Therefore, a mixin would be more appropriate, as it removes the extra widget - as is implemented and recommended in this PR.

Essentially, with those 3 points, you come to this PRs implementation. However, because the detection will need to be applied more frequently, and for longer each execution (unless you use mixins), I would say it would be less performant.

4. There is an alternative, which is perhaps closer to what you were suggesting: forcing all `children` to be of type `Layer`, where `Layer` is sealed and extended by `StandardLayer` and `AnchoredLayer`. This would mean that the detection logic could be removed, but it might impact the developer experience having to wrap all their custom widgets in one of those. I guess I'm open to this approach, it could possibly be applied to this PR as well/instead, but it also feels just a bit 'off' somehow, like a step backwards.

5. To me, it seems counter-intuitive. Maps are mostly made out of moving, non-anchored layers. Why should these layers have to opt-in, instead of the special (anchored) layers opting-out?

> (App developers should not need to figure out where to put the layers, they should all just go in a stack top to bottom; layer widget developers decide how the widget behaves)

6. This is not true. Some apps want markers on top of polygons, others might want it the other way around. The ordering is still an issue. 

> i haven't looked deeply at the translucent_pointer, but i have a feeling both of these things can get mitigated by a slightly easier design. we shouldn't need to be injecting touch behavior, if someone wants to modify a layers touch properties they should do it by wrapping the widget themselves (in my opinion; we should avoid modifying "default" touch/input capturing behavior when possible).

7. Again, I'm not sure this is true. The purpose of applying translucency to all layers is to allow each layer to be able to handle gestures, as you proposed. Without it, only the topmost layer can receive gestures, if it encompasses the entire screen (as, let's say the `CircleLayer` does). The reason I've hidden this away by default, is because I can't think of many use-cases where you wouldn't want all the layers to be able to handle gestures, and because there's many issues with people missing this point. Having users wrap every layer in the `TranslucentPointer` widget is still possible in this PR, but FM does it automatically by default.

> i think the "flutter way" is that it is okay to have a widget handle behavior change

8. I would usually agree with this. However, the current detection implementation is based off of https://api.flutter.dev/flutter/widgets/AutomaticKeepAliveClientMixin-mixin.html, which definitely changes the behaviour of widgets it is applied to. In this case, we want to manipulate as few widgets as possible (for performance), so mixins seems like a good idea.

One thing I would agree with is that the current `_LayersStack` implementation may be sub-optimal in terms of performance. I've tried to optimize, but it's very possible there's better ways to get less rebuilds on fewer layers.

It would be great if @TesteurManiak & @ibrierley could also join in, and maybe @Robbendebiene as well, as you seemed to be interested in this.

ibrierley:
I'm not sure I have anything useful to add, I think I'm 50/50 on everything (without looking at the code in depth) but a couple of thoughts..

If one of the main incentives for any decision is a performance reason, I'd probably want to verify that is the case. Flutter often does a few optimisations anyway.

I'm not convinced either way on the non-rotated argument. My gut instinct is that a non-rotated widget is clear what happens to anyone new looking at code (but I don't really like the name), and makes them think about it (rather than having to think about every individual layer differently and know about it). I was wondering about the notion of only having a StaticLayer or a MovingLayer, so there's clarity, but really I'm not sure I like that either :). I feel a bit like an "anchored" layer is a bit confusing (as we use anchors elsewhere a bit differently?)

The gesture stuff, I'm even less sure of. Sorry, not being much help here, but will give any thoughts if they pop up.

Robbendebiene:
Just quickly read through this so apologies for any potential misconceptions.
I really like the concept of @mootw  at least the way I understood it wihch seems to differ from @JaffaKetchup 

I don't really understand why detection would still be needed in this case.
Currently all layers are rotated together via a bundled transform widget. If instead every layer has to rotate itself then there shouldn't be any detection needed. While this may sound more verbose, at the end it makes the layers more flexible and the verbosity can be solved by adding a mixin, super class or "rotation widget" that can be used by layers who want to rotate with the map.
I also think this perfectly fits into the concept that every widget below flutter map has access to the controller/camera and can act accordingly to state changes. With the suggested changes the flutter map widget would just be a wrapped stack with some inherited widgets and gesture detection stuff on top, wherefore any sort of widget can simply be thrown into it.

JaffaKetchup:
Ah, I think I've got you now.

Of course, I forgot that layers can access the map's state :d. I was going along that the state would need to be injected into the layers. Apologies for my stupidity.

I do wonder whether transforming more, shorter depth, widgets is less performant than transforming fewer, stacked/longer depth, widgets. But as @ibrierley said, we should probably test the performance somehow.

In regards to the translucent gesture stuff, that's almost definitely necessary IMO. There's no other way to allow all layers to handle their gestures independently.

I'll see if I can put something together...

JaffaKetchup:
@mootw @Robbendebiene @ibrierley @TesteurManiak Ok, I have implemented that in 48c0ec6.

However, one thing to note is that a widget which uses neither the new `MobileLayerTransformer`, nor a static positioning widget (such as `Align`), will not demonstrate the correct behaviour.

We can either leave this as it is, or implement a `StaticLayerTransformer`, which uses `Align` internally, then make them implement a sealed `LayerTransformer` interface and only accept them as `children`.

ignatz:
(I've only seen the latest version of the code with the MobileLayerTransformer, so I'm only infering based on the comments thus far how they implementation may have looked before).

I do like the direction this went, i.e.
 * that decision to rotate or not rotate no longer sits with the user
 * that any child can opt into map transformations w/o because it **composes**. Imagine a meta-layer that itself builds a stack of layers internally where some may be overlays and others may transform with the map.


I can't say much about performance other than more transformations means more work. However, transformations are relatively cheap. For example, if you use a marker layer with non-rotating markers today, every marker is being rotated with the map and then rotated back (-map.angle).

(At the expense of starting a naming bike-shed, when I first heard MobileLayerTransformer I couldn't tell whether it's "rotating" or "non-rotating". It's mobile from the camera's perspective but stuck to the map. Unless you grew up in game engines it might be more intuitive to think in terms of the map than the camera :shrug: . I don't have a silver bullet. Names that pop to mind: AlignWithMap, MoveWithMap, )

JaffaKetchup:
> For example, if you use a marker layer with non-rotating markers today, every marker is being rotated with the map and then rotated back (-map.angle).

That's interesting, I didn't realise that. I wonder if it's possible then that with Many markers, performance might be improved by allowing them to rotate?

> (At the expense of starting a naming bike-shed, when I first heard MobileLayerTransformer I couldn't tell whether it's "rotating" or "non-rotating". It's mobile from the camera's perspective but stuck to the map. Unless you grew up in game engines it might be more intuitive to think in terms of the map than the camera 🤷 . I don't have a silver bullet. Names that pop to mind: AlignWithMap, MoveWithMap, )

Personally, I don't see how the name can be confusing in context with in-code and web documentation. I've tried to remove all references to 'non-rotated' and 'rotated', and instead replaced them with ['static' and 'mobile'](https://docs.fleaflet.dev/v/v6-preview/usage/layers#mobile-vs-static-layers). I can see that without context it might be taken as mobile to the map (for example, an animation that moves across the map), but I've tried to build the context with phrases like "move and rotate with the map". The issue with something like "MoveWithMap" is that it should be "MoveWithMapLayer<Transformer>", which just feels unwieldy.

ignatz:
Just wanted to share a quick thought: the only reason why FlutterMap composes layers in the first place is so that layers can access the map state/camera through inherited widget. Especially for overlays, this is often not even necessary and users can just stack themselves on top of FlutterMap.

Now, with this approach where "mobile layers" are all translated independently one could imagine a world where MapCamera is something like a ValueNotifier giving users complete compositional freedom. Strawman:

```dart
  final controller = MapController(/* */);
  return Stack(
    children: [
        Watermark(),
        Opacity(
           opacity: 0.5,
           child: FlutterMap(controller),
        ),
        PolylineLayer(controller),
        Overlay(),
    ],
  );   
}
```

Now, whether you consider the controller/camera injection simpler is debatable but it would certainly give you the most freedom, is very explicit as to what depends on the map state/camera and does away with the classification into "static/mobile" :woman_shrugging: 

> Personally, I don't see how the name can be confusing in context with in-code and web documentation. I've tried to remove all references to 'non-rotated' and 'rotated', and instead replaced them with ['static' and 'mobile'](https://docs.fleaflet.dev/v/v6-preview/usage/layers#mobile-vs-static-layers).

Documentation is great but the less one needs to rely on it the better.  My personal biases always make me think android/ios first when I hear mobile :). "Dynamic" is commonly used as the opposite of static. Maybe just "MapTransformer" with the argument your tansforming into map space. Anyway, I'm gonna shut up about naming now and will never bring it up again.

JaffaKetchup:
@ignatz (yes you've won me over on the naming 😂)

I guess it's a good point that technically, something like in your example could be done.
Personally, I just don't find that 'accessible'/'friendly' - it decouples the idea of layers and maps and somewhat 'de-capsulates' things (not technically speaking, but I'm sure you can see what I mean) - but I can see that some people wouldn't necessarily mind. I guess it just depends on what feels right to you.

(If you want to have a crack at making a mock-up of your approach, that could be quite handy.)

Maybe we should be setting up a Discord bot for reaction/roles so that we can give surveys (like voting for this kind of thing) to people who are interested without disturbing everyone.

Indeed, as I've probably said before, overlays that don't require access to the map context should not be used inside `children`. I should probably put that in the docs.

The classification of static/mobile isn't really anything. It's never actually referred to internally, except for the "transformer" which is just a wrapper for some transformations. The entire "transformer" could be skipped, just implementing the transformations manually, and it would still be 'mobile'.

---

I also wanted to highlight this:

> However, one thing to note is that a widget which uses neither the new `MobileLayerTransformer`, nor a static positioning widget (such as `Align`), will not demonstrate the correct behaviour.
We can either leave this as it is, or implement a `StaticLayerTransformer`, which uses `Align` and/or `SizedBox.expand` internally, then make them implement a sealed `LayerTransformer` interface and only accept them as children. At least the second option would make it more obvious what each widget type is, as I do feel like it's not 100% clear at the moment, but maybe you disagree.

I'm still not sure what the best approach here would be (if we are sticking with the approach that's in this PR rather than nothing or some other overhaul suggested), so your opinion would be much appreciated!

---

Naming wise, I see where you're coming from again. But to me, 'dynamic' is a technical term, or otherwise suggests that the whole 'thing'/object/state/type might change/morph, whereas 'mobile' at least confines the dynamism to just moving about. On the other hand, I guess it also suggests reactivity - and the layers do react to changes in state - and does remove any ideas about it being tied to platforms.

Just `MapTransformer` doesn't make sense as it suggests either the map is being transformed (to what?)- which it kinda is, but technically isn't (and we want to encourage a distinction between map and 'tile layer', to make debugging other peoples' issues a bit clearer for us) - or doesn't say anything at all. 

Again, a Discord survey would be useful right about now, as they are both pretty solid options.

You've got me :D. I'll try to change 'mobile' to 'dynamic' when I get a moment. It also sounds less boring and more 'cooler', so there's that as well :D

Robbendebiene:
> Indeed, as I've probably said before, overlays that don't require access to the map context should not be used inside children. I should probably put that in the docs.

Just want to add a quick (albeit small) argument for adding overlays to children of the map even when they don't access the map's state. The map can still detect pan or zoom gestures that are performed on these overlays. Sometimes this is definitely undesirable, but I like to use it for the contribution text and zoom buttons. The benefit is that you can still tap the overlays while they are not blocking any other map interaction events (like panning).

ignatz:
> I guess it's a good point that technically, something like in your example could be done.
> Personally, I just don't find that 'accessible'/'friendly' - it decouples the idea of layers and maps and somewhat 'de-capsulates' things (not technically speaking, but I'm sure you can see what I mean) - but I can see that some people wouldn't necessarily mind. I guess it just depends on what feels right to you.
> 
> (If you want to have a crack at making a mock-up of your approach, that could be quite handy.)

I'm not arguing very strongly either way. I can totally see your point and was mostly just excited to realize how your proposed change shifts responsibilities and thus could allow us to flatten some hierarchies for the better or worse.

> I also wanted to highlight this:
> 
>>     However, one thing to note is that a widget which uses neither the new MobileLayerTransformer, nor a static positioning widget (such as Align), will not demonstrate the correct behaviour.
>>     We can either leave this as it is, or implement a StaticLayerTransformer, which uses Align and/or SizedBox.expand internally, then make them implement a sealed LayerTransformer interface and only accept them as children. At least the second option would make it more obvious what each widget type is, as I do feel like it's not 100% clear at the moment, but maybe you disagree.
> 
> I'm still not sure what the best approach here would be (if we are sticking with the approach that's in this PR rather than nothing or some other overhaul suggested), so your opinion would be much appreciated!

I do think this ties nicely into your point about encapsulation above, i.e. if users would need to use a StaticLayerTransformer wrapper for overlay layers, having FlutterMap compose children would be a bit of a leaky abstraction, i.e. the user is affected by how FlutterMap internally stacks non-positioned and positioned widgets.

That said, just like you I did expect the non-positioned OverflowBox to mess with the positioning/alignment of siblings in the stack, however playing around with this change I actually couldn't get it to break. For example, removing the Align from the ZoomButtonsPlugin actually, worked as expected. Could you elaborate a bit more on what issues you were experiencing when w/o a positioning widget? Maybe this isn't actually needed in all cases?

> Indeed, as I've probably said before, overlays that don't require access to the map context should not be used inside children. I should probably put that in the docs.

While FlutterMap is responsible for stacking the widgets, you might still want to throw your static overlay in there, if nothing else for layering reasons. After all, one of the limitations your change is trying to overcome is that today you cannot do a mobile-static-mobile sandwich.

> Just want to add a quick (albeit small) argument for adding overlays to children of the map even when they don't access the map's state. The map can still detect pan or zoom gestures that are performed on these overlays. Sometimes this is definitely undesirable, but I like to use it for the contribution text and zoom buttons. The benefit is that you can still tap the overlays while they are not blocking any other map interaction events (like panning).

While I agree that it should be fine to throw static layers such as zoom buttons into FlutterMap's children, I don't see why that wouldn't be possible if you stack them externally. Concretely, your button's hit detection shouldn't interfere with the pan gesture detection underneath unless you capture the inputs.

JaffaKetchup:
> Could you elaborate a bit more on what issues you were experiencing when w/o a positioning widget?

I did have some issues where it mispositioned itself, and one where it did some weird things when rotating. I can't remember exactly what I did to get these behaviours unfortunately.

I've forgotten the exact implementation of the zoom buttons, but I believe the issue might arise when there are uncertain sizes, hence why SizedBox.expand also works sometimes?

What happens with just using a `ColoredBox` with no other widgets? As far as I can remember, FM sized it a little strangely?

> While FlutterMap is responsible for stacking the widgets, you might still want to throw your static overlay in there, if nothing else for layering reasons.

That's a good point that I didn't think of.

> I don't see why that wouldn't be possible if you stack them externally. Concretely, your button's hit detection shouldn't interfere with the pan gesture detection underneath unless you capture the inputs.

Agreed. One effect that this PR does have on existing setups is that layers become translucent to gestures. They'll now be able to handle gestures in the way you say.

ignatz:
> I did have some issues where it mispositioned itself, and one where it did some weird things when rotating. I can't remember exactly what I did to get these behaviours unfortunately.
> 
> I've forgotten the exact implementation of the zoom buttons, but I believe the issue might arise when there are uncertain sizes, hence why SizedBox.expand also works sometimes?
> 
> What happens with just using a ColoredBox with no other widgets? As far as I can remember, FM sized it a little strangely?

The zoom buttons are implemented as a column of FloatingActionButtons. I removed the alignment and used the Column's mainAxisAlignment + a row to push 'em back to the bottom right. The Row and Column should fully expand into the parent's constraints.

I also added a marker layer (with the MobileTransformer) to make sure to have both in the stack and added a semi-transparent colored container with a margin:

Screenshot_20230920-161536_flutter_map Demo

Rotating seems fine. Did you have the issues with this revision of your code? I'm a bit baffled because just by reading the Stack's documentation I would expect issues but I'm not having any. Should I be excited? :thinking: Did you maybe have the Stack within the OverflowBox at some point?

JaffaKetchup:
I can't seem to reproduce the issue now either. Maybe I managed to reproduce it then made some changes afterward that fixed it without me realising.
I guess that means it can be left how it is (other than renaming 'mobile' to 'dynamic') :D

mootw (approved):
looks good

mootw (approved):
renaming anchorPose to alignment is breaking but a better name. just need to make sure documentation exists for that change

JaffaKetchup:
Yep, it was changed because anchor was a poor name that could be considered to mean the opposite of what it actually did.

Migration info is already written (check the v6 docs).

lib/src/layer/marker_layer.dart
```diff
@@ -102,49 +103,51 @@ class MarkerLayer extends StatelessWidget {
...
-                    child: m.child,
-                  )
-                : m.child,
-          );
-        },
-      )..retainWhere((w) => w != null))
-          .cast(),
+    return MobileLayerTransformer(
+      child: Stack(
+        children: (List.generate(
```
TesteurManiak, line 108: It would probably be better to use a callback:
```dart
Iterable<Widget> createMarkers(List<Marker> markers) sync* {
  // ...
}
```

You'd be able to use yield to add a valid marker instead of returning null then doing a `retainWhere` and a `cast` (hence avoiding to iterate multiple times over your marker list).

```dart
Stack(
  children: [ ...createMarkers(markers) ],
),
```

lib/src/map/widget.dart
```diff
@@ -1,50 +1,129 @@
...
+  /// It is recommended to provide a [RichAttributionWidget] or
+  /// [SimpleAttributionWidget] to the [attribution] argument. For more
+  /// information, see their documentation.
+  FlutterMap.simple({
+    super.key,
+    required this.options,
+    required String urlTemplate,
+    required String userAgentPackageName,
+    AttributionWidget? attribution,
+  })  : children = [
```
TesteurManiak, line 74: Instead of having a dedicated constructor which could be a bother for future evolution on the main `FlutterMap()` constructor, I would recommend to have a separated `SimpleFlutterMap` or `SimpleMap` widgets that would abstract the logic of `FlutterMap()`.

TesteurManiak (changes requested):

lib/src/layer/attribution_layer/shared.dart
```diff
@@ -0,0 +1,21 @@
...
+/// Layer widget intended to attribute a source
+///
+/// Implemented by [RichAttributionWidget] & [SimpleAttributionWidget].
+///
+/// Has no effect other than as a label to group the provided layers together
+/// for the [FlutterMap.simple] constructor.
+@immutable
+sealed class AttributionWidget extends Widget {
+  const AttributionWidget._();
+}
```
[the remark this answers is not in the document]
JaffaKetchup: Doesn't need to be a `mixin` or a `class` if it doesn't need to exist anymore: e991671.

lib/src/layer/marker_layer.dart
JaffaKetchup: Added in e991671. Just to note, this was introduced in #1659, not this PR.

lib/src/map/widget.dart
JaffaKetchup: I've removed the simple constructor altogether in e991671. It was an experiment, and I'm not really sure I like it that much.

TesteurManiak (approved):

PR #1586, opened 2023-07-14
Example app improvements

rorystephenson:
WIP PR to address changes suggested in #1545.

PR #1620, opened 2023-08-23
Fixed `TileLayer.reset` failing to load new tiles

### LINKED ISSUE #1619: [BUG] `TileLayer.reset` fails to load new tiles after pruning
### What is the bug?

When invoking `reset` by stream, tiles are removed, but no new tiles are loaded.

### How can we reproduce it?

Use the example app: https://demo.fleaflet.dev/reset_tilelayer

_No response_

### Platforms

All

### Severity

JaffaKetchup:
Fixes #1619, replaces part of #1586.


TesteurManiak (approved):
LGTM 👍

PR #1622, opened 2023-08-26
Added cancellation support to `TileProvider` and surrounding mechanisms

### LINKED ISSUE #1430: [FEATURE] Cancel unnecessary tile requests

JaffaKetchup:
Closes #1430. Note that this doesn't provide an implementation of a `TileProvider` that actually cancels HTTP requests, as this is currently unavailable without Dio (see #1430's thread for more info). It just implements the infrastructure necessary to allow an implementation.

To test the cancellation support, install Dio, and try these code blocks on top of this branch. Then, run on the web, and open the browser's DevTools Network tab. Observe how some requests are marked as "(cancelled)". Note that this is more apparent with slower tile servers, so using the default OpenStreetMap tile server likely won't cause too many cancellations.

<details>
<summary>network_tile_provider.dart</summary>

```dart
import 'package:flutter/rendering.dart';
import 'package:flutter_map/src/layer/tile_layer/tile_coordinates.dart';
import 'package:flutter_map/src/layer/tile_layer/tile_layer.dart';
import 'package:flutter_map/src/layer/tile_layer/tile_provider/base_tile_provider.dart';
import 'package:flutter_map/src/layer/tile_layer/tile_provider/network_image_provider.dart';
import 'package:http/http.dart';
import 'package:http/retry.dart';

/// [TileProvider] to fetch tiles from the network
///
/// By default, a [RetryClient] is used to retry failed requests. 'dart:http'
/// or 'dart:io' might be needed to override this.
///
/// On the web, the 'User-Agent' header cannot be changed as specified in
/// [TileLayer.tileProvider]'s documentation, due to a Dart/browser limitation.
class NetworkTileProvider extends TileProvider {
  /// [TileProvider] to fetch tiles from the network
  ///
  /// By default, a [RetryClient] is used to retry failed requests. 'dart:http'
  /// or 'dart:io' might be needed to override this.
  ///
  /// On the web, the 'User-Agent' header cannot be changed as specified in
  /// [TileLayer.tileProvider]'s documentation, due to a Dart/browser limitation.
  NetworkTileProvider({
    super.headers,
    BaseClient? httpClient,
  }) : httpClient = httpClient ?? RetryClient(Client());

  /// The HTTP client used to make network requests for tiles
  final BaseClient httpClient;

  @override
  bool get supportsCancelLoading => true;

  @override
  ImageProvider getImageWithCancelLoadingSupport(
    TileCoordinates coordinates,
    TileLayer options,
    Future<void> cancelLoading,
  ) =>
      FlutterMapNetworkImageProvider(
        url: getTileUrl(coordinates, options),
        fallbackUrl: getTileFallbackUrl(coordinates, options),
        headers: headers,
        httpClient: httpClient,
        cancelLoading: cancelLoading,
      );
}
```

</details>

<details>
<summary>network_image_provider.dart</summary>

```dart
import 'dart:async';
import 'dart:ui';

import 'package:dio/dio.dart';
import 'package:flutter/foundation.dart';
import 'package:flutter/painting.dart';
import 'package:flutter_map/src/layer/tile_layer/tile_provider/base_tile_provider.dart';
import 'package:http/http.dart';

/// Dedicated [ImageProvider] to fetch tiles from the network
@immutable
class FlutterMapNetworkImageProvider
    extends ImageProvider<FlutterMapNetworkImageProvider> {
  /// The URL to fetch the tile from (GET request)
  final String url;

  /// The URL to fetch the tile from (GET request), in the event the original
  /// [url] request fails
  final String? fallbackUrl;

  /// The HTTP client to use to make network requests
  final BaseClient httpClient;

  /// The headers to include with the tile fetch request
  final Map<String, String> headers;

  final Future<void> cancelLoading;

  /// Dedicated [ImageProvider] to fetch tiles from the network
  const FlutterMapNetworkImageProvider({
    required this.url,
    required this.fallbackUrl,
    required this.headers,
    required this.httpClient,
    required this.cancelLoading,
  });

  @override
  ImageStreamCompleter loadImage(
    FlutterMapNetworkImageProvider key,
    ImageDecoderCallback decode,
  ) {
    final chunkEvents = StreamController<ImageChunkEvent>();

    return MultiFrameImageStreamCompleter(
      codec: _loadAsync(key, chunkEvents, decode),
      chunkEvents: chunkEvents.stream,
      scale: 1,
      debugLabel: url,
      informationCollector: () => [
        DiagnosticsProperty('URL', url),
        DiagnosticsProperty('Fallback URL', fallbackUrl),
        DiagnosticsProperty('Current provider', key),
      ],
    );
  }

  @override
  Future<FlutterMapNetworkImageProvider> obtainKey(
    ImageConfiguration configuration,
  ) =>
      SynchronousFuture<FlutterMapNetworkImageProvider>(this);

  Future<Codec> _loadAsync(
    FlutterMapNetworkImageProvider key,
    StreamController<ImageChunkEvent> chunkEvents,
    ImageDecoderCallback decode, {
    bool useFallback = false,
  }) async {
    final cancelToken = CancelToken();
    cancelLoading.then((_) => cancelToken.cancel());

    final Uint8List bytes;
    try {
      final dio = Dio();
      final response = await dio.get<Uint8List>(
        useFallback ? fallbackUrl ?? '' : url,
        cancelToken: cancelToken,
        options: Options(
          headers: headers,
          responseType: ResponseType.bytes,
        ),
      );
      bytes = response.data!;
    } on DioException catch (err) {
      if (CancelToken.isCancel(err)) {
        return decode(
          await ImmutableBuffer.fromUint8List(TileProvider.transparentImage),
        );
      }
      if (useFallback || fallbackUrl == null) rethrow;
      return _loadAsync(key, chunkEvents, decode, useFallback: true);
    } catch (_) {
      if (useFallback || fallbackUrl == null) rethrow;
      return _loadAsync(key, chunkEvents, decode, useFallback: true);
    }

    return decode(await ImmutableBuffer.fromUint8List(bytes));
  }
}
```

</details>

---

Also involves a heavy (and breaking for a minority of implementations) cleanup of the `TileProvider` interface, to reduce the number of private methods, reduce the amount of code that must be overridden to achieve a custom behaviour, and reduce the scope of the `TileLayer`.

### DIFF
--- a/lib/src/layer/tile_layer/tile_image_manager.dart
+++ b/lib/src/layer/tile_layer/tile_image_manager.dart
@@ -125,10 +125,16 @@ class TileImageManager {
     final tilesToReload = List<TileImage>.from(_tiles.values);
 
     for (final tile in tilesToReload) {
-      tile.imageProvider = layer.tileProvider.getImage(
-        tileBounds.atZoom(tile.coordinates.z).wrap(tile.coordinates),
-        layer,
-      );
+      tile.imageProvider = layer.tileProvider.supportsCancelLoading
+          ? layer.tileProvider.getImageWithCancelLoadingSupport(
+              tileBounds.atZoom(tile.coordinates.z).wrap(tile.coordinates),
+              layer,
+              tile.cancelLoading.future,
+            )
+          : layer.tileProvider.getImage(
+              tileBounds.atZoom(tile.coordinates.z).wrap(tile.coordinates),
+              layer,
+            );
       tile.load();
     }
   }
--- a/lib/src/layer/tile_layer/tile_provider/base_tile_provider.dart
+++ b/lib/src/layer/tile_layer/tile_provider/base_tile_provider.dart
@@ -1,4 +1,5 @@
 import 'dart:typed_data';
+import 'dart:ui';
 
 import 'package:flutter/widgets.dart';
 import 'package:flutter_map/src/layer/tile_layer/tile_coordinates.dart';
@@ -6,17 +7,24 @@ import 'package:flutter_map/src/layer/tile_layer/tile_layer.dart';
 import 'package:flutter_map/src/layer/tile_layer/tile_provider/network_tile_provider.dart';
 import 'package:meta/meta.dart';
 
-/// The base tile provider, extended by other classes such as
-/// [NetworkTileProvider] with more specialised purposes and/or requirements
+/// The base tile provider, extended by other classes with more specialised
+/// purposes and/or requirements
+///
+/// Prefer extending over implementing.
 ///
 /// For more information, see
 /// <https://docs.fleaflet.dev/explanation#tile-providers>, and
-/// <https://docs.fleaflet.dev/layers/tile-layer/tile-providers>.
+/// <https://docs.fleaflet.dev/layers/tile-layer/tile-providers>. For an example
+/// extension (with custom [ImageProvider]), see [NetworkTileProvider].
 abstract class TileProvider {
   /// Custom HTTP headers that may be sent with each tile request
   ///
-  /// Some non-networking implementations may ignore this property. [TileLayer]
-  /// will always set the 'User-Agent' based on what is specified by the user.
+  /// Non-networking implementations may ignore this property.
+  ///
+  /// [TileLayer] will usually automatically set the 'User-Agent' header, based
+  /// on the `userAgentPackageName`, but this can be overridden. On the web, this
+  /// header cannot be changed, as specified in [TileLayer.tileProvider]'s
+  /// documentation, due to a Dart/browser limitation.
   final Map<String, String> headers;
 
   /// Indicates to flutter_map internals whether to call [getImage] (when
@@ -26,12 +34,17 @@ abstract class TileProvider {
   /// will be thrown.
   ///
   /// [getImageWithCancelLoadingSupport] is designed to allow for implementations
-  /// that can cancel HTTP requests in-flight, when the underlying tile is
-  /// disposed before it is loaded. This may increase performance, and may
-  /// decrease unnecessary tile requests. See documentation on that method for
-  /// more information.
+  /// that can cancel HTTP requests or other processing in-flight, when the
+  /// underlying tile is disposed before it is loaded. This may increase
+  /// performance, and may decrease unnecessary tile requests. Note that this
+  /// only applies to the web platform. For more information, and detailed
+  /// implementation expectations, see documentation on
+  /// [getImageWithCancelLoadingSupport].
   ///
-  /// Defaults to `false`.
+  /// [getImage] does not support cancellation.
+  ///
+  /// Defaults to `false`. Only needs to be overridden where
+  /// [getImageWithCancelLoadingSupport] is in use.
   bool get supportsCancelLoading => false;
 
   /// Construct the base tile provider and initialise the [headers] property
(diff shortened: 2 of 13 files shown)


josxha:
Hi @JaffaKetchup, thanks for implementing this awesome feature!
I created a demo app, throttled the internet speed in the chrome dev tools and did some testing on web.
image

The cancellation works flawlessly.
image

I haven't looked at the complete code yet but noticed two things:

lib/src/layer/tile_layer/tile_image_manager.dart
```diff
@@ -125,10 +125,16 @@ class TileImageManager {
...
+      tile.imageProvider = layer.tileProvider.supportsCancelLoading
+          ? layer.tileProvider.getImageWithCancelLoadingSupport(
+              tileBounds.atZoom(tile.coordinates.z).wrap(tile.coordinates),
+              layer,
+              tile.cancelLoading.future,
+            )
+          : layer.tileProvider.getImage(
+              tileBounds.atZoom(tile.coordinates.z).wrap(tile.coordinates),
+              layer,
+            );
```
josxha, line 137: Why is this logic not inside the TileProvider if the tile manager asks the TileProvider what function to call anyways.

lib/src/layer/tile_layer/tile_provider/base_tile_provider.dart
```diff
@@ -1,84 +1,284 @@
...
+  ImageProvider getImageWithCancelLoadingSupport(
+    TileCoordinates coordinates,
+    TileLayer options,
+    Future<void> cancelLoading,
+  ) {
+    throw UnimplementedError(
+      'A `TileProvider` that overrides `supportsCancelLoading` to `true` must '
+      'override `getImageWithCancelLoadingSupport`',
+    );
+  }
```
josxha, line 97: Wouldn't it be possible to merge `.getImage()` and `getImageWithCancelLoadingSupport()` into one function? The implementation will probably only override one function anyways and could handle additional logic inside the function.
Your solution is probably best for backwards compatibility but it feels a bit odd to me.

JaffaKetchup: Yeah, I was aiming for backwards compatibility. I have made breaking changes in this PR, but most implementations won't be affected.
If the other maintainers think that this isn't necessary, I'm happy to simplify it.

lib/src/layer/tile_layer/tile_image_manager.dart
JaffaKetchup: See below, in summary, to reduce breaking chnages.

TesteurManiak (approved):
LGTM, maybe wwe should add your code snippets as examples in the test app to showcase the cancellation mechanism 🤔

JaffaKetchup:
@TesteurManiak That's a good idea, I'll do that. Someone else said they might make a package, but it's easy enough just to make an example for. Just to double check, you're happy with the breaking refactoring and non-breaking cancellation support of the `TileProvider`?

TesteurManiak:
> you're happy with the breaking refactoring and non-breaking cancellation support of the TileProvider?

Yeah, deprecation messages are clear and the changes are not too intrusive 👍

PR #1623, opened 2023-08-26
Fixed incorrect `Marker` anchoring when counter-rotating

### LINKED ISSUE #1500: [BUG] Markers with `rotate: true` do not adjust their anchor
### What is the bug?

Markers (and info windows implemented as markers) with `rotate: true` to counter map rotation float around unexpectedly when the map is rotated. This seems to be because the rotation does not include the anchor offset, which is thus applied in the rotated space.

Marker geometry at the anchor position remains stationary in the map space when the map is rotated. That is, if the marker geometry is pointing at a particular map location and `rotate: true`, I expect the marker to continue pointing at that map location as the map is rotated.

```dart

void main() => runApp(
      MaterialApp(
        home: Scaffold(
          body: FlutterMap(
            options: MapOptions(center: LatLng(0, 0)),
            children: [
              MarkerLayer(
                rotate: true,
                markers: [
                  Marker(
                    point: LatLng(0, 0),
                    width: 256,
                    height: 256,
                    anchorPos: AnchorPos.align(AnchorAlign.left),
                    builder: (context) => const ColoredBox(
                      color: Colors.lightBlue,
                      child: Align(
                        alignment: Alignment.centerRight,
                        child: Text('-->'),
                      ),
                    ),
                  ),
                  Marker(
                    point: LatLng(0, 0),
                    width: 256,
                    height: 256,
                    anchorPos: AnchorPos.align(AnchorAlign.right),
                    builder: (context) => const ColoredBox(
                      color: Colors.pink,
                      child: Align(
                        alignment: Alignment.centerLeft,
                        child: Text('<--'),
                      ),
                    ),
                  ),
                ],
              ),
            ],
          ),
        ),
      ),
    );
```

A workaround could be to pass the resolved `Anchor` as `rotateOrigin`.

Fixes might include applying the anchor offset as a `Transform`, or fully using `Transform`s instead of `Positioned` widgets.

The effect is more pronounced as markers are larger. In the video below, it is particularly pronounced on the info window "markers", which are sized to a max size much larger than the info window itself.

https://user-images.githubusercontent.com/13222934/234441759-041071df-2ebb-48ff-b7a0-cd5fc75f7c44.mp4

### Platforms Affected

Android, iOS, Web, Windows, MacOS, Linux, Other

### Severity

### Frequency

### Requirements

JaffaKetchup:
Fixes #1500. `Marker.rotateAlignment` is now automatically adjusted, where possible, to fix incorrect anchoring when `rotate` is `true`.

Also:
* Removed `AnchorAlign` (without deprecation), preferring `Alignment`
* Added new documentation

### DIFF
--- a/example/lib/pages/markers.dart
+++ b/example/lib/pages/markers.dart
@@ -16,24 +16,28 @@ class MarkerPage extends StatefulWidget {
 
 class MarkerPageState extends State<MarkerPage> {
   final alignments = {
-    315: AnchorAlign.topLeft,
-    0: AnchorAlign.top,
-    45: AnchorAlign.topRight,
-    270: AnchorAlign.left,
-    null: AnchorAlign.center,
-    90: AnchorAlign.right,
-    225: AnchorAlign.bottomLeft,
-    180: AnchorAlign.bottom,
-    135: AnchorAlign.bottomRight,
+    315: Alignment.topLeft,
+    0: Alignment.topCenter,
+    45: Alignment.topRight,
+    270: Alignment.centerLeft,
+    null: Alignment.center,
+    90: Alignment.centerRight,
+    225: Alignment.bottomLeft,
+    180: Alignment.bottomCenter,
+    135: Alignment.bottomRight,
   };
 
-  AnchorAlign anchorAlign = AnchorAlign.top;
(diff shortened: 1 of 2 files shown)


TesteurManiak (approved):
LGTM

PR #1626, opened 2023-08-29
Changed `FlutterMapInternalController.fitCamera`'s `source` to `MapEventSource.mapController`

### LINKED ISSUE #1563: [BUG] `fitBounds` doesn't cause new tiles to load

JaffaKetchup:
Fixes #1563. However, this is a quick fix, and does not resolve the root of the problem. I'm not sure why `MapEventSource.mapController` acts differently to `MapEventSource.fitCamera`.

### DIFF
--- a/lib/src/map/internal_controller.dart
+++ b/lib/src/map/internal_controller.dart
@@ -238,7 +238,7 @@ class FlutterMapInternalController extends ValueNotifier<_InternalState> {
       fitted.zoom,
       offset: offset,
       hasGesture: false,
-      source: MapEventSource.fitCamera,
+      source: MapEventSource.mapController,
       id: null,
     );
   }


TesteurManiak (approved):
LGTM

PR #1627, opened 2023-08-29
Changed `TileLayer`'s default (native) zoom levels

### LINKED ISSUE #1625: [BUG] `maxNativeZoom` does not work since v4
Originally reported in https://github.com/fleaflet/flutter_map/discussions/1624 by @sequanimartin

---

`TileLayer.maxNativeZoom` no longer results in the tiles at the specified zoom levels being zoomed/scaled as intended.

Instead, in v4 & v5, the map appears to break, and in v6, there is no apparent effect.

```dart
TileLayer(
    urlTemplate: "https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png",
    maxNativeZoom: 18,
);
```

JaffaKetchup:
Fixes #1625.

The issue was already apparently resolved in v6 'master', however, it did not appear that way due to a likely conflict between the default `maxZoom` and specified `maxNativeZoom`, where the `maxZoom`'s effect took priority.

Changes have been made to the defaults, as follows:
* `minZoom`: 0
* `maxZoom`: double.infinity
* `minNativeZoom`: 0 (non-nullable)
* `maxNativeZoom`: 19 (non-nullable)

Changes have also been made to the documentation to reflect more appropriate usage recommendations and remove/correct confusing information.

The internal `_clampToNativeZoom` implementation has also been massively simplified to use Dart's native `clamp` method.

### DIFF
--- a/lib/src/layer/tile_layer/tile_layer.dart
+++ b/lib/src/layer/tile_layer/tile_layer.dart
@@ -57,23 +57,40 @@ class TileLayer extends StatefulWidget {
   /// Default is 256
   final double tileSize;
 
-  // The minimum zoom level down to which this layer will be
-  // displayed (inclusive).
+  /// The minimum zoom level down to which this layer will be displayed
+  /// (inclusive)
+  ///
+  /// This should usually be 0 (as default).
   final double minZoom;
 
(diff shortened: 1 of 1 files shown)


TesteurManiak (approved):
LGTM

PR #1632, opened 2023-08-30
v6 Release Preparation

JaffaKetchup:
A variety of changes that didn't fit in other PRs, or depend on other PRs. **This should be merged last, just prior to release!**

- Added informational/recommendation logs when in debug mode
  - Prefer CanvasKit over HTML rendering on web, when running on web with HTML rendering
  - Avoid using subdomains with OSM's tile servers, when using subdomains with OSM URL
  - Try out the new cancellable tile provider plugin, when running on web with default tile provider
  - Remember to use `TileLayer.retinaMode` when `{r}` is located in the URL, when `{r}` is in the URL and the property is `null`
- Switched to plugin for cancellable tile provider demo, instead of implementation in app
- Changed default `TileLayer.panBuffer` to 1 (from 0)
- Removed v5 deprecations
- Refactoring and tidy-up
- Improved example application
- Improved documentation
- ... and other misc improvements


example/pubspec.yaml
```diff
@@ -11,14 +11,18 @@ dependencies:
...
   latlong2: ^0.9.0
   proj4dart: ^2.1.0
   url_launcher: ^6.1.10
   shared_preferences: ^2.1.1
   url_strategy: ^0.2.0
   http: ^1.1.0
-  dio: ^5.3.2
+
+dependency_overrides:
+  flutter_map_cancellable_tile_provider:
```
TesteurManiak, line 23: Does it really work to use `flutter_map_cancellable_tile_provider` without any version or path? 🤔

TesteurManiak (approved):

JaffaKetchup:
This should be merged last.

PR #1636, opened 2023-09-01
Removed `MoveAndRotateResult` in favour of a `Record<bool, bool>`

JaffaKetchup:
Technically breaking, but should be invisible for most users.


TesteurManiak (approved):
LGTM

PR #1641, opened 2023-09-06
Compute label position in map/coordinate space rather than pixel space

ignatz:
Computing polylabel positions is insanely expensive eating up some 70% of my UI thread render times.

Screenshot from 2023-09-06 17-41-02

### DIFF
--- a/example/lib/pages/polygon.dart
+++ b/example/lib/pages/polygon.dart
@@ -131,6 +131,7 @@ class PolygonPage extends StatelessWidget {
                       borderColor: Colors.purple,
                       label: "Rotated!",
                       rotateLabel: true,
+                      labelPlacement: PolygonLabelPlacement.polylabel,
                     ),
                     Polygon(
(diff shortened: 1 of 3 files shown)


TesteurManiak (approved):
LGTM, I'm wondering if we should consider spawning an isolate to compute all label positions in one go to further improve performances 🤔

ignatz:
One could. Sparse in and output seem well suited given the comms overhead.
That said, it would make Polygon construction async and thus add extra
complexity for the user unless you make labels best-effort. I'd rather add
a label position art to the Polygon constructor letting users do whatever
they want.

I also have a patch out to polylabel to reduce computational cost by
another 70℅. In it's current shape I wouldn't move to isolates for my use
case.

With this change already the frame to frame cost for computing the label's
pixel position has gone to virtually zero.

JaffaKetchup (commented):
LGTM. I think using Isolates would be too complex for now.

PR #1642, opened 2023-09-06
Added more customization to cursor/keyboard rotation

JaffaKetchup:
The implementation of cursor/keyboard rotation in #1592 set the North of the map to the angle of the cursor. This is now named `setNorth` in `CursorRotationBehaviour`.  
However, at the time I noted that an alternative might be possible, and this was requested by @androidseb (in https://github.com/fleaflet/flutter_map/pull/1592#issuecomment-1706493077): offsetting the current rotation based on the difference between the cursor's start and end angle, instead of just setting it. This has been named `offset`.

Optionally (on by default), it is now possible to set the map's rotation to the clicked angle (when triggered as normal), instead of needing a drag.

These options are now encapsulated in a dedicated object `CursorKeyboardRotationOptions`, set from `InteractionOptions` (which has also been refactored into its own separate file to reduce the scope of 'options.dart'). Options have also been renamed to be shorter, and improve clarity within documentation.

Additionally:

* Fixed a bug where flinging during cursor/keyboard rotation would start pan fling animation and exit rotation
* Fixed (some?) bugs with incorrect gesture event emission when cursor/keyboard rotating
* Removed seemingly? unnecessary variables from the cursor/keyboard rotation code
* Improved performance by not setting variables on every build unnecessarily

*Note: there are many issues with the current gesture handling system, mostly inconsistent event emission, which seems to have been somewhat complicated by a different PR which added more ways to emit events: there now appears to be 3 or 4. This PR has not fixed that, and has indeed made it more inconsistent. A larger rewrite is required to fix this.*


lib/src/gestures/flutter_map_interactive_viewer.dart
```diff
@@ -329,6 +330,16 @@ class FlutterMapInteractiveViewerState
     --_pointerCounter;
     crDegrees = crDragDegrees;
 
+    // If the keyboard/cursor rotation did not drag, then just set north = cursor
+    if (crRotationTriggered.value && crInitialDegrees == _camera.rotation) {
```
androidseb, line 334: Do we want to check for this condition as well before entering this if statement?
```
_interactionOptions.cursorRotationBehaviour == CursorRotationBehaviour.setNorth
```

JaffaKetchup: I guess we can, but I did intentionally leave it at this, as it feels like better UX to me in both modes.

JaffaKetchup:
With @androidseb's [suggestions](https://github.com/fleaflet/flutter_map/pull/1642#discussion_r1318550796) in mind, I've changed the way cursor/keyboard rotation is configured considerably. Check this PR's updated description.

(Keep in mind that **none** of these changes are technically breaking, as development/preview versions are assumed to be unstable.)

JaffaKetchup:
I'm going to break standard procedures here and merge my own PR without a second maintainer's approval. This PR only affects what was already added in a prerelease version, and we need to get v6 on the move at some point!

Happy to make changes if required by another maintainer.

PR #1647, opened 2023-09-09
Fixed `MapOptions.backgroundColor` failing to update

### LINKED ISSUE #1646: [BUG] backgroundColor doesn't get updated on rebuilds
### What is the bug?

When the `backgroundColor` passed to `FlutterMap.options` changes, the actual displayed color does not update.

### How can we reproduce it?

```dart
import 'dart:math';

void main() {
  runApp(const MainApp());
}

class MainApp extends StatefulWidget {
  const MainApp({super.key});

  @override
  State<MainApp> createState() => _MainAppState();
}

class _MainAppState extends State<MainApp> {
  Color backgroundColor = Colors.green;

  void randomizeBackgroundColor() {
    setState(() {
      backgroundColor =
          Colors.primaries[Random().nextInt(Colors.primaries.length)];
    });
  }

  @override
    return MaterialApp(
      home: FlutterMap(
        options: MapOptions(
          backgroundColor: backgroundColor,
        ),
        nonRotatedChildren: [
          Center(
            child: ElevatedButton(
              onPressed: randomizeBackgroundColor,
              child: Text(
                'Color: #${backgroundColor.value.toRadixString(16)}\n'
                'Press to randomize',
              ),
            ),
          ),
        ],
      ),
    );
  }
}

```

_No response_

### Platforms

flutter_map 6.0.0-dev.3. Android, Web.

### Severity

lonelyteapot:
Fixes #1646.

I noticed that internally `MapOptions` were only updated if they differed from the old ones. However, the `==` operator did not check for `backgroundColor`, and neither did `hashCode`.

### DIFF
--- a/lib/src/map/options.dart
+++ b/lib/src/map/options.dart
@@ -304,6 +304,7 @@ class MapOptions {
       boundsOptions == other.boundsOptions &&
       minZoom == other.minZoom &&
       maxZoom == other.maxZoom &&
+      backgroundColor == other.backgroundColor &&
       onTap == other.onTap &&
       onSecondaryTap == other.onSecondaryTap &&
       onLongPress == other.onLongPress &&
@@ -330,6 +331,7 @@ class MapOptions {
         boundsOptions,
         minZoom,
         maxZoom,
+        backgroundColor,
         onTap,
         onSecondaryTap,
         onLongPress,


JaffaKetchup (approved):
Must have missed this out when implementing, thanks for this. LGTM.

PR #1648, opened 2023-09-10
Fixed crash on uncaught "Invalid image data" exception

s6o:
When running a custom tile server, in the case of a missing tile we can get back zero bytes. Resulting in the **decode** function failure in **_loadAsync** of FlutterMapNetworkImageProvider.

Looking at the **_loadAsync** implementation, the call to the **decode** function was left out of the try/catch block. Simply moving the **decode** function call inside the try block fixes the issue - the app no longer crashes on a missing tile.

### DIFF
--- a/lib/src/layer/tile_layer/tile_provider/network_image_provider.dart
+++ b/lib/src/layer/tile_layer/tile_provider/network_image_provider.dart
@@ -84,12 +84,12 @@ class FlutterMapNetworkImageProvider
         Uri.parse(useFallback ? fallbackUrl ?? '' : url),
         headers: headers,
       );
+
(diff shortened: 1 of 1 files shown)


JaffaKetchup:
Hey @s6o, thanks for this! Would it be possible for you also to create a bug report?

s6o:
Screenshot 2023-09-10 at 11 41 59

s6o:
> Hey @s6o, thanks for this! Would it be possible for you also to create a bug report?

Hey, I've done so: [Bug report](https://github.com/fleaflet/flutter_map/issues/1649)

JaffaKetchup (approved):
LGTM, thanks for fixing this!

PR #1651, opened 2023-09-10
Fixed `keepAlive` behaviour

### LINKED ISSUE #1650: [BUG] `MapOptions.keepAlive` has no effect since v6.0.0-dev.1
### What is the bug?

The `keepAlive` option in `MapOptions` has no effect as of the first v6 prerelease, as the `AutomaticKeepAliveClientMixin` seems to have been removed in #1551.

### How can we reproduce it?

Use the Demo snippet inside a `PageView`, with something else on another page. Load the map, and move it to a different position. Then move to the second page, and reload the map page. It will be back at the start position.

Add the appropriate mixin and fields back into `FlutterMapStateContainer`.

### Platforms

All

### Severity

JaffaKetchup:
Fixes #1650.

Mixed `AutomaticKeepAliveClientMixin` back into `FlutterMapStateContainer` and added `FlutterMapStateContainer.wantKeepAlive` field which redirects to `MapOptions.keepAlive`.

### DIFF
--- a/lib/src/map/widget.dart
+++ b/lib/src/map/widget.dart
@@ -44,7 +44,8 @@ class FlutterMap extends StatefulWidget {
   State<FlutterMap> createState() => FlutterMapStateContainer();
 }
 
-class FlutterMapStateContainer extends State<FlutterMap> {
+class FlutterMapStateContainer extends State<FlutterMap>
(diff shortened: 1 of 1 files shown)


TesteurManiak (approved):
LGTM

PR #1660, opened 2023-09-24
Ensure the RichAttributionWidgetState doesn't fail if its removed from widget tree.

bramp:
While there was an assertion to check for this, a Widget being removed ~1-2 frames after its added is perfectly valid, and thus we should easily check if the widget is still mounted. While we are at it, it seemed a GlobalKey was being used to track the context. In later versions of flutter Stateful widgets have a context property for this purpose.


JaffaKetchup:
Closed in favour of #1661, which includes this commit 51fda54.

PR #1661, opened 2023-09-25
Simplified the building of `RichAttributionWidget`

bramp:
Previously the widget waited two frames to find the dimensions it was allowed, then set the minWidth to be the width it naturally wanted. I think this was to ensure the widget took the full width of the screen. It was simple to set a Container's width to infinity to achieve the same thing, with less complexity.

I may have misunderstood the purpose, but either way it seems I get the same behaviour when tested.

### DIFF
--- a/lib/src/layer/attribution_layer/rich.dart
+++ b/lib/src/layer/attribution_layer/rich.dart
@@ -153,8 +153,6 @@ class RichAttributionWidget extends StatefulWidget {
 class RichAttributionWidgetState extends State<RichAttributionWidget> {
   StreamSubscription<MapEvent>? mapEventSubscription;
 
-  Size? persistentAttributionSize;
-
   late bool popupExpanded = widget.popupInitialDisplayDuration != Duration.zero;
   bool persistentHovered = false;
 
@@ -172,19 +170,6 @@ class RichAttributionWidgetState extends State<RichAttributionWidget> {
         },
       );
     }
-
-    WidgetsBinding.instance.addPostFrameCallback(
-      (_) => WidgetsBinding.instance.addPostFrameCallback(
-        (_) {
-          if (!mounted) return;
-
-          final renderObject = context.findRenderObject();
-          if (renderObject is RenderBox) {
-            setState(() => persistentAttributionSize = renderObject.size);
-          }
-        },
-      ),
-    );
   }
 
   @override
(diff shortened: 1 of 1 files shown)


JaffaKetchup:
Yeah, I remember adding this and thinking it was less than ideal. It's quite possible that this approach was needed before, but with some of the changes since then, it's become redundant.

JaffaKetchup (changes requested):
This has broken the sizing rules of the attribution box. On wider screens, it should shrink wrap around its contents, and on narrower screens, it should consume all of the available width. This appears to have made the box always behave as if it were on a narrow screen.

bramp:
Can you provide me an example. Here are two screenshots (narrow and wide) from before my change:

Screenshot 2023-09-25 at 11 44 13 AM

Screenshot 2023-09-25 at 11 43 42 AM

Both show the box expanding across the screen. Do you see something different? I'm happy to adjust to meet whatever you think is correct.

JaffaKetchup:
Yep, so the wider screen should shrink wrap around the attribution text, instead of expanding. You can check the latest master (latest exe build on master if you're on windows), to see what it should be.

bramp:
Those screenshots were taken from the latest master (without my change).  Running the example app, in a Chrome browser.  That's why I'm confused as I don't see what you see.

JaffaKetchup:
Ah, that's strange, I checked latest main and it was different. Can you just run from a prebuild if that's possible, just to rule out any caching or something? I'll have a double check as well ASAP, but away from my PC atm.

bramp:
Ok using release from https://github.com/fleaflet/flutter_map/releases/tag/v6.0.0-dev.3 I get the behaviour you expect.
Screenshot 2023-09-25 at 12 18 52 PM

After a lot of hitting vscode, the master branch now also looks like that. So I think I had some build caching issue somewhere :(

Anyways, I'll make it act like how you want, which should take less time than debugging why my setup looked different to yours :)

bramp:
Minor tweak. It now seems to do what you would like. Please take a look.

JaffaKetchup (approved):
It's still not quite as it was :D On narrow screens, it should expand fully. On wide screens, it should not.

However, I think for the sake of simplifying the codebase, it looks fine like this.

LGTM, thanks for the contribution. I'll update the CHANGELOG.

bramp:
I'm happy to tweak it more if needed, to fill width if less than 420 pixel, otherwise its natural size.

As for changelog/attribution I don't mind. I only stumbled on this because it caused a crash in my app, and happy to help improve the project.

PR #1662, opened 2023-09-25
Fixed bugs and added tests for `FlutterMapNetworkImageProvider`

bramp:
Testing it fails and fallback appropriately under different http responses.

Currently the last test, "test load with invalid response (with successful fallback)" fails. The primary URL returns 200 OK but HTML not a image. The fallback URL is never attempted. Per the discussion on discord this may be a bug, however, it looks to fix this you need to go higher up the stack (I suspect TileImage) because the image isn't actually processed until then.

### DIFF
--- a/lib/src/layer/tile_layer/tile_provider/network_image_provider.dart
+++ b/lib/src/layer/tile_layer/tile_provider/network_image_provider.dart
@@ -86,8 +86,13 @@ class FlutterMapNetworkImageProvider
             headers: headers,
           ),
         ),
-      );
+      ).catchError((dynamic e) {
+        // ignore: only_throw_errors
+        if (useFallback || fallbackUrl == null) throw e as Object;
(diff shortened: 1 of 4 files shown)


JaffaKetchup:
Thanks for this, I'll see what I can do to fix the issue you mentioned in your description. Is it safe for me to work directly in this PR?

bramp:
Feel free to do as you need.

JaffaKetchup:
Tbh, I'm surprised the fallback isn't being attempted at decode time here: https://github.com/fleaflet/flutter_map/blob/eb3b7908d12daf4197420596ef61f34ffcc7e6fb/lib/src/layer/tile_layer/tile_provider/network_image_provider.dart#L81-L93

JaffaKetchup (approved):
LGTM, thanks for this!

Note to self: some parts of the cancellable TP need changing to reflect this, and this needs adding to CHANGELOG.

PR #1665, opened 2023-09-27
Added '{d}' 'dimension'/tile size placeholder support to `TileLayer.urlTemplate`

### LINKED ISSUE #1664: [FEATURE] Add `TileLayer` `urlTemplate` placeholder for `tileSize`

`TileLayer` accepts a `tileSize` as well as a `urlTemplate`. Mapbox supports Static Tile URLs of the form `https://api.mapbox.com/styles/v1/{username}/{style_id}/tiles/{tilesize}/{z}/{x}/{y}{@2x}`

When using Mapbox you need to specify the tileSize twice, e.g

```dart
TileLayer(
   tileSize: 512,
   urlTemplate: 'https://api.mapbox.com/styles/v1/mapbox/satellite-v9/tiles/512/{z}/{x}/{y}{r}?access_token={accessToken}',
)
```

This can be error prone.

My proposal is a new placeholder, `{d}` (for dimension) that is replaced with the tile size, so I can now provide a url like `'https://api.mapbox.com/styles/v1/mapbox/satellite-v9/tiles/{d}/{z}/{x}/{y}{r}?access_token={accessToken}'` and not worry about ensuring I write 512 twice.

A little bit of duplication

Happy to send a very modest (one-line) patch.

### Severity

JaffaKetchup:
Fixes #1664.

Will only be breaking for users already using '{d}' normally in their URL: nobody.


bramp:
To catch the issue with users already using `{d}`, you could warn if additionalOptions contains keys x/y/z/s/d and that it would override what's provided in TileCoordinates/TileLayer. Or maybe that's a support model.

JaffaKetchup:
I don't think anyone should be using '{d}'. It's very undescriptive as a custom placeholder, does not refer to anything that that is normally used for, such as filling API keys or style params. It's also not valid in a URL outside of this.

bramp:
While I don't disagree that folks should not be using, Hyrum’s Law states otherwise :).  Just trying to avoid someone being confused in future.

JaffaKetchup:
I'll add it to the Migration instructions out of an abundance of caution! 😂

JaffaKetchup:
I'm going to bypass normal procedures here, and merge without a second maintainer's approval. This is a very small change, which adds no really new functionality or ideas. v6 needs to get moving at some point!

Happy to make changes if a second maintainer requests it.

PR #1668, opened 2023-09-27
Create new example page to test the Retina and TileSize options.

bramp:
Displays a MapBox map (that supports both retina and tile sizes), and allows the user to alter them as appropriate.

However, when changing the TileSize option to 512, odd things happen to the map :/ That most likely needs investigating.


JaffaKetchup:
Yeah, the interaction between tile size in FM, tile size of tiles requested and retina mode is unclear/uncertain to say the least. This needs sorting out and documenting.

JaffaKetchup:
I also wouldn't want to use a paid tile server for this, you could find yourself with a large bill if someone decides to be a nuisance.

bramp:
The API key is MapBox's from their examples :) so not my bill. If there is a open source provider that supports @2x happy to use that.

JaffaKetchup:
I've changed this to add more config options so we can test this properly.

Here's some tests that I already did (11 of 16 possible combinations) - it should be noted that my screen does not seem to be high-dpi (the suggested line in the docs does not evaluate true for me), so the simulated retina mode just looks really small.
[screenshots.zip](https://github.com/fleaflet/flutter_map/files/12743039/screenshots.zip)

It seems as though only 256 works in FM, regardless of the setting in the URL. This appears to be bug - although I'm not sure what the correct behaviour here should be.

Note that plugins often (in my experience) use `tileSize` as an indication of the tile size, regardless of the actual size of the image - for me at least, this has not caused any noticeable issues with my FMTC plugin.

Also see https://github.com/fleaflet/flutter_map/pull/1666#issuecomment-1738049869.

bramp:
Very cool. Ok, our comments seem split across two threads now.

I think this thread should continue to focus on the example page.

We should discuss correct retina/tileSize behaviour over in https://github.com/fleaflet/flutter_map/pull/1666#issuecomment-1738049869

JaffaKetchup (changes requested):
We can't use a Mapbox demo key here, it's probably against their ToS, and dishonest if it isn't.
We need to add an input field to allow users to enter their own demo key, or find a different retina-supporting 'free' tile server.

bramp:
I searched and searched, and I can't find a free provider that offers the retina/tileSize feature.  So providing their own API key seems reasonable. I'll amend the PR.

JaffaKetchup:
Ah, thanks. I'm happy to change this, as I also need to go back and make sure my changes work on smaller screens (I always test on Windows).

JaffaKetchup:
Closing this for now as it conflicts directly with the similar example in #1673. Tile size functionality can be added to that example at a later date.

PR #1674, opened 2023-10-02
Minor cleanup to tests

bramp:
* Moved two tests into their appropriate directories. These seemed to have been missed with previous refactors.
* Remove debug print statements from previous test.


JaffaKetchup (approved):
LGTM, thanks!

PR #1676, opened 2023-10-02
Fixed PositionedTapDetector2's transformation of global to local coords

### LINKED ISSUE #1675: [BUG] Tap relative offsets are incorrect when the widget is within a scaling widget
### What is the bug?

When the map is inside a scaling widget, such as `Transform.scale(... child: FlutterMap(...))` the onTap gestures are not scaled correctly, leading to both the onTap(TapPosition tapPosition, LatLng point) being slightly the wrong place, even though the map looks correct.

This seems to be due to [PositionedTapDetector2._getTapPositions](https://github.com/fleaflet/flutter_map/blob/master/lib/src/misc/private/positioned_tap_detector_2.dart#L171-L175) translating the global coordinates to local coordinates, but not correctly scaling them.

### How can we reproduce it?

```
testWidgets('Test tap is detected where expected with scale and offset.',
      (tester) async {
    const screenSize = Size(400, 400);
    await tester.binding.setSurfaceSize(screenSize);

    Offset? lastTap;
    // The Transform.scale fills the screen, but the PositionedTapDetector2
    // occupies the center, with height of 200 (0.5 * 400) and width 400.
    final widget = Transform.scale(
      scaleY: 0.5,
      child: PositionedTapDetector2(
        onTap: (position) => lastTap = position.relative,
        child: SizedBox(
          width: screenSize.width,
          height: screenSize.height,
          child: Container(
            color: Colors.red,
          ),
        ),
      ),
    );

    await tester.pumpWidget(widget);

    // On the screen the PositionedTapDetector2 is actually 400x200, but the
    // widget thinks its 400x400.
    expect(screenSize, tester.getSize(find.byType(SizedBox)));

    Future<void> tap(Offset pos, Offset expected) async {
      lastTap = null;
      await tester.tapAt(pos);
      expect(lastTap, expected);
    }

    // Tap top left of PositionedTapDetector2 which should be 0,0.
    await tap(const Offset(0, 100), Offset.zero);

    // Tap bottom right of PositionedTapDetector2
    await tap(const Offset(400 - 1, 300 - 0.5),
        Offset(screenSize.width - 1, screenSize.height - 1));
  });
```

Yes, PR coming.

### Platforms

Tested on Chrome (Web) and Android.

### Severity

bramp:
Now, before we accept this PR, there are two possible solutions

1) ```dart
    // Transform the global coordinate to the local coordinates.
    final renderBox = context.findRenderObject() as RenderBox?;
    final relative = renderBox?.globalToLocal(global);
    ```

or

2)  ```dart
    final relative = details.localPosition;
    ```

Option 2 is clearly the simplest, but it is unclear to me why that wasn't always the case. Perhaps `localPosition` wasn't a member when this code was written. (Which btw localPosition was added to flutter ~4 years ago). @TesteurManiak last touched this code ~6 months ago, and perhaps knows why local wasn't used.


JaffaKetchup:
I can't think of a reason why we wouldn't use the second option.

I wonder if this will also open up the route to perspective with `Transform` working properly (to implement tilting)...?

bramp:
From my understanding of `localPosition` it effectively does the same as my option 1. Also Option 1 will correctly support any transformation, skew,scale,translate,rotate, etc.

I'll amend this to just do option 2. It's less code and simpler.

JaffaKetchup (approved):
LGTM, many thanks once again!

PR #1677, opened 2023-10-03
Avoid setting 'User-Agent' header in `TileProvider` when on web platform

### LINKED ISSUE #1654: [BUG] Web: Refused to set unsafe header "User-Agent"
### What is the bug?

There seems to be a regression in v5 that is trying to set the `User-Agent` header on web.

### How can we reproduce it?

Use flutter_map in web application.

_No response_

### Platforms

Flutter Web on Chrome

### Severity

JaffaKetchup:
Fixes #1654.


JaffaKetchup:
I'm merging this without secondary approval, as it's low volume and low potential impact, and because we need to get v6 on the move!

PR #1678, opened 2023-10-04
Change pixelOrigin from a rounded Point<int> to a Point<double>

bramp:
In all cases pixelOrigin was used it was converted back to a double, and this little tweak ensures that pixel coordinates are kept doubles everywhere. Which also simplifies the code a little.

### DIFF
--- a/lib/src/layer/overlay_image_layer.dart
+++ b/lib/src/layer/overlay_image_layer.dart
@@ -50,15 +50,15 @@ class OverlayImage extends BaseOverlayImage {
   @override
   Positioned buildPositionedForOverlay(MapCamera map) {
     // northWest is not necessarily upperLeft depending on projection
-    final bounds = Bounds<num>(
+    final bounds = Bounds<double>(
       map.project(this.bounds.northWest).subtract(map.pixelOrigin),
       map.project(this.bounds.southEast).subtract(map.pixelOrigin),
     );
     return Positioned(
-        left: bounds.topLeft.x.toDouble(),
-        top: bounds.topLeft.y.toDouble(),
-        width: bounds.size.x.toDouble(),
-        height: bounds.size.y.toDouble(),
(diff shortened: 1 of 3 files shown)


ibrierley:
I'd be careful about removing the rounding and having fractional pixelOrigins if thats what happens. Iirc it's not clear if that was/is needed, and keep an eye out for things like thin gaps between tiles etc when testing.

bramp:
Thanks @ibrierley , I am using this change in my local project, and all flutter map tests pass.

But I'll go though all the example pages and see if I can get gaps/etc to appear, and report back.

Btw the only place this offset seems to be used is for overlays and markers, so hopefully that limits what it could break. In my case I noticed this when placing markers, and I was trying to rule out sources of error for why it wasn't pixel perfect.

JaffaKetchup:
> In my case I noticed this when placing markers, and I was trying to rule out sources of error for why it wasn't pixel perfect.

I have a feeling this may be to do with my suboptimal implementation of #1623. 

Using the Marker page in the demo, and using `Icon.circle` to better see things, it's perfectly centered when the anchor is central. Otherwise, the distance between the cursor and edge of the circle increases with the size of the circle itself.

This may be an additional issue and not the one you are talking about, but this might also be an issue.

JaffaKetchup:
Ah ha. What's this padding doing here?

image

JaffaKetchup:
Welp, it seems to work perfectly fine if I don't use an `Icon`, and just make a circle myself. @bramp Were you using an `Icon`?

JaffaKetchup:
Ok, just for the sake of not wanting to break anything at this late stage, I think we'll close this PR for now.
However, the change in lib/src/layer/overlay_image_layer.dart is isolated and the current code is entirely redundant and internal, so your changes make sense and cannot break anything. So I'll put this in #1632.

Thanks for this!

PR #1679, opened 2023-10-05
`CircleMarker` Performance Optimizations

Alexays:
I'm working on a project that requires a map with a lot of elements, while maintaining optimum performance.
At the moment, we have one custom painter per circle, which leads to a huge drop in performance when there are thousands of circles, for example, whereas we could just as easily use a single painter.

### DIFF
--- a/example/lib/main.dart
+++ b/example/lib/main.dart
@@ -9,6 +9,7 @@ import 'package:flutter_map_example/pages/fallback_url_network_page.dart';
 import 'package:flutter_map_example/pages/home.dart';
 import 'package:flutter_map_example/pages/interactive_test_page.dart';
 import 'package:flutter_map_example/pages/latlng_to_screen_point.dart';
+import 'package:flutter_map_example/pages/many_circles.dart';
 import 'package:flutter_map_example/pages/many_markers.dart';
 import 'package:flutter_map_example/pages/map_controller.dart';
 import 'package:flutter_map_example/pages/map_inside_listview.dart';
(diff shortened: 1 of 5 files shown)


JaffaKetchup:
Hi @Alexays, thanks for your PR!

We'll have a look into this, as performance improvements are always welcome. However, the tests are now failing.

Alexays:
> Hi @Alexays, thanks for your PR!
> 
> We'll have a look into this, as performance improvements are always welcome. However, the tests are now failing.

Thanks!
Tests should be ok now.

Alexays:
Just pushed another improvements.
As accorded to Hixie here
https://www.reddit.com/r/flutterhelp/comments/oa5dh3/is_there_a_way_to_cache_a_portion_of_canvas_to/
The use of drawPoints is way faster, and it allows us to draw the same circles in batch.
I just added a group logic to keep the border and color functionalities different.
And in my tests, I finally reached 60 fps with ~10000 circles drawn vs ~12fps with the old implementation.

JaffaKetchup:
Wow that's an improvement and a half! I'll check later today. 

To be honest, circle markers have not had much work done to them since their introduction.

TesteurManiak:
Thanks for the contribution! Do you think you can also update the `CirclePage` example to showcase your improvements with a lot of circles? This will make it easier to test.

JaffaKetchup:
(maybe follow the pattern from Many Markers?)

Alexays:
Sure done :)

TesteurManiak (approved):
LGTM, I've tested the example on Flutter Web with canvaskit and the map was quite smooth even with 20.000 circles 👍

JaffaKetchup:
Many thanks :)

PR #1681, opened 2023-10-07
Fixed bug where `null` `urlTemplate` caused exception in `TileLayer` construction

JaffaKetchup:
It's not necessarily safe to assume a URL template will be provided when WMS options are missing.
Some plugins, for example, (*cough *cough FMTC) use `TileLayer` to transport other useful information, but not the URL itself.

### DIFF
--- a/lib/src/layer/tile_layer/tile_layer.dart
+++ b/lib/src/layer/tile_layer/tile_layer.dart
@@ -306,7 +306,7 @@ class TileLayer extends StatefulWidget {
     }
     if (kDebugMode &&
         retinaMode == null &&
-        wmsOptions == null &&
+        urlTemplate != null &&
         urlTemplate!.contains('{r}')) {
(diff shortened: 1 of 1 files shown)


mootw (approved):
sure, though is it bad practice whatever fmtc is doing?? (i don't want to encourage bad practice by allowing something that should be avoided)

JaffaKetchup:
Well, it might be bad practise in some people's books, but it works just fine. If it is bad practise, then it needs a proper assertion. Maybe we also need to think about whether how we're expressing it is the best way - but I can't think of any alternatives off the top of my head.

One example is converting object A to B, which requires a `TileLayer`. B has two methods. One requires the `TileLayer` to have a URL, one doesn't care what's inside the TileLayer (but B must still have it), IYSWIM.
Also used to access the default templating function, but that needs to change now I guess.

Issue #1685, opened 2023-10-10
[FEATURE] Tappable `Polyline`s


thomascoumau:
### What do you want implemented?

I wanted to implement an onTap for PolylineLayer. This tap could be specific for each polyline in the layer.

### What other alternatives are available?

There is a package TappablePolyline, however, the tap is not really functional and the layer take the size of the all map so the mapTap is inconsistently triggered

### Can you provide any other information?

I will propose a PR soon

### Severity

Annoying: Currently have to use workarounds

thomascoumau:
I have made a draft PR for now [here](https://github.com/fleaflet/flutter_map/pull/1686)

thomascoumau:
I have opened #1697 in order to prepare future PR's in both FM and FMTP repo in the future.

JaffaKetchup:
I'm going to close this for now, as this won't be included in the core.

PR #1686, opened 2023-10-10
Feature: Add a polyline Tap

thomascoumau:
Add a polyline tap that can be specific to each polyline in the the Layer. Also now the layer does not take the size of the all map but a tap outside a polyline is directly passed to the map onTap.
This PR is linked to this [issue](https://github.com/fleaflet/flutter_map/issues/1685)

### DIFF
--- a/lib/src/layer/polyline_layer.dart
+++ b/lib/src/layer/polyline_layer.dart
@@ -19,6 +19,7 @@ class Polyline {
   final StrokeCap strokeCap;
   final StrokeJoin strokeJoin;
   final bool useStrokeWidthInMeter;
+  final String? tag;
 
   LatLngBounds? _boundingBox;
 
@@ -37,6 +38,7 @@ class Polyline {
     this.strokeCap = StrokeCap.round,
     this.strokeJoin = StrokeJoin.round,
     this.useStrokeWidthInMeter = false,
+    this.tag,
   });
 
   /// Used to batch draw calls to the canvas.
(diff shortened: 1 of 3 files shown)


JaffaKetchup:
Hi @thomascoumau, and thanks for the contribution!

I'm aware that the current plugin appears to be abandoned. However, before absorbing it, I think we should just try to check this. Have you contacted him in any way outside of GitHub?

thomascoumau:
Hello @JaffaKetchup, sorry for my late response. 

I haven't contacted him yet and I think I should try it first. However, for some context i have made another PR on their repo to improve the polyline tap and It was approved  some month ago but wasn't merged and this influenced my choice. 

At first I was thinking if I should post this PR to the tappablePlugin and after some discussion with my coworkers, when we saw that an issue existed to improved the tap and arrangement of the different layers of flutterMap, we decided it was better to provide the tap logic directly to this pluggin as it would be easier to maintain. 
(we got some problems with tappablePolylines since it doesnt support the latest version of FlutterMap and create conflicts)

lib/src/layer/polyline_layer.dart
```diff
@@ -37,6 +40,7 @@ class Polyline {
     this.strokeCap = StrokeCap.round,
     this.strokeJoin = StrokeJoin.round,
     this.useStrokeWidthInMeter = false,
+    this.tag,
```
JaffaKetchup, line 43: Is there a reason we shouldn't use `key`? Or at the very least just accept any `Object?`?

thomascoumau: Yeah you are right. I think it would be better to pass an `Object?`. The tag should be used in case you want to specify the tap for one specific polyline. In my project we use Ids for polyline and trigger some event based on this. I think it is easier to handle an Object you choose constructing your polyline rather than some keys.

JaffaKetchup:
Closed due to internal conversations.

Issue #1688, opened 2023-10-10
[BUG] `CircleMarker` incorrectly fills with stroke color and has incorrect size


vitalsh:
### What is the bug?

CircleMarker fills the circle with either CircleMarker.color or CircleMarker.borderColor

### How can we reproduce it?

Looks like any circle will be filled with the stroke color regardless of the settings/parameters

### Do you have a potential solution?

circle_layer.dart : line 104 and 118
this line looks suspicious:
final radiusPaint = paint..strokeWidth = radius;
should probably be:
final radiusPaint = paint..strokeWidth = borderStrokeWidth;

### Platforms

Any

### Severity

Obtrusive: Prevents normal functioning but causes no errors in the console

JaffaKetchup:
Thanks for reporting, we'll look into it along with the other issues from v6.
Looks like this issue may have been introduced by #1679, heads up @Alexays (I'm happy to fix, but you might have a reason or way you want to, so just pinging).

vitalsh:
Thanks for the fast reaction.

 

I’m developing an app for registering the flights at aur aeroclub.

I’m using distance circles (along with flight traces) to show how far one flew away from the field (there are some restrictions about this)

 

Vitaly

mootw:
can confirm this issue, not sure why i did not pick it up during my review. i think im also seeing incorrect radius on the rendering too.

JaffaKetchup:
Potentially radius and diameter are being switched up.

Alexays:
I'm on holiday until Wednesday, I'll be able to look into it from then on, sorry for the breaking changes 🥲

JaffaKetchup:
Ah, no worries. I'll also take a look into it, but you just might have a preferred method.

Alexays:
I'll go using drawCircle only if the marker is not filled, in that way we preserve as much a we can the perf improvement of using draw Point.

JaffaKetchup:
Not quite sure what you mean, but there's no rush at all!

TesteurManiak:
@Alexays you should also use `drawCircle` if the marker is filled with color with an opacity inferior to 1.

DevTello:
Hi, when do you plan to update the version with fixes? This bug is quite critical to me

siunus:
I just created this as a custom component with code from @Alexays 
Just in case anyone needs it.

```dart
import 'dart:ui';
import 'package:flutter/widgets.dart';
import 'package:flutter_map/flutter_map.dart';
[...]

PR #1692, opened 2023-10-18
Fixed `CircleMarker`'s incorrect appearance and size

Alexays:
Fix https://github.com/fleaflet/flutter_map/issues/1688

- Radius instead of diameters
- Fixed circles with only borders/filled with a transparent color

### DIFF
--- a/example/lib/pages/circle.dart
+++ b/example/lib/pages/circle.dart
@@ -12,12 +12,21 @@ class CirclePage extends StatelessWidget {
   Widget build(BuildContext context) {
     final circleMarkers = <CircleMarker>[
       CircleMarker(
-          point: const LatLng(51.5, -0.09),
-          color: Colors.blue.withOpacity(0.7),
-          borderStrokeWidth: 2,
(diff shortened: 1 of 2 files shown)


JaffaKetchup (changes requested):
That doesn't quite look like a circle to me 😂! Did you forget something or is this not ready for review yet?

image

Alexays:
Oups sorry, fixed! @JaffaKetchup

algizmonaco:
Hi

just tried the demo and the circle still does not fill correctly. When zooming it works partially.

Regards Daniel

JaffaKetchup:
Hi @algizmonaco,
I see you're running on iOS, can you try disabling Impeller to see if that resolves the issue?

algizmonaco:
thx,

Disabling Impeller fixes the issue. keeping disabled until it is fixed. May you want to add this to readme. thx

Daniel

JaffaKetchup:
I believe the fix is already available on Flutter's 'master' branch, so hopefully it won't take too long to arrive!

PR #1694, opened 2023-10-18
Exported 'positioned_tap_detector_2' library (with `TapPosition`)

### LINKED ISSUE #1693: [BUG] v6: `TapPosition` isn't exported
### What is the bug?

In v5 `TapPosition` was exported but it seems it isn't anymore

### How can we reproduce it?

Write any code that requires TapPositions (for plugins mostly, as https://github.com/OwnWeb/flutter_map_tappable_polyline/issues/58 for example)

Wa talked about it here: https://github.com/OwnWeb/flutter_map_tappable_polyline/issues/58

Export it :)

### Platforms

all

### Severity

JaffaKetchup:
Fixes #1693.

Also exports `LatLngTween`, as that might be useful for external libraries.


TesteurManiak (approved):
LGTM!

PR #1700, opened 2023-10-23
fix: Revert "Adjusted default `MapOptions.cameraConstraint` to prevent out-of-bounds viewing"

### LINKED ISSUE #1682: [BUG] v6: Unable to smoothly rotate when colliding with constraints
### What is the bug?

User cannot zoomout or rotate by gesture if the space is not fully filled with tile.

### How can we reproduce it?

```dart

class MinimumExample extends StatelessWidget {
  @override
    return Scaffold(
      appBar: AppBar(
        title: const Text('Minimum Example'),
      ),
      body: FlutterMap(
        options: const MapOptions(
          initialCenter: LatLng(0, 0),
          initialZoom: 1,
          minZoom: 0,
          maxZoom: 19,
        ),
        children: [
          TileLayer(
            userAgentPackageName:
                'net.tlserver6y.flutter_map_location_marker.example',
            maxZoom: 19,
          ),
        ],
      ),
    );
  }
}
```
0) This code set initialZoom to 1.
1) Try to move / rotate / zoom-out by gesture, and it is not work. (Note that minZoom = 0) Only zoom-in is working in this state.
2) Zoom in to zoom ~= 2 by gesture, and you will find that you cannot zoom it out back to zoom = 1 by gesture. (Stop at zoom ~= 1.5, which is according to your device's size)
3) Zoom in a little bit (around +0.2). Try to rotate the map now, and you will find that you have a very limited angle to rotate.

_No response_

### Platforms

Android 13 [XiaoMi13i, MIUI Global 14.0.5]

### Severity

### LINKED ISSUE #1699: [BUG] v6: New default constraints cause trapped zooms on different screen sizes
### What is the bug?

In my `MapOptions` I include:

`initialZoom: 1.5,
minZoom: 1.5,`

In v5.0.0 this worked as expected.
In v6.0.0:

Seems to work as expected on Android emulator.
On Chrome, it allows further zoom in but cannot zoom out until zoom level is greater than 3 (and then only allows zoom down to 3, not 1.5 requested).
On Windows 11 (x64) zoom is frozen - cannot zoom in or out - if `initialZoom` set below 3.
I've not tested on macOS/iOS.

Setting `minZoom` to a value greater or equal to 3 works as expected (doesn't allow zoom less than whatever it's set to) on all platforms tested.

It seems like `minZoom` is internally limited to no less than 3?

I've tried changing `minNativeZoom` and `minZoom` on the `TileLayer` but this has no effect (given they both default to zero anyway this is unsurprising).

### How can we reproduce it?

Set `initialZoom` and/or `minZoom` to a value less than 3 in `MapOptions` and try and zoom on affected platforms.

_No response_

### Platforms

Windows 11 (x64), Edge, Android Emulator

### Severity

JaffaKetchup:
Quick fix #1682
Quick fix #1699 (I'm assuming)

Needs a more permanent fix in future, as the zoom gesture handling code obviously doesn't work quite right. Zoom should never be 'trapped' (#1699), and it should handle rubbing against bounds.

### DIFF
--- a/lib/src/map/options/options.dart
+++ b/lib/src/map/options/options.dart
@@ -308,10 +308,7 @@ class MapOptions {
       _cameraConstraint ??
       (maxBounds != null
           ? CameraConstraint.contain(bounds: maxBounds!)
-          : CameraConstraint.contain(
-              bounds:
-                  LatLngBounds(const LatLng(-90, -180), const LatLng(90, 180)),
-            ));
+          : const CameraConstraint.unconstrained());
 
   @override
   bool operator ==(Object other) =>


TesteurManiak (approved):
LGTM

PR #1704, opened 2023-10-25
perf!: add simplification and segment culling to `PolylineLayer` & simplification to `PolygonLayer`

mootw:
adds an optional (enabled by default with conservative values) simplification step to polygon and Polyline layers.

the default value is 1 decimal degree at zoom level 0, which is scaled based on the floored zoom level. this is similar to leaflets default simplification of 1 logical pixel. i would rather this value be based on logical pixels, but i did not know an easy way to adjust the pixel scale to be static based on the zoom level. leaflet only calculates this value after the projection and uses pixels, but as far as im aware they avoid the issue of in-between zoom levels by not calculating offsets between them and enforcing whole zoom levels. I attempted to calculate the pixel scale using the camera context but was not successful.

using logical pixels lead to the lines jittering too much, so only having them simplified in incremental steps is less obvious.


JaffaKetchup:
As discussed, we should probably also think about fixing the culling code to work within elements, particularly for polylines.
For example, a long polyline currently needs to be manually broken into separate `Polyline`s to benefit from culling. It may be a good idea to leave one 'path' (ie. a segment of a polyline) either side of the camera's viewport outside of bounds, so that a false end can never be seen.
Not sure how it could work with polygons? Is drawing a straight line when out of bounds good enough?

ignatz:
Skimming the PR, IIUC you're simply merging neighboring points if they'd end up round about on the same pixel?

Right now you're re-simplifying on every draw. Since the merging is zoom-level dependent could we do some caching, so that there's less work to be done when panning/rotating?

FWIW, I'm personally already doing quite a few zoom-level-dependent gymnastics, e.g. fading certain polygons in and out. I'd be interested in having the tooling standalone as opposed to just integrated into the poly layers. In that case I can have a single merging pass with `Polygon' = f(Polygon)`. Is that something you've considered?

ignatz:
One more drive-by comment, I realized that your merging logic computes the distance between two map coordinates as cartesian. While that's a reasonable approximation at the equator, it's not so much towards the poles. You might run into some funny business. At least whatever threshold you set, will apply very differently.

mootw:
i really want the simplification and culling margin to be based on the render pixel rather than the scaled latlng, but there  are a couple trade-offs on both sides. definitely a TODO item though.

lib/src/layer/polyline_layer.dart
```diff
@@ -58,8 +53,13 @@ class Polyline {
 @immutable
 class PolylineLayer extends StatelessWidget {
   final List<Polyline> polylines;
+  @Deprecated('has no effect anymore, polyline culling is enabled by default and controlled by margin')
   final bool polylineCulling;
 
+  /// extent outside of the viewport before culling polylines, set to null to
+  /// disable polyline culling
+  final double? polylineCullingMargin;
```
JaffaKetchup, line 61: Does this need to be customizable? Can we foresee any issues with this? I'd be in favour of just removing this as well as the `polylineCulling`. I realise that the option is needed to properly support gradients, but maybe it's worthwhile just turning it off behind the scenes if a gradient is specified, and keeping it on otherwise. I can't think of any other case where culling wouldn't be wanted.
Additionally, it might make sense to make the default (or the only) value be the width of the polyline - if that's possible, I realise we're at the layer level here - or something like 10, just because we don't want to introduce visible capping of the polylines at the edges of the screen (the capping needs to overflow off screen).

lib/src/layer/polyline_layer.dart
```diff
@@ -69,28 +69,146 @@ class PolylineLayer extends StatelessWidget {
...
     super.key,
     required this.polylines,
-    this.polylineCulling = false,
+    this.polylineCulling = true,
+    this.polylineCullingMargin = 0,
     this.simplificationTolerance = 1,
     this.simplificationHighQuality = false,
   });
 
+  bool aabbContainsLine(
```
JaffaKetchup, line 78: Should this be public?

lib/src/misc/simplify.dart
```diff
@@ -0,0 +1,120 @@
+// implementation based on
```
JaffaKetchup, line 1: (referring to the file not the line) Is there a way we can make this simplification logic available more accessible publicly (as @ignatz), ie. in a single method?
Or did we want to keep this private for some reason?

lib/src/layer/polyline_layer.dart
```diff
@@ -56,29 +53,162 @@ class Polyline {
...
   final bool polylineCulling;
 
+  /// extent outside of the viewport before culling polylines, set to null to
+  /// disable polyline culling
+  final double? polylineCullingMargin;
+
+  /// how much to simplify the polygons, in decimal degrees scaled to floored zoom
+  final double? simplificationTolerance;
+
+  /// use high quality simplification
```
JaffaKetchup, line 66: We should clearly define what high quality simplification is, and it's advantages/use-cases and disadvantages.

lib/src/layer/polygon_layer/polygon_layer.dart
```diff
@@ -95,13 +97,21 @@ class PolygonLayer extends StatelessWidget {
   /// screen space culling of polygons based on bounding box
   final bool polygonCulling;
 
+  /// how much to simplify the polygons, in decimal degrees scaled to floored zoom
+  final double? simplificationTolerance;
+
+  /// use high quality simplification
```
JaffaKetchup, line 103: We should clearly define what high quality simplification is, and it's advantages/use-cases and disadvantages.

lib/src/layer/polyline_layer.dart
```diff
@@ -56,29 +53,162 @@ class Polyline {
...
+              start = -1;
+            }
+            if (start != -1) {
+              start = i;
+            }
+          }
+        }
+        if (fullyVisible) {
+          //The whole polyline is visible
+          // print("rendered whole polyhon");
```
JaffaKetchup, line 201: Is this OK to remove? Probably shouldn't leave debugging stuff in.

```suggestion
```

JaffaKetchup (changes requested):

lib/src/layer/polyline_layer.dart
mootw: it could be. i might move it into the bounding box class as it is probably more appropriate there.

lib/src/misc/simplify.dart
mootw: no reason to keep it private! it is just an implementation of simplify.js but with flutter_map typings to make things nice. we might keep it private to allow us to change the types in the future, and there might be simplify.js implementations for dart on pub, so i think we can just keep it private.

lib/src/layer/polyline_layer.dart
mootw: the culling already does not affect gradients! I also cannot think of a use case where polyline culling would want to exist. i have only kept it to prevent an api breaking change.

we do need to keep polyline culling margin since i cannot calculate the lat long radius for meter polygons easily. otherwise we could infer it from width. but since simplification is based on lat longs right now it isn't really easy to handle. that is why i want to eventually change it to be pixel based like leaflet (not just for the sake of copying them, but i think that pixel based simplification just makes more sense).

lib/src/layer/polyline_layer.dart
JaffaKetchup: Sure. I was actually asking whether it would make sense for it to be private, it just seemed like you may have made a mistake, as it's quite an "out of the way" place to put it. But if you're moving it somewhere better suited it's probably not so much of an issue.

lib/src/layer/polyline_layer.dart
JaffaKetchup: Ok, in that case, as there's so many breaking changes going on elsewhere anyway (and I can't remember if you changed the default, but that would be breaking), let's remove the toggle.

I see what you're saying about the margin. For now then, let's just make it something sensible but non-zero, just so that there aren't any visual glitches, unless you think that's really not necessary.

lib/src/misc/simplify.dart
JaffaKetchup: Ok, I see what you're saying, we don't want to confuse FM's scope as well. But I'm still not sure hiding it is the best either (if we've already got the functionality, why make someone install another package) - but I'll leave it up to you.

In any case, I think it should be refactored into a class called something like `Simplify`, and the `simplify` method either made static or renamed to `call`, and the two roots of the algorithms made public within the class. No specific reason exactly, apart from that I personally dislike trailing functions like this, especially when there's many in a file that all have related/interlinked functions - bundling them up just feels nicer. Not sure if there's any point having a constructor or any non-statics though.

mootw: i moved simplify into a top level library. not sure if this is actually the best way to do it. im hesitant to wrap it in a class with all static members just for the sake of doing that. it's actually a linting rule in dart to avoid it and it is part of effective dart: https://dart.dev/effective-dart/design#avoid-defining-a-class-that-contains-only-static-members

see latest commit for the change

lib/src/layer/polyline_layer.dart
mootw: i think its okay to be zero for a default value, most lines aren't thick enough to be problematic (and if it is that is when you should be using the value anyways)

lib/simplify.dart
```diff
@@ -0,0 +1,121 @@
+// implementation based on
```
JaffaKetchup, line 1: Still not sure about this, I feel like it's more polluting than it was. Definitely shouldn't be under 'lib/'. Would be good if @ignatz and the rest of the maintainers could chime in here.

lib/src/layer/polygon_layer/polygon_layer.dart
```diff
@@ -108,6 +110,12 @@ class PolygonLayer extends StatelessWidget {
...
```
JaffaKetchup, line 116: We should clearly define what high quality simplification is, and it's advantages/use-cases and disadvantages.

lib/src/layer/polyline_layer.dart
```diff
@@ -57,29 +51,126 @@ class Polyline {
...
-  final bool polylineCulling;
+
+  /// extent outside of the viewport before culling polylines, set to null to
+  /// disable polyline culling
+  final double? polylineCullingMargin;
+
+  /// how much to simplify the polygons, in decimal degrees scaled to floored zoom
+  final double? simplificationTolerance;
+
+  /// use high quality simplification
```
JaffaKetchup, line 62: We should clearly define what high quality simplification is, and it's advantages/use-cases and disadvantages.

JaffaKetchup (changes requested):
Just these few things left (2 carried over from my last review).

lib/simplify.dart
mootw: @ibrierley @TesteurManiak @josxha looking for some feedback!

TesteurManiak: I agree with @JaffaKetchup, I'm not sure that we need to expose to the users. IMO, this file would be better in the `lib/src/misc/` for internal use only 🤔

JaffaKetchup: (Just to say, I was originally for exposing it, just not quite in this way. But I do now agree that maybe keeping it private makes sense.)

mootw:
still want to look into using pixels instead of zoom levels + degrees; unless we want to merge fairly quickly which I'm totally ok with how it is. Technically there is not a ton of difference between the two values, because map size is based on virtual pixels so in the end the simplification is more or less based on pixels anyways (minus any weird virtual pixel scaling). It's just easier to explain: simplify with a tolerance of 1.5 virtual pixels **versus** 1.5 decimal degrees at zoom level zero / (2^zoom) which when you multiply it back out gets you about 1.456623 virtual pixels or whatever.

JaffaKetchup (approved):
LGTM, thanks! I've made a couple of changes to the example app, let me know if you like them or not, and I've hopefully fixed the Windows app build.
Happy to leave this to the end of the week if you wanted to get the pixel-based tolerance stuff in, but do want to merge relatively soon so we can properly re-assess performance, etc.

JaffaKetchup:
One thing I've spotted is that simplification (especially at larger numbers) causes the line to 'quiver'/shake, I guess due to the culling. Is there a way this can be prevented?

JaffaKetchup:
Additionally, it looks as though the simplification may be running even when the line is entirely culled, because there is some significant jank. Can this be prevented?

JaffaKetchup:
https://github.com/fleaflet/flutter_map/assets/58115698/e5918fd7-9d14-4d8b-a056-4b7a9fb8383d

lib/src/layer/polyline_layer.dart
```diff
@@ -135,31 +158,200 @@ class PolylineLayer extends StatelessWidget {
...
-    bool polylineCulling = true,
   });
 
+  @override
+  State<PolylineLayer> createState() => _PolylineLayerState();
+}
+
+// TODO: This does not work correctly when multiple `PolylineLayer`s are in use,
+// as they all share one state. For some reason, I couldn't even get them to
+// use seperate states with different `ValueKey`s
```
JaffaKetchup, line 173: Not sure how to resolve this. Help appriciated :)

TesteurManiak (approved):
LGTM! I don't know if you've been able to fix the performance issue, but I didn't seem to encounter any while testing the example. (I might not be testing in the correct conditions though)

JaffaKetchup:
If you can't spot the poor performance, then it's worked :D

lib/src/misc/simplify.dart
```diff
@@ -0,0 +1,119 @@
...
+double _getSqDist(
+  LatLng p1,
+  LatLng p2,
+) {
+  final double dx = p1.longitude - p2.longitude;
+  final double dy = p1.latitude - p2.latitude;
+  return dx * dx + dy * dy;
+}
+
+// square distance from a point to a segment
```
josxha, line 15: Probably better to use `///` here

JaffaKetchup: It's internal anyway so it doesn't matter too much, but you're right:

```suggestion
/// square distance from a point to a segment
```

josxha (approved):
lgtm, awesome pr! 👍

PR #1706, opened 2023-10-27
perf: micro-optimize `Bounds`

ignatz:
It may be somewhat worth it just because Bounds are instantiated many times per frame. I might have seen some performance uplift qualitatively playing around in profile mode, and maybe fewer GC pauses but see for yourself or whether we consider this within margin :shrug: 

Screenshot_20231027-171344_flutter_map Demo
Screenshot_20231027-171244_flutter_map Demo

### DIFF
--- a/lib/src/misc/bounds.dart
+++ b/lib/src/misc/bounds.dart
@@ -11,9 +11,9 @@ class Bounds<T extends num> {
   final Point<T> max;
 
   factory Bounds(Point<T> a, Point<T> b) {
-    final bounds1 = Bounds._(a, b);
-    final bounds2 = bounds1.extend(a);
-    return bounds2.extend(b);
+    final (minx, maxx) = a.x > b.x ? (b.x, a.x) : (a.x, b.x);
+    final (miny, maxy) = a.y > b.y ? (b.y, a.y) : (a.y, b.y);
+    return Bounds._(Point<T>(minx, miny), Point<T>(maxx, maxy));
   }
 
   const Bounds._(this.min, this.max);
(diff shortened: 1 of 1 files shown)


JaffaKetchup (approved):
LGTM. Untested, but I think the auto tests have this under control. Thanks for the micro-optimizations, they're always appreciated!

bramp:
Flyby question. Surely this kind of optimisation is simple for the dart compiler to do, especially when constructors are constant. I'm 3 months into learning dart, so unsure how well the compiler works. Curious if you have more insight.

JaffaKetchup:
Tbh, I don't know how well the compiler does with these kinds of things either :D. But I would say that a function call and memory assignment still needs to maintain some kind of direct translation IYSWIM.
There's a place for some micro-optimizations and not others. This optimization is still very readable, and so I'm fine with it.

ignatz:
The optimizations here are more than just stylistish or helping the toolchain's optimizer. For example, whether you call Bounds() or Bounds._() has different assumptions. The former always sorts the points. It also shortens some code paths and require fewer objects to be initialized at handled at runtime by the GC.

Some languages, some toolchains do analyze the local scope and construct objects on the stack if they can't escape. For sure Go and Ocaml (maybe dart). Virtually none optimize beyond the local scope because it can quickly get out of hand. I'm only aware of Ocaml that has a few pending proposals to add "memory exclusivity" as a first-class feature to allow non-local optimizations in the future.

ignatz:
I should probably clarify that my prior hand-waving around whether the change is worth or not, was never about whether it saves cycles. It's more a question of whether the savings significant in the context of all the other work that goes into rendering a frame.

PR #1707, opened 2023-10-27
feat: add option to draw labels on top of all `Polygon`s

ignatz:


lib/src/layer/polygon_layer/label.dart
```diff
@@ -17,46 +17,38 @@ void Function(Canvas canvas)? buildLabelTextPainter(
...
-  double dy = placementPoint.dy;
+  final textSpan = TextSpan(text: labelText, style: labelStyle);
+  final textPainter = TextPainter(
+    text: textSpan,
+    textAlign: TextAlign.center,
+    textDirection: TextDirection.ltr,
+    maxLines: 2,
+  )..layout();
 
-  if (dx > 0) {
```
ignatz, line 23: This check is super weird, it removes labels when their center goes off screen but only to the left. It's jarring to look at.

lib/src/layer/polygon_layer/polygon_layer.dart
```diff
@@ -235,7 +241,7 @@ class PolygonPainter extends CustomPainter {
         }
       }
 
-      if (polygonLabels && polygon.label != null) {
+      if (polygonLabels && !drawLabelsLast && polygon.label != null) {
```
ignatz, line 244: I'd also be happy to just remove this entire block, i.e. the old behavior. I'm not sure what's the use case for drawing polygons over labels with labels sometimes sticking out partially. Or appearing faint when overdrawn by semi-transparent polygons.

lib/src/layer/polygon_layer/polygon_layer.dart
```diff
@@ -267,6 +273,31 @@ class PolygonPainter extends CustomPainter {
...
+          final painter = buildLabelTextPainter(
+            polygon.points,
+            polygon.labelPosition,
+            placementPoint: map.getOffsetFromOrigin(polygon.labelPosition),
+            points: offsets,
+            labelText: polygon.label!,
+            labelStyle: polygon.labelStyle,
+            rotationRad: map.rotationRad,
+            rotate: polygon.rotateLabel,
+            padding: 50,
```
JaffaKetchup, line 294: This `padding` is different from the `padding: 10` in the other `builtLabelTextPainter` call.

lib/src/layer/polygon_layer/polygon_layer.dart
JaffaKetchup: Personally, I think the old behaviour might help with depth perception, so I think it would be a good idea to keep it.

lib/src/layer/polygon_layer/label.dart
JaffaKetchup: Wouldn't it be better to fix this to remove labels when off screen in any direction, rather than remove this optimization altogether like you appear to have done? (Unless I've missed something that is.)

ignatz: We could.

For now, I'd rather call it: "fixing visual glitches" rather than "removing an optimization that only 'work' when labels slip out to the left" :).

Computationally it would probably still be worth it. However, note that right now we're already culling polygons. So, this optimization would only kick in for polygons that are only partially on screen with the label off screen.

Would you wanna do it now or later?

lib/src/layer/polygon_layer/polygon_layer.dart
ignatz: Good catch. It wasn't intentional.

I picked a bigger value for my app and 50 is what I settled on. Not sure if there's a single right value.

For context, the padding says at what point the label should go away when you zoom out. Specifically, the polygon gets smaller but the label keeps it size, so at some point when the label plus padding doesn't fully fit into the polygon anymore the code drops it. I found that a small padding will keep the label longer on screen but may make it look a bit crammed :shrug:

lib/src/layer/polygon_layer/polygon_layer.dart
ignatz: Maybe. Maybe if the labels weren't text.

I found it to look like a car crash at best. At worst, it's an accessibility nightmare. Nothing better than a label you wanna read with lines and fills running over it :)

lib/src/layer/polygon_layer/polygon_layer.dart
JaffaKetchup: Maybe we should go with something like 20 then, as I do remember seeing 10 as a little cramped.

lib/src/layer/polygon_layer/polygon_layer.dart
JaffaKetchup: > At worst, it's an accessibility nightmare.

Absolutely true of course. Better to colour the text like GitHub does with it's issue labels.

But for the sake of not spoiling anyone's fun, let's just keep it.

lib/src/layer/polygon_layer/label.dart
JaffaKetchup: I guess that most of the time, the label is likely to be present. I guess technically 50% of the time if all locations had equal chance.
You're right, let's leave it for now. I think culling the polygons is probably enough.

lib/src/layer/polygon_layer/polygon_layer.dart
ignatz: Happy to change, no strong feelings around any particular value.

Since you brought this up, I also remembered that I had bumped the "maxLines" from 1 to 2 for my use-case. Arguably the value is very use-case dependend, would you like me to add a configurable parameter or simply not limit the lines?

JaffaKetchup (approved):
LGTM, I'll wait on one more review.

TesteurManiak (approved):
LGTM!

PR #1708, opened 2023-10-27
perf: avoid unnecessarily creating new list on every frame

ignatz:
… every time.

In my profiles this contributed sometimes several percent. The effect is probably also exacerbated by many layers. Subjectively this felt like a noticeable impact if I trust my noisy environment.


TesteurManiak (approved):
LGTM! Waiting for a second approval before merging.

JaffaKetchup (approved):
https://github.com/fleaflet/flutter_map/assets/58115698/3131a04d-d10d-458d-89f1-c5081d317d3a

PR #1709, opened 2023-10-27
perf: reduce tile management overhead

ignatz:
Consumed some single digit percent frame time for me. Improved contributions: 

* Don't compute hash of hashes.
 * Cheap custom hash.
 * Avoid string interpolation.

Also as a drive-by get rid of the TileCoordinates extends Point + zoom coordinate in favor of more consistent composition.

### DIFF
--- a/lib/src/layer/tile_layer/tile_coordinates.dart
+++ b/lib/src/layer/tile_layer/tile_coordinates.dart
@@ -8,14 +8,11 @@ class TileCoordinates extends Point<int> {
 
   const TileCoordinates(super.x, super.y, this.z);
 
-  String get key => '$x:$y:$z';
-
   @override
   String toString() => 'TileCoordinate($x, $y, $z)';
 
   // Overridden because Point's distanceTo does not allow comparing with a point
   // of a different type.
-  @override
   double distanceTo(Point<num> other) {
     final dx = x - other.x;
     final dy = y - other.y;
@@ -24,7 +21,9 @@ class TileCoordinates extends Point<int> {
 
   @override
   bool operator ==(Object other) {
-    if (identical(this, other)) return true;
+    if (identical(this, other)) {
+      return true;
+    }
     return other is TileCoordinates &&
         other.x == x &&
         other.y == y &&
@@ -32,5 +31,8 @@ class TileCoordinates extends Point<int> {
   }
 
   @override
-  int get hashCode => Object.hash(x.hashCode, y.hashCode, z.hashCode);
+  int get hashCode {
+    // NOTE: the odd numbers are due to JavaScript's integer precision of 53 bits.
+    return x | y << 24 | z << 48;
+  }
 }
--- a/lib/src/layer/tile_layer/tile_coordinates.dart
+++ b/lib/src/layer/tile_layer/tile_coordinates.dart
@@ -13,6 +13,7 @@ class TileCoordinates extends Point<int> {
 
   // Overridden because Point's distanceTo does not allow comparing with a point
   // of a different type.
+  @override
   double distanceTo(Point<num> other) {
     final dx = x - other.x;
     final dy = y - other.y;
@@ -21,9 +22,7 @@ class TileCoordinates extends Point<int> {
 
   @override
   bool operator ==(Object other) {
-    if (identical(this, other)) {
-      return true;
-    }
+    if (identical(this, other)) return true;
     return other is TileCoordinates &&
         other.x == x &&
         other.y == y &&
--- a/lib/src/layer/tile_layer/tile_image_manager.dart
+++ b/lib/src/layer/tile_layer/tile_image_manager.dart
@@ -13,10 +13,10 @@ typedef TileCreator = TileImage Function(TileCoordinates coordinates);
 
 @immutable
 class TileImageManager {
-  final Map<String, TileImage> _tiles = {};
+  final Map<TileCoordinates, TileImage> _tiles = {};
 
   bool containsTileAt(TileCoordinates coordinates) =>
-      _tiles.containsKey(coordinates.key);
+      _tiles.containsKey(coordinates);
 
   bool get allLoaded =>
       _tiles.values.none((tile) => tile.loadFinishedAt == null);
@@ -44,7 +44,7 @@ class TileImageManager {
   }) {
     for (final coordinates in tileBoundsAtZoom.validCoordinatesIn(tileRange)) {
       _tiles.putIfAbsent(
-        coordinates.key,
+        coordinates,
         () => createTileImage(coordinates),
       );
     }
@@ -64,7 +64,7 @@ class TileImageManager {
 
     for (final coordinates in tileCoordinates) {
       final tile = _tiles.putIfAbsent(
-        coordinates.key,
+        coordinates,
         () => createTile(coordinates),
       );
 
@@ -83,7 +83,7 @@ class TileImageManager {
   /// All removals should be performed by calling this method to ensure that
   // disposal is performed correctly.
   void _remove(
-    String key, {
+    TileCoordinates key, {
     required bool Function(TileImage tileImage) evictImageFromCache,
   }) {
     final removed = _tiles.remove(key);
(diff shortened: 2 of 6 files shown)


lib/src/layer/tile_layer/tile_image_manager.dart
```diff
@@ -13,7 +13,7 @@ typedef TileCreator = TileImage Function(TileCoordinates coordinates);
 
 @immutable
 class TileImageManager {
-  final Map<String, TileImage> _tiles = {};
+  final Map<int, TileImage> _tiles = {};
```
bramp, line 16: There is now a non-zero chance of hash collision.  Maybe the low probability is worth the performance?

bramp: Maybe just key by the coordinate directly? and add a equals method.  With the cached hashcode it should avoid redundant calculations, and ensure collisions aren't a problem.

lib/src/layer/tile_layer/tile_coordinates.dart
```diff
@@ -3,25 +3,32 @@ import 'dart:math';
 import 'package:meta/meta.dart';
 
 @immutable
```
JaffaKetchup, line 5: This class is no longer immutable.

```suggestion
```

lib/src/layer/tile_layer/tile_coordinates.dart
```diff
@@ -3,25 +3,32 @@ import 'dart:math';
...
-class TileCoordinates extends Point<int> {
+class TileCoordinates {
```
JaffaKetchup, line 6: Is there a reason we're not extending `Point` anymore? This is a breaking change (along with changing the constructor to be non-`const`), and I'm not sure what the motivation is.

JaffaKetchup (changes requested):
I'm a little unsure of some of the changes made. They are very breaking.

lib/src/layer/tile_layer/tile_image_manager.dart
ignatz: That makes a lot of sense. Thanks.

I first wanted to encode the limited length x, y, z in the key, thus avoiding collisions but then switched back to a hash because I messed up integer precision in Javascript. In my defense it was late at night :hide:

PTAL

lib/src/layer/tile_layer/tile_coordinates.dart
ignatz: removed caching. The new hash is super cheap anyways

lib/src/layer/tile_layer/tile_coordinates.dart
JaffaKetchup: (The tests and lints are failing because of this change.)

ignatz: Right now only the tests should be broken. Happy to fix those up if we can find agreement.

> Is there a reason we're not extending Point anymore? 

Moistly two:
 * Compositionally, to me it's a code-smell. It's the old composition vs inheritance discussion. TileCoordinates "are" semantically not a specialization/extension of a 2D point. Even if you were to think of it more as a "mixin" relationship, the only method reused is scaleBy and even Point.distanceTo was ill-suited. Which is just another indication, that they're not parent/child but rather false friends.
 * (It's not super critical for this use-case but I know that I always get you on the performance front and I really dislike math.Point :) ) Dart implements generics as so-called reified generics. This means unlike language like C++ that unroll all their generic instantiations into specialized code, Dart (go, kotlin, ...) only emit code once that treats Point<int> and Point<double> as Point<num> internally, with the "reified" types only there to enforce type-safety along the edges.

A teenie benchmark to make my point:

```dart
class Point<T extends num>  {
  T x;
  T y;

  Point(this.x, this.y);

  Point<T> operator+(Point<T> other) {
    x = (x + other.x) as T;
    y = (y + other.y) as T;
    return this;
  }
}

class PointInt  {
  int x;
  int y;

  PointInt(this.x, this.y);

  PointInt operator+(PointInt other) {
    x += other.x;
    y += other.y;
    return this;
  }
}


void main() {
  const N = 1000000000;
  final funcs = {
    'PointInt': () {
      var x = PointInt(23, 23);
      var y = PointInt(42, 42);
      for (int i = 0; i < N; ++i) {
        x+= y;
      }
    },
    'Point<int>': () {
      var x = Point<int>(23, 23);
      var y = Point<int>(42, 42);
      for (int i = 0; i < N; ++i) {
        x+= y;
      }
    },
    // Needed, otherwise the compiler strips the generic.
    'Point<double>': () {
      var x = Point<double>(23, 23);
      var y = Point<double>(42, 42);
      for (int i = 0; i < N; ++i) {
        x+= y;
      }
    },
  };

  for (final MapEntry(key: k, value:  f) in funcs.entries) {
    final w = Stopwatch()..start();
    f();
    print('${k}: ${w.elapsed}');
  }
}
```

Make sure to compile to avoid any JIT warmup:

```
$ dart compile exe generic2.dart && ./generic2.exe                                                                                                          

PointInt: 0:00:00.285020
Point<int>: 0:00:04.055126
Point<double>: 0:00:07.007694
```

Which is a 14x difference because you're paying the virtual function overhead for each addition (which is literally the cheapest operation in itself).

I've always felt that for something as low-level as Point, implementing it as a generic was a poor choice just to save a little bit of code and binary size.

To finish my tangent, I'm sure FlutterMap would benefit from using a PointInt a PointDouble, it's just hard to quantify w/o putting a large amount of upfront work in first.

More importantly for this use-case though, I'd still challenge that TileCoordinates is a Point or benefits from being a point.

JaffaKetchup: Wow, I wasn't expecting that much of a performance improvement just be not using generics! You've definitely got me there.

My biggest issue is that this change is very breaking. We should at least try to implement feature parity with `Point`, ie. the operators and other members. But it'll still be breaking because methods that accept `Point` will no longer be able to accept `TileCoordinates`.

ignatz:
Are you thinking oft TileCoordinates as a public API and hence how it might
currently be used by others? Otherwise, it's only the tests that are
broken. You can eg run the example

JaffaKetchup:
Yep, `TileCoordinates` is public and commonly used by plugins.

lib/src/layer/tile_layer/tile_coordinates.dart
ignatz: > Wow, I wasn't expecting that much of a performance improvement just be not using generics! You've definitely got me there.

It's a very exaggerate case because additions is so cheap vs extra indirection of virtual function call. For shits and giggles, if you change my benchmark to actually return a new Point, i.e. implementing addition as non-mutating, the difference is 30+x

ignatz:
> Yep, TileCoordinates is public and commonly used by plugins.

Or is it moistly flutter_map_tile_caching? ;)

I hadn't looked at it this way. From that point of view even reimplemting the same method set would be breaking since TileCoordinates can no longer be referenced as Point<int> :/

I reverted that part of the change. It's not the most critical part anyways. From an aesthetics point of view, I'm certainly not convinced that TileCoordinates is a Point but hey :shrug:

JaffaKetchup:
> Or is it moistly flutter_map_tile_caching? ;)

No no, it's used by many plugins! ~fmtc_plus_sharing for example.~ But seriously, this has widespread use - basically any plugin that touches `TileProvider` or any tile related logic will use it at some point.

> I hadn't looked at it this way. From that point of view even reimplementing the same method set would be breaking since TileCoordinates can no longer be referenced as Point :/

Indeed, I had realised that. But there's a breaking change with a year's worth of work of migration to do, and one where most people need to do barely anything. ~Maybe we should be using x.x.x.x for versioning? 🤔~

JaffaKetchup (approved):
LGTM, many thanks!

ignatz:
> No no, it's used by many plugins!

That's what they all say :). I'm just pulling your leg.

Thanks for the review. FTR, after our conversation and my little benchmark my old resentment towards math.Point rekindled. I moved my own fork of polylabel to a custom PointDouble which saved almost 20% cycles. Arguably polylabel is more pure number crunching but still, that was way more than I had expected. Maybe something worth thinking about for the future.

PR #1712, opened 2023-10-31
fix: remove flutter_map text attribution when `showFlutterMapAttribution` is `false`

leiflinse-trivector:
When `showFlutterMapAttribution` is false, there is still a text "Made with 'flutter_map' that is displayed. This PR removes it.

bild


TesteurManiak (approved):
LGTM

JaffaKetchup (approved):
LGTM, many thanks!
This likely won't be released for a little while.

PR #1716, opened 2023-11-03
perf(polygons): cull labels separately, and cache `TextPainter`

ignatz:
For ease of reviewing, I pulled the "polygon label"-only changes out of https://github.com/fleaflet/flutter_map/pull/1714 AND add TextPainter caching (which is actually the biggest cost reduction)

I could also add the TextPainter caching to https://github.com/fleaflet/flutter_map/pull/1714 but maybe we don't make that one even more sprawling. I'd suggest to review this one first and then I'll update https://github.com/fleaflet/flutter_map/pull/1714 .

### DIFF
--- a/lib/src/layer/polygon_layer/label.dart
+++ b/lib/src/layer/polygon_layer/label.dart
@@ -7,41 +7,50 @@ import 'package:latlong2/latlong.dart';
 import 'package:polylabel/polylabel.dart';
 
 void Function(Canvas canvas)? buildLabelTextPainter({
+  required math.Point<double> mapSize,
   required Offset placementPoint,
-  required List<Offset> points,
-  required String labelText,
+  required ({Offset min, Offset max}) bounds,
+  required TextPainter textPainter,
   required double rotationRad,
   required bool rotate,
-  required TextStyle labelStyle,
   required double padding,
 }) {
-  final textSpan = TextSpan(text: labelText, style: labelStyle);
-  final textPainter = TextPainter(
-    text: textSpan,
-    textAlign: TextAlign.center,
-    textDirection: TextDirection.ltr,
-  )..layout();
+  final dx = placementPoint.dx;
+  final dy = placementPoint.dy;
+  final width = textPainter.width;
+  final height = textPainter.height;
 
-  final dx = placementPoint.dx - textPainter.width / 2;
-  final dy = placementPoint.dy - textPainter.height / 2;
-
-  double maxDx = 0;
-  var minDx = double.infinity;
-  for (final point in points) {
-    maxDx = math.max(maxDx, point.dx);
-    minDx = math.min(minDx, point.dx);
+  // Cull labels where the polygon is still on the map but the label would not be.
+  // Currently this is only enabled when the map isn't rotated, since the placementOffset
+  // is relative to the MobileLayerTransformer rather than in actual screen coordinates.
+  if (rotationRad == 0) {
+    if (dx + width / 2 < 0 || dx - width / 2 > mapSize.x) {
+      return null;
+    }
+    if (dy + height / 2 < 0 || dy - height / 2 > mapSize.y) {
+      return null;
+    }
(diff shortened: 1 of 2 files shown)


lib/src/layer/polygon_layer/label.dart
```diff
@@ -6,42 +6,54 @@ import 'package:flutter_map/src/layer/polygon_layer/polygon_layer.dart';
 import 'package:latlong2/latlong.dart';
 import 'package:polylabel/polylabel.dart';
 
+class PolygonBounds {
+  final Offset min;
+  final Offset max;
+
+  const PolygonBounds(this.min, this.max);
+}
```
JaffaKetchup, line 14: For a small class like this without methods, I'd usually prefer a `Record`: `({Offset min, Offset max})`.

lib/src/layer/polygon_layer/label.dart
```diff
@@ -6,42 +6,54 @@ import 'package:flutter_map/src/layer/polygon_layer/polygon_layer.dart';
...
 
-  final dx = placementPoint.dx - textPainter.width / 2;
-  final dy = placementPoint.dy - textPainter.height / 2;
-
-  double maxDx = 0;
-  var minDx = double.infinity;
-  for (final point in points) {
-    maxDx = math.max(maxDx, point.dx);
-    minDx = math.min(minDx, point.dx);
+  // Cull labels where the polygon is still on the map but the label would not be.
```
JaffaKetchup, line 30: It doesn't look like culling works correctly when the map is rotated. I simply added a `print` statement below line 40, and it keeps printing even when the label is offscreen.

JaffaKetchup (changes requested):

lib/src/layer/polygon_layer/label.dart
ignatz: Works for me. Changed

lib/src/layer/polygon_layer/label.dart
ignatz: Interesting. It's because the offset isn't relative to the camera but rather relative to the rotated MobileTransformationLayer. I'm not sure we currently have functionality to easily translate into screen coordinates. If it wasn't for the MobileTransformationLayer we could work with latLngToScreenPoint.

I guess since culling is merely a best effort optimization, not culling is fine. I wasn't able to produce visual glitches as is but I also couldn't convince myself that it's not generally possible. So the safest would probably to disbable culling when the map is rotated. What do you think?

JaffaKetchup: I couldn't see any visual glitches either, so I think I'm happy to just keep it on. We should add a comment though, just to point out that the issue is known.

I also couldn't see that the culling didn't work correctly on a concave polygon (the one in the example app), so I think that comment can be removed.

ignatz: > I couldn't see any visual glitches either, so I think I'm happy to just keep it on. We should add a comment though, just to point out that the issue is known.

Personally, I'd rather go the safe route. I did try extreme aspect ratios and it still didn't glitch, the effectiveness is of this optimization is greatly reduced, i.e. it accidentally culls. If we want this to work we should figure out the correct math, I'm just a bit too baked right now.

> I also couldn't see that the culling didn't work correctly on a concave polygon (the one in the example app), so I think that comment can be removed.

That comment is different though and visible on the example app. That's about the padding, i.e. and when not to draw the label even when the label placement point is on screen. Specifically, the intent was to not draw labels if the label doesn't fit into the polygon. In the example app you can see the label being drawn onto the border of the concave polygon. The original author was trying to prevent that case and for convex polygons it works that way.

JaffaKetchup: I've got it now. We'll keep it disabled then just in case, but I think it should be alright.

JaffaKetchup (approved):

Issue #1723, opened 2023-11-09
[BUG] `InteractionFlag.enableScrollWheel` has no effect


josxha:
### What is the bug?

It's currently not possible to disable scrolling without using a deprecated property.

### How can we reproduce it?

1. Open the [Interactive Flags page of the demo app](https://demo.fleaflet.dev/interactive_flags_page).
2. Ensure that the "Scroll" checkbox is not checked.
3. Scroll

### Do you have a potential solution?

The deprecated `enableScrollWheel` proterty defaults to true. The check if the gesture is enabled is performed with an OR operator:
```dart
InteractiveFlag.hasFlag(_interactionOptions.flags, InteractiveFlag.scrollWheelZoom) || 
      _interactionOptions.enableScrollWheel
```
Using false as a default value for `InteractionOptions().enableScrollWheel` will fix this issue.
Going to create a pull request.

### Platforms

all

### Severity

Obtrusive: Prevents normal functioning but causes no errors in the console

PR #1724, opened 2023-11-09
fix: enableScrollWheel interaction flag has no effect

josxha:
See https://github.com/fleaflet/flutter_map/issues/1723


JaffaKetchup:
As the next release is breaking anyway, we can go through and remove all deprecations anyway. I'm happy to do this this evening.

JaffaKetchup:
@josxha How does this relate with #1726?

JaffaKetchup:
Made redundant by #1730.

PR #1727, opened 2023-11-10
perf!: optimize CRS performance and add micro-benchmark of limited expressiveness

ignatz:
Sorry to any reviewer, this change looks wilder than it is. It should no be braking unless you were depending on Transformation which I now  made internal (should allow us to freely change APIs in the future). At the end of the day, I spend a significant effort to reduce the `math.Point` constructions and most importantly reduce virtual function call overhead.

There are also some minor simplifications:
 * I moved things that don't need to be virtual into the base class since we're sub-typing anyway
 * I moved transformation out of `Crs` and up a level since it didn't seem to fit the Proj4Crs well.

I found testing on a device a bit limiting so I ended up adding some micro-benchmarks (we can also drop 'em, no strong feelings). I always like to say: don't trust micro-benchmarks but I also found them to be quite misleading in some cases. Things that seemed faster in the micro-benchmark didn't have an effect on-device and vice versa. It's unclear to me how much this is artifacting or due to the differences in the arm/x86 compiler backends :shrug:

At the end of the day, we probably care the most that it's faster on-device. The impact of this change will ultimately depend on how many LatLng coordinates you're translating into screen coordinates. In other words, more poly(gons|lines), more impact. I used the example and added an extra 100 polygons with 50 points each. The result was roughly a **25% reduction** in UI thread times:

<table>
<tr>
<th> Before: </th>
<th> After: </th>
</tr>
<tr>
<td>

Screenshot_20231110-142305_flutter_map Demo

</td>
<td>

Screenshot_20231110-143225_flutter_map Demo

</td>
</tr>
</table>

I only wish, I could point to the micro benchmarks and see a similar uplift. For transparency:

```
BEFORE:
(duration: 0:00:04.416050, name: Concrete type: EPSG:3857.latLngToPoint())
(duration: 0:00:03.895437, name: EPSG:3857.latlngToPoint())
(duration: 0:00:04.058475, name: EPSG:3857.pointToLatLng())
(duration: 0:00:01.341018, name: EPSG:4326.latlngToPoint())
(duration: 0:00:02.095936, name: EPSG:4326.pointToLatLng())


AFTER
(duration: 0:00:02.536865, name: Concrete type: EPSG:3857.latLngToCoordinates())
(duration: 0:00:03.327685, name: Concrete type: EPSG:3857.latLngToPoint())
(duration: 0:00:04.135330, name: EPSG:3857.latLngToCoordinates())
(duration: 0:00:04.808905, name: EPSG:3857.latlngToPoint())
(duration: 0:00:04.275039, name: EPSG:3857.pointToLatLng())
(duration: 0:00:01.148857, name: EPSG:4326.latLngToCoordinates())
(duration: 0:00:01.656589, name: EPSG:4326.latlngToPoint())
(duration: 0:00:02.149977, name: EPSG:4326.pointToLatLng())
```

### DIFF
--- a/lib/src/geo/crs.dart
+++ b/lib/src/geo/crs.dart
@@ -322,6 +322,7 @@ abstract class Projection {
 
   const Projection(this.bounds);
 
+  @nonVirtual
   Point<double> project(LatLng latlng) {
     final (x, y) = projectXY(latlng);
     return Point<double>(x, y);
@@ -329,6 +330,7 @@ abstract class Projection {
 
   (double, double) projectXY(LatLng latlng);
 
+  @nonVirtual
   LatLng unproject(Point point) =>
       unprojectXY(point.x.toDouble(), point.y.toDouble());
   LatLng unprojectXY(double x, double y);
@@ -365,7 +367,7 @@ class SphericalMercator extends Projection {
   const SphericalMercator() : super(_bounds);
 
   static double projectLat(double latitude) {
-    final lat = latitude.clamp(-maxLatitude, maxLatitude);
+    final lat = _clampSym(latitude, maxLatitude);
     final sin = math.sin(lat * math.pi / 180);
 
     return r / 2 * math.log((1 + sin) / (1 - sin));
@@ -445,5 +447,8 @@ class _Transformation {
       );
 }
 
-double _inclusiveLat(double value) => value.clamp(-90, 90);
-double _inclusiveLng(double value) => value.clamp(-180, 180);
+// Num.clamp is slow due to virtual function overhead.
+double _clampSym(double value, double limit) =>
+    value < -limit ? -limit : (value > limit ? limit : value);
+double _inclusiveLat(double value) => _clampSym(value, 90);
+double _inclusiveLng(double value) => _clampSym(value, 180);
--- a/lib/src/geo/crs.dart
+++ b/lib/src/geo/crs.dart
@@ -13,24 +13,34 @@ import 'package:proj4dart/proj4dart.dart' as proj4;
 /// points of objects of different dimensions. In our case 3D and 2D objects.
 @immutable
 abstract class Crs {
-  const Crs();
+  @nonVirtual
+  final String code;
+  @nonVirtual
+  final bool infinite;
+  @nonVirtual
+  final (double, double)? wrapLng;
+  @nonVirtual
+  final (double, double)? wrapLat;
 
-  String get code;
+  const Crs({
+    required this.code,
+    required this.infinite,
+    this.wrapLng,
+    this.wrapLat,
+  });
 
   Projection get projection;
 
-  Transformation get transformation;
-
   /// Converts a point on the sphere surface (with a certain zoom) in a
   /// map point.
+  (double, double) latLngToXY(LatLng latlng, double scale);
   Point<double> latLngToPoint(LatLng latlng, double zoom) {
-    final projectedPoint = projection.project(latlng);
-    return transformation.transform(projectedPoint, scale(zoom));
+    final (x, y) = latLngToXY(latlng, scale(zoom));
+    return Point<double>(x, y);
   }
 
   /// Converts a map point to the sphere coordinate (at a certain zoom).
-  LatLng pointToLatLng(Point point, double zoom) =>
-      projection.unproject(transformation.untransform(point, scale(zoom)));
+  LatLng pointToLatLng(Point point, double zoom);
 
   /// Zoom to Scale function.
   double scale(double zoom) => 256.0 * math.pow(2, zoom);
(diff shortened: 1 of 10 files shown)


ignatz:
For the record, I was also experimenting a bit with `List<LatLng>` APIs in the hopes that the compiler would pick up on it and - I don't know - go full SIMD. For this use-case it would be really awesome, however I didn't have much success. Maybe still worth looking into it more.

.gitignore
```diff
@@ -1,4 +1,5 @@
 # Miscellaneous
+*.exe
```
JaffaKetchup, line 2:
```suggestion
```

JaffaKetchup (commented):
I'll leave this to someone else to review, as CRS really isn't my strength.
I did notice from your on-device comparison that the Raster times increased significantly? I'm assuming that was a coincidence, as these changes shouldn't place any more strain on the raster thread.

.gitignore
ignatz: > I'll leave this to someone else to review, as CRS really isn't my strength.

No worries, though there's zero math involved. This change is mostly just moving things around to avoid overhad.

>  I did notice from your on-device comparison that the Raster times increased significantly? I'm assuming that was a coincidence, as these changes shouldn't place any more strain on the raster thread.

That's probably accidental. The raster times shouldn't change at all.

I did continue my explorations. I've tried for example vector operations with Float64x2 and single-precision floats with Float32x4. None of them did too much, happy to share my benchmark results if there's more interest. However, I did find two more worthwhile optimizations: 
 * avoid Num.clamp
 * monomorphize the common EPSG3857/spherical mercator case.

This saved another 22% cpu time. Which if you compare HEAD to this PR is a ~41% reduction for my polygon testcase (your mileage may vary). You can also see that the raster thread remains unchanged.

reduction

JaffaKetchup: Wow, I'm surprised that `clamp` is that inefficient, I'll have to remember that.

ignatz: In absolute terms it's not super slow, it's just that clamping is such a cheap operation that the virtual function overhead is large in relative terms, especially when you call it in a tight loop.

ignatz:
The change isn't as bad as it may look at first glance. It's mostly a cleanup and a bit of shuffeling. Is there something I can do to help the process?

JaffaKetchup:
@ignatz Apologies for being slow, I'll try to have a look tomorrow.

JaffaKetchup:
(Failing code analysis will be fixed by #1704)

lib/src/geo/crs.dart
```diff
@@ -303,156 +309,144 @@ class Proj4Crs extends Crs {
...
-  LatLng unproject(Point point);
+  final Bounds<double>? bounds;
 
-  double _inclusive(double start, double end, double value) {
-    if (value < start) return start;
-    if (value > end) return end;
+  const Projection(this.bounds);
 
-    return value;
+  Point<double> project(LatLng latlng) {
```
JaffaKetchup, line 325: Does it ever make sense to override this? Maybe `@nonVirtual` as well?

lib/src/geo/crs.dart
```diff
@@ -303,156 +309,144 @@ class Proj4Crs extends Crs {
...
-  double inclusiveLat(double value) {
-    return _inclusive(-90, 90, value);
-  }
+  (double, double) projectXY(LatLng latlng);
 
-  @protected
-  double inclusiveLng(double value) {
-    return _inclusive(-180, 180, value);
-  }
+  LatLng unproject(Point point) =>
```
JaffaKetchup, line 332: Does it ever make sense to override this? Maybe `@nonVirtual` as well?

JaffaKetchup (changes requested):
This is what I can see from a once over, but I'm probably missing a couple other small things, so it'd be great if someone else could take a look. Apologies for being so slow, I think we're just having a little downtime after the v6 release.

lib/src/geo/crs.dart
```diff
@@ -13,24 +13,34 @@ import 'package:proj4dart/proj4dart.dart' as proj4;
 /// points of objects of different dimensions. In our case 3D and 2D objects.
 @immutable
 abstract class Crs {
```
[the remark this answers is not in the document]
ignatz: I don't know. Do you want to prevent folks from implementing their own coordinate systems? :shrug: Would definitely break any implementations that already exist.

lib/src/geo/crs.dart
```diff
@@ -13,24 +13,34 @@ import 'package:proj4dart/proj4dart.dart' as proj4;
...
+  });
 
   Projection get projection;
 
-  Transformation get transformation;
-
   /// Converts a point on the sphere surface (with a certain zoom) in a
   /// map point.
+  (double, double) latLngToXY(LatLng latlng, double scale);
   Point<double> latLngToPoint(LatLng latlng, double zoom) {
```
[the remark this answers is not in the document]
ignatz: I guess not since this file contains overrides

lib/src/geo/crs.dart
```diff
@@ -39,156 +49,147 @@ abstract class Crs {
...
-@immutable
-class Epsg4326 extends Earth {
-  @override
-  final String code = 'EPSG:4326';
+      : super(
+          code: 'EPSG:3857',
+          transformation: const _Transformation(_scale, 0.5, -_scale, 0.5),
+          projection: const SphericalMercator(),
+          infinite: false,
+          wrapLng: const (-180, 180),
```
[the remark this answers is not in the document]
ignatz: Maybe. Note that this is just how it was before: https://github.com/fleaflet/flutter_map/blob/master/lib/src/geo/crs.dart#L95

lib/src/geo/crs.dart
```diff
@@ -39,156 +49,147 @@ abstract class Crs {
...
   const Epsg4326()
-      : projection = const _LonLat(),
-        transformation = const Transformation(1 / 180, 1, -1 / 180, 0.5),
-        super();
+      : super(
+          projection: const _LonLat(),
+          transformation: const _Transformation(1 / 180, 1, -1 / 180, 0.5),
+          code: 'EPSG:4326',
+          infinite: false,
+          wrapLng: const (-180, 180),
```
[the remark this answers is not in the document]
ignatz: Same as other comment. HEAD version doesn't define any wrapLat.

lib/src/geo/crs.dart
ignatz: Done. Crossing my fingers that nobody overwrote this :)

lib/src/geo/crs.dart
ignatz: Done. Crossing my fingers that nobody overwrote this :)

lib/src/geo/crs.dart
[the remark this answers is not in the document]
JaffaKetchup: Would this prevent them from implementing their own? Afaik `base` doesn't imply `final`, it just forces extension (instead of implementation) only, which seems to be what you want, as you are using `@nonVirtual`?

lib/src/geo/crs.dart
JaffaKetchup: Have I missed something? Shouldn't people just override `projectXY`?

lib/src/geo/crs.dart
JaffaKetchup: Have I missed something? Shouldn't people just override `unprojectXY`?

lib/src/geo/crs.dart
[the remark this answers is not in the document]
ignatz: You're correct that base and final are different. My statement is based on the documentation "A base class disallows implementation outside of its own library." https://dart.dev/language/class-modifiers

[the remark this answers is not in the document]
JaffaKetchup: But `implements`ation doesn't make sense if any fields are marked `nonVirtual`. So it would seem to me as though `base` is just reinforcing what you already want.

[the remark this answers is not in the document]
ignatz: Sorry, maybe I'm terribly confused or we're talking past each other.

What I understand atm: I made a few abstract methods members and marked them as nonVirtual because they no longer need to be implemented. Yet, there are still abstract methods left that need implementing. There are several CRS implementations in this file. AFAIU, adding the "base" class-modifier won't do anything other than disallow implementations of CRS outside this module.

If that's what we want, I'll add the modifier. I don't understand how this relates to some members being marked @nonVirtual.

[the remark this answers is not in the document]
JaffaKetchup: Whoops, totally another mistake from me. I thought `@nonVirtual` will cause a lint error for classes that `implements` rather than `extends`, but nope!

JaffaKetchup (approved):
I think I'm happy enough with this now, thanks. Still would like a review from someone else! (@mootw @josxha @TesteurManiak)

Robbendebiene:
Just stumbled across this.

> I only wish, I could point to the micro benchmarks and see a similar uplift. For transparency:

It appears you are using `StopWatch` for this measurement. I also tried measuring performance this way sometime ago and got completely unexpected results. Using the [benchmarking package](https://pub.dev/packages/benchmark_harness) gave me way more reasonable results. I honestly don't know why because under the hood they are also using `StopWatch`. Perhaps you can give it a try.

ignatz:
That's an interesting suggested. I might take a look in the future (however it does sound rather unexpected given that both use stopwatch).

That said, with the two changes stacked up there's a clear improvement. This is where we stand right now:

### Microbenchmark:

* Before:
```
(duration: 0:00:04.406465, name: Concrete type: EPSG:3857.latLngToPoint())
(duration: 0:00:03.859791, name: EPSG:3857.latlngToPoint())
(duration: 0:00:04.050427, name: EPSG:3857.pointToLatLng())
(duration: 0:00:01.334883, name: EPSG:4326.latlngToPoint())
(duration: 0:00:02.050565, name: EPSG:4326.pointToLatLng())
```
* After
```
(duration: 0:00:02.009074, name: Concrete type: EPSG:3857.latLngToXY())
(duration: 0:00:02.292637, name: Concrete type: EPSG:3857.latLngToPoint())
(duration: 0:00:02.593093, name: EPSG:3857.latLngToXY())
(duration: 0:00:03.882431, name: EPSG:3857.latlngToPoint())
(duration: 0:00:03.500675, name: EPSG:3857.pointToLatLng())
(duration: 0:00:01.160899, name: EPSG:4326.latLngToXY())
(duration: 0:00:01.652765, name: EPSG:4326.latlngToPoint())
(duration: 0:00:00.649775, name: EPSG:4326.pointToLatLng())
```

Which boasts some solid speedups:

* Monomorphized: EPSG:3857.latLngToXY() vs EPSG:3857.latLngToPoint() is a **54%** reduction
* Non monomorphized EPSG:3857.latlngToPoint : **33%** reduction
* EPSG:4326 a 13% reduction (but could easily be improved the same as for EPSG:3857)
* And pointToLatLng (which we care less about): EPSG:3857: 4% and EPSG:4326: 68% (not sure if the latters absolute numbers should be trusted. Almost feels too good)


### and in practice:
image

josxha:
The `EPSG4326` page of the example app has weird behaviour. Other CRS systems seem to work correctly. Compared with `master` and the issue does not happen there. 

image

Errors in the console:
```
flutter: Exception: Invalid image data
```

ignatz:
> The `EPSG4326` page of the example app has weird behaviour. Other CRS systems seem to work correctly. Compared with `master` and the issue does not happen there.
> 
> image
> 
> Errors in the console:
> 
> ```
> flutter: Exception: Invalid image data
> ```

Yikes. I can totally tell `x` and `y` apart :hide:. My bad, should have tested more. Thanks for the catch!

josxha (approved):
lgtm!
@JaffaKetchup if you want you can merge. 
It's a breaking so we can't include it in v6.1 tho.

JaffaKetchup:
Thanks @ignatz !

lib/src/geo/crs.dart
```diff
@@ -13,24 +13,34 @@ import 'package:proj4dart/proj4dart.dart' as proj4;
...
+  const Crs({
+    required this.code,
+    required this.infinite,
+    this.wrapLng,
+    this.wrapLat,
+  });
 
   Projection get projection;
 
-  Transformation get transformation;
```
JosefWN, line 22: Great changes @ignatz! But how can I access `transformation.transform` now?

My use case is I have `Point`s in the local CRS but I want to transform them to `Point`s corresponding to a given scale. This is something that's used here and there in my codebase, since I work primarily with `Point` rather than `LatLng`.

ignatz: The main reason the transformation is no longer part of the crs is that its an implementation detail that wasn't even well defined for Proj4Crs.

It sounds like you're already married to a specific CRS if you're operating in the CRS-dependent `Point` space rather than the more independent but earth-specific CRS (not sure if you were the person doing office floors). With that in mind:

* We could make _CrsWithStaticTransform public and you could down-cast your crs.
* If you're brave you can probably cross-cast to dynamic and still call the methods.
* Or what I would do: you can define your own transform since you already seem married to a specific CRS. At the end of the day, it's a few lines of simple math and the the implementation of Transform itself has also changed away from Point<double> for performance reasons. Mathing on reified generics like `Point` introduces virtual function overhead to every simple operation.

JosefWN: It's tempting to think that latitudes and longitudes can represent every location on Earth, so why not use them? In practise, it's quite problematic to work with them in the Arctic and the Antarctic. The locations may be exotic, but there is nothing exotic about my coordinates, they are just as "earth-bound" as latitudes and longitudes, although they happen to be on a plane:
https://pro.arcgis.com/en/pro-app/3.1/help/mapping/properties/stereographic.htm
https://www.youtube.com/watch?v=pLdK1AlMpqY

In fact, if you look at maps of the Arctic and Antarctic, they are quite likely to be some variant of polar stereographic projection where the latitude lines are forming concentric circles around the pole (they are not in WGS84 for sure). One example is available in this repo: https://github.com/fleaflet/flutter_map/blob/master/example/lib/pages/epsg3413_crs.dart

With `Proj4Crs` you can map a coordinate from a plane to a sphere for example, so I'm sure there can be cases where `transform.transform` doesn't work. For us, on the other hand, we only have a 2D viewport, we need to map the coordinates onto a plane at some point before we render them (the raster sent to the GPU will be an XY-plane), regardless of the source coordinate system. Once they are projected on a plane, I would say there is no problem with `transformation.transform`, or are you referring to something specific?

Once you have a projected coordinate, regardless of whether its origin is lat/lon or something else, `transformation.transform` is a convenient utility method, is what I suppose I'm trying to say.

My goal has been to at least keep the `flutter_map` core reasonably easy to use with a custom CRS (https://github.com/fleaflet/flutter_map/pull/1295), so I only have to reimplement layer code where need be (which is still quite a handful to maintain), and it actually works, it even performs better than the stock `flutter_map`... I only use `LatLng` for the `TileLayer` 😄

ignatz: > Once they are projected on a plane, I would say there is no problem with transformation.transform, or are you referring to something specific?

Specifically to the implementation of Proj4Crs, as it was before, which uses multiple transforms.

If you have improvements for FM, maybe upstream them if nothing else it would reduce such churn in maintaining your fork. The project owners are generally very welcoming of PRs. As a non-owner I would certainly appreciate more speed ups.

FWIW, I never proposed for you to use LatLng and I don't disagree with transform being useful 🤷‍♀️. I only said it didn't fit the Proj4Crs implementation and that's a NPM-leftPad-level abstraction.

Unrelatedly, I understand that stereographic projections typically only show one hemisphere and thus not all LatLng can be mapped. That said it's nor clear to me why you couldn't work with LatLng or introduce a crs for your usecase.

JosefWN: > Unrelatedly, I understand that stereographic projections typically only show one hemisphere and thus not all LatLng can be mapped. That said it's nor clear to me why you couldn't work with LatLng or introduce a crs for your usecase.

The problem isn't that my CRS is local/limited to one hemisphere, it's the other way around: using WGS84 (lat/lon, i.e. not what I am using but what `flutter_map` promotes), is not feasible in the Arctic regions because the farther you go from the equator, the greater the projection distortions. There is a WGS84/EPSG:4326 example showing the whole world in the repo, all the way out to the poles, there you can see that although you can use the lat/lon approach in theory, you can hardly make out any details up north: https://github.com/fleaflet/flutter_map/blob/master/example/lib/pages/epsg4326_crs.dart

Just trying to say that `LatLng` is not always a feasible choice, and far from the only one when it comes to maps. I use a CRS, and the guy with the indoor maps can admittedly construct a planar dummy CRS that minimizes local distortions, something along the lines of centering on the north pole with a really tight latitude of true scale, but unlike mapping lat/lons from a sphere to a plane, our coordinates are already planar so there is no need to project anything at all. The only time I use the CRS is when a user picks a point on the map and wants to know which lat/lon that point corresponds to, which happens rarely.

By moving the projection step out of the layer itself it would not only open up for planar CRS to use the layers as-is without hacks, even if you use `LatLng` you could keep the projected `List<Point>` alive across frames, and pan and zoom all you want without re-projecting, leaving only the scaling step from point to pixel to the layer (using some equivalent to `transformation.transform`).

Anyway, sorry for the rant. I just get the feeling that `flutter_map` is slowly drifting even farther from my fork, and that if after a couple of years with the repo I feel that the custom CRS situation feels quite hairy, it must feel even worse to new users. "Why don't you just use `LatLng`?" is an often recurring question, if I asked: Why don't you just use a planar CRS? You would probably say that in your circumstances it wouldn't make any sense (although in theory you certainly _can_). It's the same for me really, `LatLng` would only serve to degrade the performance of the app and introduce illogical workarounds in the code.

ignatz: I don't mind. I'm open to all your suggestions (as a non owner) and would still encourage you to upstream as much as possible of your changes.

I'm still not sure if there's a comms issue, I never once suggested for you to use LatLng. At least now I feel like I understand your issues better, maybe? I understand your projection concerns around the poles, which is different from what you said before. Do I understand correctly that LatLng as a coordinate system would work fine for you unlike many projections such as WGS84, which introduce severe distortions. This seems very different from the issue with the office floors.

What would happen if you had your own stereographic Crs?

JosefWN: Summarizing it required a bit of background, it turned into more of an article 😅

I put it in a separate issue to open up for some discussion: https://github.com/fleaflet/flutter_map/issues/1793

PR #1728, opened 2023-11-10
feat: add hit detection to `Polyline`s

ignatz:


JaffaKetchup:
@ignatz With all of that decided, it'd be great if you could fix the conflicts here, and just add Hugo's name to the top of the new code (in a comment or something) :)

JaffaKetchup:
Also just need to confirm how this PR will interact with #1704, and which should be merged first.

ignatz:
> and just add Hugo's name to the top of the new code (in a comment or something) :)

Help me out. What would you like me to write/credit? I'll put there whatever you'd like me to put.

It's not like FM generally credits authors (which I like) nor is this code based on Hugo's code. I feel a bit silly to mention that there used to be an unrelated plugin that did something similar.

JaffaKetchup:
Yeah, we don't normally credit code unless we're absorbing it. In this case, I feel like we should thank Hugo, just something like "Thanks to Hugo H for maintaining tappability until v7, now maintained here for flexibility".

ignatz:
Sounds reaonsable if there was some sort oft absorbing happening. "Thanks to Hugo for maintaining tappability until v7, now maintained here for flexibility" feels misleading. It seems to imply a lineage where there's none. I also don't quite understand the flexibility argument. If at all, it might be worth pointing out to users that these are entirely separate implementations that behave semantically differently.

josxha:
I would agree that adding it on top of the implementation could be misleading. It makes sense to me if we want to hint that code got integrated to show the copyright but in this case it's different. 🤔 
Maybe the migration docs, release notes and/or changelog fits that message better?

JaffaKetchup:
Ok, sure, we'll leave it out of here then, and add it to the docs site.

lib/src/layer/polyline_layer.dart
```diff
@@ -112,6 +148,72 @@ class PolylinePainter extends CustomPainter {
...
+    Polyline? hit;
+
+    outer:
+    for (final p in polylines.reversed) {
+      // TODO: For efficiency we'd ideally filter by bounding box here. However
+      // we'd need to compute an extended bounding box that accounts account for
+      // the stroke width.
+      // if (!p.boundingBox.contains(touch)) {
+      //   continue;
+      // }
```
josxha, line 169: Remove the "TODO" from the comment please. I'd say this as a future performance enhancement and therefore not a "real" todo that would otherwise get highlighted by IDEs.

josxha (changes requested):
Haven't experienced any bugs while testing. Thanks @ignatz for adding a nice showcase to the example app too.
Could you check my comments?

lib/src/layer/polyline_layer.dart
```diff
@@ -54,45 +57,78 @@ class Polyline {
...
   final List<Polyline> polylines;
-  final bool polylineCulling;
+  final bool interactive;
 
   const PolylineLayer({
     super.key,
     required this.polylines,
-    this.polylineCulling = false,
+    //@Deprecated('Let's always cull')
+    bool polylineCulling = true,
```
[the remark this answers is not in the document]
JaffaKetchup: #1704 replaces all the culling logic anyway, so it's best for this to be left out of here, along with anything else that changes culling.

JaffaKetchup:
Although this is technically non-breaking, it will not be included in v6.1.0, because of its impact on the plugin.

JaffaKetchup:
OK, as @ignatz seems to be taking a break, I'm going to see if I can polish off what @josxha reviewed above, and fix the conflicts, and hopefully we can get this in a mergeable-state. @ignatz feel free to jump in and/or stop me!

josxha:
Sounds good. Not sure how to handle the `_LastHit` tho. It could be that ignatz had good reason to use a wrapper class here which I'm not aware of. We can just leave it as it is now but up to you @JaffaKetchup.

ignatz:
> OK, as @ignatz seems to be taking a break, I'm going to see if I can polish off what @josxha reviewed above, and fix the conflicts, and hopefully we can get this in a mergeable-state. @ignatz feel free to jump in and/or stop me!

I'm happy either way. I can also give it another stab, I hadn't given up on it. It just slipped lower in my priority list after the initial momentum went a bit on a circle. For context, I'm not in a rush nor blocked. I've been maintaining my own FM fork for a long while now. My goal was merely to give back to you folks and to minimize the differences I have to maintain.

> Sounds good. Not sure how to handle the _LastHit tho. It could be that ignatz had good reason to use a wrapper class here which I'm not aware of. We can just leave it as it is now but up to you @JaffaKetchup.

It's an output argument or a bit of a "hack" if you will. It's wrapper is passed by the widget to the painter, so the painter can update the wrapper's contents and thus bubble up the information to the parent widget's gesture detector. Maybe more idiomatically, I could have also passed down a buffered Stream for the hits to bubble up from the painter to the widget, however that felt a bit heavy handed w/o a clear benefit. Open to suggestions

lib/src/layer/polyline_layer.dart
```diff
@@ -20,6 +21,7 @@ class Polyline {
   final StrokeCap strokeCap;
   final StrokeJoin strokeJoin;
   final bool useStrokeWidthInMeter;
+  final void Function(LatLng point)? onTap;
```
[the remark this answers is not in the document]
JaffaKetchup: Removed `interactive` parameter.

lib/src/layer/polyline_layer.dart
```diff
@@ -54,45 +57,78 @@ class Polyline {
...
+      child: GestureDetector(
+        behavior: HitTestBehavior.deferToChild,
+        onTap: () {
+          final hit = lastHit.hit;
+          if (hit == null) return;
+
+          final onTap = hit.polyline.onTap;
+          if (onTap != null) {
+            onTap(hit.point);
+          }
```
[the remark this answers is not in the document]
JaffaKetchup: Shortened.

ignatz:
> I want to experiment with adding more options for hit behaviour - for example, calling every polyline's callback when they overlap, or making non-tappable lines transparent rather than opaque as they are now. 

Nothing against different behaviors but I would suggest to think about use-cases first and then work your way backwards to what's actually needed. What I implemented is what I thought would be a sensible default, which is up for discussion.

As to your concrete suggestions, I'm a bit hard pressed to think of good use cases. Most users wouldn't expect polylines to be translucent, since most tappable objects are not. As for transparent non-tappable polylines, maybe, but you can also always sort your lines :woman_shrugging:

JaffaKetchup:
Yeah, I think what you implemented is exactly what most people will want, I'm just thinking about flexibility so we don't have to revisit unnecessarily in future.

Ok, so an example for firing all callbacks: a hiking app where the user needs to select a single line in a remote place, where all the lines are going along one path. Having each line fire separately will then allow (in conjunction with `MapOptions.onTap`) a modal sheet to appear where the user can select the correct line.
The issue with sorting the lines to move all non-tappable ones to the bottom is that then they'll also be visually at the bottom, which may not be wanted.

Please do say if there's something I've missed - but it's not like I've ever missed anything before ever 👀😂.

I've realised why you have the separate `GestureDetector` now as well :D. `hitTest` also fires for hover events, which is why I suddenly have so many modals open (I'm keeping a list of tapped lines instead of just one). I can't finish it now, hopefully tomorrow.

ignatz:
> The issue with sorting the lines to move all non-tappable ones to the bottom is that then they'll also be visually at the bottom, which may not be wanted.

I don't think it would help with the translucency case but I'm still not sure about an obvious use case where you wanted to draw a line visually over another line to then have the user tap the one below. I'm big on errorring on the keep-it-simple side and add functionality as use-cases arise. At the end, this is super easy to implement and would probably take <10min to write the code.

> Ok, so an example for firing all callbacks: a hiking app where the user needs to select a single line in a remote place, where all the lines are going along one path. Having each line fire separately will then allow (in conjunction with MapOptions.onTap) a modal sheet to appear where the user can select the correct line.

This sounds very reasonable. FWIW, this is exactly what I'm doing with (not-upstreamed) tappable polyGOONS :). That said, if I wanted to address that use-case I wouldn't have per-polyline on-tap-callbacks but rather a single layer callback that receives a sorted list of tapped polylines. Otherwise, as you described it with `MapOptions.onTap` to de-de-multiplex the de-multiplex, just puts extra complexity on the user.

In other words, if we wanted to cover or plan for that use-case we should probably already start out with only a layer-level on-tap-callback unless we want both. WDYT?

ignatz:
Personally, I wouldn't do both. I'd go with either. I do think that `PolylineLayer.onTap` will work **much** better for the translucent case, while `Polyline.onTap` is the natural choice if you didn't want to support translucency (since you don't have to deal with keys etc). Going with `PolylineLayer.onTap` would also avoid the distinction between tappable and non-tappable polylines.

JaffaKetchup:
@ignatz Managed to solve that issue. It's not that much more complicated to support both, and support better overlap handling. Let me know what you think of the implementation now (also @josxha @mootw wdyt?) - the actual detection algorithm is the same as you wrote, I've just changed how output is handled. I'm happy to make changes or revert if you think it's too much.

Tbh, I'm also thinking that maybe we should add support for the secondary taps/clicks, hovers, and holds. It's only one more step away :D.

ignatz:
> Tbh, I'm also thinking that maybe we should add support for the secondary taps/clicks, hovers, and holds. It's only one more step away :D.

I don't wanna hold you back by any means, I'm merely wondering why we're trying to solve everything at once. The additions all seem fairly incremental, especially if you want to support both onTap approaches.

I guess the next step would be to carry all these features over to polygons :)

JaffaKetchup:
> I don't wanna hold you back by any means, I'm merely wondering why we're trying to solve everything at once. The additions all seem fairly incremental, especially if you want to support both onTap approaches.

I just feel like if we're going to release a new feature, we should be doing it well, and implementing it as comprehensively as possible first time. Saves time in future triaging and thinking of workarounds and releasing etc.
Also, both approaches are now supported in this PR, check out the latest commit.

> I guess the next step would be to carry all these features over to polygons :)

Indeed! We're open to PRs for that as well, and it would be amazing to introduce them both in v7. We can carry over the 'meta'/surrounding template/logic from here as well iyswim, if we do everything right now.

JaffaKetchup:
Hey @ignatz, would appreciate your thoughts if you have any!

ignatz:
Sorry, on what specifically? I did express thoughts before. Would you like me to review the code?

lib/src/layer/polyline_layer.dart
```diff
@@ -39,66 +79,292 @@ class Polyline {
...
+  /// Defaults to `true`.
+  final bool nonTappablesObscure;
+
+  /// Whether a tappable [Polyline] should prevent taps from being handled
+  /// on all [Polyline]s beneath it, at overlaps
+  ///
+  /// "Tappable" means that either or both of [Polyline.onTap] and
+  /// [Polyline.tapKey] has been defined.
+  ///
+  /// If `false`, then [onTap] becomes redundant to [Polyline.onTap].
```
JaffaKetchup, line 163: Whoops, that should probably be `true` not `false`.

```suggestion
  /// If `true`, then [onTap] becomes redundant to [Polyline.onTap].
```

JaffaKetchup:
You're right, I just hoped that I've found a solution that isn't too messy. Just wanted to check with you that you couldn't see any glaring issues with this approach that I've missed.

lib/src/layer/polyline_layer.dart
```diff
@@ -39,66 +79,292 @@ class Polyline {
...
+          hits.add(p);
+          if (tappablesObscure) break outer;
+        }
+      }
+    }
+
+    if (hits.isEmpty) return false;
+
+    lastHit!.hit = (
+      // Remove duplicates caused when the hit is close to two segements
```
ignatz, line 361: I don't understand this. Why would you end up with two segments of the same polyline? Why wouldn't you stop after you found the first hit segment?

lib/src/layer/polyline_layer.dart
```diff
@@ -39,66 +79,292 @@ class Polyline {
...
+      tapped: LinkedHashSet<Polyline<TapKeyType>>.from(hits).toList(),
```
ignatz, line 362: Hit is being called when ever hit detection happens, also on hovers etc. Or  in other words when you move the mouse it's called on every frame. It would be nice to avoid:

1. the deduping as mentioned above
2. and the list allocation. Could you call clear on the pre-existing list instead?

lib/src/layer/polyline_layer.dart
```diff
@@ -39,66 +79,292 @@ class Polyline {
...
+      size: Size(camera.size.x, camera.size.y),
+      isComplex: true,
+    );
+
+    return MobileLayerTransformer(
+      child: interactive
+          ? MouseRegion(
+              hitTestBehavior: HitTestBehavior.deferToChild,
+              cursor: SystemMouseCursors.click,
+              child: GestureDetector(
```
ignatz, line 212: The GestureDetector code has certainly gotten quite involved with all the features, i.e. the cross product between different modal interactions and stacking/obscuring/multi-hit options. I'm slightly worried that this looks a bit brittle...

...but I'm wondering even more if in hindsight and with the desire for so much flexibility, it would have been better to instead pass up the hit detection result so that users could set up a GestureDetector as a parent of the Polyline layer that would leave you with ultimate flexibility w/o the PolylineLayer having to anticipate all use-cases.

lib/src/layer/polyline_layer.dart
```diff
@@ -39,66 +79,292 @@ class Polyline {
...
+  /// Whether a non-tappable [Polyline] should prevent taps from being handled
+  /// on all [Polyline]s beneath it, at overlaps
+  ///
+  /// "Tappable" means that either or both of [Polyline.onTap] and
+  /// [Polyline.tapKey] has been defined.
+  ///
+  /// Applies to both [onTap] and every [Polyline.onTap].
+  ///
+  /// Defaults to `true`.
+  final bool nonTappablesObscure;
```
ignatz, line 155: :nit: as a non-native-speaker, "occlude" seems to me the more common verb for this in computer graphics :woman_shrugging:

JaffaKetchup: I suppose you're right given that "obscure" implies sight. I'll change it.

lib/src/layer/polyline_layer.dart
JaffaKetchup: As we're handling multiple `Polyline`s now, and the detection algorithm seems to rely on comparing with each segment of the `Polyline`, a hit within two or more segments will trigger a result for the same line twice.

I guess this could be fixed by tweaking the detection algorithm to not depend on segments, but rather check the entire line, but I have no idea how to do that, and you're the expert there!

image

lib/src/layer/polyline_layer.dart
JaffaKetchup: The deduping is purely for the reason mentioned above, it does not happen in order to dedupe from non-click hit detections.

I can try using a single list, I'll see if that works.

lib/src/layer/polyline_layer.dart
JaffaKetchup: I know exactly what you mean and I agree it looks a little overgrown, there's suddenly a lot of options. (Although do remember that the in-code docs take up quite a few of the diff lines :D).

However, passing the result up seems less than ideal as well - we're replacing a plugin here so we should be a drop in replacement as far as possible. It would also seem a little half-baked to me.
It would definitely be nice if we could refactor all the handling of all three gesture types, as there's a lot of code duplication there, but I don't see an immediately clear way to do that.
Is there a better middle ground to be struck?

lib/src/layer/polyline_layer.dart
JaffaKetchup: Fixed in f820807. This (amongst the other stuff) is why I asked for your input :D!

lib/src/layer/polyline_layer.dart
```diff
@@ -9,7 +12,16 @@ import 'package:flutter_map/src/misc/offsets.dart';
...
 import 'package:latlong2/latlong.dart';
 
-class Polyline {
+class _LastHit<TapKeyType extends Object> {
+  ({List<Polyline<TapKeyType>> tapped, LatLng point})? hit;
+}
+
+typedef PolylineOnTap = void Function(LatLng point);
+typedef PolylineLayerOnTap<TapKeyType extends Object> = void Function(
+    List<TapKeyType> tappedKeys, LatLng point)?;
```
josxha, line 21: Could we use List<Polygon> here? This would give users more flexibility.

josxha (changes requested):

example/lib/pages/polyline.dart
```diff
@@ -82,22 +131,75 @@ class _PolylinePageState extends State<PolylinePage> {
...
+            const Text(
+              'Tapped Polyline(s)',
+              style: TextStyle(fontSize: 20, fontWeight: FontWeight.bold),
+            ),
+            Text(
+              '${[
+                'Tapped',
+                'Long-pressed/held',
+                'Secondary-tapped/alternative-clicked'
+              ][mode]} at point: $coords',
```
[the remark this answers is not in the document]
josxha: oh this appears to be solved.😅

lib/src/layer/polyline_layer.dart
ignatz: > I guess this could be fixed by tweaking the detection algorithm to not depend on segments, but rather check the entire line, but I have no idea how to do that, and you're the expert there!

I don't know how you'd simply check the entire line instead of segments. However, as we iterate the segments sequentially and detect a hit, we stop and move on to the next polyline. No need to continue iterating segments for the same polyline.

lib/src/layer/polyline_layer.dart
JaffaKetchup: If we just pass around the `Polyline` instead of the key, then we don't need the `TapKeyType` at all - users can just use an `Object`.
I was concerned about the performance of passing the entire object along everywhere, but I guess Dart is smart enough to pass by reference. I'll replace it.

lib/src/layer/polyline_layer.dart
ignatz: > we're replacing a plugin here so we should be a drop in replacement as far as possible.

I still don't get where you got this idea from:

* FlutterMap does not need to achieve feature parity with some third-party plugin nor should it.
* The third party plugin will continue to work, there's no need for anyone to migrate.
* Even despite all your feature parity work, APIs and hit detection semantics are quite different. There's no trivial migration.
* My original PR doesn't share any ancestry with said plugin. Nobody owes anyone anything.

This sounds harsh but I come with good intentions. If you let go of the idea just a little bit,  we could instead discuss what makes the most sense for FM core rather than chasing a predefined goal.

This ties into our earlier discussions, where I wanted to error on the side of adding things incrementally as use-cases arise. You wanted to offer a more feature-full package. Both are fine. The truth is probably somewhere in the middle.

Just keep in mind that there's a real asymmetry: **It will always be easier to add new things, than take things away**

example/lib/pages/polyline.dart
```diff
@@ -26,16 +28,49 @@ class _PolylinePageState extends State<PolylinePage> {
         ),
         children: [
           openStreetMapTileLayer,
-          PolylineLayer(
+          PolylineLayer<_TapKeyType>(
+            onTap: (tappedLineKeys, coords) => _openTappedLinesModal(
+              0,
+              tappedLineKeys,
+              '(${coords.latitude.toStringAsFixed(6)}, ${coords.longitude.toStringAsFixed(6)})',
```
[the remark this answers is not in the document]
JaffaKetchup: Resolved.

lib/src/layer/polyline_layer.dart
```diff
@@ -9,7 +12,16 @@ import 'package:flutter_map/src/misc/offsets.dart';
 import 'package:flutter_map/src/misc/point_extensions.dart';
 import 'package:latlong2/latlong.dart';
 
-class Polyline {
+class _LastHit<TapKeyType extends Object> {
+  ({List<Polyline<TapKeyType>> tapped, LatLng point})? hit;
```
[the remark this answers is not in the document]
JaffaKetchup: Resolved.

lib/src/layer/polyline_layer.dart
```diff
@@ -22,6 +34,34 @@ class Polyline {
...
+  /// See [PolylineLayer.onSecondaryTap], [PolylineLayer.onTapTolerance],
+  /// [PolylineLayer.nonTappablesObscure], and [PolylineLayer.tappablesObscure]
+  /// for more information.
+  final PolylineOnTap? onSecondaryTap;
+
+  /// Custom value that identifies this particular [Polyline] when used with
+  /// [PolylineLayer.onTap]/[PolylineLayer.onLongPress]/
+  /// [PolylineLayer.onSecondaryTap] (either instead of or in addition to
+  /// [onTap]/[onLongPress]/[onSecondaryTap])
+  final TapKeyType? tapKey;
```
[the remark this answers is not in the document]
JaffaKetchup: Resolved with internal conversation. For full reasoning, "trust me" :D.

lib/src/layer/polyline_layer.dart
JaffaKetchup: Resolved.

lib/src/layer/polyline_layer.dart
JaffaKetchup: Ah, I've got you now. There was a bug where I wasn't `break`ing normally after one valid segment had been considered.

lib/src/layer/polyline_layer.dart
josxha: Both is just a pointer that gets passed on so it shouldn't make a difference in performance.
(As I learned from you) `TapKeyType` is still required that the Polyline gets recognized as interactive by the layer. 
But in case other properties of the polygon need to be accessed I think it still makes sense to have List<Polyline> returned. 🤔

JaffaKetchup: Yep, that's how I've implemented it.

lib/src/layer/polyline_layer.dart
JaffaKetchup: I completely understand what you're saying. Me saying "drop in replacement" was a little 'short-sighted'(?).
The only reason I want to give 'credit' is because his plugin is essentially becoming redundant: there's no need for anyone to migrate, but also no need for any new users, and as FM gets away from supported versions, people will migrate away. You're right though that putting it in here isn't appropriate and suggests your work descends from his. Putting it in the docs is a far better idea.

I'll try to drop the idea, let's just leave it here :D.

---

If you want to quickly mockup what you mean by

> but I'm wondering even more if in hindsight and with the desire for so much flexibility, it would have been better to instead pass up the hit detection result so that users could set up a GestureDetector as a parent of the Polyline layer that would leave you with ultimate flexibility w/o the PolylineLayer having to anticipate all use-cases.

, because I don't 100% get what you mean, I'm open to it. Perhaps doing it like that, then providing prebuilt solutions/handlers is the better solution, as we do for some other complex areas (like `TileUpdateTransformer`).

ignatz: > The only reason I want to give 'credit' is because his plugin is essentially becoming redundant: there's no need for anyone to migrate, but also no need for any new users, and as FM gets away from supported versions, people will migrate away. 

My point is, that that's only true *because* you're striving for feature parity. There's tons of 3rd party implementations of layers that have overlapping functionality with core features. There's marker layers with clustering, marker layers with popups, tile providers with offline storage, ... . None of them are redundant, nor should the FlutterMap core try to replicate full set of functionality.
By going into this review making some 3rd party plugin the bottom line we're painting ourselves into a corner. Supporting only a core subset of functionality is ok, likely even desirable. Folks can always ask us to add more features, they can always continue to use someones plugin, or build their own.

The documentation also **should not** acknowledge that there "once was a plugin that did something similar". The docs aren't a fairy tale or archeological piece. Nobody who reads the docs is looking for that information. What the docs **could or should** acknowledge is that there are these curated and working plugins with similar or extended feature sets. If a plugin is fringe, poor quality or defunct, it shouldn't be mentioned.

Look, I'm merely trying to pivot the conversation to: "These are great common use-cases we want to support from the get go. And these are great common use-cases we might want to support in the future" instead of "We need to do X because plugin Y does it".

> , because I don't 100% get what you mean, I'm open to it.

Known problem, most of the time I've no clue what I mean :)

There's the famous quote of "all problems can be solved by another layer of abstraction, except the problem of too many abstractions". Sometimes folks call for "deep modules", which just means any layer should abstract "some good" amount of functionality in order to justify its existence. Otherwise it might as well go away.

Looking at our concrete case,  I see what I would call a "thin abstraction", i.e. the abstraction the PolylineLayer is mostly wrapping GestureDetector/MouseRegion functionality through w/o providing much of an abstraction. Layering this abstractions this way also opens up the rabbit hole of wrapping more and more functionality through (there's still tons of interaction modes missing). For example, my own fork supports mouse hovers to change the color of polylines when the pointer goes over them :). This is not a call for us to wrap even more features. I deliberately ripped it out for this PR because I didn't think it was as generally useful and if there's demand we can always add in the future. Meanwhile I'll maintain it for my use-case.

An essay later, what I was proposing is to change the layering, **to push responsibilities up the stack** giving users all the flexibility w/o PolygonLayer having to support all the fringe use-cases. Imagine an API that would look more like this:

```dart
Widget build(context) {
  final lastHit = LastHit();

  void setHit(LastHit hit) => lastHit = hit;

  return GestureDetector(
    hitTestBehavior: HitTestBehavior.deferToChild,
    onTap: () {
       if (lastHit.hit == null) return;
      
       // Do something.
    },
    child: PolygonLayer(
      polylines: [ /* ... */ ],
      hitSetter: setHit,
    ),
  }
}
```

You're probably going to say that this is harder to use and you're right. We could still support simple taps internally while offering this mechanism for boon like me to get fringe hover functionality w/o having to explicitly support it... at least until there's consensus that everyone wants hover.

TesteurManiak: I tend to agree with @ignatz here, this implementation aims to offer a 1st party alternative to `flutter_map_tappable_polyline`, we don't have to provide exactly the same features nor API. It's not like the package will be deleted after this package is merged, it might become less relevant in the long term but migrating will be the responsibility of the developers who want to migrate, not ours.



> If you want to quickly mockup what you mean by
> 
> > but I'm wondering even more if in hindsight and with the desire for so much flexibility, it would have been better to instead pass up the hit detection result so that users could set up a GestureDetector as a parent of the Polyline layer that would leave you with ultimate flexibility w/o the PolylineLayer having to anticipate all use-cases.
> 
> , because I don't 100% get what you mean, I'm open to it.

I think what he meant here would be that to have a fully flexible implementation of tap events we would need to ensure that tap events emanating from the `PolylineLayer` would propagate correctly so, a `GestureDetector` would be able to catch those events by itself without needing us to add more callbacks in the API to handle some specific use cases.

JaffaKetchup: That does seem like a sensible middle ground. We can just send all the hit polylines up, then the user can decide which are tappable, and which are not, and whether to handle them all, or just the top. I agree that that is probably a better solution.
(I wonder if there is a way to eliminate the `lastHit` stuff from being public, as that is less than ideal.)

ignatz: Agreed, on `lastHit` being a bit ugly, I was mostly trying to keep it similar in my mock. 

A more idiomatic way could look like:

```dart
// Could be a class or record, don't really care
class Hit {
 List<Polylines> lines,
 LatLng point,
};

// ...

final notifier = ValueNotifier<Hit?>(null);
return PolylineLayer(
  //...
  hitNotifier: notifier,
);
``` 

With this approach you could also listen to hits.

JaffaKetchup: @ignatz Yeah I like that. How does the implementation look now?

lib/src/layer/polyline_layer.dart
JaffaKetchup: Resolved.

lib/src/layer/polyline_layer.dart
JaffaKetchup: @ignatz Yeah I like that. How does the implementation look now?

lib/src/layer/polyline_layer.dart
```diff
@@ -39,66 +69,191 @@ class Polyline {
...
+        strokeCap,
+        strokeJoin,
+        useStrokeWidthInMeter,
+        hitKey,
+      ];
+
+  /// Used to batch draw calls to the canvas
+  int get renderHashCode => Object.hashAll(_baseHashableValues);
+
+  @override
```
ignatz, line 109: Why are we doing both: giving the polyline a key identity and a hash identity? Shouldn't either do it?

:nit: this should be slightly more expensive since you're allocating a few more lists. If you restore the original renderHashCode you should be able to do Object.hash(renderHashCode, _, __, ...), i.e. it should be fine to hash the hash at the end of the day that's what they do internally as well and we don't have any cryptographic requirements.

lib/src/layer/polyline_layer.dart
```diff
@@ -39,66 +69,191 @@ class Polyline {
...
+  ///
+  /// Will notify [PolylineHit]s if any [Polyline]s were hit, otherwise will
+  /// notify `null`.
+  final PolylineHitNotifier<TapKeyType>? hitNotifier;
+
+  /// The number of pixels away from a visual line (including any width and
+  /// outline) in which a hit should still register as a hit on the line, and
+  /// trigger a notification to [hitNotifier]
+  ///
+  /// Defaults to 3.
```
ignatz, line 144: It seems quite useful to have this configurable but I would suggest to make the default 0. Larger values can make for a nicer UX when you have non-overlapping lines but they would probably be quite annoying for overlapping lines especially if they're parallel.

lib/src/layer/polyline_layer.dart
```diff
@@ -39,66 +69,191 @@ class Polyline {
...
-  PolylinePainter(this.polylines, this.map) : bounds = map.visibleBounds;
+  _PolylinePainter({
+    required this.polylines,
+    required this.camera,
+    required this.hitNotifier,
+    required this.onTapTolerance,
+  }) : bounds = camera.visibleBounds;
 
-  int get hash => _hash ??= Object.hashAll(polylines);
+  final hits = List<Polyline<TapKeyType>>.empty(growable: true);
```
ignatz, line 191: Nice optimization! Maybe add a short comment that this is to reduce allocations

lib/src/layer/polyline_layer.dart
```diff
@@ -39,66 +69,191 @@ class Polyline {
...
+        final distance = math.sqrt(_distToSegmentSquared(
+          position.dx,
+          position.dy,
+          o1.dx,
+          o1.dy,
+          o2.dx,
+          o2.dy,
+        ));
+
+        if (distance < maxDistance) {
```
ignatz, line 238: No need to it now or ever but if we add back an occlusion parameter you "break outer" here and safe a few cycles.
That said, I do think that the above TODO would address any performance concerns more holistically, i.e. only look at polygons with an overlapping bbox

lib/src/layer/polyline_layer.dart
```diff
@@ -22,6 +46,12 @@ class Polyline {
   final StrokeJoin strokeJoin;
   final bool useStrokeWidthInMeter;
 
+  /// Custom value that can identify this particular [Polyline] after hit
+  /// detection
+  ///
+  /// See [PolylineLayer.hitNotifier] for more info.
+  final TapKeyType? hitKey;
+
```
ignatz, line 54: I see, you also removed the Polyline.onTap for the simplest of use-cases. I'm cool with that, we can always add more stuff in later.

ignatz (commented):
I'm glad we argued. I think I like this better than my original proposal. Gives you all the flexibility w/o much complexity. Definitely let's me minimize my fork even more. Thanks

lib/src/layer/polyline_layer.dart
JaffaKetchup: Yeah, I was trying to avoid hashing twice, but I guess it's OK. Resolved.

lib/src/layer/polyline_layer.dart
JaffaKetchup: But you have to balance with accessibility. I think 3px is alright.

lib/src/layer/polyline_layer.dart
JaffaKetchup: This was your suggestion :D Resolved.

lib/src/layer/polyline_layer.dart
JaffaKetchup: Yeah, I'm aware of this trade-off, but wanted to minimize complicated configuration options.

The TODO seems more complicated and mathematically complex (converting stroke width to geographic space) than the work it would save. I'm hoping @mootw's upcoming PR #1704 will improve all this anyway, and will add culling within lines. Good review point for that tho, need to check whether it accounts for stroke width.

lib/src/layer/polyline_layer.dart
JaffaKetchup: To do that, you'd have to add back in a `GestureDetector`, and at that point, where do you stop :D - might as well add a hold and secondary, and then a `MouseRegion`, etc.

JaffaKetchup:
Ok, everything is ready for final review now :D
I've also written the website documentation for interactivity: https://docs.fleaflet.dev/v/v7-beta/layers/polyline-layer#interactivity. Let me know if that seems suitable and comprehensive enough.

lib/src/layer/polyline_layer.dart
ignatz: Thinking about this a bit more maybe this should be posed as "min dimension". There's no point in adding a transparent out-out-line to a line that might transparently occlude a neighbor, if the line is already wide.

JaffaKetchup: Actually yeah, a minimum would make some sense. I'll have a play around and see what's comfortable to me, then add a little bit more.

lib/src/layer/polyline_layer.dart
ignatz: Did you respond to my first or my second paragraph? I'd guess the latter.

My bigger question remains: why two different ways of identifying polylines?

(as for the second paragraph and your response, you're already hashing hashes anyway. To illustrate my point, a simple hash function would look something like:

```dart
int hash2(Object a, Object b) => a.hashCode ^ b.hashCode << 13;
```
)

lib/src/layer/polyline_layer.dart
ignatz: > The TODO seems more complicated and mathematically complex (converting stroke width to geographic space) than the work it would save. 

We'll see but just FTR, I'd be surprised if that was more expensive. You can save a lot of segments but if #1704 provides the same benefit, even better.

lib/src/layer/polyline_layer.dart
ignatz: Feels like my earlier ramblings got to you. Hope you're not getting nightmares :hugs: :hide:

lib/src/layer/polyline_layer.dart
JaffaKetchup: Render hash code is used for batching and layering operations, and doesn't include the list of coordinates, because that wouldn't be helpful.

The normal hash code should include the points, because the equality method does, and they should match.

lib/src/layer/polyline_layer.dart
JaffaKetchup: Not at all, you were right as usual. Please do keep correcting me, and I'll try to listen sooner 😂

lib/src/layer/polyline_layer.dart
ignatz: Let me rephrase, why do we need hashCode and hitKey?

JaffaKetchup: I'm not sure I understand, they are intended for entirely different purposes.

`hitKey` is just some freeform metadata for the user to assign if they want, helpful to track a line back out when handling.
`hashCode` is for the equality stuff and all the other normal uses of `hashCode`, which is not directly influenced by `hitKey`.

`hashCode` cannot just be `hitKey`, because multiple lines could have the same key, which could cause issues for the `hashCode`, if two objects are not equal but share the same hash.

(And stepping one step into the future, we've discussed removing `hitKey` and the generic typing, and just forcing users to extend `Polyline`, but a) that makes for an unnecessarily lengthy constructor and b) a type cast would be required somewhere.)

JaffaKetchup: Ah, I see where you might be getting at. Using the hashCode to identify the Polyline instead of the key at the other end. That is not ideal because the user has no direct control over the hashCode, the hashCode cannot be used to store information any more advanced than an integer, and the reason above. Think of hashCode as entirely internal as you would normally.

ignatz: 99% of the time when folks implement hashCode it's to use the Object as a key in a map.

```dart
final data = <Polyline, Data>{ /* ...*/ };

void onTap(PolylineHit hit) {
  doSomething(data[hit.lines.first]);
}
```

```dart
void onTap(PolylineHit hit) {
  doSomething(hit.lines.first.hitKey);
}
```

Feel very similar to me. Or did you add the `hashCode` for another reason? Either way, above options are both possible now.

JaffaKetchup: Ooo, that's a nice way to attach data that I guess is possible now. Maybe that alone is enough to remove `hitKey`, and that'll clean up the generic typing as well. Good idea, I'll do that.

JaffaKetchup: Removed `tapKey` in favour of this method. I'll change the docs.

lib/src/layer/polyline_layer.dart
JaffaKetchup: Resolved.

ignatz (commented):
LGTM

lib/src/layer/polyline_layer.dart
```diff
@@ -41,64 +64,185 @@ class Polyline {
...
+  final PolylineHitNotifier? hitNotifier;
+
+  /// The minimum radius of the hittable area around each [Polyline] in logical
+  /// pixels
+  ///
+  /// The entire visible area is always hittable, but if the visible area is
+  /// smaller than this, then this will be the hittable area.
+  ///
+  /// Defaults to 10.
+  final double minimumHitbox;
```
josxha, line 133: It's pixels I assume? Could we add the unit to the comment?

lib/src/layer/polyline_layer.dart
```diff
@@ -41,64 +64,185 @@ class Polyline {
...
+  _PolylinePainter({
+    required this.polylines,
+    required this.camera,
+    required this.hitNotifier,
+    required this.minimumHitbox,
+  }) : bounds = camera.visibleBounds;
 
-  int get hash => _hash ??= Object.hashAll(polylines);
+  // Avoids reallocation on every `hitTest`, is cleared every time
+  final hits = List<Polyline>.empty(growable: true);
```
josxha, line 180: Probably better to use `///` here

lib/src/layer/polyline_layer.dart
```diff
@@ -138,10 +282,11 @@ class PolylinePainter extends CustomPainter {
       paint = Paint();
     }
 
-    final origin = map.project(map.center).toOffset() - map.size.toOffset() / 2;
+    final origin =
+        camera.project(camera.center).toOffset() - camera.size.toOffset() / 2;
```
josxha, line 286: origin could be cached and later used in `hitTest`? but up to you.

josxha (approved):
I only found some small suggestions. Up to you how and if you solve them.
Great work @JaffaKetchup and @ignatz!
Tested on windows and web, works like a charm.

lib/src/layer/polyline_layer.dart
JaffaKetchup:
```suggestion
  /// Avoids reallocation on every `hitTest`, is cleared every time
  final hits = List<Polyline>.empty(growable: true);
```

JaffaKetchup: Its internal anyway so it won't make a difference. The comment also only explains why it exists, not what it does, so probably not suitable for a doc string.

lib/src/layer/polyline_layer.dart
JaffaKetchup: It's possible I suppose but more effort than reward - having to keep track of it and change it in `shouldRepaint` and also more space for bugs to sneak in.

lib/src/layer/polyline_layer.dart
```diff
@@ -41,64 +64,185 @@ class Polyline {
...
+
+  @override
+  int get hashCode => Object.hash(points, renderHashCode);
 }
 
 @immutable
 class PolylineLayer extends StatelessWidget {
-  final List<Polyline> polylines;
-  final bool polylineCulling;
+  final Iterable<Polyline> polylines;
```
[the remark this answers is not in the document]
JaffaKetchup: Resolved in 3ef7ee3.

lib/src/layer/polyline_layer.dart
ignatz: It's **not** possible. `paint` and `hitTest` have completely independent lifecycles. You can hit test without repainting and you can repaint w/o hit testing.

example/lib/pages/polyline.dart
```diff
@@ -26,78 +116,111 @@ class _PolylinePageState extends State<PolylinePage> {
...
-                  const LatLng(50.5, -7.8),
-                  const LatLng(56.5, 0.4),
-                ],
-                strokeWidth: 10,
-                color: Colors.amber,
-                borderStrokeWidth: 10,
-                borderColor: Colors.blue.withOpacity(0.5),
+              child: PolylineLayer(
+                hitNotifier: hitNotifier,
+                polylines: (hoverLines ?? []).followedBy(polylines.keys),
```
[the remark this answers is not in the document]
JaffaKetchup: @ignatz I did indeed get this the wrong way around, meant to be drawing hover lines on top. I've changed it, but that has revealed this bug which appears to be a canvas bug of some sort:

image
image

Fixing this is out of scope of this PR, but it is important. It looks like some layers are being painted and/or saved incorrectly, or something like that. #1704 might change/fix this.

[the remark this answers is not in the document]
JaffaKetchup: Resolved in fb9751d.

JaffaKetchup:
Huge thanks @ignatz! Now on to polygons when you're ready and have had a break from me :D

PR #1726, opened 2023-11-10
fix: add `doubleTapDragZoom` and `scrollWheelZoom` to `InteractiveFlag.all`

amal-stack:
Fixes #1725.

PR #1730, opened 2023-11-12
chore!: remove all leftover deprecated symbols

JaffaKetchup:
**Requires #1726 to function correctly. Makes #1724 redundant.**

Also includes a minor refactoring of example app Live Web Demo introduction dialog.


JaffaKetchup:
Would be great if this could be reviewed! It should hopefully make things simpler/less confusing for other PRs.

JaffaKetchup:
@ibrierley @mootw @TesteurManiak @josxha Would be great if this could be reviewed! It should hopefully make things simpler/less confusing for other PRs.

josxha:
Can do!

example/lib/pages/home.dart
```diff
@@ -19,80 +19,7 @@ class _HomePageState extends State<HomePage> {
...
-                bottom: 0,
-                right: 24,
-              ),
-            ),
-          );
-          await prefs.setBool(seenIntroBoxKey, true);
-        },
-      );
-    }
+    _displayWebDemoIntroDialog();
```
josxha, line 22: I'd prefer to not commit refactorings to the example project in this pr.
For more details: I wanted to do some clean up to the example project anyways and could include this as well. Mostly smaller things like using `MediaQuery.sizeOf` and so on.
This refactoring would suit there well too. What do you think?

josxha (changes requested):
Only one consideration. Other than that it looks good to me.

example/lib/pages/home.dart
josxha: But a change that we could sneak in here would be there @override lint in `_SimulateErrorImageProvider` since https://github.com/fleaflet/flutter_map/pull/1704 might take another day or two to merge. 😅

JaffaKetchup: Thanks for pointing that second one out. Removing `@override` is not the correct solution here - `load` has been removed and wouldn't be called from anywhere! I've migrated to `loadImage` in 136a5b8.

I've also removed the changes to example/lib/pages/home.dart in a96724f.

josxha:
The changes are looking good to me. 👍🏼 
Windows CD builds are currently broken. Seems like the build directory has changed?
- old: `D:\a\flutter_map\flutter_map\example\build\windows\runner\Release\example.exe`
- new: `D:\a\flutter_map\flutter_map\example\build\windows\x64\runner\Release\example.exe`

josxha (approved):

josxha:
Ready to merge 🎉

JaffaKetchup:
Ah, already fixed Inno in 706ce4, but should cause conflicts so no worries.

Issue #1729, opened 2023-11-12
[FEATURE] Revert to using single `onPanUpdate` to detect pan events


S-Man42:
### What is the bug?

Since v4 I was able to create a marker which contained a GestureDetector as builder output. This GestureDetector included an onPanUpdate event which I used to enable dragging functionality. It worked fine a long time.

Now I upgraded FlutterMap to v5 and it came out, that exactly this event is not called anymore. Funny: All other events can be called as `onTap` or even `onVerticalDragUpdate`. 

Do you have an idea what changed and how could I restore this feature?

### How can we reproduce it?

```
FlutterMap(
	children: [
		MarkerLayer(
		 markers: [		 
			Marker(
			  point: myPoint,
			  builder: (context) => GestureDetector(
				onTap: () {
					// WORKS
				},
				onPanUpdate: (details) {
					// DOES NOT WORK ANYMORE
				},
				child: myIcon
			  )
			)
		 ]
		)
	]
)
```

### Platforms

Android, web, iOS

### Severity

Obtrusive: Prevents normal functioning but causes no errors in the console

JaffaKetchup:
Hi @S-Man42,
Can you try disabling `MapOptions.applyPointerTranslucencyToLayers` to see if that makes a difference?

S-Man42:
This is not part of v5 as I see it. I believe it was integrated a bit later... But I cannot do another major upgrade at the moment. I need a v5 solution, if possible...

JaffaKetchup:
Ah, sorry, didn't see that you were still using v5. Can I ask why you are not using the flutter_map_dragmarker plugin? If there's a reason, maybe you can check the implementation there?

The only thing I can think of off the top of my head as to why this changed is due to #1455, but that should have been available in and impacting v4 as well.

S-Man42:
> The only thing I can think of off the top of my head as to why this changed is due to #1455, but that should have been available in and impacting v4 as well.

Yes, you are right, I cross-checked: We were using v3.1.0, not 4 - sorry my fault.

> Can I ask why you are not using the flutter_map_dragmarker plugin

I am currently trying. But my problem is, that I am using the flutter_map_marker_popup plugin for my markers. So I cannot see a way to combine both plugins to create a draggable popup marker. I am currently trying to implement an inherited PopupMarkerLayer with integrated code from the DragMarker plugin, but it seems not correct to me:

[...]

Issue #1715, opened 2023-11-03
[SPEC] Overhaul gesture system


josxha:
> Request for comments

### Motivation
As mentioned by @JaffaKetchup in https://github.com/fleaflet/flutter_map/issues/1529#issue-1718239936, "much of the [gesture handling] code that does this was written in the first few releases, and has not been significantly changed since. [...] Gesture handling at this low-level often differs slightly between platforms, which means that manual testing succeeds on one platform, but a bug occurs on another."

### Rewrite Proposals
- Break down and structure the [FlutterMapInteractiveViewer](https://github.com/fleaflet/flutter_map/blob/19649100ae223b5785c32755fb1dc4bafdab7313/lib/src/gestures/flutter_map_interactive_viewer.dart) class to improve readablity and reduce its complexity.
- Remove the usage of [PositionedTapDetector2](https://github.com/fleaflet/flutter_map/blob/master/lib/src/gestures/positioned_tap_detector_2.dart) and merge its GestureDetector with the one in FlutterMapInteractiveViewer.
- Use the high level GestureDetector as much as possible. This ensures the use of flutters GestureArena.
- Get rid of RawGestureDetector to avoid platform specific bugs (see issue tracker here https://github.com/fleaflet/flutter_map/issues/1529).
- ~~Use boolean fields in [InteractiveFlag](https://github.com/fleaflet/flutter_map/blob/19649100ae223b5785c32755fb1dc4bafdab7313/lib/src/gestures/interactive_flag.dart) and [MultiFingerGesture](https://github.com/fleaflet/flutter_map/blob/19649100ae223b5785c32755fb1dc4bafdab7313/lib/src/gestures/multi_finger_gesture.dart). This simplifies internal usage, is more beginner friendly in our public API and has no performance penalty. I'll write some code examples in the next comment.~~
- ~~https://github.com/fleaflet/flutter_map/issues/1732~~
- open for extendibility like https://github.com/fleaflet/flutter_map/issues/1432

### If feasible do the following changes:
- https://github.com/fleaflet/flutter_map/issues/1729

### More resources
- [Flutter Deep Dive: Gestures. Flutter provides some really amazing… | by Nash | Flutter Community | Medium](https://medium.com/flutter-community/flutter-deep-dive-gestures-c16203b3434f)
- [Taps, drags, and other gestures | Flutter](https://docs.flutter.dev/ui/interactivity/gestures)
[...]

Issue #1602, opened 2023-08-02
[BUG] Double-tap-drag gesture doesn't cause new tiles to load after completion


JaffaKetchup:
### What is the bug?

Performing the double-tap-drag gesture (`MapEventSource.doubleTapHold`) does not cause new tiles to load after the gesture.

It's the correct behaviour for tiles not to load DURING the gesture, however - they should only load afterward.

### How can we reproduce it?

Perform the double-tap-drag gesture on any map to zoom in or out. The previously visible tiles will remain visible, and no new tiles will be loaded.

### Do you have a potential solution?

The example application's Interactive Flags test page does not appear to show this event after/during the gesture, meaning it is likely that none is properly emitted during zooming.

### Platforms

All

### Severity

Obtrusive: Prevents normal functioning but causes no errors in the console

flodaniel:
I am on `flutter_map: 5.0.0` and the loading does also not occur once the gesture is completed. I use the gesture to zoom out and after completion the entire map is grey and no server calls to the tiles server are made. Once the user does a simple drag the load happens.

JaffaKetchup:
Thanks @flodaniel, I've updated the original issue, as I can reproduce that now as well (I struggled to for some reason previously, but that was probably coincidence).

josxha:
> I always thought this was more of a feature and not a bug.

That's a nice point of view. 😅 

> Of course after the gesture is finished the tiles should load, but surprisingly that always worked for me.

The problem is that another gesture input is needed until the tiles load after you zoomed with this gesture. I guess *most* users would do that intuitively if they see that the map doesn't load althrough not every user.

JaffaKetchup:
Agreed, I think it makes sense for tiles to only load after completion with this gesture.

josxha:
Some information on how to reproduce the bug:

1. open the interactive flags page of the demo app
2. deactivate all movement gestures, you can activate everything else but at least `+ drag`
3. perform a double tap + drag gesture

https://github.com/fleaflet/flutter_map/assets/34318751/ea048a3d-2876-4ea3-b93d-831e8ab3bc2a

JaffaKetchup:
This appears to be fixed when performing it on the home page of the example application, but it does still occur in the conditions described by https://github.com/fleaflet/flutter_map/issues/1602#issuecomment-1837264417.

JaffaKetchup:
I'm going to close this as a duplicate of #1813.

Issue #1429, opened 2023-01-12
[BUG] iOS: Pinch zoom broken (a)


FeofanGreek:
### What is the bug?

Zoom not working when swiping with two fingers

### What is the expected behaviour?

there should have been a change in scale

### How can we reproduce this issue?

```dart
when moving across the screen with the pads of two fingers, the map is rotated and shifted, but zooming does not occur
```

### Do you have a potential solution?

No

### Can you provide any other information?

https://user-images.githubusercontent.com/57085995/212029989-d534cbc4-029a-4a00-9bcf-5f3d9a3797db.mov

### Platforms Affected

iOS

### Severity

Obtrusive: Prevents normal functioning but causes no errors in the console

### Frequency

Rarely: Occurs every so often

### Requirements

- [X] I agree to follow this project's [Code of Conduct](https://github.com/fleaflet/flutter_map/blob/master/CODE_OF_CONDUCT.md)
- [X] My Flutter/Dart installation is unaltered, and `flutter doctor` finds no relevant issues
- [X] I am using the [latest stable version](https://pub.dev/packages/flutter_map) of this package
- [X] I have checked the FAQs section on the documentation website
- [X] I have checked for similar issues which may be duplicates

ivanzhadnov:
I have the same problem, can someone suggest a solution?

ibrierley:
Wonder if this is related to mouse zoom issue as well with iOS https://github.com/fleaflet/flutter_map/issues/1414

ibrierley:
and https://github.com/fleaflet/flutter_map/issues/1354

FeofanGreek:
Don't work zoom with fingers multi_finger_gesture.dart

JaffaKetchup:
Hey @FeofanGreek, if you're still interested, can you try this out with the latest 'master' and Flutter 3.10?

JosefWN:
I think we might need some more context if this issue persists on `v5.0.0`, last I tried on `v4.0.0` on iOS it worked.

Do you see the issue both in the emulator and in the deployed app?

ibrierley:
I was wondering if this was related to https://github.com/fleaflet/flutter_map/pull/1543 ?

JosefWN:
It could be, especially if only the emulator is impacted since the touch behavior on MacOS is a bit different from iOS. I don't run the emulator, I only test on actual devices, where I haven't seen this issue personally. Better test on `v5.0.0` before digging deeper at least.

[...]

PR #1733, opened 2023-11-21
multi!: rewrite gesture handling

josxha:
### Main Issue:
- Fix https://github.com/fleaflet/flutter_map/issues/1715

#### Move gestures
- [x] drag
- [x] flingAnimation
- [x] twoFingerMove (old: pinchMove)
#### Zoom gestures
- [x] twoFingerZoom (old: pinchZoom)
- [x] doubleTapZoomIn (old: doubleTapZoom)
- [x] doubleTapDragZoom
- [x] scrollWheelZoom
- [x] trackpadZoom
#### Rotate gestures
- [x] twoFingerRotate (old: rotate)
- [x] ctrlDragRotate (old: hidden gesture)
#### Gesture callbacks
- [x] onTap
- [x] onLongPress
- [x] onSecondaryTap
- [x] onSecondaryLongPress (new)
- [x] onTertiaryTap (new)
- [x] onTertiaryLongPress (new)
- [x] onPointerDown
- [x] onPointerUp
- [x] onPointerCancel
- [x] onPointerHover 

### Changes along the way:
- [x] rewrote `FlutterMapInteractiveViewer` as `MapInteractiveViewer`
- [x] created gestures as objects
- [x] remove `PositionedTapDetector2`
- [x] add new callbacks for `secondaryLongPress`, `tertiaryTap`, `tertiaryLongPress` gestures
- [x] fix https://github.com/fleaflet/flutter_map/issues/1602
- [x] fix https://github.com/fleaflet/flutter_map/issues/1729
- [x] fix https://github.com/fleaflet/flutter_map/issues/1767
- [x] remove typdefs for `PointerDownCallback`, `PointerUpCallback`, `PointerCancelCallback`, `PointerHoverCallback`, `MapEventCallback`, `IsKeyCursorRotationTrigger`
- [x] turn `MoveAndRotateResult` into a class
- [x] Changed `InteractionOptions.flags` into `InteractionOptions.enabledGestures` that still supports flags
- [x] Remove `debugMultiFingerGestureWinner`, `enableMultiFingerGestureRace`, `rotationWinGestures`, `pinchMoveWinGestures` and `pinchZoomWinGestures` (use case unclear)
- [x] move directory `/gestures` to `/map/gestures` since it part of it
- [x] rename `InterativeFlagsPage` to more general `GesturesPage`
- [x] remove hidden gesture `setNorthOnClick` (press CTRL + click to rotate to 0°)
- [x] remove `CursorRotationBehaviour`
- [x] remove `CursorKeyboardRotationOptions`, merge gesture with `EnabledGestures` and keys option with `InteractiveOptions`
- [x] Use more descriptive names for `InteractiveFlags`: pinchMove -> twoFingerMove, doubleTapZoom -> doubleTapZoomIn,  rotate -> twoFingerRotate
- [x] Remove `MapEventSource.cursorKeyboardRotation` in favor of `ctrlDragRotateStart`, `ctrlDragRotateEnd` and `ctrlDragRotate`


josxha:
#### To do
- [ ] `CursorKeyboardRotationOptions`
- [ ] `InteractionOptions`

josxha:
Can't believe it, this PR is after 3 weeks finally in a reviewable state. 😄 I stretched the freedom of subjective changes pretty sure not only once. In case you disagree with any of it or have something in mind that is missing, please let me know.

Notes:
- I hope that I catched every relevant change in the first comment. If you find anything that is not mentioned, please edit the comment. This will hopefully help writing the changelog later on.
- I'm still not sure if the gesture callbacks would fit better under the `interactiveOptions`. therefore I'm open for suggestions.
- `MapEvents` and `MapEventSource` should get some clean up in the future too. For example often the source information is redundant because an event has its own type of class. It's considered out of scope for this PR since it's a change to the event system and not gesture system.
- Currently only few deprecations made it in this PR. I don't think that we are able to provide deprecations for everything in a viable way but let me know what you think requires deprecations.

--- 

@erikschul This pull request should stop all animations caused by other gestures (issue https://github.com/fleaflet/flutter_map/issues/1767). If you want you can check and verify. Looking forward for feedback. :) (Your other issue will be part of a an other pull request.)

lib/src/map/options/enabled_gestures.dart
```diff
@@ -0,0 +1,254 @@
...
+    required this.twoFingerZoom,
+    required this.doubleTapZoomIn,
+    required this.doubleTapDragZoom,
+    required this.scrollWheelZoom,
+    required this.twoFingerRotate,
+    required this.ctrlDragRotate,
+  });
+
+  /// Shortcut constructor to allow all gestures that don't rotate the map.
+  const EnabledGestures.noRotation()
```
JaffaKetchup, line 23: This could be an indicator that we need more categorization of these gestures, so larger swathes can be toggled at once. For example, movement. Again, maybe enums, maybe mixins, maybe inheritance, maybe just all possible constructors - not sure.

lib/src/map/options/enabled_gestures.dart
```diff
@@ -0,0 +1,254 @@
...
+  /// Enable rotation by pressing the CTRL key and dragging with the cursor
+  /// or finger.
+  final bool ctrlDragRotate;
+
+  /// Returns true of any gesture with more than one finger is enabled.
+  bool hasMultiFinger() => twoFingerMove || twoFingerZoom || twoFingerRotate;
+
+  /// Wither to change the value of some gestures. Returns a new
+  /// [EnabledGestures] object.
+  EnabledGestures withFlag({
```
JaffaKetchup, line 122: Should probably use the standard `copyWith` name for clarity and consistency.

lib/src/map/options/enabled_gestures.dart
```diff
@@ -0,0 +1,254 @@
...
+  ///
+  /// The associated [MapEventSource] is [MapEventSource.doubleTapHold].
+  final bool doubleTapDragZoom;
+
+  /// Enable zooming with a mouse scroll wheel
+  final bool scrollWheelZoom;
+
+  /// Enable rotation by pressing the CTRL key and dragging with the cursor
+  /// or finger.
+  final bool ctrlDragRotate;
```
JaffaKetchup, line 115: Same as above, not sure of the name here. Applies to documentation as well.

lib/src/map/options/enabled_gestures.dart
```diff
@@ -0,0 +1,254 @@
...
+/// you have two options:
+///   a. Add your own flags:
+///      [InteractiveFlag.drag] | [InteractiveFlag.twoFingerRotate]
+///   b. Remove unnecessary flags from all:
+///     [InteractiveFlag.all] &
+///       ~[InteractiveFlag.flingAnimation] &
+///       ~[InteractiveFlag.twoFingerMove] &
+///       ~[InteractiveFlag.twoFingerZoom] &
+///       ~[InteractiveFlag.doubleTapZoomIn]
+abstract class InteractiveFlag {
```
JaffaKetchup, line 188: Am I missing something, or is having this in addition to `EnabledGestures` redundant?

lib/src/map/options/interaction_options.dart
```diff
@@ -0,0 +1,97 @@
...
+
+  /// The velocity how fast the map should zoom when using the scroll wheel
+  /// of the mouse.
+  /// Defaults to 0.01.
+  final double scrollWheelVelocity;
+
+  /// Override this option if you want to use custom keys for the CTRL+drag
+  /// rotate gesture.
+  /// By default the left and right control key are both used.
+  final List<LogicalKeyboardKey> ctrlRotateKeys;
```
JaffaKetchup, line 51: This was one of the things I was looking to avoid in my early unfinished rewrites from a while back. Ideally, there should be no configuration of a gesture seperate to where that gesture is enabled or disabled.

Maybe that means every gesture flag/option needs to not be a boolean anymore, and needs to be an entire object `GestureOptions`, with inheritable constructors `enabled()` and `disabled()`, and each option can have it's own arguments. In combination with the recommendation above to establish better grouping of options, this would allow for 'labels'/'tags' of categories, or something similar. I'm aware of this potential impact on performance here though of having loads of new objects, so am open to benchmarks and/or alternative suggestions.

lib/src/map/options/map_options.dart
```diff
@@ -60,14 +31,20 @@ class MapOptions {
...
+
+  final void Function(PointerDownEvent event, LatLng point)? onPointerDown;
+  final void Function(PointerUpEvent event, LatLng point)? onPointerUp;
+  final void Function(PointerCancelEvent event, LatLng point)? onPointerCancel;
+  final void Function(PointerHoverEvent event, LatLng point)? onPointerHover;
+
   final PositionCallback? onPositionChanged;
-  final MapEventCallback? onMapEvent;
+
+  final void Function(MapEvent event)? onMapEvent;
```
JaffaKetchup, line 47: As mentioned above, are we sure loads of callbacks is the best way of doing things here?

lib/src/map/options/map_options.dart
```diff
@@ -77,13 +54,13 @@ class MapOptions {
   /// Only use this if your map isn't built immediately (like inside FutureBuilder)
   /// and you need to access the controller as soon as the map is built.
   /// Otherwise you can use WidgetsBinding.instance.addPostFrameCallback
-  /// In initState to controll the map before the next frame.
+  /// In initState to control the map before the next frame.
```
JaffaKetchup, line 57: Might need to change the documentation here to match some of my previous invesigations of bug reports and help questions, and new recommendations on the docs site. Don't worry about that in this PR though.

(PS. thanks for catching my spelling bug below :D)

lib/src/map/options/map_options.dart
```diff
@@ -106,6 +83,8 @@ class MapOptions {
 
   final InteractionOptions interactionOptions;
 
+  /// The options of the closest [FlutterMap] ancestor. If this is called from a
```
JaffaKetchup, line 86: Is this unfinished?

test/flutter_map_test.dart
```diff
@@ -223,7 +215,12 @@ class TestRebuildsApp extends StatefulWidget {
 class _TestRebuildsAppState extends State<TestRebuildsApp> {
   MapController _mapController = MapController();
   Crs _crs = const Epsg3857();
-  int _interactiveFlags = InteractiveFlag.all;
+
+  /// double tap gestures delay the tap gestures, disable them here
+  EnabledGestures _interactiveFlags = const EnabledGestures.all(
```
JaffaKetchup, line 220: As above, this is another indicator that we need better categorization here.

JaffaKetchup (changes requested):
Wow, thanks!

Two things to consider in general: documentation and class modifiers.
With so many new classes, we really need to document every single one, no matter how obvious the doc string might be. I'm happy to go through if needed, but I just need some clarification on the whole new system infrastructure as said in one of the comments - perhaps a quick overall graphic/flowchart could be helpful, and useful in public documentation as well.
Class modifiers are a bit tricker, as we need to carefully consider their external impact - especially so when retroactively adding them (which is partially why we haven't yet), but they can be really useful for enforcing the correct desired patterns and maintainability.
IMO adding a new gesture or service or whatever should feel like writing an almost self contained addon, and be strongly enforced as such. Both these things help with this.

I haven't tested any of the changes yet, but I've scraped through the code itself. Most of it seems to seem like a good idea and I only have small minor comments, but there's a few major bits I think could be improved on in terms of inheritance with services, and gesture options. All of my reviews are open to discussion, especially about any patterns we may want to consider implementing.

(Round 1)

example/lib/pages/gestures_page.dart
```diff
@@ -4,50 +4,50 @@ import 'package:flutter_map_example/misc/tile_providers.dart';
...
 }
 
-class _InteractiveFlagsPageState extends State<InteractiveFlagsPage> {
+class _GesturesPageState extends State<GesturesPage> {
   static const availableFlags = {
     'Movement': {
       InteractiveFlag.drag: 'Drag',
       InteractiveFlag.flingAnimation: 'Fling',
-      InteractiveFlag.pinchMove: 'Pinch',
+      InteractiveFlag.twoFingerMove: 'Pinch',
```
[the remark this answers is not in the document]
josxha: Agree, got me confused mutliple times while testing. How about "Two finger drag"?

lib/src/map/controller/events/map_event_source.dart
```diff
@@ -0,0 +1,29 @@
+/// Event sources which are used to identify different types of
+/// [MapEvent] events
+enum MapEventSource {
```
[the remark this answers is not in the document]
josxha: I agree that the event system should get revised. But in my opinion the "event system" and "gesture system" are two different things therefore I'd prefer to have those changes not bundled in this pull request.
My main point is that some MapEventSources have their own dedicated Event class which should be avoided. But the specification should be not discussed here because off topic. We can however delay this pull request until we have reworked the gesture system and integrate it here.

lib/src/map/controller/events/map_events.dart
```diff
@@ -243,6 +258,24 @@ class MapEventDoubleTapZoomEnd extends MapEvent {
...
+class MapEventDoubleTapDragZoomEnd extends MapEvent {
+  const MapEventDoubleTapDragZoomEnd({
+    required super.source,
+    required super.camera,
+  });
+}
+
 /// Event which is fired when map is being rotated
 @immutable
 class MapEventRotate extends MapEventWithMove {
```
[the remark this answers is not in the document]
josxha: Agree, waiting until https://github.com/fleaflet/flutter_map/pull/1733#discussion_r1428962255 is resolved to have a name for the new events.

lib/src/map/controller/map_controller_impl.dart
```diff
@@ -236,11 +239,14 @@ class MapControllerImpl extends ValueNotifier<_MapControllerState>
     }
 
     if (degree == camera.rotation) {
-      return const (moveSuccess: false, rotateSuccess: false);
+      return const MoveAndRotateResult(
```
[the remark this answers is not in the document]
josxha: Can do, but instead of an AND I'd use an OR because as far as I understood the return is false if the current camera already matched the requested chances and therefore true if any change has made to the camera, not specifically the requested (in case of max map bounds).
Chaining both results together with OR would mean "some change has been made" and not specifically "the requested change".

Side note: I'd prefer to have only one function for all move, zoom and rotate gestures and the function like a `changeCamera` function with optional parameters for rotation, center and zoom.

lib/src/map/gestures/map_interactive_viewer.dart
```diff
@@ -0,0 +1,268 @@
...
+    updateGestures(null, _interactionOptions.enabledGestures);
+  }
+
+  @override
+  void dispose() {
+    widget.controller.removeListener(reload);
+    super.dispose();
+  }
+
+  void reload() {
```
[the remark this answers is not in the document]
josxha: When would you override the `setState` function of `MapInteractiveViewerState`?
If you do and have a good reason to so I think the reload function would require to use the overridden function anyway.

lib/src/map/gestures/services/base_services.dart
```diff
@@ -0,0 +1,49 @@
...
+part 'ctrl_drag_rotate.dart';
+part 'double_tap.dart';
+part 'double_tap_drag_zoom.dart';
+part 'drag.dart';
+part 'long_press.dart';
+part 'scroll_wheel_zoom.dart';
+part 'tap.dart';
+part 'two_finger.dart';
+
+abstract class BaseGestureService {
```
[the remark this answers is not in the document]
josxha: Welp you're right, I lost focus on documentation. 
I'll revisit the code to add documentation comments whereever possible.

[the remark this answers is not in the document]
josxha: I can add any class modifier that you want. 
I personally prefer to keep it simple unless we don't have a benefit like pattern matching or the `abstract` modifier of course.

lib/src/map/gestures/services/base_services.dart
```diff
@@ -0,0 +1,49 @@
...
+
+  DelayedGestureService({required super.controller});
+
+  void setDetails(TapDownDetails newDetails) => details = newDetails;
+
+  void reset() => details = null;
+}
+
+/// Return a rotated Offset
+Offset _rotateOffset(MapCamera camera, Offset offset) {
```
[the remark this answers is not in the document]
josxha: This function is used by the drag gesture and two finger gestures.
Waiting until https://github.com/fleaflet/flutter_map/pull/1733#discussion_r1428962688 is resolved.

lib/src/map/gestures/services/ctrl_drag_rotate.dart
```diff
@@ -0,0 +1,42 @@
+part of 'base_services.dart';
+
+class CtrlDragRotateGestureService extends BaseGestureService {
```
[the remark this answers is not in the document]
josxha: I saw that the `shift` key should get included in the list anyways so I agree with you.
`CursorKeyboardRotation` lacks in my opinion the descriptive character, `KeyDragRotation` feels better to me however I would add a second verb for better understanding that there is no key to be dragged around.
How about `KeyPressDragRotate` or `KeyDownDragRotate`? Slightly longer but in my opinion better to understand.

This name change would be applied to every occurrence of "ctrlDragRotate".

lib/src/map/gestures/services/ctrl_drag_rotate.dart
```diff
@@ -0,0 +1,42 @@
...
+  bool isActive = false;
+  final List<LogicalKeyboardKey> keys;
+
+  CtrlDragRotateGestureService({
+    required super.controller,
+    required this.keys,
+  });
+
+  void start() {
```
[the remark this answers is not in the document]
josxha: waiting for https://github.com/fleaflet/flutter_map/pull/1733#discussion_r1428962522 to be resolved.

lib/src/map/gestures/services/double_tap.dart
```diff
@@ -0,0 +1,52 @@
+part of 'base_services.dart';
+
+class DoubleTapGestureService extends DelayedGestureService {
+  DoubleTapGestureService({required super.controller});
+
+  /// A double tap gesture tap has been registered
+  void submit() {
```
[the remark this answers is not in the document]
josxha: *Merging the inherited discussions here.*

I'd like to have them inherited too any did so once but reverted the change. The problem was that all `submit` functions create a different event type. We can however create a function to emit the event in every child class and let the parent handle the rest. 

Side note: If we rework use the event system to just have one camera move event that rules them all we don't need the additional function at all.

lib/src/map/gestures/services/double_tap.dart
```diff
@@ -0,0 +1,52 @@
...
+    controller.stopAnimationRaw();
```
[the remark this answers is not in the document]
josxha: Waiting until https://github.com/fleaflet/flutter_map/pull/1733#discussion_r1428962688 is resolved.

lib/src/map/options/enabled_gestures.dart
josxha: Is there really a meaningful amount of users that need to toggle all movement or all zooming in a group that it makes it worth the additional shortcut code for every variant? No rotation is a common use case, clearly written in the docs and still gets asked from time to time therefore a shortcut makes sense in my opinion. For the other gesture groups however it's not much more work to toggle those flags themselfs. `EnabledGestures` provides users with a gesture whitelist and blacklist approach that is in my opinion sufficient. 
Could you provide my with an use case for extended grouping of gestures?

lib/src/map/options/enabled_gestures.dart
josxha: Agree.

lib/src/map/options/enabled_gestures.dart
josxha: discussion merged with https://github.com/fleaflet/flutter_map/pull/1733#discussion_r1428962255

lib/src/map/options/enabled_gestures.dart
josxha: While writing the scope of this pull request I recignized that some really like the bitfield approach to define gestures. That's why I kept it in for everyone who prefers to define gestures that way. In case a user doesn't use this functionality it will get treeshakend.

lib/src/map/options/map_options.dart
josxha: Missed that line while refactoring, good spot.

lib/src/map/gestures/map_interactive_viewer.dart
```diff
@@ -0,0 +1,268 @@
...
+
+  InteractionOptions get _interactionOptions => _options.interactionOptions;
+
+  @override
+  void initState() {
+    super.initState();
+    widget.controller.interactiveViewerState = this;
+    widget.controller.addListener(reload);
+
+    // callback gestures for the application
```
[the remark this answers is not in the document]
josxha: The problem is that most of the gestures have differnt parameters. If we use only one callback or a stream the user needs to cast the event type and we need additional classes to wrap all parameters together. We can go that way but I'm not sure if it would make things easier for us and for the users.

lib/src/map/gestures/map_interactive_viewer.dart
```diff
@@ -0,0 +1,268 @@
...
+        _doubleTapDragZoom != null ||
+        _twoFingerInput != null;
+
+    return Listener(
+      onPointerDown: _options.onPointerDown == null
+          ? null
+          : (event) => _options.onPointerDown!.call(
+                event,
+                _camera.offsetToCrs(event.localPosition),
+              ),
```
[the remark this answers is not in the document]
josxha: When implementing I tried to listen only to those events that are required. I only know about the performance impact it has on the three double tap callbacks. An so I thought if we need to have it for some, we might as well just have it for all.

test/flutter_map_test.dart
josxha: discussion merged with https://github.com/fleaflet/flutter_map/pull/1733#discussion_r1428964013

example/lib/pages/markers.dart
```diff
@@ -110,9 +110,11 @@ class _MarkerPageState extends State<MarkerPage> {
               options: MapOptions(
                 initialCenter: const LatLng(51.5, -0.09),
                 initialZoom: 5,
-                onTap: (_, p) => setState(() => customMarkers.add(buildPin(p))),
+                onTap: (_, p) {
+                  setState(() => customMarkers.add(buildPin(p)));
+                },
                 interactionOptions: const InteractionOptions(
-                  flags: ~InteractiveFlag.doubleTapZoom,
+                  enabledGestures: EnabledGestures.noRotation(),
```
TesteurManiak, line 117: Why not use `EnabledGestures.all(doubleTapZoomIn: false)` like in the file `latlng_to_screen_point.dart`? Or maybe it should be the other way around and use `EnabledGestures.noRotation()` as well?

lib/src/map/options/enabled_gestures.dart
```diff
@@ -0,0 +1,254 @@
+import 'package:flutter_map/flutter_map.dart';
+import 'package:meta/meta.dart';
+
+@immutable
+class EnabledGestures {
+  /// Private constructor, use the constructors [InteractiveFlags.all] or
+  /// [InteractiveFlags.none] instead to enable or disable all gestures by default.
+  /// If you want to define your enabled gestures using bitfield operations,
+  /// use [InteractiveFlags.bitfield] instead.
+  const EnabledGestures._({
```
TesteurManiak, line 10: Not sure why this constructor should be kept private 🤔

josxha: We can make this constructor public too. 👍🏼 Currently users need to choose if they want to use a whitelist or blacklist approach on allowed gestures but that's not really necessary.

example/lib/pages/markers.dart
josxha: Oh using the noRotation constructor is wrong in the first place. I think the reason to disable double tap gestures was to remove the click delay but I make a mistake while migrating.
Are you fine to remove the custom `interactionOptions` here completely and disable `doubleTapDragZoom` in `latlng_to_screen_point.dart`? Or do you want to have rotation disabled in more examples in general?

lib/src/map/options/map_options.dart
josxha: merging discussion with https://github.com/fleaflet/flutter_map/pull/1733#discussion_r1428964941

lib/src/map/gestures/map_interactive_viewer.dart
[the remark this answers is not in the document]
josxha: We can remove those callbacks if you want. Currently all gestures will get emitted by the event system anyway. 
I personally think that callbacks are easier to understand and implement for users and the their implementation overhead seems quite low to me. It's probably a question of how beginner friendly we want our API to be.

lib/src/map/options/map_options.dart
josxha: discussion merged with https://github.com/fleaflet/flutter_map/pull/1733#discussion_r1428960703

lib/src/map/options/interaction_options.dart
josxha: After thinking about it my mind is still split. 

#### Pro gesture options as separate classes
- All gesture options of a gesture are separated from other gesture options
- Option to enable or disable complete gesture groups

#### Contra gesture options as separate classes
- Separation between enabled options and option overrides
- Huge class overhead and more complex map options (13+ classes, 9 additional instances)
- More complex implementation of the map options for users 
    - Either users need to save the gesture options as field variables
    - or something like `gestureEnabeld ? SomeGesture.enabled(additionalOption: 123) : SomeGesture.disabled()`

lib/src/map/options/enabled_gestures.dart
josxha: While wrtiting https://github.com/fleaflet/flutter_map/pull/1733#discussion_r1429228899 an idea came in my mind that would allow this functionality in a really easy way. Really don't know why this hasn't come into my mind earlier:
```dart
const EnabledGestures.fromGroups({
  required bool move,
  required bool zoom,
  required bool rotate,
}) : this._(
        drag: move,
        twoFingerMove: move,
        flingAnimation: move,
        doubleTapDragZoom: zoom,
        doubleTapZoomIn: zoom,
        scrollWheelZoom: zoom,
        twoFingerZoom: zoom,
        twoFingerRotate: rotate,
        ctrlDragRotate: rotate,
      );
```

In addition, there could be another wither, something like `copyWithGroups()` and that's it.

lib/src/map/options/interaction_options.dart
JaffaKetchup: > gestureEnabeld ? SomeGesture.enabled(additionalOption: 123) : SomeGesture.disabled()

To avoid this we can create an unnamed constructor with all additional options, and an `enabled` flag.

> In combination with the recommendation above to establish better grouping of options, this would allow for 'labels'/'tags' of categories, or something similar.

I've realised this statement, whilst true, probably won't help on the users side. Your suggestion in https://github.com/fleaflet/flutter_map/pull/1733#discussion_r1429233040 is still probably the best.

lib/src/map/options/enabled_gestures.dart
JaffaKetchup: Yeah, I do like that. Maybe a better name would be with `byGroup`?

lib/src/map/gestures/services/double_tap.dart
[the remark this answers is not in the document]
JaffaKetchup: > If we rework use the event system to just have one camera move event that rules them all we don't need the additional function at all.

How would this look?

> We can however create a function to emit the event in every child class and let the parent handle the rest.

This wouldn't however have the advantage of https://github.com/fleaflet/flutter_map/pull/1733#discussion_r1428962917, which I think is quite important.

lib/src/map/gestures/services/ctrl_drag_rotate.dart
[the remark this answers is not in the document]
JaffaKetchup: How about `KeyTriggerDragRotate`? A bit of a mouthful perhaps. Not sure what's the best tbh.

lib/src/map/gestures/map_interactive_viewer.dart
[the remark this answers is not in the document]
JaffaKetchup: I'm just looking at what we've done with the polyline hit detection. It's perhaps not quite as simple as a callback, but it is the most flexible option by far, and not too much more difficult to implement, especially with some helpful documentation and demos.
What do others think?

lib/src/map/gestures/map_interactive_viewer.dart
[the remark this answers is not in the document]
JaffaKetchup: I'm just looking and seeing that `reload` is just calling `setState` if it's safe to. Therefore to hide that bit of abstraction from people looking at the code, and prevent them from having to check what `reload` does, it might make sense. Maybe I'm just overcomplicating things - let's leave it for now.

lib/src/map/controller/map_controller_impl.dart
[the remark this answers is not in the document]
JaffaKetchup: > Chaining both results together with OR would mean "some change has been made" and not specifically "the requested change".

Yep, that seems more sensible.

> Side note: I'd prefer to have only one function for all move, zoom and rotate gestures and the function like a changeCamera function with optional parameters for rotation, center and zoom.

That sounds sensible.

lib/src/map/controller/events/map_event_source.dart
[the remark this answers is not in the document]
JaffaKetchup: > in addition I really like to get rid of the event id property as they don't get used by gestures anyway.

It can be passed through from the `MapController` (or at least it should be if it can't atm), and is very useful in some cases.

> But in my opinion the "event system" and "gesture system" are two different things therefore I'd prefer to have those changes not bundled in this pull request.

I'm not so sure that they're that unrelated, but we can keep them seperate. We just really need to revisit it, as that is also a weak point in the entire gesture area.

example/lib/pages/gestures_page.dart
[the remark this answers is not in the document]
JaffaKetchup: Yeah, bit long, but not really anything better. I'm good with that.

lib/src/map/options/enabled_gestures.dart
josxha: Sounds good, renamed the constructor.

example/lib/pages/gestures_page.dart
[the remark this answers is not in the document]
josxha: Using it to "Two finger drag" for now, if we come up with something better we can change it again of course.

lib/src/map/controller/map_controller_impl.dart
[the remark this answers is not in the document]
josxha: Removed `MoveAndRotateResult`.

I'd prefer to have a combined `changeCamera` method not as part of this pr.

lib/src/map/options/enabled_gestures.dart
josxha: Made the default constructor public.

example/lib/pages/markers.dart
TesteurManiak: Removing the `interactionOptions` property here and disabling `doubleTapDragZoom` in the other file seems fine to me 👍

lib/src/map/controller/events/map_event_source.dart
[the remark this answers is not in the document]
josxha: > It can be passed through from the MapController (or at least it should be if it can't atm), and is very useful in some cases.

Would it make sense to change the id type to `Object`? That way users would input everything they like and don't have to format and parse Strings.

lib/src/map/gestures/services/double_tap.dart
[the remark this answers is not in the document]
josxha: These are the the current method signatures (removed unused parameters):

- BaseGestureService
  - LongPressGestureService, SecondaryLongPressGestureService, TertiaryLongPressGestureService
    - `submit(LongPressStartDetails)`
  - ScrollWheelZoomGestureService
    - `submit(PointerScrollEvent)`
  - CtrlDragRotateGestureService
    - `start()`
    - `update(ScaleUpdateDetails)`
    - `end()`
  - DoubleTapDragZoomGestureService, DragGestureService, TwoFingerGesturesService
    - `start(ScaleStartDetails)`
    - `update(ScaleUpdateDetails)`
    - `end(ScaleEndDetails)`
  - DelayedGestureService
    - `setDetails(TapDownDetails)`
    - DoubleTapGestureService, TapGestureService, SecondaryTapGestureService, TertiaryTapGestureService
      - `submit()`

We can create a parent for these two:
- LongPressGestureService, SecondaryLongPressGestureService, TertiaryLongPressGestureService
- DoubleTapGestureService, TapGestureService, SecondaryTapGestureService, TertiaryTapGestureService

However there are unique gesture services like `ScrollWheelZoomGestureService` and `CtrlDragRotateGestureService` I don't know how we could use `@mustCallSuper`. Do you have an idea how to implement this so it gets enforced on every gesture service?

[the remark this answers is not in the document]
JaffaKetchup: Ok, I see. I'm not sure the current situation ideally expresses what inherits what.

There should be at least an intermediate `SingleShotGestureService` (which has `submit`) and `ProgressableGestureService` (which has `start`, `update`, `end`). We can use generic typing where necessary.  
This allows us to define base implementations of those methods which call `controller.stopAnimationRaw();` and enforce with `@mustCallSuper` & `@mustBeOverriden`, and the `base` class modifier to prevent implementations.  
`DelayedGestureService` might then be better as a mixin which provides `TapDownDetails`, or as a subclass of `SingleShotGestureService`, but I'm not sure about whether either of these would work, let alone which one is best.  
Have I missed something as to why this could not work? I probably have.

> However there are unique gesture services like `ScrollWheelZoomGestureService` and `CtrlDragRotateGestureService`

What makes these unique that a generic type wouldn't solve? And why doesn't `CtrlDragRotateGestureService` get args for `start` and `end`?

> `DoubleTapDragZoomGestureService`

Shouldn't this be a delayed gesture, as it relies on tap detection?

lib/src/map/options/interaction_options.dart
josxha: I did some prototyping. This is what I came up with:
#### each gesture one by one:
```dart
// option 1
const opt1 = InteractionOptions(
  enabledGestures: EnabledGestures.all(
    flingAnimation: flingEnabed,
  ),
  ctrlRotateKeys: [
    ...InteractionOptions.defaultCtrlRotateKey,
    LogicalKeyboardKey.keyA,
  ],
  scrollWheelVelocity: 1,
  twoFingerMoveThreshold: 2,
  twoFingerRotateThreshold: 3,
  twoFingerZoomThreshold: 4,
);

// option 2
final opt2 = const InteractionOptions.all(
  ctrlRotate: CtrlTriggerRotateGesture.enabled(
    triggerKeys: [
      ...InteractionOptions.defaultCtrlRotateKey,
      LogicalKeyboardKey.keyA,
    ],
  ),
  scrollWheelZoom: ScrollWheelZoomGesture.enabled(velocity: 1),
  twoFingerMove: TwoFingerMoveGesture.enabled(threshold: 2),
  twoFingerRotate: TwoFingerRotateGesture.enabled(threshold: 3),
  twoFingerZoomThreshold: TwoFingerZoomThreshold.enabled(threshold: 4),
  fling: FlingAnimationGesture(enabled: flingEnabed),
  // all other gestures enabled by the .all() constructor
);
```

#### Using grouped gestures
```dart
// option 1
const opt1 = InteractionOptions(
  enabledGestures:
      EnabledGestures.byGroup(move: true, zoom: false, rotate: false),
  scrollWheelVelocity: 1,
  twoFingerMoveThreshold: 2,
  twoFingerZoomThreshold: 4,
);

// option 2
final opt2 = const InteractionOptions.byGroups(
  zoom: ZoomGestures(
    scrollWheelVelocity: 1,
    twoFingerZoomThreshold: 4,
  ),
  move: MoveGestures(
    twoFingerMoveThreshold: 2,
  ),
  rotate: RotateGestures.disabled(),
);

```

Is this option 2 the the design idea you had?

example/lib/pages/markers.dart
josxha: > Removing the interactionOptions property here and disabling doubleTapDragZoom in the other file seems fine to me 👍

I wasn't aware that the `markers.dart` screen makes use of the onTap callback, sorry about that. If the user clicks on the map a marker on that position gets added. Therefore I kept double tap gestures disabled on that page. Let me know if you still prefer to have your requested changes implemented or if you are fine with how it is now.

JaffaKetchup:
TODO phase 2 (I think):

- [ ] Figure out https://github.com/fleaflet/flutter_map/pull/1733#discussion_r1428960703
- [ ] Add documentation throughout and maybe a nice little simple flowchart of what relations are here, what is responsible for what, and what triggers what (https://github.com/fleaflet/flutter_map/pull/1733#discussion_r1428961223)
- [ ] (not priority) Consider class modifiers. If we're always implementing something, make it an `interface`. If we should only `extend` something, make it `base` or `sealed`. Useful cheat sheet: https://miro.medium.com/v2/resize:fit:828/format:webp/1*C3lYhML0IzrddcZRKkZtEg.png
- [ ] Choose a name for https://github.com/fleaflet/flutter_map/pull/1733#discussion_r1428962255
- [ ] Figure out https://github.com/fleaflet/flutter_map/pull/1733#discussion_r1428962688
- [ ] (for Jaffa first) Figure out https://github.com/fleaflet/flutter_map/pull/1733#discussion_r1429228899
- [ ] Follow up: https://github.com/fleaflet/flutter_map/pull/1733#discussion_r1429194762

lib/src/map/gestures/services/ctrl_drag_rotate.dart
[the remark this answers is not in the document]
josxha: fixed in https://github.com/josxha/flutter_map/commit/905abbd402a6e3b94488df75c5e32d425936a84a

lib/src/map/gestures/services/double_tap.dart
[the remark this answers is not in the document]
josxha: The advantage of the current implementation is that we treat every gesture as individually as possible to have a simple while extendable class design.
We could use inheritance, generics or mixins. However what is the benefit that justifies the complexity, overhead less flexibility? What does this solve that we are currently not solving?

---

The name `DelayedGestureService` is a bit misleading. It is meant as a base service where the `TapDownDetails` get set first and gets submitted later on. The `DoubleTapDragZoomGestureService` uses a start(), update(), end() logic which doesn't get represented by the `DelayedGestureService`. I renamed it to `BaseDetailsGestureService`.

lib/src/map/gestures/services/base_services.dart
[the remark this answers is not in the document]
josxha: Added documentation in https://github.com/josxha/flutter_map/commit/6d90dd3d7a6f516a3552ec9036392875484731e6.

josxha:
I added changes like documentation and stuff. An open topic is how to handle https://github.com/fleaflet/flutter_map/pull/1733#discussion_r1429228899, most other topics should be included or discussed as delayed to future PRs.

lib/src/map/gestures/services/base_services.dart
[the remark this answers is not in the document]
josxha: I can move the `_rotateOffset` function as an either static or normal method inside of `BaseService` if that helps.
As it is only used inside of these gesture service classes I wouldn't want to move it to /misc.
Do you have some other ideas of where to put it?

lib/src/map/options/enabled_gestures.dart
josxha: Closing this as resolved since the merged discussion got resolved. 
ctrlDragRotate got renamed to keyTriggerDragRotate.

lib/src/map/options/enabled_gestures.dart
josxha: Do you want to have the bitfield operations removed or to keep them for those that prefer them?

lib/src/map/options/map_options.dart
josxha: Closing this conversation since the discussion it got merged with is resolved.

test/flutter_map_test.dart
josxha: The other conversation discussed the grouping of gestures by it's affect on `MapCamera` (zoom, move, rotate). This conversation means to have a categorization for gesture types like double taps.

It would be possible to add another constructor `EnabledGestures.byCategories(List<enum GestureCategory>)` with zooming, rotating, dragging, doubleTapping as fields. However it's unclear to me if a reasonable amount of users would use this.

lib/src/map/controller/events/map_events.dart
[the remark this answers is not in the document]
JaffaKetchup: Not sure what needs to be done here?

lib/src/map/gestures/services/base_services.dart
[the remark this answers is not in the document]
JaffaKetchup: We'll leave it here for now.

lib/src/map/controller/events/map_events.dart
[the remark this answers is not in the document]
josxha: Not sure either. `MapEventRotate` currently gets emitted for `keyTriggerDragRotate` and `twoFingerRotate` gestures. Both can be distingulished however by checking the event source. This discussion fits probably better in https://github.com/fleaflet/flutter_map/issues/1779.

lib/src/map/gestures/services/base_services.dart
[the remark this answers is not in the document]
JaffaKetchup: Leaving modifiers for now.

lib/src/map/options/enabled_gestures.dart
JaffaKetchup: We should have only one system, otherwise it can be confusing.

lib/src/map/options/interaction_options.dart
josxha: As further discussed internally we are keeping it like this for this pull request with the option to make changes to it later on if a better design comes up than the current.

test/flutter_map_test.dart
josxha: marking this conversation as resolved as https://github.com/fleaflet/flutter_map/pull/1733#discussion_r1428964013 is resolved.

example/lib/pages/markers.dart
josxha: Resolving this conversation for now, please feel free to reopen in case changes are required.

lib/src/map/gestures/services/double_tap.dart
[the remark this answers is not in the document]
josxha: Pull request that would resolve this thread in https://github.com/josxha/flutter_map/pull/2.

josxha:
Parity issues with `master` (tested in https://github.com/fleaflet/flutter_map/issues/1429#issuecomment-1896992462):
- [x] Stop map in motion by putting finger on map
- [ ] Rotate gesture on touch pad
- [ ] Pinch to zoom on touch pad (in/out)

lib/src/map/gestures/services/double_tap.dart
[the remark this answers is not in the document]
josxha: pr merged

josxha:
### Marking this pull request as ready to review again.

#### Some thoughts:
- Touchpad (aka. Trackpad) rotate gesture to rotate the map hasn't been included because I'm not able to implement this without a Macbook. It could be that `PointerPanZoomUpdateEvent.rotation` provides the necessary rotation data but I can't tell.
- The two finger scroll gesture hasn't been included (is not supported on master either). It seems to me that this event gets turned into a pan gesture with a `ScaleUpdateDetails.pointerCount` of 1.
- Trackpad behaviour is extremely platform dependent. Web uses the legacy trackpad service, modern windows and macos devices use the new trackpad service.
- The legacy trackpad service doesn't zoom on the cursor but on the center on the map.

lib/src/map/gestures/map_interactive_viewer.dart
```diff
@@ -0,0 +1,308 @@
...
+            // - On **Web**, where not enough data is provided by browser APIs.
+            //
+            // https://docs.flutter.dev/release/breaking-changes/trackpad-gestures#description-of-change
+            _trackpadLegacyZoom?.submit(event);
+            break;
+        }
+      },
+      // Trackpad gestures on most platforms since flutter 3.3 use
+      // these onPointerPanZoom* callbacks.
+      // See https://docs.flutter.dev/release/breaking-changes/trackpad-gestures
```
JaffaKetchup, line 157: Do you think we should provide a link to these docs and/or a short explanation in our docs?

lib/src/map/options/enabled_gestures.dart
```diff
@@ -0,0 +1,287 @@
...
+  /// Enable zooming with the device trackpad / touchpad
+  final bool trackpadZoom;
+
+  /// Enable rotation by pressing the defined keyboard key (by default CTRL key)
+  /// and dragging with the cursor
+  /// or finger.
+  final bool keyTriggerDragRotate;
+
+  /// Returns true of any gesture with more than one finger is enabled.
+  bool hasMultiFinger() => twoFingerMove || twoFingerZoom || twoFingerRotate;
```
JaffaKetchup, line 127: Prefer a getter here? Maybe under another name?

```dart
  bool get hasMultiFinger => twoFingerMove || twoFingerZoom || twoFingerRotate;
```

_(`suggestion` not made to prevent introduction of bugs elsewhere)_

lib/src/map/options/map_options.dart
```diff
@@ -70,13 +97,13 @@ class MapOptions {
   /// Only use this if your map isn't built immediately (like inside FutureBuilder)
   /// and you need to access the controller as soon as the map is built.
   /// Otherwise you can use WidgetsBinding.instance.addPostFrameCallback
-  /// In initState to controll the map before the next frame.
+  /// In initState to control the map before the next frame.
```
JaffaKetchup, line 100: (For @JaffaKetchup) I need to fix the documentation here, as this method is now the recommended way to attach a controller if needed at initial build for reliability: see https://docs.fleaflet.dev/v/v7-beta/usage/programmatic-control/controller#usage-inside-initstate (link likely to change). Might as well do it in this PR, I'll make the change before merge.

lib/src/map/options/enabled_gestures.dart
```diff
@@ -0,0 +1,287 @@
...
+    required this.doubleTapZoomIn,
+    required this.doubleTapDragZoom,
+    required this.scrollWheelZoom,
+    required this.twoFingerRotate,
+    required this.keyTriggerDragRotate,
+    required this.trackpadZoom,
+  });
+
+  /// Shortcut constructor to allow all gestures that don't rotate the map.
+  const EnabledGestures.noRotation()
```
JaffaKetchup, line 27: Personally, I would remove this constructor, as it's now easy to do with `byGroup` (especially if they have optional parameters). WDYT?

lib/src/map/options/enabled_gestures.dart
```diff
@@ -0,0 +1,287 @@
...
+        twoFingerMove: twoFingerMove ?? this.twoFingerMove,
+        doubleTapZoomIn: doubleTapZoomIn ?? this.doubleTapZoomIn,
+        doubleTapDragZoom: doubleTapDragZoom ?? this.doubleTapDragZoom,
+        scrollWheelZoom: scrollWheelZoom ?? this.scrollWheelZoom,
+        twoFingerRotate: twoFingerRotate ?? this.twoFingerRotate,
+        keyTriggerDragRotate: keyTriggerDragRotate ?? this.keyTriggerDragRotate,
+        trackpadZoom: trackpadZoom ?? this.trackpadZoom,
+      );
+
+  const EnabledGestures.byGroup({
```
JaffaKetchup, line 156: Maybe this should have optional params, `true` by default? Would allow removal of `noRotation` constructor. WDYT?

lib/src/map/options/enabled_gestures.dart
```diff
@@ -0,0 +1,287 @@
...
+/// you have two options:
+///   a. Add your own flags:
+///      [InteractiveFlag.drag] | [InteractiveFlag.twoFingerRotate]
+///   b. Remove unnecessary flags from all:
+///     [InteractiveFlag.all] &
+///       ~[InteractiveFlag.flingAnimation] &
+///       ~[InteractiveFlag.twoFingerMove] &
+///       ~[InteractiveFlag.twoFingerZoom] &
+///       ~[InteractiveFlag.doubleTapZoomIn]
+abstract class InteractiveFlag {
```
JaffaKetchup, line 216: As in https://github.com/fleaflet/flutter_map/pull/1733#discussion_r1428964382, we should probably remove this. I'm not sure if there's a convienient method of deprecation here, but this redundancy just complicates things IMO. WDYT?

JaffaKetchup (changes requested):
Still untested, but here's review round 2 :D. Mostly just smaller stuff now, and there's less than it looks at first glance.
I would like for @mootw @TesteurManiak or @ibrierley to also review this PR, and any other community reviews also appricated! It's quite a big PR, so want to make sure I/we haven't missed anything.

lib/src/map/options/enabled_gestures.dart
josxha: I don't care either way but @mootw had a strong preference for this. I let the discussion up to both of you.

lib/src/map/options/enabled_gestures.dart
josxha: Having no map rotation is probably one of the most used map options. Making it as easy as can be therefore makes sense to me.
The additional constructor doesn't has any performance or mentionable maintainence penalty.

lib/src/map/options/enabled_gestures.dart
josxha: the name `byGroup` would be a disable-all-by-default constructor for my understanding. It wouldn't be reasonable for me that calling this constructor without parameters would enable all gestures.
Default parameter values are a great thing but in this case its 3 boolean parameters we can assume the user wants to set at least one anyways when using it.

lib/src/map/options/map_options.dart
josxha: off topic, please do in a separate pr or in https://github.com/fleaflet/flutter_map/pull/1776

lib/src/map/options/enabled_gestures.dart
josxha: Removed this method because it was only used in `MapInteractiveViewer.updateGestures` and could cause confusion between multi point touchscreen and mutli point trackpad gestures.

lib/src/map/controller/map_controller.dart
```diff
@@ -105,7 +104,7 @@ abstract class MapController {
   ///
   /// The operation was successful if both fields of the resulting record are
   /// `true`.
```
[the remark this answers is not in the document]
josxha: fixed

lib/src/map/controller/map_controller.dart
```diff
@@ -121,7 +120,7 @@ abstract class MapController {
...
```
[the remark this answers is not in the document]
josxha: fixed

lib/src/map/controller/events/map_event_source.dart
```diff
@@ -0,0 +1,30 @@
...
+  custom,
+  scrollWheel,
+  nonRotatedSizeChange,
+  tertiaryTap,
+  tertiaryLongPress,
+  secondaryLongPressed,
+  keyTriggerDragRotateStart,
+  keyTriggerDragRotateEnd,
+  keyTriggerDragRotate,
+  trackpad,
```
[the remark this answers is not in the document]
josxha: I matched the name with the naming of the `onTwoFinger` sources (`twoFingerStart`, `onTwoFinger`, `twoFingerEnd`).

Ideally the MapEventSource should describe where the event came from and not what event has happend. But this should be discussed in https://github.com/fleaflet/flutter_map/issues/1779.

lib/src/map/options/enabled_gestures.dart
JaffaKetchup: > we can assume the user wants to set at least one anyways when using it

Yeah, so if they want to enable one, then they need to set the others as well. If they want to disable one, they need to set the others as well. If they want to enable two, they still need to disable the last, and vice versa. I think this just backs up my argument more.

> the name byGroup would be a disable-all-by-default constructor for my understanding. It wouldn't be reasonable for me that calling this constructor without parameters would enable all gestures.

OK, then we can default to `false`. Or rename to `disableByGroup`. Either way, I feel pretty strongly about this.

lib/src/map/gestures/map_interactive_viewer.dart
josxha: I brief explaination of every gesture in the docs would probably be useful especially because there are gestures that are not well known (e.g. double tap drag to zoom). 
The linked documentation is more for better understanding when working with the internal code. Therefore either not mentioned in our docs or as "for more information in this see here"-thingy.

lib/src/map/options/enabled_gestures.dart
josxha: Optimally we double have both to let users decide if they want to follow a whitelist or blacklist approach. We could do a similar design as the individual gestures by providing a `byGroup`, `disableByGroup` and `enableByGroup` constructor. 

Otherwise I'd prefer false as default value for the `byGroup` constructor.

What do you prefer?

JaffaKetchup: Ok, let's support all 3. Although we should keep it consistent between `all`/`none` and `enabled`/`disabled`: it's up to you which you think is better.

JaffaKetchup: Then remove `noRotation`?

josxha: > Ok, let's support all 3. Although we should keep it consistent between all/none and enabled/disabled: it's up to you which you think is better.

Added `allByGroup` and `noneByGroup` contructors.

> Then remove noRotation?

Still no fan of this shortcut constructor, huh? 😄 Alright, I removed it.

lib/src/map/options/enabled_gestures.dart
JaffaKetchup: I believe Moo may now be busy for a while with other matters. Can you link to his discussion?

JaffaKetchup: Ok, let's leave it in for now.

JaffaKetchup (changes requested):
- [ ] One bug with cursor keyboard rotation. Rotation is not tracking the cursor properly. I remember having a similar issue when implementing the algorithm initially. Maybe double checking that nothing in the implementaiton itself has changed, and every variable is being changed/reset at the correct points.

- [ ] Cursor keyboard rotation also no longer sets rotation to the angle of cursor immediately when clicking.

https://github.com/fleaflet/flutter_map/assets/58115698/74cc4727-7a31-49c7-8617-b5ecf21a9c35

---

- [ ] One bug with fling event end/finished event not being emmitted (when fling is cancelled with tap (as shown in video), or when finished naturally)

---

- [ ] Double tap zoom no longer emits correct event.

PR           |  Master
!pr  |  master

---

- [ ] Can no longer fling if drag gesture is disabled. On master, enabling fling without drag would start a fling if the normal conditions were met, but wouldn't drag/move if not, and wouldn't move whilst normal dragging.

---

- [ ] Double tap + drag gesture now loads tiles whilst zooming. master behaviour isn't correct either. The correct behaviour is that new tiles shouldn't load and old ones shouldn't be pruned whilst moving, and the refresh should occur when the gesture completes.

---

- [ ] On tap events and callbacks are no longer triggered.

---

- [ ] Can't seem to perform the new trackpad zoom gesture. Can you explain how to do it?

---

Still untested on any touch devices. Will perform that testing later.

josxha:
Branch moved, see new pr here: https://github.com/fleaflet/flutter_map/pull/1809

PR #1734, opened 2023-11-22
chore: refactor example app

josxha:
- [x] use "new" language features like `super.key`
- [x] turn things like the drawer into their own widgets
- [x] use `CancellableNetworkTileProvider` everywhere possible since it's the recommended way and to avoid console warnings
- [x] remove the `CancellableTileProviderPage`
- [x] restructure plugin files
- [x] ~~use lint rules from the package for the example project too~~ Match the lint rules with the package automatically
- [x] Bump `flutter_lints` to `^3.0.0`


example/analysis_options.yaml
```diff
@@ -27,4 +27,18 @@ linter:
...
+    prefer_asserts_with_message: true
+    missing_whitespace_between_adjacent_strings: true
+    avoid_multiple_declarations_per_line: true
+    noop_primitive_operations: true
+    avoid_void_async: true
+    avoid_types_on_closure_parameters: true
+    unnecessary_null_checks: true
+    unnecessary_parenthesis: true
+    use_if_null_to_convert_nulls_to_bools: true
+    matching_super_parameters: true
```
TesteurManiak, line 44: Shouldn't we add `flutter_lints` as a dev_dependency to the example instead? This would avoid to maintain the rule set 🤔

example/lib/main.dart
```diff
@@ -50,8 +49,6 @@ class MyApp extends StatelessWidget {
       ),
       home: const HomePage(),
       routes: <String, WidgetBuilder>{
-        CancellableTileProviderPage.route: (context) =>
-            const CancellableTileProviderPage(),
```
TesteurManiak, line 54: I'm not sure that we want to use `CancellableNetworkTileProvider` everywhere, the whole point of having a separated example is to showcase its benefits compared to the default one. Moreover, the warnings are only leveraged on the web, you wouldn't have them on desktop or mobile.

TesteurManiak (changes requested):

example/lib/widgets/drawer/floating_menu_button.dart
```diff
@@ -0,0 +1,35 @@
...
+      child: SafeArea(
+        child: Container(
+          decoration: BoxDecoration(
+            color: Theme.of(context).colorScheme.background,
+            borderRadius: BorderRadius.circular(999),
+          ),
+          padding: const EdgeInsets.all(8),
+          child: Row(
+            children: [
+              Builder(
```
TesteurManiak, line 20: I don't think that `Builder` is needed here, the `context` of `FloatingMenuButton` should already have access to the Scaffold.

example/analysis_options.yaml
josxha: We are already using `flutter_lints` in the example project. See in line 1:
```yaml
include: package:flutter_lints/flutter.yaml
```
There are additional linters specified in the `flutter_map/analysis_options.yaml` file that I thought should match the example project at `flutter_map/example/analysis_options.yaml`.

The only rules I **haven't** copied over are
- `avoid_redundant_argument_values` because it can make sence to specify the default paramters in the example app as showcase and for explainational purpose.
- `prefer_single_quotes` since the project currently uses both `"` and `'`. I could change this tho for more concistency.

But I agree that it would be nice to use some trick here do match the rules with the package if somehow possible.

TesteurManiak: Oh my bad you're right!
We might want to update our version of flutter_lints because from what I see, the latest release exposes some new rules such as `use_super_parameters`, so we could remove it from our `analysis_options`.

josxha: Agree, completely missed that it has a new major version. 😅 

Btw see my changes to `example/analysis_options.yaml`. We are now matching the package lint rules automatically by using:
```yaml
include: ./../analysis_options.yaml
```

example/lib/main.dart
josxha: Wouldn't it be better to have a page that doesn't use the `CancellableNetworkTileProvider` or (probably better) where users have the option to toggle the usage of it?
If we don't use the `CancellableNetworkTileProvider` it floods the console with warnings when running in debug mode on the web:
image
This makes it harder to check for other potiential warnings and logs while testing.

This has probably been discussed before but wouldn't it make sence to embrace the tile cancellation on every platform? While it leads to delayed tile loading on the web pretty quickly it **does** impact the speed on other platforms as well by using bandwidth. If the user has a bad mobile network connection or only 3G this has a big impact too. Moreover it decreases the data usage and saves up bandwidth for the tile provider.

Back to our example application: I think it makes sense to utilise these benefits here as good as possible as showcase for "best practise". 

Would you be fine with a page that lets users toggle the usage of the `CancellableNetworkTileProvider`?

example/lib/widgets/drawer/floating_menu_button.dart
josxha: Good point, fixed.

JaffaKetchup:
I'll have a review of this as well, a little later.

example/lib/main.dart
TesteurManiak: > This has probably been discussed before but wouldn't it make sence to embrace the tile cancellation on every platform?

It's already the case but here's what @JaffaKetchup has written about that on the package:

> Although HTTP request abortion is supported on all platforms, it is especially useful on the web - and therefore recommended for web apps. This is because the web platform has a limited number of simulatous HTTP requests, and so closing the requests allows new requests to be made for new tiles.
> On other platforms, the other benefits may still occur, but may not be as visible as on the web.

---

> Would you be fine with a page that lets users toggle the usage of the CancellableNetworkTileProvider?

Sure! It would still showcase the difference with/without a cancellable tile provider so I'm fine with this.

josxha: fixed

TesteurManiak (commented):
LGTM! (waiting for @JaffaKetchup's review before approving)

JaffaKetchup:
@TesteurManiak @ignatz Let me know what you think. If you're happy @TesteurManiak feel free to approve and merge.

example/lib/widgets/notice_banner.dart
```diff
@@ -0,0 +1,83 @@
+import 'package:flutter/material.dart';
+import 'package:url_launcher/url_launcher.dart';
+
+enum NoticeBannerType {
```
josxha, line 4: If we use named constructors for the `NoticeBanner` we don't need an extra enum.

JaffaKetchup: Good idea, I'll do that.

JaffaKetchup: @josxha Completed in 40a5019.

ignatz:
I only skimmed through the many lines. I only found a few pick-your-own-destiny nits. Looks like a great brush up. Thank you Josha

TesteurManiak (approved):

josxha:
Thanks @ignatz  I appreciate it :)

PR #1738, opened 2023-11-26
refactor!: merge `MapInternalController` with `MapControllerImpl`

josxha:
## Motivation

- `FlutterMapImpl` has become an internal object in `flutter_map` version 5. Since then there are currently two abstration levels of the map controller (`MapController` uses `MapControllerImpl` uses `MapInternalController`).
- Most of the time `MapControllerImpl` calls its underlying `MapInternalController` and passes it's arguments though in addition to setting some defaults.
- Two internal controller layers require complex bindings to glue them together.

## Changes in this pull request

- Merge the `MapInternalController` with `MapControllerImpl`
- Allow "raw" methods (naming inspired by the flutter sdk) to be called internally with all parameters while still hiding them from the public API.
- Remove `FlutterMap` prefix, use a shorter `Map` prefix instead
  - Rename `FlutterMapInteractiveViewer`to `MapInteractiveViewer`  
  - Rename `FlutterMapNetworkImageProvider` to `MapNetworkImageProvider`
  - Rename `FlutterMapInheritedModel` to `MapInheritedModel`

## Roadmap
1. Refactor the map controller logic (this PR)
2. Add animation logic to the internal map controller (follow up PR)
3. Continue work on the gesture system rewrite that needs to have animations (https://github.com/fleaflet/flutter_map/pull/1733)


TesteurManiak:
I don't mind the name changes, but I would still recommend to create deprecated typedefs for the old names just to not instantly breaks code from the previous version and inform users that they should migrate to the new names, for example:
```dart
@Deprecated('Use MapInteractiveViewer instead')
typedef FlutterMapInteractiveViewer = MapInteractiveViewer;
```

josxha:
Typedefs are a great idea for deprecations.👍🏼 Do we need to deprecate internal code though? 🤔 The renamed classes are not exported and therefore not directly exposed outside of the package.

TesteurManiak:
> Typedefs are a great idea for deprecations.👍🏼 Do we need to deprecate internal code though? 🤔 The renamed classes are not exported and therefore not directly exposed outside of the package.

True if all classes are internal, no need for deprecation. I thought that some of them were exposed, but if it's not the case then don't bother 👍

lib/src/layer/tile_layer/tile_provider/network_image_provider.dart
```diff
@@ -11,8 +11,7 @@ import 'package:http/http.dart';
 /// Note that specifying a [fallbackUrl] will prevent this image provider from
 /// being cached.
 @immutable
-class FlutterMapNetworkImageProvider
-    extends ImageProvider<FlutterMapNetworkImageProvider> {
+class MapNetworkImageProvider extends ImageProvider<MapNetworkImageProvider> {
```
JaffaKetchup, line 14: Additionally as @TesteurManiak said, some typedef deprecations are at least needed. `(Flutter)MapNetworkImageProvider` may be commonly used externally, for example.

JaffaKetchup (changes requested):
I think it's pretty much good (except for my point above) except for these couple of things.

IMO in general, I prefer to avoid changing names when it doesn't add clarity or better match what it's assigned to. But I realise that's just my personal way of doing things, so I won't let that block this.

josxha:
> It's probably worthwhile reading through the history of these objects, in the discussion at https://github.com/fleaflet/flutter_map/pull/1548 and the https://github.com/fleaflet/flutter_map/pull/1551 PR

Thanks for the links! In that case we should probably invite @rorystephenson to this discussion as well.
I try to summarize some points while reading along:
- Have MapController as public API and hide internal methods and variables from external and plugin access.
- "`MapControllerImpl` is given direct access to another internal controller which manipulates the map state." *(Sadly the reasons for this haven't been discussed further.)*
- Making `MapController` a sealed class has been mentioned but not implemented, *maybe worth adding here.*
- We want to have an immutable FlutterMapState
- "All changes to the map state are [...] managed by [the] FlutterMapStateController."
- "MapController is a completely separate concept from a widget's state. The controller is intended to let you control the state, not access and directly manipulate it, which is important in reactive programming. You want to be able to tell the map to "move here" or "change the camera to these bounds" and then react to any state changes that result (if you desire). Directly manipulating the state, for example to change it's center/zoom value, creates coupling.

As far as I can tell this PR doesn't violates the principles mentioned in the other discussions. This pull request doesn't even breaks public APIs, it's all internal.

The concept of two internal layers for the controller has indeed been introduced in https://github.com/fleaflet/flutter_map/pull/1551 but was not mentioned in the discussion. @rorystephenson, could you give some details about the design choice? I'm curious to hear what you think about this PR too, maybe I'm overlooking something here.

lib/src/layer/tile_layer/tile_provider/network_image_provider.dart
josxha: For real? `(Flutter)MapNetworkImageProvider` is not exported by the package therefore shouldn't be public in my opinion.
Are you referring to the unchanged `NetworkTileProvider` that is part of the public API?

josxha: I can add deprecation typedefs for internal classes, but in what szenario are they used externally? They wouldn't have any usage internal too because the code is migrated to the new class names.

lib/src/map/controller/map_controller_impl.dart
```diff
@@ -1,66 +1,135 @@
...
+
+  @override
+  Stream<MapEvent> get mapEventStream => _mapEventStreamController.stream;
 
-  MapCamera get camera => value.camera;
+  MapOptions get options {
+    assert(
+      value.options != null,
+      'You need to have the FlutterMap widget rendered at least once before using the MapController.',
+    );
```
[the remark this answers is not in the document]
josxha: True, seems to be a leftover - fixed.

lib/src/layer/tile_layer/tile_provider/network_image_provider.dart
JaffaKetchup: Ah, sorry, thought I'd already responded here. No, I got it wrong, must have been half asleep when I reviewed this: of course FMNIP isn't used externally, that would be NTP. You're right, no reason to have typedefs for internal deprecations.

JaffaKetchup (approved):
If we're sure this isn't impacting any advanced use-cases where people might be implementing their own highly-specialised controller, I'm OK with this.
If this isn't breaking, does it need the ! in the title?

josxha:
> If we're sure this isn't impacting any advanced use-cases where people might be implementing their own highly-specialised controller, I'm OK with this.

I'm quite confident here because of the changes made in https://github.com/fleaflet/flutter_map/pull/1551 (users don't extend from MapController but use it as parameter like `flutter_map_animations`)

> If this isn't breaking, does it need the ! in the title?

Oh, yes that's correct - changing it.

josxha:
hi @JaffaKetchup I merged the master branch.
Additionally i wanted to use `radians2Degrees` instead of `radianToDeg()` first but then removed it again in favor of https://github.com/fleaflet/flutter_map/pull/1763. Do you want to have another quick look at it before we merge?

PR #1742, opened 2023-12-01
refactor!: multiple fixes & additions to `NetworkTileProvider` (and underlying `ImageProvider`)

### LINKED ISSUE #1698: [BUG] `NetworkTileProvider` has poorer performance and new unhandled exceptions since v6
### What is the bug?

Sometimes not all the tiles are being loaded/drawn.
After moving/dragging the map all the tiles are immediately drawn/refreshed

### How can we reproduce it?

It might be me with all the code in my project, but try with these tiles:
TileLayer(
        urlTemplate: "https://{s}.basemaps.cartocdn.com/rastertiles/voyager/{z}/{x}/{y}.png",
        subdomains: ['a', 'b', 'c', 'd'],
        tileProvider: NetworkTileProvider(),
        keepBuffer: 10,
        maxZoom: 23,
        retinaMode: false);

A workaround: add a method in MapController to refresh/redraw the tiles on demand.

### Platforms

All

### Severity

### LINKED ISSUE #1703: [FEATURE] Improve error handling of built-in image provider

As discussed in #1698.

* Reduce the number of errors that trigger breakpoints when debugging
* Achieve an effect similar to what is seen in Flutter's `NetworkImage`

_No response_

The behaviour has changed in v6 from v5 due to changed error handling. In combination with using `loadBytes` instead of lower-level IO, this triggers more errors, but allows us to keep our code simple and platform independent.

It is also worth saying that the current implementation may not be performant in comparison to a standard `NetworkImage`, but more testing is need to validate this. See #1698 for an issue designated for reporting/tracking the performance of the tile provider.

### Severity

JaffaKetchup:
Fixes #1698 (and fixes #1703) as much as possible.

This transitions to a more consistent internal implementation of the `FlutterMapNetworkImageProvider`, both in that it uses the same asynchronous styles throughout, and follows roughly what is recommended by the `ImageProvider` documentation.
It also fixes a bug or two, and aims to improve performance as much as is feasibly possible - however, because we cannot depend on dart:io without a lot of extra code, we cannot use some efficiency methods such as `consolidate...whatever`, but I think there'll be little differences with the tiny images of tiles. An option has been added to just ignore exceptions in the HTTP IO portion of the image provider.

This will be reflected in the `CancellableNetworkTileProvider` in its v2, by PR fleaflet/flutter_map_cancellable_tile_provider#4.


josxha (approved):
lgtm, tested on windows and web

PR #1743, opened 2023-12-02
perf: avoid caching a single multiplication

mootw:
avoids caching a single multiplication operation. converts all toRadians and other helpers to use vector_math constant. (part of reducing dependency on latlong2).


TesteurManiak (approved):
LGTM!

JaffaKetchup (changes requested):
Some files import '_64', some don't? Is that intentional?

josxha:
As far as I can see had `vector_math` as a dependency but it was not used anywhere?
This PR introduces the use of it for the `degrees2Radians` constant. Are there additional use cases in the future? If not I'd prefer to remove the dependency entirely and add the const to flutter_map directly. 

Long terms I think it should be used from a future latlong package but not the current implementation of `latlong2`.

JaffaKetchup:
I would agree with @josxha here - adding a dependency just for this seems a little bit too much.

mootw:
vector_math is a core flutter dependency, and we already use it in camera.dart for the matrix transform at line 301, we just don't explicitly depend on it. Matrix4 is exposed by `package:flutter/material.dart`

JaffaKetchup (approved):
LGTM! I suppose that we indirectly depend on it anyway, so no harm in depending on it directly as well.

Would be great if you could resolve the conflicts as well :D.

josxha (approved):
thanks for the explanation. pr lgtm

JaffaKetchup:
(@josxha feel free to cherry pick)

JaffaKetchup:
> np! also, we could consider using the aabb2 and other methods from vector_math in the future for things like bounding boxes and transforms!

Yep, that would make sense.

PR #1745, opened 2023-12-02
revert: #1731

JaffaKetchup:
With new plans underway to eventually remove the 'latlong2' dependency altogether, this should help avoid user's needing two migrations.

Should be cherry picked into v6.1.0.


josxha (approved):
lgtm

PR #1747, opened 2023-12-02
fix: `MapPosition.hashCode` value distribution

josxha:
This PR adds a better value distribution of for `MapPosition.hashCode`.
### Old
Adding the 3 hashCodes together has the effekt that relativly bigger values are more likely to happen. This increases the chance of hash code collision.
```dart
  @override
  int get hashCode => center.hashCode + bounds.hashCode + zoom.hashCode;
```

### New
If we use `.hashCode` anyway let's do the following change:
```dart
  @override
  int get hashCode => Object.hash(center, bounds, zoom);
```

Can and should get cherry picked to `v6.1.0-rc`

### DIFF
--- a/lib/src/misc/position.dart
+++ b/lib/src/misc/position.dart
@@ -17,7 +17,7 @@ class MapPosition {
   });
 
   @override
-  int get hashCode => center.hashCode + bounds.hashCode + zoom.hashCode;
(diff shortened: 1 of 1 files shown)


TesteurManiak (approved):
LGTM, I'm wondering why it was like this in the first place 😅

JaffaKetchup (approved):

PR #1763, opened 2023-12-07
perf: use `radians2Degrees` const from `vector_math`

josxha:
This PR is a follow up to https://github.com/fleaflet/flutter_map/pull/1743.
It makes use of the `radians2Degrees` from the `vector_math` package.


josxha:
These were the leftover usages of these functions from latlong2:
- all `radianToDeg()` calls were removed in #1743
- all `degreeToRadian()` calls get removed with this PR

I currently know very little about the `vector_math` package but we can keep this PR open and turn it into an general "make better use of `vector_math`" PR.

JaffaKetchup (approved):
LGTM. Feel free to merge this now, we can integrate more vector_math later.

PR #1776, opened 2023-12-12
chore: document public APIs (part 1)

josxha:
This is a work in progress to document public APIs.

A lot of documentation has been added (master is currently on 603 undocumented members, this pull request crunches this number down to 85).
Marking this pull request as ready to review now because it feels like chasing a moving target with this pr.

### If any reviewer has additions or wants to have some things changed, please commit directly to the branch if possible. There are documenations that could probably provide more details. The focus for this pull request was a bit more on quantity than on completeness.

PR #1777, opened 2023-12-12
refactor!: generic cleanup & minor performance improvements

josxha:
Some smaller changes alongside https://github.com/fleaflet/flutter_map/pull/1776.
See the [commit history](https://github.com/fleaflet/flutter_map/pull/1777/commits) for the list of changes.


JaffaKetchup (approved):
Mostly LGTM.
One thing that I'm a little unsure about are the changes to `MoveAndRotateResult`. The performance gain is theoretically negligible, and it is the perfect candidate for a `Record`. If left as a class, it should probably have at least the equality/`hashCode` methods.

josxha:
@JaffaKetchup reverted the changes to `MoveAndRotateResult` as it will get changed in https://github.com/fleaflet/flutter_map/pull/1733 anyways.

PR #1795, opened 2024-01-13
perf!: add simplification caching to `PolygonLayer` & other performance improvements

JaffaKetchup:
Reflected the same changes made to `PolylineLayer`. Also includes some minor performance enhancements and consistency changes to both.

Updated example project and improved consistency and styling. Added `PerformanceOverlay` widget to performance stress testing pages.

Also includes some other minor performance improvements (eg. wrapped `RepaintBoundary` around widget tree), and deprecates `Polygon.isFilled`.

### DIFF
--- a/example/lib/pages/many_circles.dart
+++ b/example/lib/pages/many_circles.dart
@@ -4,6 +4,7 @@ import 'package:flutter/material.dart';
 import 'package:flutter_map/flutter_map.dart';
 import 'package:flutter_map_example/misc/tile_providers.dart';
 import 'package:flutter_map_example/widgets/drawer/menu_drawer.dart';
+import 'package:flutter_map_example/widgets/number_of_items_slider.dart';
 import 'package:latlong2/latlong.dart';
 
 const maxCirclesCount = 20000;
@@ -26,7 +27,8 @@ class ManyCirclesPageState extends State<ManyCirclesPage> {
       source.nextDouble() * (end - start) + start;
   List<CircleMarker> allCircles = [];
 
-  int _sliderVal = maxCirclesCount ~/ 10;
+  static const int _initialNumOfCircles = maxCirclesCount ~/ 10;
+  int numOfCircles = _initialNumOfCircles;
 
   @override
   void initState() {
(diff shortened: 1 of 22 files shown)


JaffaKetchup:
> I assume .geojson.noformat is indicate that the file should not get treatend as a browsable file by tools or the user?
I'd suggest to gzip the geojson file. This shrinks the file size from 6.92MB down to 1.85MB, is build into dart:io and optimized for fast decoding.
Caveats: While browsers use gzip to speed up the fetching of websites, dart:html does not support gzip. Maybe we can still find a way around this limitation? (If it's no quick fix, don't worry about it since it can be changed later.)

I did consider compressing it, but I think the effort required to keep it working on web and quickly could offset the benefit. The '.noformat' part is just a 'random' string to prevent auto-formatters from formatting the JSON and bloating the file size with empty whitespace.

> The Polygon fill color currencly only works when simplification is disabled.

Good spot, will need to investigate that.

> The Polygon Layer screen showcased features like dotted borders, labels and so on before. While the polyline PR added simplification and culling to the existing page, this polygon PR removes the other polygons. Therefore could we showcase this feature inside a different screen?

I wasn't considering what I was doing fully when I removed the original polygons :D. I'll bring them back to where they were.

JaffaKetchup:
@josxha

> The Polygon fill color currencly only works when simplification is disabled.

Fixed, was an issue with the `copyWithNewPoints` method. Also fixed a related equality issue.

> The Polygon Layer screen showcased features like dotted borders, labels and so on before. While the polyline PR added simplification and culling to the existing page, this polygon PR removes the other polygons. Therefore could we showcase this feature inside a different screen?

I've split out both the polyline and polygon stress tests into their own page, grouped them with the Many Markers/Circles pages, redesigned those last two for consistency, and added the `PerformanceOverlay` widget.

example/lib/pages/many_circles.dart
```diff
@@ -4,6 +4,7 @@ import 'package:flutter/material.dart';
 import 'package:flutter_map/flutter_map.dart';
 import 'package:flutter_map_example/misc/tile_providers.dart';
 import 'package:flutter_map_example/widgets/drawer/menu_drawer.dart';
+import 'package:flutter_map_example/widgets/number_of_items_slider.dart';
 import 'package:latlong2/latlong.dart';
 
 const maxCirclesCount = 20000;
```
josxha, line 10: Use `maxCirclesCount` as max value for the slider.

josxha (changes requested):
Good pull request! I only found one small thing that could be changed.

josxha (approved):
lgtm

PR #1800, opened 2024-01-15
perf: add `Canvas.drawVertices` render pathway for `PolygonLayer` & fix bundled drawing

JaffaKetchup:
`Canvas.drawVertices` is a low level, very performant, canvas method. See https://www.youtube.com/watch?v=pD38Yyz7N2E for more info.

Will use https://pub.dev/packages/dart_earcut, and based roughly off 22b6cd3.

---

See documentation at https://docs.fleaflet.dev/v/v7-beta/layers/polygon-layer#performant-rendering-with-drawvertices-internal-disabled.

I've seen mixed results in terms of performance. It seems to generally greatly increase (up to 2x) performance on the GeoJson example (which I've also updated in this PR), with simplification enabled. However, with simplification disabled, performance has tanked - the triangulation algorithm simply isn't fast enough for all 138k points.

For this reason, and also the other considerations (mentioned within docstrings and on the docs site), I've disabled it by default.

Interestingly, from some limited testing on my budget Android device, the Impeller engine seems to not break things - and in fact, improves it a fair amount! We'll wait and see how well it works when it's finished. I'm in fact wondering whether the improvements Impeller gives dwarfs the improvements here, but more testing at another point is needed to determine that - and it wouldn't necessarily be a bad thing :D.


mdmm13:
Just wanted to write a note of thanks to @JaffaKetchup for taking a lead on this. flutter_map may not be able to use the GPU like flutter-maplibre-gl or other native implementations, but on this path, performance will likely become comparable and any remaining differences negligible.

Maybe the learnings will even inspire vector maps to go that route and then there will be no way around flutter_map for anyone doing mapping in Flutter 🚀

So - THANK YOU!

JaffaKetchup:
Hey @mdmm13, huge thanks :) Credit where credit is due, most of these perf improvements were brainstormed by @mootw and @ignatz. But we appriciate your thanks, and we hope that flutter_map leads the way for mapping in Flutter soon as well!

mdmm13:
> Hey @mdmm13, huge thanks :) Credit where credit is due, most of these perf improvements were brainstormed by @mootw and @ignatz. But we appriciate your thanks, and we hope that flutter_map leads the way for mapping in Flutter soon as well!

Apologies, was searching Discord for drawVertices, but didn't see anything, so I thought it was you.

In that case - THANK YOU to all three of you @mootw, @ignatz and @JaffaKetchup. What you guys are doing with flutter_map is fantastic.

Once we have any meaningful revenue, we'll happily start sponsoring.

JaffaKetchup:
I believe this is now in a reviewable and (hopefully) usuable state!

As well as the normal code reviews, I'd appricate any feedback from anyone else (ping @mdmm13)! See https://docs.fleaflet.dev/v/v7-beta/getting-started/installation#from-github.com for installation instructions from GitHub, and docs at https://docs.fleaflet.dev/v/v7-beta/layers/polygon-layer#performant-rendering-drawvertices for more info.

JulianBissekkou:
@JaffaKetchup 
Thanks for improving the rendering speed of flutter_map. Do you roughly know how much faster this adjustment made the rendering?
Are there more parts of the map that can be improve by drawing verticies instead?

JaffaKetchup:
> Do you roughly know how much faster this adjustment made the rendering?

It's hard to say. On the GeoJson example (with simplification 0.5), it's effectively halved the raster thread times, with a slight increase on the UI thread. However, if you disable simplification, then it drastically increases the UI thread times, to the point where performance is worse than if it wasn't used at all - triangulating 138,000 points every frame isn't cheap. I would encourage you to try it with your own data set.

> Are there more parts of the map that can be improve by drawing verticies instead?

No, not really. The map is already raster images, we can't improve this. Maybe the vector plugin could get some benefit out of this, but we can't. Lines are too complex to apply this logic to, and there's not usually much point.

JaffaKetchup:
> have you guys ever discussed taking on vector as part of flutter_map or where's the place you discuss general strategy for the flutter_map package/ ecosystem?

We have briefly discussed it, but for now, we have no plans in any form to implement vector stuff on our side. We usually discuss this sort of stuff privately or semi-privately (with a concerned group of people) on the Discord server, but we're happy to hear any suggestions and discuss a little bit further with everyone involved. I'm happy to make a dedicated channel, but it is not a focus for now, and I don't want to tread on @greensopinion's toes, or add any more work to his shoulders.

> or by some magic Impeller?

Impeller is looking to be really promising here, I'm eager to see how well it works when stable.

greensopinion:
> we're happy to hear any suggestions and discuss a little bit further with everyone involved. I'm happy to make a dedicated channel, but it is not a focus for now, and I don't want to tread on @greensopinion's toes, or add any more work to his shoulders.

Great to hear about ideas. Vector maps are great in theory but they are a lot of work to get right, and poor performance of the current implementation makes it hard to justify investing more without a plan to address that.

ignatz:
> > we're happy to hear any suggestions and discuss a little bit further with everyone involved. I'm happy to make a dedicated channel, but it is not a focus for now, and I don't want to tread on @greensopinion's toes, or add any more work to his shoulders.
> 
> Great to hear about ideas. Vector maps are great in theory but they are a lot of work to get right, and poor performance of the current implementation makes it hard to justify investing more without a plan to address that.

@greensopinion : Big fan of your vector tiles work. I feel your frustration but on the upside I don't see why fundamentally we wouldn't be able to make it work :partying_face: .

AFAICT, the concrete proposal being floated here is to use `drawVertices` here: https://github.com/greensopinion/dart-vector-tile-renderer/blob/main/lib/src/features/fill_renderer.dart#L44-L49. That could be a nice win given that raster performance is the most elusive.

There's also some wishful thinking that impeller might help in the future with some of the render performance issues. Personally, I think that's a plausible hypothesis. I'd not be surprised if a lot of the stutter is caused by shader recompilation. Fortunately, we should already be able to validate that hypothesis by either running on iOS or using the android alpha.

josxha:
`Canvas.drawVerticies` is really promising! I did some prototying too and was able to get it to ~60fps with the test dataset.

drawVertices obviously is the way to go here. 👍

https://github.com/fleaflet/flutter_map/assets/34318751/187e6af9-b750-40a3-9794-609e6fcad07e

JaffaKetchup:
Was that using this PR or something else? It seems to be missing the simplification slider.

JaffaKetchup:
Also, I want to be clear: Impeller seems to be helping already, at least the Android implementation! It definitely has a great impact, I just wonder whether it will get even better before release.

josxha:
> Was that using this PR or something else? It seems to be missing the simplification slider.

No as said in my last post I did some prototying myself too. :) I removed the simplification slider to have a better view on the polygons when the window size is small.

mdmm13:
Apologies, @JaffaKetchup for hijacking your PR for this discussion, believe it's super valuable though. Might be worth moving this to a discussion thread for when you close this PR :-)

@greensopinion - you wrote a while ago that aside from [threading and incrementally rendering less important layers like text over multiple frames](https://github.com/greensopinion/flutter-vector-map-tiles/issues/93#issuecomment-1268617838), that raster would be the main bottleneck and that [rendering raster with vector detail](https://github.com/greensopinion/flutter-vector-map-tiles/issues/120#issuecomment-1857195661) is the way to go. Is that still accurate based on your experiments with the latter?

@JaffaKetchup, @ignatz from your experiments with drawVertices, can you tell/guess whether this could be enough to get vector performance up drastically?

Alternatives like [flutter_gpu](https://github.com/flutter/flutter/wiki/Flutter-GPU) (saw some [great news](https://twitter.com/algebrandon/status/1735907945467294020) a few weeks back!) and Impeller may be quite some time into the future ... which may also be the answer. Just trying to figure out if drawVertices might solve it in the meantime.

ignatz:
Let's move vector tile related performance discussions to https://github.com/greensopinion/flutter-vector-map-tiles/issues/120.

Just quickly responding to your points...

> @greensopinion - you wrote a while ago that aside from https://github.com/greensopinion/flutter-vector-map-tiles/issues/93#issuecomment-1268617838, that raster would be the main bottleneck and that https://github.com/greensopinion/flutter-vector-map-tiles/issues/120#issuecomment-1857195661 is the way to go. Is that still accurate based on your experiments with the latter?

I know you didn't ask me but I just looked into it. Raster performance is certainly the bottleneck. Though there's quite a bit of optimization opportunities. Just with draw batching I more than halfed raster times while playing around with it in the morning (see the other thread). drawVertices, Impeller, and zoom-level based geometry simplifications would all help to bring them down further. 

> @JaffaKetchup, @ignatz, @josxha from your experiments with drawVertices, can you tell/guess whether this could be enough to get vector performance up drastically?

I'm pretty sure that with some investment the vector performance could be stellar. I don't see why it couldn't. Web clearly demonstrates that vector renders are viable.

> Alternatives like [flutter_gpu](https://github.com/flutter/flutter/wiki/Flutter-GPU) (saw some [great news](https://twitter.com/algebrandon/status/1735907945467294020) a few weeks back!) and Impeller may be quite some time into the future ... which may also be the answer. Just trying to figure out if drawVertices might solve it in the meantime.

I generally like the flutter GPU layering proposal, though it will be **years** before this will be somewhat available and portable. When working with the flutter folks, I was promised Impeller in 2019 and it launched in 2022 for iOS only. Besides if you just wanna draw into a texture using OpenGL there's also https://pub.dev/packages/three_dart (looks a bit abandoned)

JaffaKetchup:
See https://github.com/flutter/flutter/issues/131345#issuecomment-1925338105 for some performance comparisons.

josxha:
@JaffaKetchup I pushed my basic attempt with drawVertices here: https://github.com/josxha/flutter_map/tree/draw-vertices

Seeing the current discussions in https://github.com/flutter/flutter/issues/131345#issuecomment-1925338105 I'd be interessted to see some benchmarks here (don't have a device that supports impeller myself 😅).

As said I think it's really not compatible with the current `PolygonLayer` because mine is a completely bare bone implementation. It has basically no of the features the current PolygonLayer has. The idea was to see how much peformance is possible 
 by Skia and without other polygon optimizations like simplification or even the earcut algorithm (althrough the basic method I use to create vertices is probably cheaper).
It is focussed on these principles:
- use as few calls on the cancas as possible to avoid an expensive back and forth with Skia
- use as far and as efficient as possible typed data to avoid the creation of ephemeral objects
- avoid records because of their performance penalty
- break down the build logic to O(n).
- use simple caching

I hope this helps. (:

JaffaKetchup:
I've tested out your method @josxha. Whilst the triangulation algorithm doesn't really work quite right (orange is yours, dark is real), it is a whole lot faster. 

image

I think it shows what the Flutter engineer in the other thread was saying: the majority of the time comes from drawing the outlines.

JaffaKetchup:
I've just updated the example to allow for the thickness of the border line to be more carefully fine-tuned (None, 1px, 4px, 9px), to better demonstrate the performance falloff.

JaffaKetchup:
I just found quite a major bug in the `renderHashCode` that I introduced. I was attempting to use the `holePointsList` in it, which of course (a) isn't important, and (b) will never result in the same hash. This meant every polygon and border was being drawn individually, and batching was essentially broken. I've resolved the issue, and there's been a massive performance improvement.

lib/src/layer/polygon_layer/label.dart
```diff
@@ -85,7 +85,7 @@ LatLng _computePolylabel(List<LatLng> points) {
     // point with more distance to the polygon's outline. It's given in
     // point-units, i.e. degrees here. A bigger number means less precision,
     // i.e. cheaper at the expense off less optimal label placement.
-    precision: 0.000001,
+    precision: 0.0001,
```
josxha, line 88: Based on [this](https://wiki.openstreetmap.org/wiki/Precision_of_coordinates) 4 decimal places can alter up to about 11 meters on the equator. Are we sure that we want to have this change?

If we lower the percision we should at least give users the possibility to configure this value.

lib/src/layer/polygon_layer/painter.dart
```diff
@@ -91,11 +122,27 @@ class _PolygonPainter extends CustomPainter {
       // ignore: deprecated_member_use_from_same_package
       if (polygon.isFilled ?? true) {
         if (polygon.color != null) {
```
josxha, line 124: Not a change of this pull request but am I correct that these two if statements can be merged together?

lib/src/layer/polygon_layer/painter.dart
```diff
@@ -256,10 +302,13 @@ class _PolygonPainter extends CustomPainter {
...
   }
 
-  void _addLineToPath(ui.Path path, List<Offset> offsets) {
+  void _addLineToPath(Path path, List<Offset> offsets) {
     path.addPolygon(offsets, true);
   }
 
+  // TODO: Fix bug where wrapping layer in some widgets (eg. opacity) causes the
+  // features to not move unless this is `true`, but `true` significantly impacts
+  // performance
```
josxha, line 311: Is this only the case for drawVertices? 

If so we should 
- add this to the documentation of `performantRendering`
- reconsider if `performantRendering` should be true by default of rather false (aka use this but be aware of the potential consequences).

lib/src/layer/polygon_layer/polygon.dart
```diff
@@ -49,11 +49,26 @@ class Polygon {
...
 
+  /// {@macro fm.PolygonLayer.performantRendering}
+  ///
+  /// Value meanings (defaults to `true`):
+  ///
+  /// - `true` : respect layer-level value (disabled by default)
+  /// - `false`: disabled, ignore layer-level value
+  ///
+  /// Also see [PolygonLayer.performantRendering].
+  final bool performantRendering;
```
josxha, line 70: Do we actually have users that want to switch the rendering logic per polygon but have them together in one PolygonLayer? I'm just worried that this introduces unnecessary complexity without common or rare usage.

If so there should be a hint in the API docs to `PolygonLayer.performantRendering` (not just in the online documentation).

lib/src/layer/polygon_layer/polygon_layer.dart
```diff
@@ -126,10 +152,41 @@ class _PolygonLayerState extends State<PolygonLayer> {
...
+                  points.length * 2,
+                  (ii) => ii % 2 == 0
+                      ? points.elementAt(ii ~/ 2).x
+                      : points.elementAt(ii ~/ 2).y,
+                  growable: false,
+                ),
+                // Not sure how just this works but it seems to :D
+                holeIndices: culledPolygon.holePoints.isEmpty
+                    ? null
+                    : [culledPolygon.points.length],
```
josxha, line 179: Are we fine with having `holeIndices` like it is now?

lib/src/layer/polygon_layer/polygon_layer.dart
```diff
@@ -165,9 +222,9 @@ class _PolygonLayerState extends State<PolygonLayer> {
             tolerance: tolerance,
             highQuality: true,
           ),
-          holePoints: holes == null
-              ? null
-              : List<List<DoublePoint>>.generate(
+          holePoints: holes.isEmpty
+              ? []
+              : List.generate(
```
josxha, line 227: Probably redundant, List.generate(length: 0) shouldn't be a expensive operation.

lib/src/layer/polyline_layer/painter.dart
```diff
@@ -280,6 +288,9 @@ class _PolylinePainter<R extends Object> extends CustomPainter {
   LatLng _unproject(DoublePoint p0) =>
       camera.crs.projection.unprojectXY(p0.x, p0.y);
 
+  // TODO: Fix bug where wrapping layer in some widgets (eg. opacity) causes the
+  // features to not move unless this is `true`, but `true` significantly impacts
+  // performance
```
josxha, line 293: Same thoughts as for the PolygonLayer, but seeing this comment here i assume it's unrelated to drawVertices?

josxha (commented):
Added my thoughts as comments, let me know what you think.
Other than that I think it looks good to me.

example/lib/pages/polygon_perf_stress.dart
```diff
@@ -38,7 +38,7 @@ class _PolygonPerfStressPageState extends State<PolygonPerfStressPage> {
 
   @override
   void dispose() {
-    geoJsonLoader.ignore();
+    geoJsonParser.ignore();
```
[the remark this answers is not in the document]
JaffaKetchup: `ignore` is a method on all `Future`s from 'dart:async' that "signal[s] that the result of the future is no longer important to the program, not even if it's an error".

example/lib/pages/polygon_perf_stress.dart
```diff
@@ -83,10 +84,146 @@ class _PolygonPerfStressPageState extends State<PolygonPerfStressPage> {
...
+                                  ),
+                                ),
+                              ],
+                            ),
+                          ),
+                        ),
+                      ),
+                      // Not ideal that we have to re-parse the GeoJson every
+                      // time this is changed, but the library gives no easy
+                      // way to change it after
```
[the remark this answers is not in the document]
JaffaKetchup: Agreed. It could be theoretically worked around by just creating a whole new set of `Polygon`s using something like `copyWith`, but this is just easier, and provides a facility to dynamically update the GeoJson in future.
I might change it later when/if I add a `Polygon.copyWith` when the parameters have settled down.

example/lib/pages/polygon_perf_stress.dart
```diff
@@ -83,10 +84,146 @@ class _PolygonPerfStressPageState extends State<PolygonPerfStressPage> {
...
+                                    borderThickness == 1)
+                                  const Tooltip(
+                                    message:
+                                        'Screen has a high DPR: 1px border may be more than 1px.',
+                                    child: Icon(
+                                      Icons.warning,
+                                      color: Colors.amber,
+                                    ),
+                                  ),
+                                const SizedBox.shrink(),
```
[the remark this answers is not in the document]
JaffaKetchup: Yes, the surrounding `Span` adds two spacers on either side, providing a double space.

lib/src/layer/polygon_layer/label.dart
JaffaKetchup: With the current precision, it sometimes caused lag so bad that it literally crashed the debug engine and app.
We should ideally make it an option, but that's not for this PR. I'll add a TODO.

lib/src/layer/polygon_layer/painter.dart
JaffaKetchup: Yes, they can be. I left it like this to make it clearer/easier to remove once the deprecation is removed. I can change it (and elsewhere) if you prefer that.

lib/src/layer/polygon_layer/painter.dart
JaffaKetchup: This is unrelated to `drawVertices`, and is a bug I noticed/was reported whilst developing this. Also, "significant" was probably OTT, but there is still a small performance difference.

lib/src/layer/polygon_layer/polygon_layer.dart
JaffaKetchup: If it ain't broke, don't fix it :D
Tbh, I still don't really understand why it works, and I think it might be a 'feature' of dart_earcut that doesn't exist in the originals.

lib/src/layer/polyline_layer/painter.dart
JaffaKetchup: See above.

lib/src/layer/polygon_layer/polygon_layer.dart
JaffaKetchup: Resolved in 9475c7a.

lib/src/layer/polygon_layer/polygon.dart
JaffaKetchup: I've removed the feature level argument altogether. Resolved in 9475c7a.

example/lib/pages/polygon_perf_stress.dart
```diff
@@ -83,10 +84,146 @@ class _PolygonPerfStressPageState extends State<PolygonPerfStressPage> {
...
+                                    final thickness = pow(i, 2);
+                                    return ChoiceChip(
+                                      label: Text(
+                                        thickness == 0
+                                            ? 'None'
+                                            : '${thickness}px',
+                                      ),
+                                      selected: borderThickness == thickness,
+                                      shape: const StadiumBorder(),
+                                      onSelected: (selected) async {
```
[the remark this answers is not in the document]
JaffaKetchup: I don't want to spend too much time splitting it apart now, but I've made some changes to condense it, and extracted the method that updates the GeoJson. It's a bit more acceptable. Resolved in 9475c7a.

example/lib/pages/polygon_perf_stress.dart
```diff
@@ -83,10 +84,146 @@ class _PolygonPerfStressPageState extends State<PolygonPerfStressPage> {
...
+                          decoration: BoxDecoration(
+                            color: Theme.of(context).colorScheme.background,
+                            borderRadius: BorderRadius.circular(32),
+                          ),
+                          child: Padding(
+                            padding: const EdgeInsets.only(
+                              left: 16,
+                              right: 16,
+                              top: 4,
+                              bottom: 4,
```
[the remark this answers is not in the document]
JaffaKetchup: Resolved in 9475c7a.

example/lib/pages/polygon_perf_stress.dart
```diff
@@ -83,10 +84,146 @@ class _PolygonPerfStressPageState extends State<PolygonPerfStressPage> {
...
+                            color: Theme.of(context).colorScheme.background,
+                            borderRadius: BorderRadius.circular(32),
+                          ),
+                          child: Padding(
+                            padding: const EdgeInsets.only(
+                              left: 16,
+                              right: 16,
+                              top: 8,
+                              bottom: 8,
+                            ),
```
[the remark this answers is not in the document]
JaffaKetchup: Resolved in 9475c7a.

example/lib/pages/polygon_perf_stress.dart
```diff
@@ -83,10 +84,146 @@ class _PolygonPerfStressPageState extends State<PolygonPerfStressPage> {
...
-              onChangedTolerance: (v) =>
-                  setState(() => simplificationTolerance = v),
+            child: RepaintBoundary(
+              child: Column(
+                children: [
+                  SimplificationToleranceSlider(
+                    initialTolerance: _initialSimplificationTolerance,
+                    onChangedTolerance: (v) =>
+                        setState(() => simplificationTolerance = v),
+                  ),
```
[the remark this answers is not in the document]
JaffaKetchup: Quite right, I should've done that :D. Resolved in 9475c7a, alongside `NumOfItemsSlider`.

example/lib/pages/polygon_perf_stress.dart
```diff
@@ -21,14 +23,12 @@ class PolygonPerfStressPage extends StatefulWidget {
 class _PolygonPerfStressPageState extends State<PolygonPerfStressPage> {
   static const double _initialSimplificationTolerance = 0.5;
   double simplificationTolerance = _initialSimplificationTolerance;
+  static const bool _initialUsePerformantDrawing = true;
+  bool usePerformantRendering = _initialUsePerformantDrawing;
+  static const double _initialBorderThickness = 1;
+  double borderThickness = _initialBorderThickness;
```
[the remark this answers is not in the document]
JaffaKetchup: In conjunction with the changes bought about by the response to the review below, it all makes a lot more sense now. Thanks for pointing it out, quite bad style on my part! Resolved in 9475c7a.

JaffaKetchup:
I've also changed the name from `performantRendering` to `useAltRendering`, and updated the documentation, to remove the quick assumption that it always results in better performance.

mohammedX6:
image

Using 
```
filledPath.reset();
borderPath.reset();
```

is better than creating new Path each time, please see this : https://github.com/flutter/flutter/issues/83872

lib/src/layer/polygon_layer/projected_polygon.dart
```diff
@@ -25,7 +25,11 @@ class _ProjectedPolygon {
           ),
           holePoints: () {
             final holes = polygon.holePointsList;
-            if (holes == null) return null;
+            if (holes == null ||
+                holes.isEmpty ||
+                holes.every((e) => e.isEmpty)) {
+              return <List<DoublePoint>>[];
+            }
```
josxha, line 32: `holes.every()` is an expensive check.
It could be put into the list generation beneath.
Something like `holes.where((e) => e.isNotEmpty).map((e) => ...).toList(growable: false)` should do the same but more efficient.

josxha (approved):
Just one expensive check, other than that we should be good to go.

lib/src/layer/polygon_layer/projected_polygon.dart
JaffaKetchup: I'm not too fussed about this. It should only ever be expensive when the user specifies a list like so `[[], [], []]`, which there is no real reason to do. I've also put it last in the list to take advantage of the lazy logic gates.

josxha: I see your point. In that case the performance penalty is not as much as I've thought.

Issue #1793, opened 2024-01-11
Different CRS and their impact on how you work with maps


JosefWN:
### What do you want implemented?

# Background

I made this ticket to try to explain, to the best of my abilities, how I use `flutter_map` and some of the challenges it has posed. I probably should have made it sooner but I knew it would be a lengthy explanation and I've been a bit busy the past year.

To me, choosing which CRS to use is similar to choosing which programming language to use; what's downright impossible in one CRS could be a breeze in another. If you switch from C to Dart, your first thought likely wouldn't be to look into disabling the garbage collector, but rather how to leverage garbage collection and other language features to the greatest extent. Similarly, in leveraging the strengths and avoiding the pitfalls of the CRS you have chosen, you will work with your coordinates differently than you would if you had chosen another CRS.

Since there aren't that many families of projections to choose from however, you don't necessarily have the luxury of choice. So when you hear people complaining about seemingly exotic CRS, it's not likely that they _wanted_ to work with them, but simply that they _had_ to, which I will explain further below.

Spurred by: https://github.com/fleaflet/flutter_map/pull/1750
Relates to: https://github.com/fleaflet/flutter_map/issues/1386, https://github.com/fleaflet/flutter_map/pull/1295, https://github.com/fleaflet/flutter_map/issues/191 / https://github.com/fleaflet/flutter_map/issues/211

# Technical background

To be able to make sense of latitudes and longitudes they need to belong to some geometry resembling that of the Earth's. This geometry, called a geoid, needs a spherical coordinate system defining the origin of latitude/longitude and how the axes are oriented. The geoid, along with the definition of the coordinate system, is collectively called a datum.

As the Earth formed it was hotter and softer than it is today. Rotating around its own axis, the planets mass was pulled toward the equator, flattening the regions around the poles, like a spinning pizza dough. Part of WGS84 is a datum that defines the coordinate system of the Earth on an "oblate spheroid", a flattened sphere more generally referred to as an ellipsoid. The WGS84 datum is what GNSS/GPS systems and most projections use to understand how they should relate to the latitudes and longitudes.

This alignment on the WGS84 _datum_ is convenient, because it ensures that when you give me a latitude and a longitude, it means exactly the same thing to you as it does to me, regardless of CRS. You could use a spherical datum/approximation to speed up calculations because of the symmetry (at the cost of precision), but the projection of the source data is pretty much always derived from the WGS84 datum.

[...]

PR #1801, opened 2024-01-15
perf!: cache projection of polygon points & CRS improvements

ignatz:
RFC based on #1793

### DIFF
--- a/lib/src/geo/crs.dart
+++ b/lib/src/geo/crs.dart
@@ -67,13 +67,6 @@ abstract class CrsWithStaticTransformation extends Crs {
   @override
   final Projection projection;
 
-  @override
-  (double, double) transform(double x, double y, double scale) =>
-      _transformation.transform(x, y, scale);
-  @override
-  (double, double) untransform(double x, double y, double scale) =>
-      _transformation.untransform(x, y, scale);
-
   const CrsWithStaticTransformation._({
     required _Transformation transformation,
     required this.projection,
@@ -82,6 +75,13 @@ abstract class CrsWithStaticTransformation extends Crs {
     super.wrapLng,
     super.wrapLat,
   }) : _transformation = transformation;
+
+  @override
+  (double, double) transform(double x, double y, double scale) =>
+      _transformation.transform(x, y, scale);
+  @override
+  (double, double) untransform(double x, double y, double scale) =>
+      _transformation.untransform(x, y, scale);
 
   @override
   (double, double) latLngToXY(LatLng latlng, double scale) {
--- a/lib/src/geo/crs.dart
+++ b/lib/src/geo/crs.dart
@@ -29,10 +29,10 @@ abstract class Crs {
     this.wrapLat,
   });
 
-  // Project a spherical LatLng coordinate into planar space (unscaled).
+  /// Project a spherical LatLng coordinate into planar space (unscaled).
   Projection get projection;
 
-  // Scale planar coordinate to scaled map point.
+  /// Scale planar coordinate to scaled map point.
   (double, double) transform(double x, double y, double scale);
   (double, double) untransform(double x, double y, double scale);
 
--- a/lib/src/geo/crs.dart
+++ b/lib/src/geo/crs.dart
@@ -58,7 +58,9 @@ abstract class Crs {
   Bounds<double>? getProjectedBounds(double zoom);
 }
 
+/// Internal base class for CRS with a single zoom-level independent transformation.
 @immutable
+@internal
 abstract class CrsWithStaticTransformation extends Crs {
   @nonVirtual
   @protected
--- a/lib/src/geo/crs.dart
+++ b/lib/src/geo/crs.dart
@@ -29,10 +29,15 @@ abstract class Crs {
     this.wrapLat,
   });
 
+  // Project a spherical LatLng coordinate into planar space (unscaled).
   Projection get projection;
 
-  /// Converts a point on the sphere surface (with a certain zoom) in a
-  /// map point.
+  // Scale planar coordinate to scaled map point.
+  (double, double) transform(double x, double y, double scale);
+  (double, double) untransform(double x, double y, double scale);
+
+  /// Converts a point on the sphere surface (with a certain zoom) to a
+  /// scaled map point.
   (double, double) latLngToXY(LatLng latlng, double scale);
 
   Point<double> latLngToPoint(LatLng latlng, double zoom) {
@@ -54,32 +59,39 @@ abstract class Crs {
 }
 
 @immutable
-abstract class _CrsWithStaticTransformation extends Crs {
+abstract class CrsWithStaticTransformation extends Crs {
   @nonVirtual
   @protected
-  final _Transformation transformation;
+  final _Transformation _transformation;
 
   @override
   final Projection projection;
 
-  const _CrsWithStaticTransformation({
-    required this.transformation,
+  @override
+  (double, double) transform(double x, double y, double scale) =>
+      _transformation.transform(x, y, scale);
+  @override
+  (double, double) untransform(double x, double y, double scale) =>
+      _transformation.untransform(x, y, scale);
+
+  const CrsWithStaticTransformation._({
+    required _Transformation transformation,
     required this.projection,
     required super.code,
     required super.infinite,
     super.wrapLng,
     super.wrapLat,
-  });
+  }) : _transformation = transformation;
 
   @override
   (double, double) latLngToXY(LatLng latlng, double scale) {
     final (x, y) = projection.projectXY(latlng);
-    return transformation.transform(x, y, scale);
+    return _transformation.transform(x, y, scale);
   }
 
   @override
   LatLng pointToLatLng(Point point, double zoom) {
-    final (x, y) = transformation.untransform(
+    final (x, y) = _transformation.untransform(
       point.x.toDouble(),
       point.y.toDouble(),
       scale(zoom),
@@ -93,8 +105,8 @@ abstract class _CrsWithStaticTransformation extends Crs {
 
     final b = projection.bounds!;
     final s = scale(zoom);
-    final (minx, miny) = transformation.transform(b.min.x, b.min.y, s);
-    final (maxx, maxy) = transformation.transform(b.max.x, b.max.y, s);
+    final (minx, miny) = _transformation.transform(b.min.x, b.min.y, s);
+    final (maxx, maxy) = _transformation.transform(b.max.x, b.max.y, s);
     return Bounds<double>(
       Point<double>(minx, miny),
       Point<double>(maxx, maxy),
@@ -104,9 +116,9 @@ abstract class _CrsWithStaticTransformation extends Crs {
 
 // Custom CRS for non geographical maps
 @immutable
-class CrsSimple extends _CrsWithStaticTransformation {
+class CrsSimple extends CrsWithStaticTransformation {
   const CrsSimple()
-      : super(
+      : super._(
           code: 'CRS.SIMPLE',
           transformation: const _Transformation(1, 0, -1, 0),
           projection: const _LonLat(),
@@ -118,11 +130,11 @@ class CrsSimple extends _CrsWithStaticTransformation {
 
 /// The most common CRS used for rendering maps.
 @immutable
-class Epsg3857 extends _CrsWithStaticTransformation {
+class Epsg3857 extends CrsWithStaticTransformation {
   static const double _scale = 0.5 / (math.pi * SphericalMercator.r);
 
   const Epsg3857()
-      : super(
+      : super._(
           code: 'EPSG:3857',
           transformation: const _Transformation(_scale, 0.5, -_scale, 0.5),
           projection: const SphericalMercator(),
@@ -132,12 +144,15 @@ class Epsg3857 extends _CrsWithStaticTransformation {
 
   @override
   (double, double) latLngToXY(LatLng latlng, double scale) =>
-      transformation.transform(SphericalMercator.projectLng(latlng.longitude),
-          SphericalMercator.projectLat(latlng.latitude), scale);
+      _transformation.transform(
+        SphericalMercator.projectLng(latlng.longitude),
+        SphericalMercator.projectLat(latlng.latitude),
+        scale,
+      );
 
   @override
   Point<double> latLngToPoint(LatLng latlng, double zoom) {
-    final (x, y) = transformation.transform(
+    final (x, y) = _transformation.transform(
       SphericalMercator.projectLng(latlng.longitude),
       SphericalMercator.projectLat(latlng.latitude),
       scale(zoom),
@@ -148,9 +163,9 @@ class Epsg3857 extends _CrsWithStaticTransformation {
 
 /// A common CRS among GIS enthusiasts. Uses simple Equirectangular projection.
 @immutable
-class Epsg4326 extends _CrsWithStaticTransformation {
+class Epsg4326 extends CrsWithStaticTransformation {
   const Epsg4326()
-      : super(
+      : super._(
           projection: const _LonLat(),
           transformation: const _Transformation(1 / 180, 1, -1 / 180, 0.5),
           code: 'EPSG:4326',
@@ -221,6 +236,13 @@ class Proj4Crs extends Crs {
       scales: finalScales,
     );
   }
+
+  @override
+  (double, double) transform(double x, double y, double scale) =>
+      _getTransformationByZoom(zoom(scale)).transform(x, y, scale);
+  @override
+  (double, double) untransform(double x, double y, double scale) =>
+      _getTransformationByZoom(zoom(scale)).untransform(x, y, scale);
 
   /// Converts a point on the sphere surface (with a certain zoom) in a
   /// map point.
--- a/lib/src/layer/polygon_layer/polygon_layer.dart
+++ b/lib/src/layer/polygon_layer/polygon_layer.dart
@@ -5,6 +5,7 @@ import 'package:collection/collection.dart';
 import 'package:flutter/foundation.dart';
 import 'package:flutter/widgets.dart';
 import 'package:flutter_map/flutter_map.dart';
+import 'package:flutter_map/src/misc/offsets.dart';
 import 'package:flutter_map/src/misc/simplify.dart';
 import 'package:latlong2/latlong.dart' hide Path;
 import 'package:polylabel/polylabel.dart'; // conflict with Path from UI
@@ -116,15 +117,20 @@ class _PolygonLayerState extends State<PolygonLayer> {
   }
 
   List<Polygon> _computeZoomLevelSimplification(int zoom) =>
-      _cachedSimplifiedPolygons[zoom] ??= widget.polygons
-          .map(
-            (polygon) => polygon.copyWithNewPoints(
-              simplify(
-                polygon.points,
-                widget.simplificationTolerance / math.pow(2, zoom),
-                highestQuality: true,
-              ),
+      _cachedSimplifiedPolygons[zoom] ??= List<Polygon>.generate(
+        widget.polygons.length,
+        (i) {
+          final polygon = widget.polygons[i];
+          return polygon.copyWithNewPoints(
+            // TODO: Ideally we'd simplify in projected space to minimize issues with map distortion.
+            // TODO: Simplify polygon holes as well.
+            simplify(
+              polygon.points,
+              widget.simplificationTolerance / math.pow(2, zoom),
+              highestQuality: true,
             ),
-          )
-          .toList();
+          );
+        },
+        growable: false,
+      );
 }
--- a/lib/src/layer/polygon_layer/polygon_layer.dart
+++ b/lib/src/layer/polygon_layer/polygon_layer.dart
@@ -62,50 +62,50 @@ class PolygonLayer extends StatefulWidget {
 }
 
 class _PolygonLayerState extends State<PolygonLayer> {
-  final _cachedSimplifiedPolygons = <int, List<Polygon>>{};
+  List<_ProjectedPolygon>? _polygons;
+  final _cachedSimplifiedPolygons = <int, List<_ProjectedPolygon>>{};
 
   @override
   void didUpdateWidget(PolygonLayer oldWidget) {
     super.didUpdateWidget(oldWidget);
 
-    // IF old yes & new no, clear
-    // IF old no & new yes, compute
-    // IF old no & new no, nothing
-    // IF old yes & new yes & (different tolerance | different lines), both
-    //    otherwise, nothing
-    if (oldWidget.simplificationTolerance != 0 &&
-        widget.simplificationTolerance != 0 &&
-        (!listEquals(oldWidget.polygons, widget.polygons) ||
-            oldWidget.simplificationTolerance !=
-                widget.simplificationTolerance)) {
-      _cachedSimplifiedPolygons.clear();
-      _computeZoomLevelSimplification(MapCamera.of(context).zoom.floor());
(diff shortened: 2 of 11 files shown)


JosefWN:
Cool! Will try out the PR with my CRS for a better feel soon, just adding some initial thoughts.

lib/src/geo/crs.dart
```diff
@@ -29,10 +29,15 @@ abstract class Crs {
     this.wrapLat,
   });
 
+  // Project a spherical LatLng coordinate into planar space (unscaled).
   Projection get projection;
 
-  /// Converts a point on the sphere surface (with a certain zoom) in a
-  /// map point.
+  // Scale planar coordinate to scaled map point.
+  (double, double) transform(double x, double y, double scale);
```
JosefWN, line 36: Perhaps this could take and return a `DoublePoint(x, y)`? It would make it more convenient to use, although slightly less versatile.

lib/src/misc/simplify.dart
```diff
@@ -3,53 +3,66 @@
...
-  LatLng p1,
-  LatLng p2,
-) {
-  final double dx = p1.longitude - p2.longitude;
-  final double dy = p1.latitude - p2.latitude;
-  return dx * dx + dy * dy;
+// Custom point due to math.Point<double> being slow. Math operations tend to
+// have 20+x penalty for virtual function overhead given the reified nature of
+// Dart generics.
+class DoublePoint {
```
JosefWN, line 9: Good catch, that's a heavy penalty! Context: https://github.com/dart-lang/sdk/issues/53912

We could also:
```dart
class Point extends proj4.Point {
  Point({required super.x, required super.y});

  Point operator +(Point other) => Point(x: x + other.x, y: y + other.y);
}
```
If we want to be a bit more succinct (although it would conflict with `math.Point`, so the `math` imports would need to be named, a matter of taste).

Aligning our point type with `proj4.Point` could perhaps be convenient, but not strictly necessary (it also uses named arguments, for good or for bad).

Another option is to call it `PointDouble`, might be easier to catch with auto-completion if someone is looking for a point.

JaffaKetchup: Definitely in favour of `PointDouble`, it matches closer to what is expected with using generics.

test/layer/tile_layer/tile_bounds/crs_fakes.dart
```diff
@@ -14,14 +14,16 @@ class FakeInfiniteCrs extends Crs {
...
-  Point<double> latLngToPoint(LatLng latlng, double zoom) {
-    return const Epsg3857().latLngToPoint(latlng, zoom);
-  }
+  (double, double) latLngToXY(LatLng latlng, double scale) =>
+      const Epsg3857().latLngToXY(latlng, scale);
 
   @override
-  (double, double) latLngToXY(LatLng latlng, double scale) {
-    return const Epsg3857().latLngToXY(latlng, scale);
-  }
```
JaffaKetchup, line 24: Why has this been removed?

JaffaKetchup (changes requested):
Wow, some great changes, thanks :) Few minor changes, and I'll make a couple basic formatting changes.
After the research into `Record`s which revealed they were innefficient, should we be using them? I believe you were there or acknowledged it, but I'll try to dig it out if not.

lib/src/geo/crs.dart
```diff
@@ -53,32 +58,39 @@ abstract class Crs {
 }
 
 @immutable
-abstract class _CrsWithStaticTransformation extends Crs {
+abstract class CrsWithStaticTransformation extends Crs {
```
JosefWN, line 61: Maybe we should comment what a CRS with a static transform is (or more interestingly, exemplify when the transform is non-static).

I don't know off-hand when that occurs. One thing I could possibly think of is compensating for tectonic shift? Say you have a statue in your garden pinned with a lat/lon, and 10 years down the line, your coordinate is still pointing to the same place on Earth, but the tectonic plate has moved, effectively changing the location of your statue.

I think _dynamic_ CRS like EPSG:3857 and EPSG:4326 need a _coordinate epoch_, a timestamp associated with every coordinate, to be able to compensate for the tectonic shift that has occurred. The movement is calculated relative to a _frame reference epoch_, where tectonic plate locations/velocities are known. The movements are small (cm-level per year) and only relevant for fairly precise measurements, or old measurements relative to the frame reference.

Some local CRS are referred to as _static_ because they are _plate-fixed_, i.e. move with the tectonic plates, in which case you don't need to compensate.

It doesn't seem this is supported by `Proj4Dart` though.

lib/src/layer/polygon_layer/polygon_layer.dart
```diff
@@ -64,50 +66,57 @@ class PolygonLayer extends StatefulWidget {
...
-      _computeZoomLevelSimplification(MapCamera.of(context).zoom.floor());
+    if (widget.simplificationTolerance != 0 &&
+        oldWidget.simplificationTolerance == widget.simplificationTolerance &&
+        listEquals(oldWidget.polygons, widget.polygons)) {
+      // Reuse cache.
+      return;
     }
+
+    _cachedSimplifiedPolygons.clear();
+    _cachedProjectedPolygons = null;
```
JaffaKetchup, line 84: This logic has been broken (changing the tolerance no longer 'refreshes' the layer correctly/at all).

lib/src/layer/polygon_layer/polygon_layer.dart
```diff
@@ -118,16 +127,37 @@ class _PolygonLayerState extends State<PolygonLayer> {
...
+    double tolerance,
+    int zoom,
+  ) =>
+      List<_ProjectedPolygon>.generate(
+        polygons.length,
+        (i) {
+          final polygon = polygons[i];
+          final holes = polygon.holePoints;
+
+          return _ProjectedPolygon._(
```
JaffaKetchup, line 141: Does this not mean that (if simplification is enabled), every polygon is converted to a `_ProjectedPolygon` twice (once prior to simplification then again in simplification), and all the expensive loops that means?
Wouldn't it be better to perform simplification at the same time as the first conversion, so two sets of loops don't need to be done?

JaffaKetchup (changes requested):

lib/src/geo/crs.dart
JosefWN: Also, an easy way to go from Proj4-string to a custom `CrsWithStaticTransformation` would be very nice. My CRS is not known at compile time:
```dart
Proj4Crs.fromFactory(
  code: 'EPSG:xyz',
  proj4Projection: proj4.Projection.add(
      'EPSG:xyz',
      proj4Str,
    ),
);
```

JaffaKetchup:
This PR appears to be split into two parts, CRS improvements and Polygon improvements. Should they be seperated into seperate PRs? (no worries if that's too time consuming.)

lib/src/misc/simplify.dart
JosefWN: Also, if the PR is accepted, pre-merge, will we go over the use of `math.Point` in the repo and replace with this point implementation? Keeping two separate point implementations intended to represent the same thing seems like it would fragment the API a bit.

test/layer/tile_layer/tile_bounds/crs_fakes.dart
ignatz: It was never needed (there's a default impl) and it's in a test so I removed it while adding impls for the new interfaces.

lib/src/geo/crs.dart
ignatz: I will close this simply because you opened another discussion on PointDouble and I'd like to consolidate.

lib/src/misc/simplify.dart
ignatz: > Aligning our point type with proj4.Point could perhaps be convenient, but not strictly necessary (it also uses named arguments, for good or for bad).

I'd prefer not to. If PointDouble remains internal it doesn't matter and if it becomes public we hand over ownership of a key API to another project.

> Another option is to call it PointDouble, might be easier to catch with auto-completion if someone is looking for a point.

Happy to rename don't feel very strong about names.

> Also, if the PR is accepted, pre-merge, will we go over the use of math.Point in the repo and replace with this point implementation? Keeping two separate point implementations intended to represent the same thing seems like it would fragment the API a bit.

Note that at the moment PointDouble is pretty much internal. If we were to make it public, I agree that we shouldn't have N point flavors in our APIs.

ignatz:
> After the research into Records which revealed they were innefficient, should we be using them? I believe you were there or acknowledged it, but I'll try to dig it out if not.

Yeah, pretty much. I was torn. I started with a Record (especially since some of the CRS methods already use records) but at the end:
 * they have extra memory overhead (that was the interesting find)
 * they are not as convenient in terms of extensible API (even with extension methods).
 * and records are immutable and mutability is a fence we might wanna jump some day.

Though the last points is something we should certainly discuss more in terms of safety if we decide to make PointDouble public. At the moment I'm only really using it internally, i.e. we can change it w/o a major version change.

> This PR appears to be split into two parts, CRS improvements and Polygon improvements. Should they be seperated into seperate PRs? (no worries if that's too time consuming.)

Once could, though I'd argue that CRS changes are very minor (it's mostly just adding (un)transform). If we'd split them we'd add new APIs w/o a user.

lib/src/layer/polygon_layer/polygon_layer.dart
ignatz: changing the tolerance will clear the cache. Am I missing something?

JaffaKetchup: Try dragging the slider in the example app. It doesn't change the tolerance after the slider is released. That's why I called the calculation methods here before - although there may be a better fix.

lib/src/layer/polygon_layer/polygon_layer.dart
JaffaKetchup: > That said, I do believe it's more correct to simplify projected polygons than it is to simplify polygons. Calculating all the simplifications and distances on a spherical coordinates pre-projection is kind of a recipe for projection weirdness. Are two points that are close on the surface of a sphere still close after projecting them onto the screen? (Remember, I initially had some concerns with the simplifications around the poles, simplifying projected points solves that issue... and yes we should change polylines as well)

Yep, 100% agreed with this. Need to do the same with triangulation as well.

However, we do one loop and one double-nested loop per conversion. We convert twice, because we convert to projected, then convert projected to projected via simplification. Would it not be better if we simplify within the loops of the initial projection, to save one conversion cycle?

lib/src/geo/crs.dart
ignatz: Don't forget to compensate for relativistic contractions and elongations based on the reference frame of the observer :)

Jokes aside, I feel that you're reading too much into the name. I don't think there was ever an intend to support time evolution.
To counter your point, the only non-static-transform CRS is the Proj4Crs. Maybe I chose the name (not sure) to contrast the fact that Proj4Crs supports an array of transformations depending on the zoom level. Arguably "Static" might not be the best name. Either way I'm only changing visibility here.

> Also, an easy way to go from Proj4-string to a custom CrsWithStaticTransformation would be very nice

I don't see why you wouldn't be able to implement your own Proj4 Crs or extend the existing one (though the existing one isn't a CrsWithStaticTransform but I also don't see why that would matter).

JosefWN: > Either way I'm only changing visibility here.

Yes, in exposing it I figured it could be more clearly documented than something strictly internal (although for internal use it could also help, since the name is a bit confusing).

ignatz: Fair. What name would you suggest?

(I'm also open to other suggestions. I'm only exposing it to monomorphize some code. We could also drop this optimization or maybe go some other path. Open to suggestions)

lib/src/layer/polygon_layer/polygon_layer.dart
ignatz: I see. Btw I hadn't seen the sress test before - cool stuff.

The logic works correctly, however the units changed and therefore absolute distances. In other words [0; 2] is now too tight of a tollerance. I'd suggest a more reasonable new default tolerance of ~150,000

lib/src/layer/polygon_layer/polygon_layer.dart
ignatz: > However, we do one loop and one double-nested loop per conversion. We convert twice, because we convert to projected, then convert projected to projected via simplification. Would it not be better if we simplify within the loops of the initial projection, to save one conversion cycle?

There are many simplifications, one per zoom level. If you simplify for one zoom level during projection, you have to reproject again for another zoom level.

lib/src/layer/polygon_layer/polygon_layer.dart
JaffaKetchup: I see. Why has the unit changed, and what is it now? Is there any way we can change it back? We were in-keeping with other libraries units and recommended ranges.

lib/src/geo/crs.dart
JosefWN: > Fair. What name would you suggest?

I don't know tbh, still haven't fully understood the difference.

> (I'm also open to other suggestions. I'm only exposing it to monomorphize some code. We could also drop this optimization or maybe go some other path. Open to suggestions)

I think a broader approach extending beyond a single layer could be beneficial for the handling of points, and that might warrant some careful consideration, as opposed to adding a new point type for a specific case as part of addressing something else.

> Don't forget to compensate for relativistic contractions and elongations based on the reference frame of the observer :)

It's a increasing in popularity, especially for regions like Australia that move relatively fast 😄

lib/src/layer/polygon_layer/polygon_layer.dart
ignatz: we went for X a.k.a longitude from (-180; 180) to whatever the projection spits out (similar for Y a.k.a latitude).

... as I type it out I do acknowledge that at this point a good default would be CRS dependent. So it might be best to derive an effective tollerance instead based on the projection. Should be easy.

lib/src/geo/crs.dart
ignatz: > It's a increasing in popularity, especially for regions like Australia that move relatively fast 😄

Them Aussies, maybe we should flip their latitudes :).

> I think a broader approach extending beyond a single layer could be beneficial for the handling of points, and that might warrant some careful consideration, as opposed to adding a new point type for a specific case as part of addressing something else.

Are we mixing conversations? Where do you see the connection between the:
 * CRS base class name
 * and/or a singular fixed transformation for e.g. Epsg3857
 * and internal point type
?

lib/src/layer/polygon_layer/polygon_layer.dart
JaffaKetchup: I'm assuming you say that because it depends on how close two points are once projected now (and projection can 'move' the points now)? Is that a decent understanding?

If so, I'm good with that, it's not like the layers have a `const` constructor that would be ruined. However, we need to find a succinct, relatively non-technical way to explain the unit to the user. It was previously "decimal degrees scaled by floored zoom" I think, which still isn't perfect, but at least you can "feel" what the numbers do as it's a small scale: it would be good to keep this small scale.

ignatz:
Thanks for the reformatting. Quick remark, if you guys feel strongly about style i'd suggest installing or writing linter rules for it. Overall that will reduce churn, enforce the preferences more consistently across the code-base, and tell authors what you're looking for.

lib/src/geo/crs.dart
ignatz: What I think I inferred is that this conversation has fully moved away from "CrsWithStaticTransformation" and is now fully about the point type?

To additionally throw some assumptions in there:  "you care about the internal types because you're already planning to open up these internal layers in your fork"?

Otherwise...

> So cache them to your heart's content, but that isn't necessarily related to introducing a new point type to monomorphize the code, or how are the two inseparable?

All of this is separable, that's why it felt like this conversation is slipping between unrelated topics.

> What I'm trying to say is, it's a bit of a local hack, even if it's an internal point. It could be worth keeping math.Point until we have a better/broader option. That doesn't imply there will be a better option, but it keeps things a bit more streamlined.

But we're already mixing, e.g. we're using record types in some places. These are internal APIs, we should do what's best for our users (keeping our code in a maintainable state is clearly also good for the users).

> If math.Point is a bottleneck it should probably be seen as such in a broader context than just a single layer.

I don't understand the "if". Yes, math.Point is a bottleneck. I'd love to get rid of it, however changing public APIs and thus forcing every user to migrate is a very different story than optimizing our code paths internally.

lib/src/layer/polygon_layer/polygon_layer.dart
JaffaKetchup: Yeah I don't feel that unit either, but you could get a feel for the scale :) 0.5 was like x2 simplification, 1 was x4, 2 was x8 IMO.

Yes, I think pixel distances would be ideal. That would 'remove' the need for scaling the tolerance value based on zoom level, as that would happen inherently. I suppose thats also now easier that the polylines/gons are in pixel space already now.
The best approach would also be to take into account the thickness of the line (a really thin line will 'show' more simplification than a wide one), but that might be too complex to implement and difficult to calibrate, so ignore that.

JaffaKetchup:
> Quick remark, if you guys feel strongly about style i'd suggest installing or writing linter rules for it.

Yeah we've been slowly adding them (although not for a while). One I feel strongly about is trailing commas, but I know that's debated :D

lib/src/layer/polygon_layer/polygon_layer.dart
ignatz: > I suppose thats also now easier that the polylines/gons are in pixel space already now.

Yes. We're almost in pixel space. If you'd replace the "project" with "project+transform" in the effective tolerance calculation that's what you'd get.

> The best approach would also be to take into account the thickness of the line (a really thin line will 'show' more simplification than a wide one), but that might be too complex to implement and difficult to calibrate, so ignore that.

Should be fairly easy to do. But maybe in a follow-up? For now the tolerances should behave fairly similar to what they were before.

There's other things one could do in pixel space, e.g. if the polygon on screen is only 1x1 pixel in size, why even simplify just cull the thing, etc.

JaffaKetchup: > if the polygon on screen is only 1x1 pixel in size

That should be rare enough that it's not an issue :D

> Should be fairly easy to do. But maybe in a follow-up? For now the tolerances should behave fairly similar to what they were before.

Yeah, don't worry about this extension for now. But besides that, using pixel tolerances for simplification should be done here, as that seems like a very good idea and you say we're almost there?

ignatz: > That should be rare enough that it's not an issue :D

In many ways simplification already helps with that case. Whether or not it's rare :shrug:

Screenshot from 2024-01-17 13-32-17

> Yeah, don't worry about this extension for now. But besides that, using pixel tolerances for simplification should be done here, as that seems like a very good idea and you say we're almost there?

Done.

JaffaKetchup: Great, thanks. I'll have a play around in a few hours and choose a new sensible default (assuming it needs changing now that the units have changed).

JaffaKetchup:
Ok, so I think this just needs the conversation with @JosefWN resolved, and the same changes as applied to the polygons applied to the polylines as well (as I believe some of the changes have affected polyline simplification, but the other changes have not been made).

ignatz:
> Ok, so I think this just needs the conversation with @JosefWN resolved,

Feel free to weigh in, always welcome a tie breaker. As far as I can tell, there are two conversations interleaved:

 * As we're making `_CrsWithStaticTransformation` package visible, should we give it a better name.
 * Should we use math.Point instead of PointDouble? I'll ignore the conversation around whether we should use PointDouble in the public APIs because that's a much bigger and breaking change.

Personally, I'm happy to do any rename, however feel strongly about not using math.Point especially in internal APIs since it goes against the performance optimizations we're trying to do here :shrug: 

> as I believe some of the changes have affected polyline simplification, but the other changes have not been made

This change should not affect polyline. We should be able to make these changes independently.

JaffaKetchup:
> As we're making _CrsWithStaticTransformation package visible, should we give it a better name.

I'll opt out of discussing this one, I'm not even remotely qualified :D

> Should we use math.Point instead of PointDouble? I'll ignore the conversation around whether we should use PointDouble in the public APIs because that's a much bigger and breaking change.

I think you're right, given that the entire point of these changes is performance, we might as well squeeze as much out of this as possible. It's not ideal, and could be confusing, but it's documented well enough.

> This change should not affect polyline. We should be able to make these changes independently.

I'm not sure whether it affects polylines directly - the `simplify` methods have been changed to accept a pixel tolerance. In any case, the same changes need to be made to the polylines for consistency to reduce confusion between the two.

ignatz:
> I'm not sure whether it affects polylines directly - the simplify methods have been changed to accept a pixel tolerance. In any case, the same changes need to be made to the polylines for consistency to reduce confusion between the two.

Nothing should have changed for polylines. Simplify is unit agnostic. At the moment:
 * Polygons feed projected points and and untransformed tolerance
 * polygons feed latlng points and latlng tolerance.

I was hoping to do polylines in a follow up. WDYT?

JaffaKetchup:
Ah OK, my misunderstanding, in that case I'm happy for it to be delayed out of this PR, but it does need to be done before the v7 release.

JosefWN:
> As we're making _CrsWithStaticTransformation package visible, should we give it a better name.

Perhaps at least a doc comment stating the intended purpose of the class?

But I will leave this PR in the hands of the maintainers, there isn't that much more I can add.

JaffaKetchup:
> Perhaps at least a doc comment stating the intended purpose of the class?

That does seem reasonable.

ignatz:
> Perhaps at least a doc comment stating the intended purpose of the class?

Added a docstring, annotated it explicitly internal, and am not exporting it. Hope that addresses any concerns.

lib/src/geo/crs.dart
ignatz: Will mark this conversation as resolved based on you feedback. Please re-open if you feel the issue has not been addressed.

lib/src/misc/simplify.dart
ignatz: Will mark this conversation as resolved based on you feedback. Please re-open if you feel the issue has not been addressed.

JaffaKetchup (approved):
I think LGTM, thanks!

josxha (changes requested):
Tested the pull request on windows and found no bugs.

In my opinion all functions in `simplify.dart` should have at least a doc string for better documentation. 
https://github.com/fleaflet/flutter_map/pull/1801/commits/e170a1a31b9299312157c0eea94bce41aad810dd and https://github.com/fleaflet/flutter_map/pull/1801/commits/9e07d1fedaf1bfbea9bbb684539fa739de642ebb removed two doc strings without replacement.

Other than that the pr looks good to me.

JaffaKetchup:
I committed e170a1a and 9e07d1f. I removed the doc strings because
a) they are internal
b) the methods have relatively self documenting names:
`// simplification using Ramer-Douglas-Peucker algorithm` on `simplifyDouglasPeucker`
`/// high quality simplification uses the Ramer-Douglas-Peucker algorithm otherwise it just merges close points` on `simplify` when the method body is short and makes this specific comparison: `highQuality ? simplifyDouglasPeucker(nextPoints, sqTolerance) : simplifyRadialDist`

I'm happy to add them back in another PR @ignatz.

josxha:
If you want to documentation in a follow up pull request it's fine for me - althrough it shouldn't get delayed to much or someone unlucky in the future needs to dig through these complex undocumented functions.

josxha (approved):

Issue #1760, opened 2023-12-06
[BUG] Exception "`MapCamera` is no longer within the `cameraConstraint` after an option change" when using particular camera setups


dumabg:
**Current recommended workaround:**
Ensure `initialCenter` and `initialZoom` are configured in such a way that, if `cameraConstraint` & `initialCameraFit` were undefined, the map view would fit inside the desired camera constraint & fit. 

---

### What is the bug?

A FlutterMap with a cameraConstraint CameraConstraint.contain(bounds: widget.bounds) gives the assert error "MapCamera is no longer within the cameraConstraint after an option change"

### How can we reproduce it?

Put  a ValueListenableBuilder and a FlutterMap with a cameraConstraint. Change the ValueListenableBuilder value to trigger a new repaint.

### Do you have a potential solution?

I've comment the assert and seems that it works

### Platforms

Android

### Severity

Erroneous: Prevents normal functioning and causes errors in the console

JaffaKetchup:
Hi @dumabg,
Can I just double check what version you are running?

thorito:
It happens the same to me.

pubspect.yaml:
  flutter_map: ^6.1.0
  flutter_map_cancellable_tile_provider: ^2.0.0
  flutter_map_marker_cluster: ^1.3.4

My code:

```
 MapOptions _buildMapOptions() => MapOptions(
    initialCenter: widget._parametersMap.currentPosition.center,
    initialZoom: widget._parametersMap.initialZoom,
    maxZoom: widget._parametersMap.maxZoom,
    minZoom: widget._parametersMap.minZoom,
    initialRotation: widget._parametersMap.rotation,
    cameraConstraint: widget._parametersMap.cameraConstraint,
    initialCameraFit: markers.isNotEmpty &&
            UserPreferences.getInstance().currentPosition == null
        ? markers.getCameraFit(
            minZoom: widget._parametersMap.minZoom,
            maxZoom: widget._parametersMap.initialZoom,
          )
        : null,
    keepAlive: true,
    onMapReady: () {
      // logger.d('onMapReady !!');
    },
    onPositionChanged: (MapPosition position, bool hasGesture) {
      // logger.d('onPositionChanged hasGesture: $hasGesture,
      // position: $position');

      if (hasGesture) {
        ref.read(mapChangedProvider.notifier).changePosition();
      }
    },
    onTap: (_, __) => widget._popupController.hideAllPopups(),
  );

  @override
  Widget build(BuildContext context) {
[...]

PR #1804, opened 2024-01-21
chore: close issues with no user response

josxha:
- This pull request marks issues with the `waiting for user response` label automatically `stale` after 3 weeks. This is a reminder for the user to answer a question or provide requested information.
- After another week with no activity / user response, the issue gets closed automatically.
- If the user responds within the week the `waiting for user response` and `stale` label gets automatically removed.

### DIFF
--- a/.github/workflows/issues-no-response.yml
+++ b/.github/workflows/issues-no-response.yml
@@ -1,0 +1,25 @@
+name: Close issues with no user response
+on:
+  schedule:
+    - cron: "0 0 * * *"
+
+jobs:
+  close-issues:
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+      pull-requests: write
+    steps:
+      - uses: actions/stale@v5
+        with:
+          days-before-issue-stale: 21
+          days-before-issue-close: 7
+          stale-issue-label: "stale"
+          stale-issue-message: "This issue is stale because it has been marked with `waiting for user response` but hasn't received any activity since 3 weeks."
+          close-issue-message: "This issue was closed because it has been inactive another week since being marked as stale."
+          days-before-pr-stale: -1
+          days-before-pr-close: -1
+          remove-stale-when-updated: true
+          any-of-labels: "waiting for user response"
+          labels-to-remove-when-unstale: "waiting for user response, stale"
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/issues-no-response.yml
+++ b/.github/workflows/issues-no-response.yml
@@ -21,5 +21,5 @@ jobs:
           days-before-pr-close: -1
           remove-stale-when-updated: true
           any-of-labels: "waiting for user response"
-          labels-to-remove-when-unstale: "waiting for user response, stale"
+          labels-to-remove-when-unstale: "waiting for user response,stale"
           repo-token: ${{ secrets.GITHUB_TOKEN }}


JaffaKetchup (commented):
I have mixed feelings about this. We previously had this stale bot (admittedly without the tag exception rules), but it was a bit of a pain. Even if 50% of bug reports without enough information are invalid, that still leaves 50% that would be better investigated. Leaving them open allows other people to find issues and add their own info.
Would like some input from the others (@ibrierley, @TesteurManiak, @mootw).

josxha:
(There is currently one problem with this PR: The `waiting for user response` will not get removed if the issue is not stale.)

---

> We previously had this stale bot (admittedly without the tag exception rules), but it was a bit of a pain.

I'm aware that this was used in the past. While it was configured to close all issues in the past, this pr gives automation for the `waiting for user response` label. It will only handle those and ignore all other issues.

Having some kind of automation around `waiting for user response` is required to ensure the label gets removed when it recieves an update. For my understanding `waiting for user response` is a useful label if an issue can't be reproduced or requires the answer to an question to be able to get solved.

https://github.com/fleaflet/flutter_map/issues/1760 is a good example what happens if we don't have this automation.

> Even if 50% of bug reports without enough information are invalid, that still leaves 50% that would be better investigated.

It wouldn't make sense for us to tag those other 50% of valid bug reports with `waiting for user response`, therefore I can't see your argument here.

> Leaving them open allows other people to find issues and add their own info.

It is still possible for other people to adopt such issues. As soon as an update gets posted the `waiting for user response` would get removed.

JaffaKetchup:
@josxha Is this ready?

josxha:
Not yet. I'll add this:

> There is currently one problem with this PR: The waiting for user response will not get removed if the issue is not stale.

.github/workflows/daily.yml
```diff
@@ -0,0 +1,25 @@
...
+    permissions:
+      issues: write
+      pull-requests: write
+    steps:
+      - uses: actions/stale@v5
+        with:
+          days-before-issue-stale: 21
+          days-before-issue-close: 7
+          stale-issue-label: "stale"
+          stale-issue-message: "This issue has been marked that it requires additional information or answers to open questions to be able to get resolved. However, there hasn't been any response within the last 3 weeks.\n If this issue still resists, please provide the requested information. This issue will be automatically closed in one week if there is no response."
```
JaffaKetchup, line 18: Improved conciseness and fixed spelling mistake.

```suggestion
          stale-issue-message: "This issue requires additional information in order to be resolved. However, there hasn't been any response within the last 3 weeks.\n If this issue still persists, please provide the requested information. This issue will be automatically closed in one week if there is no response."
```

.github/workflows/daily.yml
```diff
@@ -0,0 +1,25 @@
...
+          #close-issue-message: "This issue was closed because it has been inactive another week since being marked as stale."
```
JaffaKetchup, line 19:
```suggestion
          close-issue-message: "This issue was closed because it has been inactive for 4 weeks, and still requires information in order to be resolved."
```

JaffaKetchup (changes requested):
Two minor changes to wording, let me know what you think of them.

.github/workflows/daily.yml
josxha: I had it disabled because the reminder message has basically all the information.
Changing it to your suggestion.

JaffaKetchup (approved):
LGTM! Thanks :)

PR #1806, opened 2024-01-21
refactor: split layers into libraries

josxha:
This pull request refactors the map layers into libraries that haven't refactored yet.


JaffaKetchup (approved):
LGTM! Untested but this shouldn't make any difference.

PR #1809, opened 2024-01-23
multi!: rewrite gesture handling

josxha:
Moved the `gesture-rewrite` branch to the flutter_map repo to have pull requests for this branch on here too.
For previous discussions see https://github.com/fleaflet/flutter_map/pull/1733.

### Changes in addition to the ones listed in https://github.com/fleaflet/flutter_map/pull/1733:
- tiles load continuously for `doubleTapDragZoom` gesture
- fling gesture has been turned into a gesture option for the drag gesture
- now all tap gesture only emit events when they have a registered callback (before double tap gestures haven't while regular taps have emitted an event).
- all gesture animation gets cancelled on a pointer down event.
- changes to the trackpad (aka. touchpad) gestures


josxha:
### Bugs reported by @JaffaKetchup:

- [ ] One bug with cursor keyboard rotation. Rotation is not tracking the cursor properly. I remember having a similar issue when implementing the algorithm initially. Maybe double checking that nothing in the implementaiton itself has changed, and every variable is being changed/reset at the correct points.

- [ ] Cursor keyboard rotation also no longer sets rotation to the angle of cursor immediately when clicking.

https://github.com/fleaflet/flutter_map/assets/58115698/74cc4727-7a31-49c7-8617-b5ecf21a9c35

---

- [ ] One bug with fling event end/finished event not being emmitted (when fling is cancelled with tap (as shown in video), or when finished naturally)

---

- [ ] Double tap zoom no longer emits correct event.

PR           |  Master
!pr  |  master

---

- [ ] Can no longer fling if drag gesture is disabled. On master, enabling fling without drag would start a fling if the normal conditions were met, but wouldn't drag/move if not, and wouldn't move whilst normal dragging.

---

- [ ] Double tap + drag gesture now loads tiles whilst zooming. master behaviour isn't correct either. The correct behaviour is that new tiles shouldn't load and old ones shouldn't be pruned whilst moving, and the refresh should occur when the gesture completes.

---

- [ ] On tap events and callbacks are no longer triggered.

---

- [ ] Can't seem to perform the new trackpad zoom gesture. Can you explain how to do it?

---

Still untested on any touch devices. Will perform that testing later.

JaffaKetchup (changes requested):
1) Double tap + drag zoom gesture emits `MapEventMove` rather than a zoom gesture
2) Double tap + drag zoom gesture emits 'Unknown' event on completion - need to update example app

https://github.com/fleaflet/flutter_map/assets/58115698/c53be987-1749-497b-bb10-8433316f14fb

---

3) Fling gesture can no longer be toggled in gesture app

---

4) Two-finger drag checkbox doesn't appear to make any difference on Windows 11 trackpad (Lenovo). Is two-finger drag intended for mobile?

5) Trackpad zoom gesture toggle has no effect, and cannot zoom with trackpad. Instead, scroll gesture controls the trackpad zoom.

image

https://github.com/fleaflet/flutter_map/assets/58115698/a09758f1-25e7-4a8d-b79c-f0124a55357f

---

6) The new system where events aren't emitted unless a callback is directly assigned to listen to that event kinda breaks the whole point of the `onMapEvent` callback.

7) Defining/undefining a specific callback does not change whether the event was emitted from the original state (state is not changed properly).

---

8) Double tap zoom gesture has no delay when performed on mouse (which is good), but there is delay when gesture performed on trackpad.

---

I'm a little concerned that we're splitting the mouse, touchpad, and mobile/touchscreen gestures apart. As far as I'm concerned the touchpad gestures should be the same as the mobile/touchscreen gestures. Is that what you're aiming for as well? For example, the pinch zoom gesture is the same pinch that's done on the trackpad.

josxha:
> 5. Trackpad zoom gesture toggle has no effect, and cannot zoom with trackpad. Instead, scroll gesture controls the trackpad zoom.

This is expected behaviour. Since the driver release date is 2006 I assume that your touchpad doesn't support the new gestures of flutter.
`Depending on the platform and specific trackpad model, the new system might not be used, if not enough data is provided to the Flutter engine by platform APIs. This includes on Windows, where trackpad gesture support is dependent on the trackpad’s driver, and the Web platform, where not enough data is provided by browser APIs, and trackpad scrolling must still use the old PointerScrollSignal system.` https://docs.flutter.dev/release/breaking-changes/trackpad-gestures#description-of-change

> 4. Two-finger drag checkbox doesn't appear to make any difference on Windows 11 trackpad (Lenovo). Is two-finger drag intended for mobile?

There was up to this point no intention for a two-finger drag on trackpads and I'm not able to perform this gesture on other map libraries. Is this another flutter_map specific gesture?

> 6. The new system where events aren't emitted unless a callback is directly assigned to listen to that event kinda breaks the whole point of the onMapEvent callback.

Please note that events haven't been emitted concistently. normal tap gestures have while the double tap hasn't. The implementation in this pr only registers callbacks if they are needed and therefore only emits an event if the user listens to it. Maybe I understood the `onMapEvent` callback wrong. For my understanding this callback provides all events that happen to the user, not every event that would be possible.

> 7. Double tap zoom gesture has no delay when performed on mouse (which is good), but there is delay when gesture performed on trackpad.

I experience the same behaviour for master. Have you really experienced different behaviours between master and this pr?

---

> I'm a little concerned that we're splitting the mouse, touchpad, and mobile/touchscreen gestures apart. As far as I'm concerned the touchpad gestures should be the same as the mobile/touchscreen gestures. 

Touchpads and touchscreens have different gestures. For example: Run the app from the master branch and do a doubleClickDragZoom gesture. On touchpad this zooms the map in and out to allow zooming with one finger. On touchpad however this gesture moves viewport which is the expected touchpad behaviour. If your goal is it to provide the same gestures on trackpad, touchpad and mouse, a high level implementation is not sufficient and we need to use a RawGestureDetector.

Is that what you're aiming for as well? For example, the pinch zoom gesture is the same pinch that's done on the trackpad.

The classic windows zoom gesture is to move two fingers vertically on the touchpad. You can use this gesture on the web live demo: https://demo.fleaflet.dev/ More advanced touchpads and mac touchpads support the pinch zoom gesture as well. But we can't assume that this gesture is supported by any touchpad.

josxha:
Just a quick update to this pull request for everyone who's following along. It was decided internally to collect all gestures with their supported platforms and input devices to get a better overview. I created a spreadsheet for this:
https://docs.google.com/spreadsheets/d/10N41ueX9Pr0H3gc89hFQLbRwcqKpeOMXNVrYR-FR0g4/edit?usp=sharing
(note that there are multiple sites)

Looking forward to some feedback or comments to be able to continue.

josxha:
Closing this pr as it is quite outdated.

Issue #1748, opened 2023-12-02
[SPEC] gesture system enhancements


josxha:
### Motivation
The user experience can be smoothend if we add zoom level transitions and animation while and after zooming. 

> **Open for comments.** It needs to be decided what should get added and what shouldn't or only optionally.

### Potential features
1. Use transparency to blend between zoom levels while zooming
    - Leaflet and the legacy Google raster map use this.
    - The vector map by google does not uses zoom layer blending.
2. Animated zooming when using the scroll wheel
    - Both GoogleMaps and Leaflet use this
    - https://github.com/flutter/flutter/pull/139392
    - [scroll_position_with_single_context.dart#L255-L291](https://github.com/Piinks/flutter/blob/8c4a69ad818a405087522f904d4560cd1152f56f/packages/flutter/lib/src/widgets/scroll_position_with_single_context.dart#L255-L291)
3. Option to animate to a full zoom level after gesture input stops, not active by default.
    - GoogleMaps raster tiles are animating to the next full zoom level in the same direction.
    - Leaflet seems to animate to the nearest full zoom level
4. Use canvas rather than widgets for tiles. 
    - GoogleMaps prefered to for better control over the performance.
    - This might have different results with flutter.
5. Don't request tiles if a zoom level gets skipped when the user zooms fast. 
    - This gets calculated by the excepted time a zoom level will be visible according to the zoom speed.
    - Don't show downscaled tiles with a zoom level difference >1. (map would look cluttered, large amount of tiles to draw hurts performance)
6. Preloading of tiles at surrounding zoom levels
    - Ensures a more smooth transition
    - Greater data consumption
7. https://github.com/fleaflet/flutter_map/issues/1768
8. https://github.com/fleaflet/flutter_map/issues/1432
9. Animate the rotation of the the ctrl key + click gesture. (ref https://github.com/fleaflet/flutter_map/pull/1809)

### Scope for `flutter_map`
| ID | Feature | Default behaviour | Optional | Requirements & additional information |
| -- | -- | -- | -- | -- |
| 1 | Use transparency to blend between zoom levels while zooming | ✅ | ❌ | |
| 2 | Animated zooming when using the scroll wheel | ✅ | ❌ | |
| 3 | Animate to a full zoom level after pinch gesture ends | ✅ | ✅ | |
| 5 | Don't request tiles if a zoom level gets skipped when the user zooms fast | ✅ | ❌ | only when not noticable |
[...]

PR #1810, opened 2024-01-23
feat(gestures): add key trigger click rotate gesture

josxha:
>  (1) One bug with cursor keyboard rotation. Rotation is not tracking the cursor properly. I remember having a similar issue when implementing the algorithm initially. Maybe double checking that nothing in the implementaiton itself has changed, and every variable is being changed/reset at the correct points.

- This pr adds the `keyTriggerClickRotate` gesture to the gesture rewrite.
- The gesture takes slightly longer to execute than on master to filter out drag gestures while the CTRL/specified key is pressed.
- The rotation should be animated, added this as an idea to https://github.com/fleaflet/flutter_map/issues/1748.

### DIFF
--- a/lib/src/map/gestures/map_interactive_viewer.dart
+++ b/lib/src/map/gestures/map_interactive_viewer.dart
@@ -42,6 +42,7 @@ class MapInteractiveViewerState extends State<MapInteractiveViewer>
   DragGestureService? _drag;
   DoubleTapDragZoomGestureService? _doubleTapDragZoom;
   KeyTriggerDragRotateGestureService? _keyTriggerDragRotate;
+  KeyTriggerClickRotateGestureService? _keyTriggerClickRotate;
 
   MapControllerImpl get _controller => widget.controller;
 
@@ -295,6 +296,13 @@ class MapInteractiveViewerState extends State<MapInteractiveViewer>
           KeyTriggerDragRotateGestureService(controller: _controller);
     } else {
       _keyTriggerDragRotate = null;
+    }
+
+    if (newGestures.keyTriggerClickRotate) {
+      _keyTriggerClickRotate =
+          KeyTriggerClickRotateGestureService(controller: _controller);
+    } else {
+      _keyTriggerClickRotate = null;
     }
 
     if (newGestures.doubleTapDragZoom) {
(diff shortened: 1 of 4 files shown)


JaffaKetchup:
> The gesture takes slightly longer to execute than on master to filter out drag gestures while the CTRL/specified key is pressed.

Can you elaborate here? If the trigger key is held, there should be no other gestures or `onTap`/`onDoubleTap`s captured, and therefore no delay.

josxha:
Not sure if I should call this a feature or a limitation of the new implementation...
One of the biggest goals of the gesture rewrite was to be as high level as possible to avoid any potential platform specific bugs.
Because we are using the `GestureDetector` to listen for every gesture input it supports and `GestureDetector` has delays some callbacks until they are semi-verified or verified we have a slight delay here (the known double tap delay).
We could disable the double tap callbacks while the triggerKey is pressed, however this would require an additional widget rebuild. Additionally we would need a on-trigger-key-release callback that could lead to problems if the targeted windows switches before the key gets released. 
While the delay is noticable at the moment I think it's not that bad once the rotation is animated. I think I've seen such delays for commerical maps as well if I'm not wrong 🤔 

What do you think? Do you want to have some workaround code to have no delay at all?

josxha:
Added the `keyTriggerDragRotate` gesture fix to this pr because both are quite similar.

JaffaKetchup:
> We could disable the double tap callbacks while the triggerKey is pressed, however this would require an additional widget rebuild

I don't know, but this can't be too much of an issue. We're constantly doing this all the time (as far as I understand, if you mean `setState`).

> Do you want to have some workaround code to have no delay at all?

Yes, as above. This delay makes the controls feel sluggish - gestures should give instant feedback, otherwise the user may attempt to perform them again.

JaffaKetchup (changes requested):
> We could disable the double tap callbacks while the triggerKey is pressed, however this would require an additional widget rebuild

I don't know, but this can't be too much of an issue. We're constantly doing this all the time (as far as I understand, if you mean setState).

> Do you want to have some workaround code to have no delay at all?

Yes, as above. This delay makes the controls feel sluggish - gestures should give instant feedback, otherwise the user may attempt to perform them again.

josxha:
Closing this pr as it is quite outdated.

PR #1814, opened 2024-02-01
feat(gestures): fling animation as drag gesture option

josxha:
(5) ~~Can no longer fling if drag gesture is disabled. On master, enabling fling without drag would start a fling if the normal conditions were met, but wouldn't drag/move if not, and wouldn't move whilst normal dragging.~~ 
Turn the fling gesture into an option for the drag gesture.

### DIFF
--- a/lib/src/map/options/map_gestures.dart
+++ b/lib/src/map/options/map_gestures.dart
@@ -13,7 +13,6 @@ class MapGestures {
   /// use [InteractiveFlags.bitfield] instead.
   const MapGestures({
     required this.drag,
-    required this.flingAnimation,
     required this.twoFingerMove,
     required this.twoFingerZoom,
     required this.doubleTapZoomIn,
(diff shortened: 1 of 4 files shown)


lib/src/map/options/map_gestures.dart
```diff
@@ -265,9 +252,6 @@ abstract class InteractiveFlag {
   /// Enable panning with a single finger or cursor
   static const int drag = 1 << 0;
 
-  /// Enable fling animation after panning if velocity is great enough.
-  static const int flingAnimation = 1 << 1;
```
JaffaKetchup, line 269: May want to consider moving the latter bits down, but not important (especially if we are going to remove all this anyway).

JaffaKetchup (approved):
Untested, will test as part of main PR. Code LGTM. Thanks!

lib/src/map/options/map_gestures.dart
josxha: We can if you want, although would be another breaking change. Let me know if you want me to change them.

PR #1815, opened 2024-02-01
fix(gestures): double tap zoom events, fling events

josxha:
#### (4) Double tap zoom no longer emits correct event.

PR           |  Master
!pr  |  master

#### (3) One bug with fling event end/finished event not being emmitted (when fling is cancelled with tap (as shown in video), or when finished naturally)

## Information
- Fixed the emitted map events for `doubleTapZoomIn` and the fling animation of the `drag` gesture.
- I added callbacks for when an animation has ended or has been cancelled. It turned out that they are not used but having some callbacks for both events could make sense for the future. @JaffaKetchup you can decide if we keep or remove them.

### DIFF
--- a/lib/src/map/controller/map_controller_impl.dart
+++ b/lib/src/map/controller/map_controller_impl.dart
@@ -508,8 +508,8 @@ class MapControllerImpl extends ValueNotifier<_MapControllerState>
   bool get isAnimating => _animationController.isAnimating;
 
   /// cancel all ongoing animation
-  void stopAnimations() {
-    _animationController.stop();
+  void stopAnimations({bool cancelled = true}) {
+    _animationController.stop(canceled: cancelled);
     _moveAnimation = null;
     _rotationAnimation = null;
     _zoomAnimation = null;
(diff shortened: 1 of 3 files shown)


lib/src/map/controller/map_controller_impl.dart
```diff
@@ -418,21 +424,24 @@ class MapControllerImpl extends ValueNotifier<_MapControllerState>
     required Curve curve,
     required bool hasGesture,
     required MapEventSource source,
+    AnimationEndedCallback? onAnimatedEnded,
+    AnimationCancelledCallback? onAnimationCancelled,
```
JaffaKetchup, line 428: Could there be any use in having these callbacks exposed through `MapOptions`? Again, need to consider if we're adding too many callbacks and whether there's a better way to do it (eg. more event types, remove callbacks except `onMapEvent`) - but that's a discussion for another time.

JaffaKetchup (approved):
Untested, will test as part of main PR. Code LGTM. Thanks!

lib/src/map/controller/map_controller_impl.dart
josxha: Alowing the user to listen for it makes indeed sense but isn't this already possible with the event system?
Side note: these callbacks are not persistent, they are only for one animation and get cleared afterwards.

PR #1816, opened 2024-02-01
refactor: make `FlutterMap.options` an optional parameter

josxha:
This is a really tiny change but a huge quality of life change - at least for me. 😅

### DIFF
--- a/lib/src/map/widget.dart
+++ b/lib/src/map/widget.dart
@@ -20,7 +20,7 @@ class FlutterMap extends StatefulWidget {
   const FlutterMap({
     super.key,
     this.mapController,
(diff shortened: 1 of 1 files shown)


JaffaKetchup (approved):
LGTM, although we may want to consider whether the current default values are useful. Thanks!

VittorioParagallo:
I'm not sure it depends from this, with versions 7 i got an assertion exception "flutter map 'newOptions != value.options': Should not update options unless they change"
This happens when:
- i show a map
- open a dialog
- i close the dialog and the error pop up.
reverted on v6.1, no more troubles.

Issue #1314, opened 2022-07-20
[FEATURE] Add EPSG3413 (polar) projection support to `ScaleLayerPlugin` in example app


JosefWN:
### What is the bug?

The `ScaleLayerPlugin` shows incorrect scale, sometimes negative.

### What is the expected behaviour?

The `ScaleLayerPlugin` should work the same regardless of projection.

### How can we reproduce this issue?

Add a `ScaleLayerPlugin` to the example `example/lib/pages/epsg3413_crs.dart`. When you pan using this projection the scale should not change, and it should never go negative.

Best would be to draw a couple of objects of known size on different lat/lons (especially above/below the pole) and compare them to the scalebar to ensure that the correct scale is given at different zoom levels.

### Platforms Affected

iOS, Web, MacOS

### Severity

Minimum: Allows normal functioning

### Frequency

Consistently: Always occurs at the same time and location

### Requirements

- [X] I agree to follow this project's [Code of Conduct](https://github.com/fleaflet/flutter_map/blob/master/CODE_OF_CONDUCT.md)
- [X] My Flutter/Dart installation is unaltered, and `flutter doctor` finds no relevant issues
- [X] I am using the [latest stable version](https://pub.dev/packages/flutter_map) of this package
- [X] I have checked the [Common Issues](https://docs.fleaflet.dev/usage/common-issues) documentation page
- [X] I have checked for similar issues which may be duplicates

josxha:
Changing to a feature request because the ScaleLayerPlugin is meant as a showcase for a plugin and isn't use with epsg3413 in the example app.

JaffaKetchup:
The example no longer has its own scalebar, instead it has been moved into FM as a core component: `Scalebar`.

PR #1821, opened 2024-02-07
feat: add scale bar layer

josxha:
### To do
- [x] add scale bar as map layer
- [x] fix left padding partly outside screen when padding 0
- [x] fix top and left padding not the same
- [x] `lineWidth` is now `strokeWidth`
- [x] add `lineHeight` parameter
- [x] default to black as color
- [x] fix text rendererd partially outside of screen if the scalebar is very small
- [x] scalebar alignment
- [x] fix label overflow on right screen side
- [x] fix alignment of scalebar if it is smaller than the text

### Known problems
- When performing a fling animation against the top or bottom border, the scalebar increaes. I don't think that this is a bug of this pr, that's why it's not included here.


josxha:
I looked a bit into https://github.com/fleaflet/flutter_map/issues/1314 but wasn't able to get it working.

lib/src/layer/scalebar/scalebar.dart
```diff
@@ -0,0 +1,226 @@
...
+        ..scaleDistance = distance,
+    );
+  }
+
+  /// Calculates the length of the scalebar
+  static LatLng _calculateEndingGlobalCoordinates({
+    required LatLng start,
+    required double startBearing,
+    required double distance,
+  }) {
```
josxha, line 68: I kept this function separate to reduce the complexity of the scalebar.dart file because this file gets viewed more often than other related files. IMO it should be structured to be less overwhelming.

example/lib/pages/scalebar.dart
```diff
@@ -1,32 +1,28 @@
...
         ),
         children: [
           openStreetMapTileLayer,
-          const FlutterMapScaleLayer(
-            lineColor: Colors.black,
-            lineWidth: 3,
-            textStyle: TextStyle(color: Colors.black, fontSize: 14),
-            padding: EdgeInsets.all(10),
+          Scalebar(
+            painter: SimpleScalebarPainter(padding: const EdgeInsets.all(12)),
```
josxha, line 25: There should be some level of abstraction here. The user doesn't need to know that a Painter gets used internally. Padding is probably a parameter that every Scalebar should have therefore provided to the Scalebar widget directly.
Being able to provide an Painter could make it more open on the one side but you prevented an open extendability by the `base` class modifier. Don't get me wrong, I don't think that we should provide this as well but keeping the all Painters internally all together sounds like the better approach to me.

If we want to support multiple Scalebar styles I'd argue that mimicking the way Flutter does it would be good: `Switch()` and `Switch.adaptive()` or `ListView()` and `ListView.builder()`.
 If transfered to our Scalebar: `Scalebar()` would be the simple (no need to use an extra simple here) and further extensions would be created as `Scalebar.dualUnit`, `Scalebar.google` an so on.
This has the advantage that the painter can stay internally but theoretically created already in the constructor and we have full control of the parameters each Scalebar should support.

JaffaKetchup: I implemented it like this to allow for users to create their own painter based on the maths in the `Scalebar` widget.
`base` does not prevent external extension, it just forces the use of `extends` rather than `implements`. Can remove the class modifier altogether if you'd prefer that.

I'm also up for changing the name away from `Painter` to give the illusion of abstraction. Maybe `Style`?

Yes, things like `padding` could possibly be moved outside of the painter, but I didn't want to interfere with anything in these commits, just rearrange.

lib/src/layer/scalebar/scalebar.dart
JaffaKetchup: I've undone this change, but still made the function private.

example/lib/pages/scalebar.dart
josxha: Alright let's make it extentable.

josxha: I looked into this a bit further and came across more problems if we want to let users provide their own painters:
1. The `Scalebar` widget contains not much logic itself, it mostly calculates the width of the scalebar for a metric scale. This means that we can't really have painters based on other measure systems. Providing users access to `_calculateEndingGlobalCoordinates` would give users more flexibility while limiting our public API.
2. The `Scalebar` is build for horizontal sclebars in mind, things would need to get renamed and generalized otherwise.
3. Marking fields as `internal parameter` that shouldn't get used outside isn't good style.
```dart
  /// Internal parameter, do not use externally
  @protected
  abstract double scaleWidth;

  /// Internal parameter, do not use externally
  @protected
  abstract int scaleDistance;
```
4. Most of the times we tell users to build a plugin for similar cases. Wouldn't this be the better solution here too? We can encorage this by open up the calculation from `_calculateEndingGlobalCoordinates`.

JaffaKetchup: Yeah, ok, you're right. Feel free to revert my changes, or let me know if you want me to do it.
But I do still think alignment options should be provided.

josxha: Yes I'll add alignment. 👍

lib/src/layer/scalebar/painter/base.dart
```diff
@@ -0,0 +1,5 @@
+part of '../scalebar.dart';
+
+abstract class ScalebarPainter extends CustomPainter {
```
JaffaKetchup, line 3: Why does this exist if there's only one extension and no possibility for users to create/use their own? Forward planning?

JaffaKetchup (approved):
LGTM I think. I also realised that this only supports one unit (metric), but maybe we can change this later.

lib/src/layer/scalebar/painter/base.dart
josxha: Yes it's a bit of forward planning. The Scalebar widget needs to know from the painter on how large it should be.

josxha:
> I also realised that this only supports one unit (metric), but maybe we can change this later.
Yes, I experimented with an imperial scale a bit but it's a bit tricky since the calculations are metric and the the stop points of the scale bar need to be adjusted.

Are we ready to merge?

JaffaKetchup:
Yeah I think ready to merge. Maybe something like Google Maps double unit scale bar would work well in future.

josxha:
> Maybe something like Google Maps double unit scale bar would work well in future.

Yep, would be a nice addition.

S-Man42:
I really love to see this feature. However, it seems not to be correct.

The scale shows a distance of 1000km in the horizontal distance.

Screenshot_20240819_210047_Brave

In fact, it is not more than 650km. 

I guess it results from the problem, that the mercator projection yield different distances throughout the map for same pixel distances depending on latitude. 

But in this case it is not quite clear which latitude it shows. Does it measures the rhumb line or geodetic lengths? Currently it may show wring results as you can see.

I guess it could be a good idea to configure the scale this way, that it shows the distance of the start and end points lying exactly under the bar. 

Second point would be to decide which length you really display. The shortest length (geodetic) between start and end point of the scale, but in that case it is not a straight line and the half marker would not be at the drawed position. If one would show the length of the straight line as the scale bar shows, you would need the length of the rhumb line. (in mercator projection).

PR #1822, opened 2024-02-10
feat!: add interactive `Polygon`s, rework `Polyline` `hitTest` result, and deprecate/disable auto application of `TransulcentPointer`

### LINKED ISSUE #385: [FEATURE] Interactive `Polygon`s
Hi,

I'm trying to reimplement an existing Leaflet-based app using flutter. In that app, the developer used Polygons to display the outline of a building, which when clicked/tapped shows additional information. and changes the polygon.

I'm trying to replicate that function, however I noticed there is no onTap Listener, at least for Polygons. Is there a different way to listen to tap events?

### LINKED ISSUE #1740: [BUG] `MapOptions.applyPointerTranslucencyToLayers` causes `Positioned` wrappers to fail
### What is the bug?

I have been using `Positioned` extensively to place buttons around the map. This was validated by [some of the examples](https://github.com/fleaflet/flutter_map/blob/master/example/lib/widgets/drawer/floating_menu_button.dart). But those widgets [are required](https://api.flutter.dev/flutter/widgets/Positioned-class.html) to be placed in a Stack with no `RenderObjects` inbetween.

In 6.0.0 widgets, instead of being put in a Stack, [are all wrapped](https://github.com/fleaflet/flutter_map/commit/4ba3d02827dd0e3b6b20fc58b0b62d3f58160b54#diff-3d1ec1773715ac2564805c37018286dc1fee2c4087990e430611e27a7641d098L112) in `TranslucentPointer` objects. Which leads to positioning not working, and multiple error notices in a log.

Can this thing you wanted to do be done differently? Or, are there alternatives to positioning widgets inside the map?

I thought of putting those on top of the map instead, but in that case those widgets (e.g. zoom buttons) won't have access to map state.

### How can we reproduce it?

Use the `Positioned` widget as a top-level one for a map layer.

Make `applyPointerTranslucencyToLayers` flag not wrap widgets, instead of setting a property of `TranslucentPointer`?

Alternatively, a way to properly upgrade such plugins needs to be documented. `Align` widget doesn't work for bottom-aligned buttons, for example.

### Platforms

Android 12

### Severity

JaffaKetchup:
### Added interactive `Polygon`s
Resolves #385. Requires #1800 to be merged first.
This uses a similar format as to `Polyline` interactivity.

---

### Reworked `PolylinePainter.hitTest` outputs
Helps to fix #1740.
- Hit testing is now performed even if there is no attached `hitNotifier`
- Any attached notifier is now notified about a successful hit even if there were no `hitValue`s
- `hitTest` internal output no longer depends on state of `hitNotifier`, to improve definitity without below...

---

### Deprecated `MapOptions.applyPointerTranslucencyToLayers` & changed default to `false`, and removed usage `TransulcentPointer.translucent` property
Fixes #1740.
- This option caused unexpected behaviour in some cases, especially as it was enabled by default
- #1740 was caused because disabling it did not unwrap children, it just used the `translucent` property
- `TranslucentPointer` itself has been retained as it is a useful resource to further customize hit testing behaviour, and `CircleLayer` and `OverlayImageLayer` still have no hit testing provision
- Online documentation will be updated to explain new hit testing behaviour and rules

### DIFF
--- a/lib/src/map/widget.dart
+++ b/lib/src/map/widget.dart
@@ -100,12 +100,10 @@ class _FlutterMapStateContainer extends State<FlutterMap>
           Positioned.fill(
             child: ColoredBox(color: widget.options.backgroundColor),
           ),
-          ...widget.children.map(
-            // ignore: deprecated_member_use_from_same_package
-            (child) => widget.options.applyPointerTranslucencyToLayers
-                ? TranslucentPointer(child: child)
-                : child,
-          ),
+          // ignore: deprecated_member_use_from_same_package
+          ...widget.options.applyPointerTranslucencyToLayers
+              ? widget.children.map((child) => TranslucentPointer(child: child))
+              : widget.children,
         ],
       ),
     );
--- a/lib/src/map/widget.dart
+++ b/lib/src/map/widget.dart
@@ -31,9 +31,6 @@ class FlutterMap extends StatefulWidget {
   /// [Align] or another method to position themselves). Widgets/layers may or
   /// may not identify which type they are in their documentation, but it should
   /// be relatively self-explanatory from their purpose.
-  ///
-  /// [TranslucentPointer] will be wrapped around each child by default, unless
-  /// [MapOptions.applyPointerTranslucencyToLayers] is `false`.
   final List<Widget> children;
 
   /// Configure this map's permanent rules and initial state
@@ -104,10 +101,10 @@ class _FlutterMapStateContainer extends State<FlutterMap>
             child: ColoredBox(color: widget.options.backgroundColor),
           ),
           ...widget.children.map(
-            (child) => TranslucentPointer(
-              translucent: widget.options.applyPointerTranslucencyToLayers,
-              child: child,
-            ),
+            // ignore: deprecated_member_use_from_same_package
+            (child) => widget.options.applyPointerTranslucencyToLayers
+                ? TranslucentPointer(child: child)
+                : child,
           ),
         ],
       ),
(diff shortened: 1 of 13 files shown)


mohammedX6:
It's so nice you using polybool library 😄

JaffaKetchup:
Hey @mohammedX6,
We've decided not to use 'polybool' in this PR. It did exactly what we needed, and is a good package, but we were worried about bringing on such a general purpose package when we only need one tool from it. I hope you can understand!

lib/src/map/widget.dart
```diff
@@ -104,10 +101,10 @@ class _FlutterMapStateContainer extends State<FlutterMap>
...
           ),
           ...widget.children.map(
-            (child) => TranslucentPointer(
-              translucent: widget.options.applyPointerTranslucencyToLayers,
-              child: child,
-            ),
+            // ignore: deprecated_member_use_from_same_package
+            (child) => widget.options.applyPointerTranslucencyToLayers
+                ? TranslucentPointer(child: child)
+                : child,
```
josxha, line 107: If `applyPointerTranslucencyToLayers` isn't used, we don't have to map all children.

josxha (commented):
Haven't tested yet. Changes look straight forward, only found one thing.

lib/src/map/widget.dart
JaffaKetchup: Whoops, my oversight. Fixed.

josxha (approved):
lgtm 👍

PR #1825, opened 2024-02-13
Fix: repaint polygons/polylines on changes.

Alexays:
Hi!
Just tested the master version, and I noticed that my polygons no longer moved with the map, so here's the little fix.


JaffaKetchup (approved):
Huge thanks, LGTM :D!

PR #1826, opened 2024-02-14
fix exception if hitValue is null

mohammedX6:
This PR ensure that hitValue must not be null.

without checking it will throw this error :

image


JaffaKetchup:
Hi @mohammedX6,
Thanks for bringing this to my attention, I made a mistake when carrying something over from the polyline system. This PR does not correctly fix the issue unfortunately, I've done that in #1828, but thanks for contributing anyway :)

mohammedX6:
your welcome :)

PR #1828, opened 2024-02-14
fix: prevent null exception when undefined/null `Polygon.hitValue`

JaffaKetchup:
This works in the polyline layer, but I missed out a line copy when carrying some changes over.
Thanks to @mohammedX6 for bringing this to my attention. Replaces #1826.

### DIFF
--- a/lib/src/layer/polygon_layer/painter.dart
+++ b/lib/src/layer/polygon_layer/painter.dart
@@ -92,7 +92,7 @@ class _PolygonPainter<R extends Object> extends CustomPainter {
       // Second check handles case where polygon outline intersects a hole,
       // ensuring that the hit matches with the visual representation
       if ((isInPolygon && !isInHole) || (!isInPolygon && isInHole)) {
-        _hits.add(polygon.hitValue!);
+        if (polygon.hitValue != null) _hits.add(polygon.hitValue!);
         hasHit = true;
       }
     }


josxha (approved):
lgtm

Sorry that I haven't seen this in my review. 

edit: will do some additional testing, before I approve 😁

josxha:
When you zoom in far like showcased in the video and then hover over the polygons the app crashes.


https://github.com/fleaflet/flutter_map/assets/34318751/e909d301-abce-4e16-9dd8-d953a130f3d9

Not sure tho if this is related to this bug or another bug.

JaffaKetchup:
Oooh, that's quite nasty, wonder if it's a memory leak or something, or whether the hover outlines are triggering it.

JaffaKetchup:
I can reproduce the increase in frame times, but not the crash (I assume that's device dependent). I'll try and do some more debugging.
 
image

JaffaKetchup:
This is unrelated to tap detection, as it appears to be present prior to the introduction of this. It is present in commit f31829f94d5295ff2f09e7b82671177f1605bc10.
I think this is safe to be reviewed & merged.

mohammedX6:
maybe just setting the `shouldRepaint` to true is the safest option now ?

JaffaKetchup:
Ah, the commit I mentioned above isn't necessarily the commit that caused it, just the previous one that I can reproduce the issue on.

JaffaKetchup:
Report made in #1829.

PR #1832, opened 2024-02-20
refactor: add assertions, remove redundant constructor, unnecessary private field

josxha:
See the commit history, each change has it's own commit.


JaffaKetchup (approved):
LGTM!

PR #1834, opened 2024-02-20
refactor!: make `LatLngBounds` mutable & depend less on 'latlong2'

josxha:
This pr is probably a bit controversial. 
- It refactors LatLngBounds to store max and min latitude and longitude instead of the LatLngs of the corners
- It cleans up the methods to make them produce less ephemeral objects
- It renames the previous default contructor to `LatLngBounds.fromCorners()` and adds a new default raw constructor for better performance. Even though values get provided directly it uses asserts to catch potential implementation errors.

Very similar would be Bounds<T> if I'm not mistaken.


lib/src/geo/latlng_bounds.dart
```diff
@@ -1,86 +1,125 @@
...
+
+  /// The longitude west edge of the bounds
+  double west;
+
+  /// Create a [LatLngBounds] instance from raw edge values.
+  ///
+  /// Potentially throws assertion errors if the coordinates exceed their max
+  /// or min values or if coordinates are meant to be smaller / bigger
+  /// but aren't.
+  LatLngBounds({
```
JaffaKetchup, line 26: Instead of making this the new default constructor, can we call this `fromRawEdges`, and make the `fromCorners` un-named again? Just trying to minimize this breaking change.

JaffaKetchup (changes requested):
LGTM, except for the one thing around constructor naming.

lib/src/geo/latlng_bounds.dart
josxha: Can do, but instead of `fromRawEdges` I'd use the `unsafe` to be more in line with `Bounds.unsafe`. Sounds alright?

josxha:
Ready for another review. I think this pull request is now non-breaking.

JaffaKetchup (approved):
LGTM! Still breaking, but only for implementers of `LatLngBounds`, which should be no-one.

PR #1838, opened 2024-02-22
feat: 100% documented public members

josxha:
- Adds doc strings to all missing public members
- With this pr `public_member_api_docs` will get active. This enforces future PRs to keep the 100% documentation score of public members. Private members won't be affected.

### DIFF
--- a/lib/src/gestures/interactive_flag.dart
+++ b/lib/src/gestures/interactive_flag.dart
@@ -25,7 +25,7 @@ abstract class InteractiveFlag {
       scrollWheelZoom |
       rotate;
 
-  /// Use this flag if you want to disable all gesture interactions.
+  /// Disable all gesture interactions
   static const int none = 0;
 
   /// Enable panning with a single finger or cursor
--- a/lib/src/gestures/interactive_flag.dart
+++ b/lib/src/gestures/interactive_flag.dart
@@ -25,6 +25,7 @@ abstract class InteractiveFlag {
       scrollWheelZoom |
       rotate;
 
+  /// Use this flag if you want to disable all gesture interactions.
   static const int none = 0;
 
   /// Enable panning with a single finger or cursor
--- a/lib/src/layer/attribution_layer/rich/source.dart
+++ b/lib/src/layer/attribution_layer/rich/source.dart
@@ -9,9 +9,11 @@ import 'package:meta/meta.dart';
 sealed class SourceAttribution extends StatelessWidget {
   const SourceAttribution._({super.key, this.onTap});
 
-  /// This callback gets called if the user clicks or taps on the attribution
-  /// source. Most of the time you would want to open a website with the terms
-  /// of service of the tile provider.
+  /// This callback gets called when the user clicks or taps on the
+  /// attribution source.
+  /// 
+  /// Most tile providers will require you to link to their terms of
+  /// service, which can be done through this callback.
   final VoidCallback? onTap;
 
   Widget _render(BuildContext context);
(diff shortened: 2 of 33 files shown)


lib/src/gestures/interactive_flag.dart
```diff
@@ -25,6 +25,7 @@ abstract class InteractiveFlag {
       scrollWheelZoom |
       rotate;
 
+  /// Use this flag if you want to disable all gesture interactions.
```
JaffaKetchup, line 28: Avoid 'you', and avoid explaining WHY this should be used in the first sentence of a doc. Instead explain WHAT is does.

https://dart.dev/effective-dart/documentation#prefer-starting-function-or-method-comments-with-third-person-verbs

```suggestion
  /// Disable all gesture interactions
```

lib/src/layer/attribution_layer/rich/source.dart
```diff
@@ -9,6 +9,9 @@ import 'package:meta/meta.dart';
 sealed class SourceAttribution extends StatelessWidget {
   const SourceAttribution._({super.key, this.onTap});
 
+  /// This callback gets called if the user clicks or taps on the attribution
+  /// source. Most of the time you would want to open a website with the terms
+  /// of service of the tile provider.
```
JaffaKetchup, line 14:
```suggestion
  /// This callback gets called when the user clicks or taps on the
  /// attribution source.
  /// 
  /// Most tile providers will require you to link to their terms of
  /// service, which can be done through this callback.
```

lib/src/layer/polygon_layer/polygon.dart
```diff
@@ -26,6 +26,9 @@ class Polygon<R extends Object> {
   /// The color of the [Polygon] outline.
   final Color borderColor;
 
+  /// Weather holes on the [Polygon] should have a border too or not.
```
JaffaKetchup, line 29:
```suggestion
  /// Whether holes should have borders
```

lib/src/layer/tile_layer/tile_coordinates.dart
```diff
@@ -2,6 +2,8 @@ import 'dart:math';
 
 import 'package:meta/meta.dart';
 
+/// The tile coordinates are based on maths' [Point] class but extended with
+/// the zoom level.
```
JaffaKetchup, line 6: Consider a brief explanation of how tile coordinates work, or a link to the documentation site: https://docs.fleaflet.dev/getting-started/explanation#slippy-map-convention?

lib/src/layer/tile_layer/tile_image_view.dart
```diff
@@ -3,6 +3,7 @@ import 'dart:collection';
 import 'package:flutter_map/flutter_map.dart';
 import 'package:flutter_map/src/layer/tile_layer/tile_range.dart';
 
+/// A
```
JaffaKetchup, line 6: Needs finishing.

JaffaKetchup (changes requested):
Mostly LGTM, just a few things to think about...

lib/src/layer/tile_layer/tile_image_view.dart
josxha: woops, that's an oversight. Thanks for bringing this up 😄 fixed

lib/src/layer/tile_layer/tile_coordinates.dart
josxha: Added more informations. Feel free to improve or enhance. Not sure how to describe best so I added the link to the docs as well.

JaffaKetchup (approved):
LGTM! Huge thanks, a lot of work here.

Issue #1447, opened 2023-02-17
[FEATURE] Support dashed `Polyline`s (in addition to dotted)


WillyAdrianus:
### What do you want implemented?

if it is possible add dashed polyline with parameter isDashed: true and dashedWidth: {int}

### Platforms Affected

Other

### Severity

Obtrusive: No workarounds are available, and this is essential to me

### Requirements

- [X] I agree to follow this project's [Code of Conduct](https://github.com/fleaflet/flutter_map/blob/master/CODE_OF_CONDUCT.md)
- [X] I am using the [latest stable version](https://pub.dev/packages/flutter_map) of this package
- [X] I have checked for similar feature requests which may be duplicates

MendleM:
I spent a little time today working on this. I'll get around to it in the coming weeks. If someone gets to it before me, if the lines could have miter joins like this design (from Figma if you want to play around with dashes), that'd be incredible:

<img width="345" alt="image" src="https://user-images.githubusercontent.com/20160314/222955117-da617766-c16c-45f7-85fe-0adcc9aab6ef.png">

JaffaKetchup:
Hey @MendleM, did you ever get around to this? No problem if not!

JaffaKetchup:
Hey @monsieurtanuki, this looks excellent, we'd love to see it!

JaffaKetchup:
Closed by #1855.

PR #1845, opened 2024-03-07
feat: add `dottedSpacingFactor` to customize dotted polyline spacing

gnassro:
Now we can customize the space between dots if `isDotted=true`


josxha:
Thanks for the pull request @gnassro! 
I did a quick test and your changes seem to work well.

There are some small things I would prefer to have changed:
- We have an open [feature request for dashed polylines](https://github.com/fleaflet/flutter_map/issues/1447). So it wood be good to have the name a bit more general like "segmentSpacingFactor" or somthing similar. Then we can use the parameter for dashed polyine too once the feature arrives.
- The CI checks show that the code is not formatted, you can do this with `dart format .`, but if you have difficulties I can do it for you if you want.


Then there are some general things I discovered while testing that <ins>you don't need to care about</ins> but are worth writing down:
- changing isDotted doesn't get picked up by hot reload
- It would be in general nice to have the dotted Polyline features represented in the example app, but I think that putting all on the polyline page would look cluttered.

gnassro:
@josxha the only thing that I didn't understand is, what do you mean by `changing isDotted doesn't get picked up by hot reload` ?

JaffaKetchup (approved):
LGTM, thanks :). Will wait for a second review before merging.

josxha:
Sorry for the delay, changes looking good. Thanks for updating the example app too.

> the only thing that I didn't understand is, what do you mean by changing isDotted doesn't get picked up by hot reload ?

This is a bug that is not related to this pull request. If you 
- run the app in debug mode
- change `isDotted: true,` to `isDotted: false,`
- trigger a hot reload

Then the polyline doesn't change it's appearance from a dotted to a normal line. A quick guess is because the polyline hashCode gets cached it doesn't regocnizes the change - but anyways as said it's not related to this pr.

Thanks a lot for the contribution!

josxha (approved):

PR #1846, opened 2024-03-08
fix: remove legacy moving_markers and stateful_markers page from example app

josxha:
~~fix Timer on the moving_markers page, bug discovered in~~
~~https://discord.com/channels/951867686378409984/1215676233346711633/1215677536004997271~~

~~Bug can be seen here: https://demo.fleaflet.dev/moving_markers~~

Remove legacy pages.


TesteurManiak:
I'm not sure of what this page is supposed to showcase 😅 
If it's just printing info of a marker, I'd tend to think that it can be removed.

josxha:
> I'm not sure of what this page is supposed to showcase 😅
If it's just printing info of a marker, I'd tend to think that it can be removed.

Not sure either, maybe to test that the map detects and rebuilds the MarkerLayer if the markers change?
If we remove this this page, should we remove https://demo.fleaflet.dev/stateful_markers too? Or what is the usage of that showcase?

JaffaKetchup:
They are probably remnants of the old widget/option system. If you find all 172 remnants left you get a special prize 😂.

josxha:
I removed both pages from the example app.

TesteurManiak (approved):
LGTM 👍

Issue #1771, opened 2023-12-10
[FEATURE] (Re-)Support absolute `Marker` alignment


JaffaKetchup:
### What do you want implemented?

Since v6, `Marker`s can only be aligned relatively (on a scale of 0-1) through `Alignment`. It would be good for there to be an alternative constructor that takes a number of pixels from each edge, and the size of the `Marker`, and calculates the relative `Alignment`.

### What other alternatives are available?

As an example, given the width (`w`) and number of pixels to offset from the left (`d`), the formula `1.0 - 2 * d / w` can be used, as suggested by @Zverik in https://github.com/fleaflet/flutter_map/pull/1659#issuecomment-1849013832.

### Severity

Annoying: Currently have to use workarounds

monsieurtanuki:
@JaffaKetchup @josxha May I work on this issue?

JaffaKetchup:
Hey @monsieurtanuki,
Sure, we appricate any help we can get :)

PR #1847, opened 2024-03-09
feat: add static `Marker.computePixelAlignment` method to calculate absolute alignment

monsieurtanuki:
What I've done:
* Created a `factory Marker.withPixelAlignment` in order to set pixel-precise marker alignment, cf. #1771
* Updated the "Marker Layer" example to show that new feature with non)-standard "U-turn-left" icons
Screenshot_1710006953

What can still be done:
* I'm open to name change or to constructor against `factory` - but I wanted to keep the `const` and I'm not sure we can do that if _computing_ the alignment
* I can create another example, as reusing the "Marker Layer" for 2 distinct purposes can be confusing
* Now `Marker` has a specific pixel-alignment constructor that `MarkerLayer` does not have. Not sure if that would make sense to create a new `MarkerLayer` constructor with the left+top parameters being used as default.

Impacted files:
* `marker.dart`: added `factory Marker.withPixelAlignment`
* `markers.dart`: now a click on the map creates a new "U-turn-left" marker with custom alignment


JaffaKetchup:
> Now Marker has a specific pixel-alignment constructor that MarkerLayer does not have. Not sure if that would make sense to create a new MarkerLayer constructor with the left+top parameters being used as default.

This is a good point, and it makes me wonder whether it would be better just to have a more simple implementation where `_computeAlignment` becomes a static public method on `Marker` named `computePixelAligment`. Then we wouldn't need to have any new constructors, and everything would be much simpler. What do you think?

monsieurtanuki:
> This is a good point, and it makes me wonder whether it would be better just to have a more simple implementation where `_computeAlignment` becomes a static public method on `Marker` named `computePixelAligment`. Then we wouldn't need to have any new constructors, and everything would be much simpler. What do you think?

Looks good to me! I've just pushed it.
Perhaps this may deserve a specific example, e.g. where the `MarkerLayer` has an `Alignment` computed by `Marker.computePixelAligment` for U-turn-left icons, and U-turn-left `Marker`s without specified `Alignment`.

josxha (approved):
I personally liked the factory constructor better since it was more self explaining to me.
Seeing that the [relative alignment gets calculated into top/left pixel values](https://github.com/fleaflet/flutter_map/blob/a36c59a92a5b0df6ca87a22792ce91e220d5e58c/lib/src/layer/marker_layer/marker_layer.dart#L51-L55) anyways I wonder if it would be better to make pixel alignment the default and give users the option for relative alignment?

But since non-realtive marker alignment is a much needed feature so I'm down for a static method.

monsieurtanuki:
> I personally liked the factory constructor better since it was more self explaining to me.

I agree with you. It can still be done, in this PR or a next one.
The added value of the public `static` method is that it can be used for both `MarkerLayer`s and individual `Marker`s. Once you've understood it at the low level, you can use it at the high level.

> Seeing that the [relative alignment gets calculated into top/left pixel values](https://github.com/fleaflet/flutter_map/blob/a36c59a92a5b0df6ca87a22792ce91e220d5e58c/lib/src/layer/marker_layer/marker_layer.dart#L51-L55) anyways I wonder if it would be better to make pixel alignment the default and give users the option for relative alignment?

Possible, but I'd like to avoid breaking changes.
Besides, it looks like no solution (`Alignment` vs pixel) is the perfect one, and we may dance tango between both, cf. #1659, several times a year.

> But since non-realtive marker alignment is a much needed feature so I'm down for a static method.

Cool

> Edit: It would be really helpful to hint to this static method and its usage from the default Marker constructor.

I've just added
```dart
/// [alignment] may be computed using [computePixelAlignment].
````

I don't know how it works here: feel free to merge my PR when relevant.

JaffaKetchup (approved):
LGTM, thanks for this :)
I think it's good that we've distilled it down into the most vital component, we can always make the factories at a later point, as you said.

PR #1850, opened 2024-03-12
feat!: added `TileUpdateTransformers.debounce` & reverted #1840

JaffaKetchup:
After some internal consideration, we believe we may have been slightly rushed in merging #1840.  
As I hinted to in my review comment, the underlying infrastrucutre to perform this debouncing mechanism was already available, just little known, relatively undocumented, and not fully complete.

This mechanism is the `TileUpdateTransformer`, a relatively power-user feature that allows the normal stream of `TileUpdateEvent`s emitted one-to-one with `MapEvents` to be transformed into a new stream. A `debounce`ing transformer has been implemented, which has the same effect as #1840, but without the new infrastrucutre that requires maintainence in future.

The example application has also been updated with a new demo page to show this.

The in-code documentation around the transformers has been greatly improved, and the online documentation will also see a new section on these transformers added.

This PR manually reverts all changes from #1840, but allows for the same behaviour with relatively little difference in end-user usage, and improved internal maintainability.

### DIFF
--- a/example/lib/pages/debouncing_tile_update_transformer.dart
+++ b/example/lib/pages/debouncing_tile_update_transformer.dart
@@ -16,8 +16,8 @@ class DebouncingTileUpdateTransformerPage extends StatefulWidget {
 
 class _DebouncingTileUpdateTransformerPageState
     extends State<DebouncingTileUpdateTransformerPage> {
-  int changeEndKeyRefresher = 0;
-  double durationInMilliseconds = 20;
+  int _changeEndKeyRefresher = 0;
+  double _durationInMilliseconds = 20;
 
   @override
   Widget build(BuildContext context) {
@@ -59,12 +59,12 @@ class _DebouncingTileUpdateTransformerPageState
                   ),
                   children: [
                     TileLayer(
-                      key: ValueKey(changeEndKeyRefresher),
+                      key: ValueKey(_changeEndKeyRefresher),
                       urlTemplate:
                           'https://tile.openstreetmap.org/{z}/{x}/{y}.png',
                       userAgentPackageName: 'dev.fleaflet.flutter_map.example',
                       tileUpdateTransformer: TileUpdateTransformers.debounce(
-                        Duration(milliseconds: durationInMilliseconds.toInt()),
+                        Duration(milliseconds: _durationInMilliseconds.toInt()),
                       ),
                     ),
                   ],
@@ -89,17 +89,17 @@ class _DebouncingTileUpdateTransformerPageState
                           ),
                           Expanded(
                             child: Slider.adaptive(
-                              value: durationInMilliseconds,
+                              value: _durationInMilliseconds,
                               onChanged: (v) =>
-                                  setState(() => durationInMilliseconds = v),
+                                  setState(() => _durationInMilliseconds = v),
                               onChangeEnd: (v) =>
-                                  setState(() => changeEndKeyRefresher++),
+                                  setState(() => _changeEndKeyRefresher++),
                               min: 0,
                               max: 500,
                               divisions: 100,
-                              label: durationInMilliseconds == 0
+                              label: _durationInMilliseconds == 0
                                   ? 'Instant/No Debounce'
-                                  : '${durationInMilliseconds.toInt()} ms',
+                                  : '${_durationInMilliseconds.toInt()} ms',
                             ),
                           ),
                         ],
--- a/example/lib/pages/debouncing_tile_update_transformer.dart
+++ b/example/lib/pages/debouncing_tile_update_transformer.dart
@@ -1,0 +1,117 @@
+import 'package:flutter/material.dart';
+import 'package:flutter_map/flutter_map.dart';
+import 'package:flutter_map_example/widgets/drawer/menu_drawer.dart';
+import 'package:flutter_map_example/widgets/notice_banner.dart';
+import 'package:latlong2/latlong.dart';
+
+class DebouncingTileUpdateTransformerPage extends StatefulWidget {
+  static const String route = '/debouncing_tile_update_transformer_page';
+
+  const DebouncingTileUpdateTransformerPage({super.key});
+
+  @override
+  State<DebouncingTileUpdateTransformerPage> createState() =>
+      _DebouncingTileUpdateTransformerPageState();
+}
+
+class _DebouncingTileUpdateTransformerPageState
+    extends State<DebouncingTileUpdateTransformerPage> {
+  int changeEndKeyRefresher = 0;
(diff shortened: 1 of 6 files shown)


lib/src/layer/tile_layer/tile_update_transformer.dart
```diff
@@ -75,4 +98,68 @@ abstract class TileUpdateTransformers {
...
+  /// ['package:stream_transform'](https://pub.dev/documentation/stream_transform/latest/stream_transform/RateLimit/debounce.html),
+  /// with `leading` and `trailing` set `true`.
+  ///
+  /// Also see [throttle].
+  ///
+  /// ---
+  ///
+  /// {@macro tut-ignore_tap}
+  static TileUpdateTransformer debounce(Duration duration) {
+    const leading = true;
```
josxha, line 121: what's the reason to handle leading events?

JaffaKetchup: Good spot, mis-reasoned about that. Resolved.

ReinisSprogis:
Hi. @JaffaKetchup Thanks for reaching out. I just tried out TileUpdateTransformer in my app. I tried to compare both TileUpdateTransformer with 32ms duration and loadingDelay with 32ms duration. I hope I am not biased, but felt like loadingDelay was smoother on zoom and pan. I did deploy both variations multiple times. It is not exactly the same behavior with TileUpdateTransformer. It still was carrying on to load tiles while panning. Still better than without any debounce at all. Felt more like middle ground between completely stopping loading and loading all the time. However before you posted this I had some concerns regarding loadingDelay. I think there still is some unnecessary work done while tiles are not loading if using loadingDelay.  Basically need to look into that all TileLayer and check what functions are running on each map event. And if all functions within TileLayer build method are necessary to run on each map event. 
In conclusion,  feels like loadingDelay is performing better, but need to figure out why and if TileUpdateTransformer can be improved to get better results. But of course final decision is up to you.  I got feeling that something is not right somewhere else. Something is blocking UI thread while tiles are loading. And this is what is causing lag I believe. CPU usage is very high, this could be because how Tiles are actually widgets. Maybe if drawing them as shader textures using Vertices.raw in CustomPainter could reduce CPU usage and lag.


This was tile layer configuration I tested with


```dart
TileLayer(
  urlTemplate:
   'https://tile.openstreetmap.org/{z}/{x}/{y}.png',
   userAgentPackageName: 'app.example.com',
  tileProvider: CancellableNetworkTileProvider(),
   // tileUpdateTransformer: TileUpdateTransformers.throttle(const Duration(milliseconds: 32)),
    tileDisplay: const TileDisplay.instantaneous(),
    loadingDelay: const Duration(milliseconds: 32),
),

```

JaffaKetchup:
I think I understand what you mean, and it's something I overlooked. This debounces on the input `TileUpdateEvent` stream, which is basically the `MapEvent` stream, not the outputted stream. I don't know why they would be this different, but it would explain what you mean, and what you can see in this video.

If you do lots of small gestures, more tiles will load, because there are lots of breaks between map events. However, a fling gesture still will not load all the tiles.

What do others think?

https://github.com/fleaflet/flutter_map/assets/58115698/57dc0b03-8ea7-4e39-971c-a1d14c593513

corepuncher:
Question: Does this debouncer also stall the network requests?

My BIGGEST issue right now is that the app sends out a slew of tile requests for the "intermediate" zoom levels before the zoom is complete. I have confirmed this by watching my server / incoming requests. 

For example, if I start at zoomLevel 4, and slowly (but continuously) zoom in to zoomLevel 8, I will see every request for every zoom level (4,5,6,7,8) show up on my server.

The problem is that this overwhelms my app when I have many TileLayer() loaded, because of the multiplication factor and sheer number of tiles and connections managed. In the end, the device forcibly closes the app, but not sure which aspect of this is the reason.  I know that in devtools, my memory usage does not spike, so it must be something else. Does the cache manager for the tiles have a limit?

JaffaKetchup:
> Does this debouncer also stall the network requests?

It debounces the emission of tile update events. So not network requests directly, but tile loading as a whole - so it should have this effect.

> For example, if I start at zoomLevel 4, and slowly (but continuously) zoom in to zoomLevel 8, I will see every request for every zoom level (4,5,6,7,8) show up on my server.

Theoretically, this should happen less. Note that the cancelling tile provider will also help with this - your server might see every request (dependent on the number of requests and the devices simultaneous connection limit), but the device will not wait for a response, freeing the connection 'slots' for more recent zoom levels.

> Does the cache manager for the tiles have a limit?

The `ImageCache` from Flutter does have a limit. By default, it's "up to 1000 images, and up to 100 MB": https://api.flutter.dev/flutter/painting/ImageCache-class.html.

example/lib/pages/debouncing_tile_update_transformer.dart
```diff
@@ -0,0 +1,117 @@
...
+                    cameraConstraint: CameraConstraint.contain(
+                      bounds: LatLngBounds(
+                        const LatLng(-90, -180),
+                        const LatLng(90, 180),
+                      ),
+                    ),
+                  ),
+                  children: [
+                    TileLayer(
+                      key: ValueKey(changeEndKeyRefresher),
```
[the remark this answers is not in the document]
JaffaKetchup: See https://github.com/fleaflet/flutter_map/blob/debouncing-tile-update-transformer/lib/src/layer/tile_layer/tile_layer.dart#L211.
I had a look at changing this, but it would require re-initialising the tile event stream often. This caused some issues. Happy to look into changing this again if it's important.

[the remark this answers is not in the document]
josxha: How about some simple change like this?
```dart
key: ValueKey('TileLayer-$changeEndKeyRefresher')
```

example/lib/widgets/notice_banner.dart
```diff
@@ -20,6 +20,15 @@ class NoticeBanner extends StatelessWidget {
...
         backgroundColor = const Color(0xFFB8F397);
 
+  const NoticeBanner.informational({
+    super.key,
+    required this.text,
+    required this.url,
+    required this.sizeTransition,
+  })  : icon = Icons.info_outline,
+        foregroundColor = const Color(0xFF072100),
+        backgroundColor = Colors.lightBlue;
```
josxha, line 30: .lightBlue is a bit hard on the eyes, can we use something different like lightBlueAccent please?

Unbenannt

josxha (commented):
Only one more little thing...

josxha (approved):
lgtm 👍🏼

Issue #1669, opened 2023-09-27
[BUG] `tileSize` parameter is broken


bramp:
### What is the bug?

The TileProvider(tileSize: ) argument does not seem to produce desired maps when set to a value different than 256.

[JaffaKetchup](https://github.com/JaffaKetchup) did testing in https://github.com/fleaflet/flutter_map/pull/1668#issuecomment-1738056062 here are the appropriate screenshots that show the issues:

![0101](https://github.com/fleaflet/flutter_map/assets/160627/7880185e-d65a-4053-b1eb-35b862a16099)
![0100](https://github.com/fleaflet/flutter_map/assets/160627/95e6a742-aeeb-4630-adfa-2ca115a5bc02)

These maps are all meant to be centered on England, and you can see that wrapping around the global does not work correctly.

If we also look at MapBox's Static API, a [512 tile](https://api.mapbox.com/styles/v1/mapbox/satellite-v9/tiles/256/5/15/10?access_token=pk.eyJ1IjoiYnJhbXAiLCJhIjoiY2o4b3NnamZnMDhoMjMzcno4M3JrNm4wMSJ9.R8ove2k7wvRVVdEnlTUu4A), is the same as the [256 tile](https://api.mapbox.com/styles/v1/mapbox/satellite-v9/tiles/512/5/15/10?access_token=pk.eyJ1IjoiYnJhbXAiLCJhIjoiY2o4b3NnamZnMDhoMjMzcno4M3JrNm4wMSJ9.R8ove2k7wvRVVdEnlTUu4A) with the same x, y and z, just at a higher resolution. 

### How can we reproduce it?

Checkout the example Retina/TileSize page: https://github.com/fleaflet/flutter_map/pull/1668 or create a map with

```dart
TileLayer(
   tileSize: 512
)
```

### Do you have a potential solution?

Sorry no.

### Platforms

Chrome

### Severity

Erroneous: Prevents normal functioning and causes errors in the console

JaffaKetchup:
Just to note, they are not all necessarily supposed to be centered on England. However, when changing the FM tile size setting, the map does just 'jump' to a different location, which is incorrect.

V3ntus:
Having this issue as well. Another thing is that this changes where markers or polylines for example are rendered. When using a 512 tile server and setting the `tileSize` to 512, my polylines are rendered in the middle of a sea:
[...]

PR #1853, opened 2024-03-13
Tile rendering migration to CustomPainter

ReinisSprogis:
Overview
This pull request introduces a significant optimization to the map rendering process within the Flutter map. By migrating the rendering of map tiles from a Stack widget with Positioned children to a CustomPainter implementation, was achieved a remarkable improvement in performance. Initial tests demonstrate up to a 50% reduction in CPU load, leading to a smoother user experience even without the use of debouncers.

Changes Made
Map Tile Rendering: Transitioned from using Stack and Positioned widgets for displaying map tiles to drawing them directly in a CustomPainter. This approach leverages the same parameters previously used for the Stack implementation, ensuring a straightforward migration with minimal adjustments required.

Next Steps
While the initial experiment has yielded positive results, I recognize the importance of comprehensive testing and validation. The next steps will involve:

Extended Testing: Conducting a broader range of tests to validate performance improvements across various devices and use cases.
Code Review and Refinement: Get feedback from the team to refine the implementation and address any potential issues or improvements.
Documentation: Updating relevant documentation to reflect the new map rendering approach and guide future development.

### DIFF
--- a/lib/src/layer/tile_layer/tile.dart
+++ b/lib/src/layer/tile_layer/tile.dart
@@ -4,6 +4,7 @@ import 'package:flutter/widgets.dart';
 import 'package:flutter_map/flutter_map.dart';
 
 /// The widget for a single tile used for the [TileLayer].
+@Deprecated('Use TileModel instead')
 @immutable
 class Tile extends StatefulWidget {
   /// [TileImage] is the model class that contains meta data for the Tile image.
--- a/lib/src/layer/tile_layer/tile_model.dart
+++ b/lib/src/layer/tile_layer/tile_model.dart
@@ -1,0 +1,36 @@
+import 'dart:math';
+import 'package:flutter/widgets.dart';
+import 'package:flutter_map/src/layer/tile_layer/tile_builder.dart';
+import 'package:flutter_map/src/layer/tile_layer/tile_image.dart';
+
+/// The widget for a single tile used for the [TileLayer].
+/// This replaces the `Tile` widget in the `flutter_map` package making it deprecated.
+/// Previously the `Tile` widget was a `StatefulWidget`. That is not needed with rendering tiles in Canvas.
+class TileModel {
+  /// [TileImage] is the model class that contains meta data for the Tile image.
+  final TileImage tileImage;
+
+  /// The [TileBuilder] is a reference to the [TileLayer]'s
+  /// [TileLayer.tileBuilder].
+  final TileBuilder? tileBuilder;
+
+  /// The tile size for the given scale of the map.
+  final double scaledTileSize;
+
+  /// Reference to the offset of the top-left corner of the bounding rectangle
+  /// of the [MapCamera]. The origin will not equal the offset of the top-left
+  /// visible pixel when the map is rotated.
+  final Point<double> currentPixelOrigin;
+
+  /// The key for the tile.
+  /// I think this is needed to keep track of the tiles and prune them when they are not needed.
+  final ObjectKey key;
+
+  /// Creates a new instance of [Tile].
+  const TileModel(
+      {required this.scaledTileSize,
+      required this.currentPixelOrigin,
+      required this.tileImage,
+      required this.tileBuilder,
+      required this.key});
+}
--- a/lib/src/layer/tile_layer/tile_painter.dart
+++ b/lib/src/layer/tile_layer/tile_painter.dart
@@ -1,0 +1,45 @@
+import 'dart:ui' as ui;
+import 'package:flutter/rendering.dart';
+import 'package:flutter_map/src/layer/tile_layer/tile_model.dart';
+
+///Draws the tile images on the canvas.
+class TilePainter extends CustomPainter {
+  /// The list of tile models.
+  List<TileModel> tiles;
+
+  /// Constructs a `TilePainter` with the given list of `TileModel` objects.
+  TilePainter({required this.tiles});
+
+  @override
+  void paint(Canvas canvas, Size size) {
+    for (final tile in tiles) {
+      //Simplify tile positions and sizes
+      final double left = tile.tileImage.coordinates.x * tile.scaledTileSize -
+          tile.currentPixelOrigin.x;
+      final double top = tile.tileImage.coordinates.y * tile.scaledTileSize -
+          tile.currentPixelOrigin.y;
+      final double width = tile.scaledTileSize;
+      final double height = tile.scaledTileSize;
+      //Draw tiles if they have an image
+      if (tile.tileImage.imageInfo != null) {
+        final ui.Image image = tile.tileImage.imageInfo!.image;
+        final Paint paint = Paint();
+        paint.isAntiAlias = true;
+        paint.filterQuality = FilterQuality
+            .high; //Can be made configurable to reduce quality and get more performance.
+        canvas.drawImageRect(
+          image,
+          Rect.fromLTWH(0, 0, image.width.toDouble(),
+              image.height.toDouble()), // Source rectangle
+          Rect.fromLTWH(left, top, width, height), // Destination rectangle
+          paint,
+        );
+      }
+    }
+  }
+
+  @override
+  bool shouldRepaint(covariant TilePainter oldDelegate) {
+    return oldDelegate.tiles != tiles;
+  }
+}


lib/src/layer/tile_layer/tile_painter.dart
```diff
@@ -0,0 +1,45 @@
...
+  @override
+  void paint(Canvas canvas, Size size) {
+    for (final tile in tiles) {
+      //Simplify tile positions and sizes
+      final double left = tile.tileImage.coordinates.x * tile.scaledTileSize -
+          tile.currentPixelOrigin.x;
+      final double top = tile.tileImage.coordinates.y * tile.scaledTileSize -
+          tile.currentPixelOrigin.y;
+      final double width = tile.scaledTileSize;
+      final double height = tile.scaledTileSize;
```
Alexays, line 22: Can be moved to the if scope to avoid useless compute if imageInfo is null.

ReinisSprogis: Thanks! Well spotted. I adjusted for that.

ReinisSprogis:
By experimenting with Painter I found some problem. It seems like there are sub pixel gap between tiles. I tried to scale each tile by slight amount to get rid of gap, but realized that this is not a good approach. Minimum scale I could apply to get rid of gaps was to scale it by 1.002. But that means on 1000 kilometers error would be 2 kilometers. These gaps is very noticeable on dark theme map and minimally noticeable on light map. I believe this is due to some floating point rounding error. But I am sure it's something we can overcome. 
visible grid
tile gaps

bramp:
These gaps are a known issue with Flutter, see
* https://github.com/flame-engine/flame/issues/1152
* https://github.com/flutter/flutter/issues/137946
* https://github.com/flutter/engine/pull/35981
* https://github.com/flutter/flutter/issues/14288

For lots of discussion. However https://verygood.ventures/blog/solving-super-dashs-rendering-challenges-eliminating-ghost-lines-for-a-seamless-gaming-experience might be the most useful for resolving this.

ReinisSprogis:
Thanks @bramp Seems like just setting isAntiAlias = false resolved it.

ReinisSprogis:
@JaffaKetchup Hi. Any thoughts/progress on this?

lib/src/layer/tile_layer/tile_painter.dart
```diff
@@ -0,0 +1,62 @@
...
+  /// List of [TileModel]s to draw to the canvas
+  List<TileModel> tiles;
+
+  /// Paint to use when drawing each tile
+  ///
+  /// Defaults to [defaultTilePaint].
+  final Paint? tilePaint;
+
+  /// Default [tilePaint]er
+  static Paint get defaultTilePaint => Paint()
```
JaffaKetchup, line 16: Can you explain these choices?

If the input tile size is equal to the output tile size, which should be always AFAIK, then no scaling should ever be applied, as far as I understand. Therefore, there should be no reason to use `FilterQuality`, and no anti-aliasing required I think.

I'm not entirely sure though, just wanted to get a good explanation of why these were chosen.

lib/src/layer/tile_layer/tile_model.dart
```diff
@@ -0,0 +1,39 @@
...
+
+/// Model for tiles displayed by [TileLayer] and [TilePainter]
+class TileModel {
+  /// Controls how this tile is replaced by another tile
+  ///
+  /// See also:
+  ///
+  ///  * [Widget.key]
+  ///  * The discussions at [Key] and [GlobalKey].
+  final ObjectKey key;
```
JaffaKetchup, line 14: I'm not convinced why this is needed? It's never evaluated at any point, and there's no equality methods in this class that this can influence.

(We should probably implement equality here as well.)

lib/src/layer/tile_layer/tile_model.dart
```diff
@@ -0,0 +1,39 @@
...
+
+  /// [TileImage] is the model class that contains meta data for the Tile image.
+  final TileImage tileImage;
+
+  /// The [TileBuilder] is a reference to the [TileLayer]'s
+  /// [TileLayer.tileBuilder].
+  final TileBuilder? tileBuilder;
```
JaffaKetchup, line 21: `TileBuilder` is never evaluated, and has been replaced by `tilePaint`. It should be deprecated and removed from here.

lib/src/layer/tile_layer/tile_painter.dart
```diff
@@ -0,0 +1,62 @@
...
+
+/// Draws [TileModel]s onto a canvas at the correct position
+class TilePainter extends CustomPainter {
+  /// List of [TileModel]s to draw to the canvas
+  List<TileModel> tiles;
+
+  /// Paint to use when drawing each tile
+  ///
+  /// Defaults to [defaultTilePaint].
+  final Paint? tilePaint;
```
JaffaKetchup, line 13: Does `tilePaint` give enough flexibility? `TileBuilder` used to give loads of flexibility, and users could do basically anything they wanted. We might want to consider what users want to do a bit more.

lib/src/layer/tile_layer/tile_layer.dart
```diff
@@ -602,6 +644,9 @@ class _TileLayerState extends State<TileLayer> with TickerProviderStateMixin {
       onLoadError: _onTileLoadError,
       onLoadComplete: (coordinates) {
         if (pruneAfterLoad) _pruneIfAllTilesLoaded(coordinates);
+        setState(() {
+          //Refresh the widget to display the loaded tile
+        });
```
JaffaKetchup, line 649: This feels like a workaround; maybe it isn't though? Why is it required now and not before?

lib/src/layer/tile_layer/tile.dart
```diff
@@ -4,6 +4,7 @@ import 'package:flutter/widgets.dart';
 import 'package:flutter_map/flutter_map.dart';
 
 /// The widget for a single tile used for the [TileLayer].
+/// If proceed with the implementation of the CustomPainter and the TileModel this Widget should be removed.
```
JaffaKetchup, line 7: We should deprecate this.

JaffaKetchup (changes requested):
Hey, sorry for the delay. I think this looks good overall, but there's some things to consider.

lib/src/layer/tile_layer/tile_painter.dart
ReinisSprogis: What I understood is that when zooming in or out it scales current tile until new tiles are loaded, that is why need to scale image to fit that tile. Initially I did not set filterQuality and had very low quality image. Setting anti-aliasing to false removes ghost lines between tiles.

lib/src/layer/tile_layer/tile_model.dart
ReinisSprogis: Looks like `ObjectKey` can be removed and equality implemented.

lib/src/layer/tile_layer/tile_model.dart
ReinisSprogis: Yes

lib/src/layer/tile_layer/tile_painter.dart
ReinisSprogis: I am not sure what would be missing comparing with TileBuilder. User could get image from tile and modify it?

lib/src/layer/tile_layer/tile_layer.dart
ReinisSprogis: Previously each tile rebuilt it's own state when image was loaded. Needed a place to call` setState()` to refresh TilePainter. 
Seems like `onLoadComplete` is appropriate place to call that.

lib/src/layer/tile_layer/tile.dart
ReinisSprogis: I deprecated it initially, but tests didn't pass anymore, as class `_TileState extends State<Tile>`  was using deprecated `Tile` class. So I am not sure how to manage this? Just delete class altogether?

lib/src/layer/tile_layer/tile_painter.dart
JaffaKetchup: Ah, of course, I forgot about zooming! OK, `FilterQuality` makes sense then.

In regards to the thin lines, we used to have issues with those without canvas, because we weren't careful enough when it came to rounding operations. May be worth double checking any positioning/rounding maths. But I suppose it makes sense for now.

lib/src/layer/tile_layer/tile_painter.dart
JaffaKetchup: Modifying an incoming tile's bytes can be done with a `TileProvider`, so there's no gap there - but previously you could 'overlay'/'wrap' widgets, etc. See https://demo.fleaflet.dev/tile_builder.

I understand that positioning widgets over this kind of defeats the point of a canvas, and is one of the reasons we previously didn't use canvas, to allow this kind of easy tweaking. However, we should really provide the ability to overlay widgets at tile positions. Note that it's not possible to paint a widget to a canvas.

lib/src/layer/tile_layer/tile.dart
JaffaKetchup: Deprecation is transitive. Mark both classes as deprecated, and it should work. However, just don't worry about this for now, and I'll do it at the end to make sure it's correct.

JaffaKetchup:
(Just a thought, need to test whether this fixes #1669.)

lib/src/layer/tile_layer/tile_painter.dart
ReinisSprogis: I see. I looked into it, then there is now way to avoid Stack. I suppose could then preserve Stack and Tile approach and use that if `TileBuilder` is not null else use CustomPainter?

JaffaKetchup: Not necessarily. I suppose a good compromise would be to allow a point in the painting process that can be influenced by the user, ie. they can paint to the same canvas, but are constrained to a smaller `Size`, for each tile. Although much less convienient for the user, this will allow as much flexibility as possible.

I guess it's also worth verifying those performance improvements on a few different devices, just to see if we're making a correct decision here, and the pros outweigh the cons, IYSWIM.

lib/src/layer/tile_layer/tile_layer.dart
ReinisSprogis: Now looking deeper into this I noticed that fadeIn animation is running, but not animating opacity.  Perhaps can rebuild state within animation controller listener and set opacity to image in TilePainter?

ReinisSprogis: Opacity is updating, but if there is no rebuild it gets stuck on that opacity level. 
https://github.com/fleaflet/flutter_map/assets/69913791/f12260bf-bd5a-4e4e-9849-dd865896f98b

Alexays: I think an `if (mounted)` should be added in case the onLoadComplete is called after the widget has been destroyed.

lib/src/layer/tile_layer/tile.dart
josxha: `Tile` is not exported publically so it should just get removed with this pr. It got completely replaced it with `TileModel`

lib/src/layer/tile_layer/tile_painter.dart
josxha: tile_builder has been nice to debug tile loading. I wouldn't prefer to have widgets mixed in with the canvas. Most if not everything that has been done with widgets should be possible directly on the canvas too. 
In addition to the `tilePaint` parameter I'd suggest to add a callback that let's users draw custom things on the canvas that gets called after the tiles draws.

Alexays:
Need help with anything to get ahead?

ReinisSprogis:
Hi @Alexays I am bit busy with project I am currently working on. I would love to get back on this, but not sure when. Feel free to take a look what you can improve. Need to figure out how to rebuild painter for Tile fade animation or just take a look what can be improved. I have been using it in production and didn't have any issues so far. So that's a good sign I guess.

JaffaKetchup:
Also just need to sort out the details around the replacement for `TileBuilder`.

JaffaKetchup:
I'm going to try to work on this next, whenever I get time. I think this is a really promising improvement, thanks @ReinisSprogis! Is it OK to work directly on this fork?

@josxha I think providing a callback for users to draw on a small part of the canvas over each tile is an adequate solution as well.

ReinisSprogis:
Hi. Sounds good! I did experimenting some time ago with TileBuilder, and got it somewhat working. If TileBuilder was not null, then I used Stack and stacked it over CustomPainter (TilePainter) otherwise used TilePainter directly. But not sure if this is the best solution.

JaffaKetchup:
To bring this up to date without having to resolve conflicts, and to preserve this branch, I've opened #1908 seperately. I'll add @ReinisSprogis as a collaborator before merging.

PR #1854, opened 2024-03-14
fix: prevent crash when zooming far into `Polygon`s

### LINKED ISSUE #1829: [BUG] High zoom results in increasing frame times when using dotted features
### What is the bug?

Originally reported by @josxha in #1828.

Zooming in on a polygon (a lot) causes a gradual decrease in performance, and can result in a crash when the device can no longer handle it.
It appears to be unrelated to the recent addition of interactivity to polygons.

https://github.com/fleaflet/flutter_map/assets/34318751/e909d301-abce-4e16-9dd8-d953a130f3d9
 
image

image

### How can we reproduce it?

Use the "Polygon Layer" page in the example page, and zoom into a polygon a long way. Frame rate will progressively drop.

_No response_

### Platforms

All

### Severity

Fatal: Causes the application to crash

ReinisSprogis:
fix #1829 
Problem was that as zooming in, path was getting very long and many dots was calculated eventually getting into millions and crashing. This resolves it by drawing points only on visible segment. Directly painting circles rather them on path. Thought this might be better. Any suggestions for improvement are welcome.  Would appreciate some testing as I haven't done much.

### DIFF
--- a/lib/src/layer/polygon_layer/painter.dart
+++ b/lib/src/layer/polygon_layer/painter.dart
@@ -314,9 +314,6 @@ class _PolygonPainter<R extends Object> extends CustomPainter {
       final spacing = polygon.borderStrokeWidth * 1.5;
       _addDottedLineToPath(
           canvas, paint, offsets, borderRadius, spacing, canvasSize);
-      // final Float32List points = Float32List.fromList(offsets.expand((o) => [o.dx, o.dy]).toList());
-      // final vertices = Vertices.raw(VertexMode.triangleFan, points);
-      // canvas.drawVertices(vertices, BlendMode.src, _getBorderPaint(polygon));
     } else {
       _addLineToPath(path, offsets);
     }
--- a/lib/src/layer/polygon_layer/painter.dart
+++ b/lib/src/layer/polygon_layer/painter.dart
@@ -447,7 +447,6 @@ class _PolygonPainter<R extends Object> extends CustomPainter {
     if (offsets.isEmpty) {
       return;
     }
-    int x = 0; // Counter for diagnostics
 
     // Calculate for all segments, including closing the loop from the last to the first point
     final int totalOffsets = offsets.length;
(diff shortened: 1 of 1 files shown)


JaffaKetchup (approved):
Hey @ReinisSprogis,
I think this works great for polygons, thanks. LGTM!

I've also realised that the same issue occurs with polylines, we'll need to apply the same fix there.

ReinisSprogis:
Hi. @JaffaKetchup Yes! I suppose would be better if @monsieurtanuki could look into it as he would navigate better in code he just wrote.

monsieurtanuki:
@ReinisSprogis @JaffaKetchup Sounds reasonable. I have to refactor part of the code into a class.

monsieurtanuki:
> Problem was that as zooming in, path was getting very long and many dots was calculated eventually getting into millions and crashing. This resolves it by drawing points only on visible segment.

Makes sense. Probably relevant specifically for dotted lines, where the dots have little space between them, but should be true for dashed lines too.

> Directly painting circles rather them on path. Thought this might be better.

To be tested indeed about displaying dots: which is better/faster between
* `path.addOval`
* `path.moveTo` / `path.lineTo` on the same spot
* `canvas.drawCircle`

Same performance question for lines (`path.lineTo` vs. `canvas.drawLine`). Would deserve a specific issue.

ReinisSprogis:
> > Problem was that as zooming in, path was getting very long and many dots was calculated eventually getting into millions and crashing. This resolves it by drawing points only on visible segment.
> 
> Makes sense. Probably relevant specifically for dotted lines, where the dots have little space between them, but should be true for dashed lines too.
> 
> > Directly painting circles rather them on path. Thought this might be better.
> 
> To be tested indeed about displaying dots: which is better/faster between
> 
> * `path.addOval`
> * `path.moveTo` / `path.lineTo` on the same spot
> * `canvas.drawCircle`
> 
> Same performance question for lines (`path.lineTo` vs. `canvas.drawLine`). Would deserve a specific issue.

Testing in https://demo.fleaflet.dev/polyline Same lag occurs while zooming in into dashed line and dotted line. Need to draw segment only up to canvas edge.

monsieurtanuki:
@ReinisSprogis @JaffaKetchup I've just found an elegant way to display only what matters.
Not sure if it will be easy to implement all `PatternFit`s though - I may drop one of them (`extendFinalDash`).

Btw, with this new mehod:
* it will be easier to check for performances between `canvas.drawLine` and `path.lineTo`
* but as we don't display that much segments and dots we don't need to run perf tests about it in a first approach

PR #1861, opened 2024-03-27
fix(example): ensure zoom buttons zoom by a full zoom level only

monsieurtanuki:
Removed some bounding controls so that zooms get increased / decreased by 1.


JaffaKetchup (approved):
LGTM, can't see why this wouldn't be better.

PR #1869, opened 2024-04-11
feat: add `BlendMode` support to `CircleLayer`

Y0ngg4n:
Currently overlapping Polygons are adding up in stroke color. To fix this it adds the BlendMode.src to all so this should not happen anymore.
Before:

image

### DIFF
--- a/lib/src/layer/circle_layer/painter.dart
+++ b/lib/src/layer/circle_layer/painter.dart
@@ -103,6 +103,7 @@ class CirclePainter extends CustomPainter {
       for (final radius in pointsByRadius.keys) {
         final pointsByRadiusColor = pointsByRadius[radius]!;
         final radiusPaint = paint..strokeWidth = radius * 2;
+        radiusPaint.blendMode = BlendMode.luminosity;
         _paintPoints(canvas, pointsByRadiusColor, radiusPaint);
       }
     }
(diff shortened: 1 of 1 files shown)


mootw (commented):
do you have an example after picture or some code to test with? Also any particular reason to hard code the BlendMode vs pass it in as a parameter to the layer? Thanks for your contribution!

Y0ngg4n:
@mootw i could not test it because of dependency hell. But it uses the same blend mode as in the polygons so the behaviour should be the same.

lib/src/layer/circle_layer/painter.dart
```diff
@@ -92,6 +93,7 @@ class CirclePainter extends CustomPainter {
       for (final radius in pointsByRadius.keys) {
         final pointsByRadiusColor = pointsByRadius[radius]!;
         final radiusPaint = paint..strokeWidth = radius * 2;
+        radiusPaint.blendMode = BlendMode.src;
```
Alexays, line 96: You should move it to first assignment (paintPoint) to avoid useless re-assignment.

lib/src/layer/circle_layer/painter.dart
```diff
@@ -103,6 +105,7 @@ class CirclePainter extends CustomPainter {
...
```
Alexays, line 108: You can remove it if previous comment resolved.

JaffaKetchup (changes requested):
Hey @Y0ngg4n, thanks for this so far!

Some changes do need to be made before we can merge this though:

 * We will require an example (or an update to the exisiting one), or at least some code to test on this occassion
 * It may also be a good idea to incorporate this into the polygon and polyline layers, although they use some more complex logic, so it may not work very performantly (may require more calls to `saveLayer`). This may be out of the scope of this PR?
 * And as said, it should become a property on both the `CircleLayer` and `CircleMarker`, with the same decision logic as other properties.

JaffaKetchup:
I'm going to close this for now. This is a good option to introduce to all the painting layers, but that needs a new PR.

PR #1871, opened 2024-05-01
fix: correct display of `StrokePattern.solid` at high zoom levels

monsieurtanuki:
### What
* Changed the way to display the solid style segments (polygon and polyline), in order to fix a display side-effect.
* Some performance test should be performed to see if we're not significantly impacted, as we added a pre-computation step with potentially many allocations.
* That fix works only for lines, not for areas.

### Impacted files
* `polygon_layer/painter.dart`: now using the new `SolidPixelHiker` for solid style display
* `polyline_layer/painter.dart`: now using the new `SolidPixelHiker` for solid style display
* `pixel_hiker.dart`: new `SolidPixelHiker` class

### Screenshots
Example on high zoom level: the `path`s are unstable beyond a certain threshold, and as a side effect some painting are corrupted (in that case, the border). 
| before | after |
| -- | -- |
| Screenshot_1714549210 | Screenshot_1714555249 |

### DIFF
--- a/lib/src/layer/misc/line_patterns/pixel_hiker.dart
+++ b/lib/src/layer/misc/line_patterns/pixel_hiker.dart
@@ -237,6 +237,45 @@ class DashedPixelHiker extends _PixelHiker {
   }
 }
 
+/// Pixel hiker that lists the visible solid segments to display on the way.
+@internal
+class SolidPixelHiker extends _PixelHiker {
+  /// Standard Solid Pixel Hiker constructor.
+  SolidPixelHiker({
+    required super.offsets,
+    required super.closePath,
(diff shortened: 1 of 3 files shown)


josxha:
Thanks for the pull request, lgtm!

josxha (approved):

monsieurtanuki:
Thank you @josxha for your review!

We still need:
* to create a new issue about polygon areas not being correctly displayed in high zoom levels (25+) because of flutter limitations (path coordinates > 64M), and I have an algorithm for that
* to check the impact on performances - on one hand we allocate memory for pre-computations, on the other hand we display smaller polylines and polygon areas

Issue #1864, opened 2024-04-07
[BUG] v7-beta: `Polyline` becomes distorted & invisible (but remains interactive) when zoomed far in


RuoLiDevelop:
### What is the bug?

https://demo.fleaflet.dev/polyline
At zoom maximum
Hover over polyline while scrolling the mouse wheel
Display exception occurs

### How can we reproduce it?

![CF6D31E2-98B0-4E33-8150-01397AD5E816](https://github.com/fleaflet/flutter_map/assets/47424927/836f5d8c-ba51-453d-ae2f-52ffc0c9f580)

### Platforms

macOs chrome 

### Severity

Minimum: Allows normal functioning

RuoLiDevelop:
[screen-capture.webm](https://github.com/fleaflet/flutter_map/assets/47424927/2f7addd5-c17d-41df-9e98-26b0e9941fdf)

JaffaKetchup:
Hey @RuoLiDevelop,
Thanks for reporting this. I don't believe any exception occurs, as the polyline remains interactive the entire time, which means it is being painted, but there's definitely something strange going on: as you showed, it changes thickness, looks very strange, then disappears.
Zooming out does reintroduce the line correctly.
I can only assume there's some maths we're doing somewhere that doesn't like the precision required at this zoom level? Or the new code to cull polyline segments that are out of the viewport is not working quite right at this zoom level?

RuoLiDevelop:
Hi @JaffaKetchup 
Thank you for your reply. I'm sorry, but for some reason, I haven't delved too deeply into the Flutter_map code. In theory, limiting the zoom level can solve this problem.

JaffaKetchup:
No worries, no need to delve into the code unless you want to!

monsieurtanuki:
> Or the new code to cull polyline segments that are out of the viewport is not working quite right at this zoom level?

@JaffaKetchup I've definitely noticed side-effects at high zoom levels, where segments are being dismissed.
@RuoLiDevelop Would you test with `cullingMargin: null`?

RuoLiDevelop:
hi @monsieurtanuki 
I don't understand. ` cullingMargin: null`

RuoLiDevelop:
I noticed minification zoom level,create line segments between unrelated polygons when the distance between polygons is long.

RuoLiDevelop:
> I noticed minification zoom level,create line segments between unrelated polygons when the distance between polygons is long.

[...]

PR #1875, opened 2024-05-06
fix: 1874 - fixed side-effect crash for final dot / final dash

monsieurtanuki:
Fixed a side-effect crash for both final dot and final dash.
Still, the display may look funny on final dash mode with `PatternFit.extendFinalDash`, but that's probably how it's supposed to work (e.g. not well at the end in special cases).

### DIFF
--- a/lib/src/layer/misc/line_patterns/pixel_hiker.dart
+++ b/lib/src/layer/misc/line_patterns/pixel_hiker.dart
@@ -52,9 +52,13 @@ class DottedPixelHiker extends _PixelHiker {
     // side-effect of the last dot
     if (!closePath) {
       if (patternFit != PatternFit.none) {
-        final last = result.last;
-        if (last != offsets.last) {
+        if (result.isEmpty) {
           addVisibleOffset(offsets.last);
+        } else {
+          final last = result.last;
+          if (last != offsets.last) {
+            addVisibleOffset(offsets.last);
+          }
         }
       }
     }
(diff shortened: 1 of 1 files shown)


josxha:
Thanks a lot for this quick fix, @monsieurtanuki! I can confirm that this fixes the bug.

josxha (approved):

monsieurtanuki:
Thank you @josxha for your review and for having put enough details in the issue (video + log) for a quick fix.

Still a bit worried about possibly slower performances of the new dash/dot/solid algorithm: is there a sort of unit test dedicated to polygon performances?

There may also be minor interferences between culling and dash/dot algorithm: if geopoints are preventively dismissed as "too far away", the dash pattern change (e.g. as starting from a different geopoint).

JaffaKetchup:
The best way to test performance is to build in release or profile, then use the Polygon/Polylines Stress Test in the example app. Disable "Performant Rendering", and try changing the thickness of the line. Then compare the performance graphs at the bottom (not available on web) to an older version.

monsieurtanuki:
From data I've just gathered, the previous code is slightly better in extreme conditions, but the current PR is similar in normal conditions.

Optimization tracks (beyond this PR):
* use the "natural" paths for polylines with decent coordinates (below 64M) and switch to the precomputed version beyond
* the polyline cropping I've already coded for [osmdroid](https://github.com/osmdroid/osmdroid) (cf. https://github.com/fleaflet/flutter_map/issues/1864#issuecomment-2090337203)

But I would recommend merging the current PR quickly, as it fixes a crash for dashed and dotted

| x | latest PR | previous code |
| -- | -- | -- |
| optim | Screenshot_20240507_103052 | Screenshot_20240507_104342 |
| no optim |  Screenshot_20240507_103119 | Screenshot_20240507_104403 |

josxha:
> But I would recommend merging the current PR quickly, as it fixes a crash for dashed and dotted

Agree 👍🏼

PR #1876, opened 2024-05-07
feat: optimization of "solid" polygon/polyline display

monsieurtanuki:
### What
Optimized the algorithm for "solid" polygon/polyline display.
The trick is to use `addPolygon` instead of numerous `moveTo`/`lineTo`. It looked like removing redundant `moveTo`s was quite effective, too.

Following are some perf tests. We're getting closer to the old solid algorithm (that we had to dismiss because of bugs in 25+ zooms).
| x | old solid algo | new solid algo | current PR |
| -- | -- | -- | -- |
| Raster (ms) | 206.2 | 578.9 | 208.6 |
| UI (ms) | 42.1 | 53.2 | 49.6 |

### Impacted files
* `polygon_layer/painter.dart`: minor refactoring
* `polyline_layer/painter.dart`: minor refactoring
* `pixel_hiker.dart`: optimized the "solid pixel hiker" relying on `addPolygon` instead of numerous `moveTo`/`lineTo`

### Screenshots
Screenshot_20240507_160145

### DIFF
--- a/lib/src/layer/misc/line_patterns/pixel_hiker.dart
+++ b/lib/src/layer/misc/line_patterns/pixel_hiker.dart
@@ -250,12 +250,24 @@ class SolidPixelHiker extends _PixelHiker {
           patternFit: PatternFit.none,
         );
 
-  /// Returns all visible segments.
-  List<VisibleSegment> getAllVisibleSegments() {
-    final List<VisibleSegment> result = [];
-
+  /// Adds all visible segments to [paths].
+  void addAllVisibleSegments(final List<Path> paths) {
     if (offsets.length < 2) {
-      return result;
+      return;
+    }
+
+    double? latestX;
+    double? latestY;
+    List<Offset> polygons = [];
+
+    void addPolygons() {
+      if (polygons.isEmpty) {
+        return;
+      }
+      for (final path in paths) {
+        path.addPolygon(polygons, false);
+      }
+      polygons = [];
     }
 
     for (int i = 0; i < offsets.length - 1 + (closePath ? 1 : 0); i++) {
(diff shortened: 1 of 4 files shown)


JaffaKetchup (changes requested):
Hey @monsieurtanuki,

Thanks for continuing to work on this! I've been testing this PR and have noticed that there's an issue on the master branch and on this PR, that could be related. Not sure whether it's worth fixing in a different PR, or this PR, but needs to be fixed for sure.

image

It looks like the culling algortihm no longer accounts for the radius of the line, so it's possible to see the radius of the curve at the edges of the viewport.

monsieurtanuki:
Hi @JaffaKetchup!

Probably not related to culling, but to drawing segments clipped to screen rectangle instead of a continuous polygon.
Could be fixed by pretending that the screen is slightly bigger: an extra half radius on each side of the clipping screen rectangle.

monsieurtanuki:
Hi @JaffaKetchup!
I've just fixed the issue, that was more obvious for solid lines but impacted dashed and dotted lines too.
As expected, the idea was just to take into account the stroke width when we clip segments into the screen.
Instead of clipping into
  `[0, 0, canvas.width, canvas.height]`
we have to clip into
  `[-strokeWidth/2, -strokeWidth/2, canvas.width+strokeWidth/2, canvas.height/strokeWidth/2]`.
Et voilà!

| x | before | after |
| -- | -- | -- |
| polyline | Screenshot_1716189708 | Screenshot_1716189734 |
| polygon |  Screenshot_1716189913 | Screenshot_1716189938 |

JaffaKetchup:
I'll try to review this tomorrow :)

JaffaKetchup (approved):
LGTM I think! Thanks once again :)

monsieurtanuki:
Thank you @JaffaKetchup for your reviews and feedbacks!

PR #1879, opened 2024-05-09
fix: `PolylineLayer` throws exception: "The west longitude can't be smaller than the east longitude"

josxha:
Some linestrings caused the PolylineLayer to throw an assertion exception.
```console
Assertion failed: file:///Users/matthiaskoch/.pub-cache/git/flutter_map-930e8b60e125e630e3c85ab373c0ebb8e9b08d70/lib/src/geo/latlng_bounds.dart:82:16
east >= west
"The west longitude can't be smaller than the east longitude"

packages/flutter_map/src/geo/latlng_bounds.dart 82:28                             unsafe
packages/flutter_map/src/layer/polyline_layer/polyline_layer.dart 171:41          [_aggressivelyCullPolylines]
packages/flutter_map/src/layer/polyline_layer/polyline_layer.dart 138:11          build
```

This bug was reported by borg_1of3 on discord: https://discord.com/channels/951867686378409984/1235305884620689558/1235305884620689558

The bug sneaked into the code base in https://github.com/fleaflet/flutter_map/pull/1834.

### DIFF
--- a/lib/src/geo/latlng_bounds.dart
+++ b/lib/src/geo/latlng_bounds.dart
@@ -6,6 +6,18 @@ import 'package:vector_math/vector_math_64.dart';
 /// Data structure representing rectangular bounding box constrained by its
 /// northwest and southeast corners
 class LatLngBounds {
+  /// minimum latitude value, south
+  static const double minLatitude = -90;
+
+  /// maximum latitude value, north
+  static const double maxLatitude = 90;
+
+  /// minimum longitude value, west
+  static const double minLongitude = -180;
+
+  /// maximum longitude value, east
+  static const double maxLongitude = 180;
+
   /// The latitude north edge of the bounds
   double north;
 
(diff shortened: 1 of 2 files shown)


mootw (approved):
lgtm

fmt-Println-MKO:
thanks for the quick fix :)

josxha:
@mootw thanks for the review and @monsieurtanuki thanks for the comment. I added the requested refactoring, want to take another look before we merge?

fmt-Println-MKO (approved):

lib/src/geo/latlng_bounds.dart
```diff
@@ -61,21 +73,21 @@ class LatLngBounds {
     required this.south,
     required this.east,
     required this.west,
-  })  : assert(
-            north <= 90, "The north latitude can't be bigger than 90: $north"),
-        assert(north >= -90,
+  })  : assert(north <= maxLatitude,
+            "The north latitude can't be bigger than 90: $north"),
```
monsieurtanuki, line 77:
```suggestion
            "The north latitude can't be bigger than $maxLatitude: $north"),
```

fmt-Println-MKO (approved):

lib/src/geo/latlng_bounds.dart
josxha: Thanks, you're absolutely right, the const fields should be used here too. 👍

JaffaKetchup (approved):
LGTM! Should also make it easier for external uses (for example where a `Polygon` is painted over the entire world, and a hole created at the desired location).

PR #1883, opened 2024-05-15
fix: remove deprecated members (for Flutter 3.22)

josxha:
> info: 'background' is deprecated and shouldn't be used. Use surface instead. This feature was deprecated after v3.18.0-0.1.pre. (deprecated_member_use at [flutter_map_example] lib\pages\polygon_perf_stress.dart:99)


JaffaKetchup (approved):
LGTM!

PR #1886, opened 2024-05-19
feat: added interactivity to `CircleLayer` & refactored interactivity out into seperate classes

JaffaKetchup:
Uses same framework as in `PolylineLayer`/`PolygonLayer`, but less complex (no pre-projection).

Additionally, to reduce duplication and make it easier to change behaviour in future, all these layers now use a single system, and are only made responsible for hit testing a single element (via an implemented method invoked by the shared `hitTest` implementation).

- [x] Basic implementation
- [x] Example
- [x] Testing


josxha (commented):
Could it be that the order of circles is reversed?
image


The only other thing I'd preferably change is the name of the `layer/misc` directory since it's now more a `layer/common` or `layer/shared` directory.
Other than that it looks good to me. Did some quick testing and all works like intended. Good job👍

JaffaKetchup:
I believe it is working correctly. Like a `Stack`, the element at the bottom of the list is rendered on top, and this is reflected in the hit dialog in my testing.

I've renamed the directory to 'shared' :)

example/lib/pages/circle.dart
```diff
@@ -1,49 +1,187 @@
...
+
+              if (listEquals(hitValues, _prevHitValues)) return;
+              _prevHitValues = hitValues;
+
+              final hoverCircles = hitValues.map((v) {
+                final original = _circles[v]!;
+
+                return CircleMarker<HitValue>(
+                  point: original.point,
+                  radius: original.radius + (15 / 2) - 1,
```
josxha, line 90: Is this `(15 / 2) - 1` calculation only to set the hover circles bigger than the regular circles?

JaffaKetchup: Yep, just shifts the hover outline out (so its 15px out, not 7.5 px out & 7.5 px in). Just a visual thing, can remove if you want.

josxha:
```suggestion
                  radius: original.radius + 6.5,
```

JaffaKetchup: I would perhaps not suggest doing this. It will be optimized out by the compiler, there's no extra cost, and it kinda makes it a little bit more clear what's going on leaving it extended. But I don't really care that much :D

josxha: True, but because this is our example app where people can view example implementations I prefer to have it's implementation clean. (:

josxha (approved):
lgtm, thanks!

PR #1897, opened 2024-05-30
fix: downgraded some dependencies to re-introduce compatbility with Flutter 3.19

monsieurtanuki:
Slightly downgraded package dependencies so that it works in flutter 3.19.

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -34,7 +34,7 @@ dependencies:
   http: ^1.2.1
   latlong2: ^0.9.1
   logger: ^2.3.0
-  meta: ^1.12.0
+  meta: ^1.11.0
   polylabel: ^1.0.1
   proj4dart: ^2.1.0
   vector_math: ^2.1.4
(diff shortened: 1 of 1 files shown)


JaffaKetchup:
Thanks again :) However, I'm just wondering whether we should bump the minimum SDK version to 3.19, or further downgrade the depedency constraints, because I guess this could still cause issues for those on a version prior to 3.19.

monsieurtanuki:
> Thanks again :) However, I'm just wondering whether we should bump the minimum SDK version to 3.19, or further downgrade the depedency constraints, because I guess this could still cause issues for those on a version prior to 3.19.

@JaffaKetchup The idea for a public package like `flutter_map` would be to let the user configurations decide what is best by using low version numbers and the `^` syntax, except for versions that are really needed (e.g. bug fixes) or versions that are highly recommended (e.g. `latlong2: ^0.9.1`)
That would mean sort of rollbacking your previous `pubspec.yaml`, right?

JaffaKetchup (approved):
This is fine for now, we can downgrade all the dependencies to their latest major version (or later if we need functionality in a minor/patch) later. Thanks 👍

PR #1904, opened 2024-06-03
feat: implement efficient(-ish) change detection for `PolygonLayer` & `PolylineLayer`

### LINKED ISSUE #1894: [BUG] `Polygon`s and `Polyline`s don't update when `points` modified
### What is the bug?

After adding new points to polygon, it's not updating automatically. You should explicitly add UniqueKey to Polygon widget. 

### How can we reproduce it?

To reproduce you can view min sample in [github](https://github.com/tologonkudaiberdiuulu/map_draw). To solve you can uncomment "// key: UniqueKey()," part in map_page.dart file

No

### Platforms

All

### Severity

JaffaKetchup:
The main aim of this change is to enable each element to update independently only when required.

The pre-projection and pre-simplification logic have been refactored into their own state which can be mixed in (and is to `PolylineLayer` and `PolygonLayer`).

Performance results:
 - There is little overall change (perhaps +<1ms on UI/build)
 - On the stress test, update times have not changed much, but may have increased a bit - as expected
 - It is expected that in many cases, this will improve efficiency of updates, because only the elements that have updated will need to be re-projected and re-simplified
 - TLDR; there should be some overall noticeable change, except for extremely large samples where all elements change at once, where the behaviour can be manually controlled instead (see below) - middle-size and smaller samples where only a couple elements change, or normal usage, should be better off (also see below)

Additionally, this fixes #1894. Unless `useDynamicUpdate` is disabled, in which case, `Key`s have been recommended to refresh the widget manually. This approach has been recommended in some cases: see the in-code docs.

The default `simplificationTolerance` has also been lowered to 0.3. This gives perfectly acceptable performance (very similar to 0.5 as before), but reduced visual artifacts.

### DIFF
--- a/lib/src/layer/shared/layer_projection_simplification/widget.dart
+++ b/lib/src/layer/shared/layer_projection_simplification/widget.dart
@@ -5,7 +5,8 @@ import 'package:flutter_map/src/layer/shared/layer_projection_simplification/sta
 /// A [StatefulWidget] that includes the properties used by the [State] component
 /// which mixes [ProjectionSimplificationManagement] in
 @immutable
-abstract base class PSSupportedWidget extends StatefulWidget {
+abstract base class ProjectionSimplificationManagementSupportedWidget
+    extends StatefulWidget {
   /// Whether to apply the auto-update algorithm to re-paint the necessary
   /// [Polygon]s when they change
   ///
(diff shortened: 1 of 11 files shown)


lib/src/layer/shared/layer_projection_simplification/widget.dart
```diff
@@ -0,0 +1,48 @@
+import 'package:flutter/material.dart';
+import 'package:flutter_map/flutter_map.dart';
+import 'package:flutter_map/src/layer/shared/layer_projection_simplification/state.dart';
+
+/// A [StatefulWidget] that includes the properties used by the [State] component
+/// which mixes [ProjectionSimplificationManagement] in
+@immutable
+abstract base class PSSupportedWidget extends StatefulWidget {
```
josxha, line 8: Not a huge acronyms for class names myself and would prefer the long variant because it's not really clear without reading the documentation what this class is for. But you can keep it that way if you prefer this more.

josxha (approved):

PR #1907, opened 2024-06-05
perf: expose `isPointInPolygon` and make 40% faster (at least in JIT mode)

ignatz:
I mostly looked at the function because I wanted to use it myself but it was private. So I thought, if I expose it then I could at least put a ribbon on it by adding a benchmark and make it go wroom.

Before:

  (duration: 0:00:05.998949, name: In circle)
  (duration: 0:00:06.866919, name: Not in circle)

After:

  (duration: 0:00:03.649496, name: In circle)
  (duration: 0:00:04.611599, name: Not in circle)

Note, I opportunistically touched crs to remove the dart:ui dependency. This way it can be compiled with dart (w/o flutter) rendering the instructions in the benchmark correct again. Unfortunately, pointInPolygon is not so fortunate and needs flutter due to the dependency on ui.Offset. That's why I could only run it with "flutter test" in JIT mode.

### DIFF
--- a/benchmark/point_in_polygon.dart
+++ b/benchmark/point_in_polygon.dart
@@ -1,0 +1,75 @@
+import 'dart:async';
+import 'dart:math' as math;
+import 'dart:ui';
+
+import 'package:flutter_map/src/misc/point_in_polygon.dart';
+import 'package:logger/logger.dart';
+
+class NoFilter extends LogFilter {
+  @override
+  bool shouldLog(LogEvent event) => true;
+}
+
+typedef Result = ({
+  String name,
+  Duration duration,
+});
+
+Future<Result> timedRun(String name, dynamic Function() body) async {
+  Logger().i('running $name...');
+  final watch = Stopwatch()..start();
+  await body();
+  watch.stop();
+
+  return (name: name, duration: watch.elapsed);
+}
+
+// NOTE: to have a more prod like comparison, run with:
+//     $ dart compile exe benchmark/crs.dart && ./benchmark/crs.exe
+//
+// If you run in JIT mode, the resulting execution times will be a lot more similar.
+Future<void> main() async {
+  Logger.level = Level.all;
+  Logger.defaultFilter = NoFilter.new;
+  Logger.defaultPrinter = SimplePrinter.new;
+
+  final results = <Result>[];
+  const N = 3000000;
+  const points = 1000;
+
+  results.add(await timedRun('In circle', () {
+    final polygon = List.generate(points, (i) {
+      final angle = math.pi * 2 / points * i;
+      return Offset(math.cos(angle), math.sin(angle));
+    });
+
+    const point = math.Point(0, 0);
+
+    bool yesPlease = true;
+    for (int i = 0; i < N; ++i) {
+      yesPlease = yesPlease && pointInPolygon(point, polygon);
+    }
+
+    assert(yesPlease, 'should be in circle');
+    return yesPlease;
+  }));
+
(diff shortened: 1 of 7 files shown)


benchmark/point_in_polygon.dart
```diff
@@ -0,0 +1,75 @@
...
+      final angle = math.pi * 2 / points * i;
+      return Offset(math.cos(angle), math.sin(angle));
+    });
+
+    const point = math.Point(0, 0);
+
+    bool yesPlease = true;
+    for (int i = 0; i < N; ++i) {
+      yesPlease = yesPlease && pointInPolygon(point, polygon);
+    }
```
JaffaKetchup, line 51: Is there any reason we don't do an early break here (and same below)?

lib/src/misc/point_in_polygon.dart
```diff
@@ -0,0 +1,25 @@
+import 'dart:math' as math;
+import 'dart:ui';
+
+/// Checks whether point [p] is within the specified closed [polygon]
+///
+/// Uses the even-odd algorithm.
+bool pointInPolygon(math.Point p, List<Offset> polygon) {
```
JaffaKetchup, line 7: Nit: should probably be called `isPointInPolygon`, but doesn't really matter.

JaffaKetchup (commented):
LGTM, apart from one thing. Thanks!
In general in the past we've tried to avoid becoming a toolbox for these kinds of things, but seeing as we've recently exposed the simplification tools, I don't really have an issue this.

lib/src/layer/polygon_layer/painter.dart
```diff
@@ -39,42 +39,41 @@ base class _PolygonPainter<R extends Object>
...
+            final projectedHoleCoords = getOffsetsXY(
               camera: camera,
               origin: hitTestCameraOrigin,
               points: points,
-            ).toList(),
-          )
-          .toList();
+            );
 
-      if (projectedHoleCoords.firstOrNull != projectedHoleCoords.lastOrNull) {
```
ignatz, line 68: I don't think this does what the author intended. This pushes a list and not a coordinate

ignatz:
Sorry, I tagged on a bunch of other changes after looking a bit more at the implementation. Added a few comments. Hope it makes sense.

lib/src/layer/polygon_layer/painter.dart
```diff
@@ -39,42 +39,41 @@ base class _PolygonPainter<R extends Object>
...
-    }
+            if (projectedHoleCoords.first != projectedHoleCoords.last) {
+              projectedHoleCoords.add(projectedHoleCoords.first);
+            }
 
-    final isInPolygon = _isPointInPolygon(point, projectedCoords);
-    final isInHole = hasHoles &&
-        projectedHoleCoords
-            .map((c) => _isPointInPolygon(point, c))
-            .any((e) => e);
```
ignatz, line 77: Doing the any so late just means that we're doing more work than needed if if any of the holes other than the last happens to yield true.

JaffaKetchup: I'm not sure this is true? `map` outputs a lazy `Iterable`, so I would expect `any` to lazily consume it value by value, and exit early as necessary. But I have no idea. Can we do a small benchmark?

ignatz: Nobody said we can't both be right. You're right but you're still building the entire `projectedHoleCoords` upfront, which may not be needed. This is no longer the caase.

JaffaKetchup (approved):
LGTM. Thanks :)

PR #1885, opened 2024-05-16
chore(example): added WASM support to example

JaffaKetchup:
Supersedes #1867 by @josxha.

- Reset web index.html config file to initial state in Flutter 3.22
- Also adds required headers to Firebase config (hopefully)

Would like to add an indicator using dart-lang/sdk/issues/55694 when it becomes available, to confirm usage of WASM.

PR #1909, opened 2024-06-07
 update Flutter initialization script to use flutter_bootstrap.js

ReinisSprogis:
As per https://docs.flutter.dev/platform-integration/web/bootstrapping


JaffaKetchup:
We're planning for #1885, just waiting on some stuff. Thanks anyway!

Issue #1892, opened 2024-05-27
[BUG] Map Controller shows an assert when in a ListView


Zverik:
### What is the bug?

I have a FlutterMap in a ListView. It displays fine. When I scroll down so the map is not visible (and the state is lost), and then scroll back up, the map is broken (at least in the debug build), and there is an assertion in the console: "Should not update options unless they change", originating from `map_controller_impl.dart:342`.

I guess the widget gets deleted when not in view, then built again, and the same mapController gets reattached. Is that correct?

### How can we reproduce it?

Sorry I don't have a short code at hand, my laptop acts up. I encountered this when running Every Door: find a POI and tap on it, and then scroll down, opening "more fields", to the bottom. And then up. [Here](https://github.com/Zverik/every_door/blob/main/lib/screens/editor.dart#L537) is how the object created.

### Do you have a potential solution?

Remove the assertion?..

### Platforms

Android 14 (Sony Xperia 10 IV).

### Severity

Erroneous: Prevents normal functioning and causes errors in the console

JaffaKetchup:
Thanks for reporting. This is very similar to #1760. However, it is occuring here under new conditions.

Zverik:
Well it worked with version 6.1.0, but broke with 7.0.0.

Tested with a production build: the map disappears, and there is an exception in the log:

```
[07:35:32] Flutter: LateInitializationError: Field '_interactiveViewerState@832162146' has already been initialized.
[07:35:32] #0      LateError._throwFieldAlreadyInitialized (dart:_internal-patch/internal_patch.dart:185)
#1      MapControllerImpl._interactiveViewerState= (package:flutter_map/src/map/controller/map_controller_impl.dart)
#2      MapControllerImpl.interactiveViewerState= (package:flutter_map/src/map/controller/map_controller_impl.dart:47)
#3      MapInteractiveViewerState.initState (package:flutter_map/src/gestures/map_interactive_viewer.dart:103)
#4      StatefulElement._firstBuild (package:flutter/src/widgets/framework.dart:5618)
#5      ComponentElement.mount (package:flutter/src/widgets/framework.dart:5463)
...
```

JaffaKetchup:
That is something we weren't expecting to see. I wasn't aware that anything had changed in that area of the code.

JaffaKetchup:
[...]

PR #1915, opened 2024-06-13
feat: allow reassignment of `MapController` to `FlutterMap` multiple times

josxha:
### Description

Allows the MapController to get assigned multiple times to a view (only one a every time).
Can't tell much about potential side effects. Did some quick tests and everything worked alright so far but additional testing would be good.

### Issue
- https://github.com/fleaflet/flutter_map/issues/1892

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ Please consider [donating](https://docs.fleaflet.dev/supporters#support-us) or [
 
 This CHANGELOG does not include every commit and/or PR - it is a hand picked selection of the most important ones. For a full list of changes, please check the GitHub repository releases/tags.
 
-## [7.0.2] - 2024/07/XX
+## [7.0.2] - 2024/07/02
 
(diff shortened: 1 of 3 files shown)


pagetronic:
Perfect form me.

JaffaKetchup (approved):
LGTM I think. Thanks :)

Tbh I'm not 100% confident that all of the users experiencing this issue are using the `MapController` as per the documentation in the first place, especially when external state is involved, users seem to commonly get the setup wrong by initialising/constructing/defining the `MapController` in the state initially, then passing that to the map, rather than the reverse: https://docs.fleaflet.dev/usage/programmatic-interaction/external-custom-controllers#usage-within-a-state-system-model. Therefore, I'm not planning to change the documentation for this: I'm not sure how this will affect the need for those instructions, but they should ensure more stability than other options for the future.

josxha:
> Tbh I'm not 100% confident that all of the users experiencing this issue are using the MapController as per the documentation in the first place, especially when external state is involved

Yes, I agree. All the reported issues that I've seen on this were caused by incorrect usage of the MapController. But anyways, now we have unofficial support for those implementations. 👍

Thanks for the code review (:

PR #1922, opened 2024-06-24
chore: relax dependency on 'package:logger'

monsieurtanuki:
Unless we have a strong need for a higher version of `logger`, we can lower the dependency to `^2.0.0`
* `2.0.0` having been released in July 2023
* `2.3.0` being the latest version (May 2024)

It's used only for debug or benchmarks.


JaffaKetchup (approved):
Thanks :)

PR #1925, opened 2024-06-27
fix: fixed multiple issues in `PolygonLayer` and `PolylineLayer`

### LINKED ISSUE #1921: [BUG] Unstable `PolygonLayer` & `PolylineLayer`
### What is the bug?

The users clicks min. 4 points on the map to draw an area.
The polygon will not be drawn at all zoom levels and cannot be deleted.
(This worked fine on flutter map version 6.0.1, but no more on 7.0.0/7.0.1 .

### How can we reproduce it?

I have uploaded the code for reproduction [Link to github](https://github.com/IAmHereAtGithub/try_flutter_map/tree/master).
I recorded the behavior in this video:

https://github.com/fleaflet/flutter_map/assets/89143144/292bfada-6b12-49e1-8736-de4baa7c17d4

Manual reproduction:
1. Run the project on an android emulator.
2. Click on the pen icon to get in the edit mode.
3. Draw an area by clicking min. 4 times in the map.
4. Click on the pen icon to get out of the edit mode.
5. You will see no polygon between the circles (bug).
6. Zoom out and/or zoom in, the polygon will be shown.
7. Click on the X on the right bottom button, this should clear the circle and polygon lists.
8. The polygon (depending on the zoom level) will not be cleared and will stay forever.

I tried to debug flutter map. It looks like that in the state.dart , the _cachedSimplifiedElements don´t recognized that a new polygon is created in the zoom level.
And general the _cachedProjectedElements will not getting updated if the user cleared the polygon.

### Platforms

e.g. Android emulator Pixel 7

### Severity

### LINKED ISSUE #1924: [BUG] Holes in `Polygon`s do not always appear
### What is the bug?

The points of holes in `Polygon`s must be specified in a winding direction opposite to the winding direction used to specify the normal points - otherwise, a hole is not cut.

The bug was introduced in #1901.

### How can we reproduce it?

Reverse the order of the inner hole points in the example application.

image

Using proper `Path.combine` techniques in the painter appears to fix the issue:

```dart
filledPath = Path.combine(
	PathOperation.xor,
	filledPath,
	Path()..addPolygon(holeOffsets, true),
);
```

However, this is not supported on the Web HTML backend: https://github.com/flutter/flutter/issues/44572.

Instead, the following line works, but was inadvertantely removed and the introduced bug was not found in testing:

```dart
filledPath.fillType = PathFillType.evenOdd;
```

### Platforms

All

### Severity

JaffaKetchup:
Fixes #1924 and fixes #1921.

Effectively reverts some broken commits.

Note that `useDynamicUpdate` has been deprecated, but its functionality has changed - it now has no purpose. However, this will not be a breaking change, as `useDynamicUpdate` was broken when initially introduced, so there should be no legitimate users.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,16 @@ Please consider [donating](https://docs.fleaflet.dev/supporters#support-us) or [
 
 This CHANGELOG does not include every commit and/or PR - it is a hand picked selection of the most important ones. For a full list of changes, please check the GitHub repository releases/tags.
 
+## [7.0.2] - 2024/07/XX
+
+> Note that this version causes a technically breaking change by removing `PolygonLayer.useDynamicUpdate` & `PolylineLayer.useDynamicUpdate`, introduced in v7.0.1. However, the implementations for these was broken on introduction, and their intended purpose no longer exists. Therefore, these should not have been used in any capacity, and should not affect any projects.
+
+Contains the following user-affecting bug fixes:
+
+- Fixed significant performance issues with `PolygonLayer` & `PolylineLayer` inadvertently introduced by v7.0.1 - [#1925](https://github.com/fleaflet/flutter_map/pull/1925) for [#1921](https://github.com/fleaflet/flutter_map/issues/1921)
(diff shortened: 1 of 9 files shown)


mootw (approved):
looks good. I think we should just remove the value outright and cause a break. given the feature shipped broken, i only see depreciating it as changing its behavior unexpectedly anyways since it does not do anything anymore. and removing the value will alert anyone who was using the broken feature to fix it.

ill personally handle any issues related to this break.

PR #1908, opened 2024-06-06
refactor: migrate tile rendering from widgets to `CustomPainter`

JaffaKetchup:
Superceeds #1853.

PR #1928, opened 2024-07-04
refactor: migrate tile rendering from widgets to `CustomPainter`

JaffaKetchup:
Superceeds #1853 & continues #1908 targeting `master`.


ReinisSprogis:
Hi. Any insights on this? That CPU usage is still very high. This was cutting it down by half or more. I don't know if there are any other plans to reduce it.

JaffaKetchup:
I couldn't find an improvement in performance as big as was suggested on a non-web platform, but I'll go back and see if I can do some more detailed testing (as it wasn't massively detailed before). There wasn't an immediate plan to merge this, but it's definitely something we can consider in the short-to-mid term.

ReinisSprogis:
It is the WEB platform that has slow performance.

JaffaKetchup:
We're still looking towards potentially merging this, but we will need to just verify that loosing the ability to render widgets over tiles is acceptable. I'm not 100% sure it is without a good workaround.

JaffaKetchup:
Hey @ReinisSprogis :)

We are actually planning to move ahead in some form with a new primarily canvas-based tile layer as we move towards better support & integration for other tile types. 

Of course this issue if overlaying widgets onto the canvas is still unresolved. So what you are talking about is a very good point to consider in any new implementation in that we might require lower level and more state-integrated control of widget tiles. I guess it would be the same for a canvas layer as well. I'm not sure yet how everything will fit together, but we'll try to cover as many bases as possible ofc.

The idea is that we create a base tile layer and a few implemented variants, but it becomes much more extensible for end users to implement a custom tile layer if required.

PR #1940, opened 2024-08-05
fix!: added integer `tileDimension` & deprecated `tileSize`

### LINKED ISSUE #1936: [BUG] `{d}` `urlTemplate` placeholder improperly converted to string on non-web platforms
### What is the bug?

When using the `{d}` in the `urlTemplate` (introduced in #1665) it defaults to a float value which gets converted to string (e.g. `256.0`). This is obviously not supported by many tile providers.  

### How can we reproduce it?

Use the following option: `urlTemplate: 'https://api.mapbox.com/styles/v1/mapbox/streets-v12/tiles/{d}/{z}/{x}/{y}{r}?access_token=$mapboxPublicToken'`. The tile server will respond with 422 status codes.
Problem doesn't occur on web due to differences in types and therefore `toString()`.

I don't see why the resolution needs to be a `double`, maybe it can be changed to an integer type.

### Platforms

Android 14

### Severity

JaffaKetchup:
Fixes #1936.

There's no apparent reason that `tileSize` would be anything but an integer, since it is not possible to get tiles in fractional pixels. The WMS layer already assumed this. This PR relies on this assumption, and a different fix would be needed for #1936 if this assumption is not valid.

This is breaking, particularly for those using more internal functionality. However, for most users, it should not be breaking. Those which are using `tileSize` explicitly currently will have a deprecation message, but functionality will not change. Additionally, the 'temporary' fix for #1936 (just converting the number to an integer before a string) is also implemented if not using the new `tileDimension`.


JaffaKetchup:
We'll leave this to review and merge before a major upgrade.

josxha (dismissed):

TesteurManiak:
As a way of making this less breaking, you could deprecate `tileSize` for a `tileDimension` parameter?

JaffaKetchup:
> As a way of making this less breaking, you could deprecate `tileSize for a `tileDimension` parameter?

I don't see why that wouldn't work, I'd be happy to do that if you think that's better.

JaffaKetchup:
I've done what @TesteurManiak suggested.

TesteurManiak (approved):
LGTM 👍

PR #1942, opened 2024-08-05
fix: polygon hit detection when map is rotated

### LINKED ISSUE #1934: [BUG] Polygon hitbox does not take into account map rotation
### What is the bug?

When the map is in the initial position, and when we are trying to select a polygon(shape made on the map) its working fine, but when we try to rotate it , the hit box or when i try to select anything, we are not able to select the polygon.
When camera rotation is 0, hit detection works as intended, if camera rotation is != 0 hit detection is not working as intended.
PolyLinelayer layer is working as intended, PolygonLayer is the issue.
I added a listener to _hitNotifier but is is detecting the same as the onTap function on GestureDetector.

[√] Flutter (Channel stable, 3.22.2, on Microsoft Windows [Version 10.0.19045.4529], locale en-US)
[√] Windows Version (Installed version of Windows is version 10 or higher)
[√] Android toolchain - develop for Android devices (Android SDK version 35.0.0)
[X] Chrome - develop for the web (Cannot find Chrome executable at .\Google\Chrome\Application\chrome.exe)
    ! Cannot find Chrome. Try setting CHROME_EXECUTABLE to a Chrome executable.
[X] Visual Studio - develop Windows apps
    X Visual Studio not installed; this is necessary to develop Windows apps.
      Download at https://visualstudio.microsoft.com/downloads/.
      Please install the "Desktop development with C++" workload, including all of its default components 
[√] Android Studio (version 2024.1)
[√] VS Code (version 1.91.1)
[√] Connected device (3 available)    
[√] Network resources

! Doctor found issues in 2 categories.

### How can we reproduce it?

```dart
GestureDetector(
          onTap: () => _onTap(_hitNotifier.value!.hitValues.firstOrNull),
          child: PolygonLayer(
            hitNotifier: _hitNotifier,
            polygons: polygons,
          ),
        ),
```

No

### Platforms

Android 14

### Severity

monsieurtanuki:
Fixes #1934.

Removed an optimization that prevented the hit test to work for polygons on rotated maps.

### DIFF
--- a/lib/src/layer/polygon_layer/painter.dart
+++ b/lib/src/layer/polygon_layer/painter.dart
@@ -52,7 +52,12 @@ base class _PolygonPainter<R extends Object>
     required math.Point<double> point,
     required LatLng coordinate,
   }) {
-    final polygon = projectedPolygon.polygon;
+    // TODO: We should check the bounding box here, for efficiency
+    // However, we need to account for map rotation
+    //
+    // if (!polygon.boundingBox.contains(touch)) {
+    //   continue;
+    // }
 
     final projectedCoords = getOffsetsXY(
       camera: camera,
--- a/lib/src/layer/polygon_layer/painter.dart
+++ b/lib/src/layer/polygon_layer/painter.dart
@@ -53,9 +53,6 @@ base class _PolygonPainter<R extends Object>
     required LatLng coordinate,
   }) {
     final polygon = projectedPolygon.polygon;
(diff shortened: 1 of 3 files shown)


JaffaKetchup (approved):
Thanks for looking into this :)

(Just had to fix some lint issues, one unrelated but preventing merge.)

monsieurtanuki:

One optimization path would be to keep the code I've just removed _only for map rotation == 0_.

Issue #1813, opened 2024-01-28
[BUG] Tiles sometimes do not load until extra event


coogle:
### What is the bug?

I'm not sure what exactly to report as the bug, but I have a situation where I am using an animation to move my map programmatically using the following code. The animation works pretty well, but at the end of the animation I end up with a totally gray map until I interact with it again (e.g. slightly zoom out or move the map by dragging).

All of the other layers render except the tile layer. I'm not seeing any network traffic downloading the tiles, but that said when the map does render after manual intervention it loads instantly (so it already had the tiles).

```dart
    void _smoothMoveMapTo(Location destination, double destZoom)
    {
        final latTween = Tween<double>(
            begin: context.read<MapBloc>().mapController.center.latitude,
            end: destination.latitude
        );

        final lonTween = Tween<double>(
            begin: context.read<MapBloc>().mapController.center.longitude,
            end: destination.longitude
        );

        final zoomTween = Tween<double>(
            begin: context.read<MapBloc>().mapController.camera.zoom,
            end: destZoom
        );

        final animation = CurvedAnimation(
            parent: _animationCtl,
            curve: Curves.fastOutSlowIn
        );

        _animationCtl.addListener(() => context.read<MapBloc>()
            .mapController.move(
                Location(latTween.evaluate(animation), lonTween.evaluate(animation)),
                zoomTween.evaluate(animation)
            )
        );

        _animationCtl.forward(from: 0);
    }
```

### How can we reproduce it?

See provided animation controller which moves a map.

### Do you have a potential solution?

As a workaround, or perhaps to fix a bug in my code because I missed something, I'm trying to figure out how to trigger whatever is happening when I manually interact with the map after moving to my destination to actually redraw it. Notably, I've tried ending the animation with one last move ... including on a slight delay .. to no effect.

### Platforms

Flutter Web

### Severity

Minimum: Allows normal functioning

josxha:
As far as I know, the tile layer checks if tiles need to be purged or loaded when a `MapEventWithMove` is emitted. move, rotate and so on all trigger a child object of this event and therefore update the tiles.

---

I'm sorry but I'm not sure if I can reproduce your described problem with your code snipped. 

However in the example app if you trigger a second animation before the first has ended, the tiles won't be loaded.
Could this be related to your described issue?

https://github.com/fleaflet/flutter_map/assets/34318751/cc5e9800-cf69-43bd-85e8-5a4bc1b5936a

[...]

PR #1943, opened 2024-08-07
fix: eagerly listen to `TileLayer.reset` stream to ensure handler is fired

### LINKED ISSUE #1808: [BUG] `TileLayer.reset` internal listener never triggers
### What is the bug?

I was trying to use reset to trigger reload one of my TileLayer, but I found that the subsciption of reset never be initialized.
In lib/src/layer/tile_layer.dart, the subscription of reset dosen't be triggered properly:

```Dart
class _TileLayerState extends State<TileLayer> with TickerProviderStateMixin {
  bool _initializedFromMapCamera = false;

  final _tileImageManager = TileImageManager();
  late TileBounds _tileBounds;
  late var _tileRangeCalculator =
      TileRangeCalculator(tileSize: widget.tileSize);
  late TileScaleCalculator _tileScaleCalculator;

  // We have to hold on to the mapController hashCode to determine whether we
  // need to reinitialize the listeners. didChangeDependencies is called on
  // every map movement and if we unsubscribe and resubscribe every time we
  // miss events.
  int? _mapControllerHashCode;

  StreamSubscription<TileUpdateEvent>? _tileUpdateSubscription;
  Timer? _pruneLater;

  late final _resetSub = widget.reset?.listen((_) {
    _tileImageManager.removeAll(widget.evictErrorTileStrategy);
    _loadAndPruneInVisibleBounds(MapCamera.of(context));
  });

  // This is called on every map movement so we should avoid expensive logic
  // where possible.
  @override
  void didChangeDependencies() {
    super.didChangeDependencies();

    final camera = MapCamera.of(context);
    final mapController = MapController.of(context);
.......
```

Because keyword "late" was used for _resetSub, and it never be accessed which means it will not be initialized. I tried to add one line "print(_resetSub);", then this subscription works. I think that might be bug.

### How can we reproduce it?

1. Include a `TileLayer` in your Flutter application.
2. Pass a proper `reset` stream for the `TileLayer`, exactly same logic with the flutter map demo app.
3. Trigger a reset event to observe the issue. (add break point inside the reset subsription will find the program doesn't get into it)

Access it somewhere might be able to solve this problem.

### Platforms

macOS, web

### Severity

JaffaKetchup:
Also added refreshing of reset listener when reset stream changes within widget lifecycle.

Fixes #1808. However, the behaviour is still not correct, because the tile loading logic is incorrect. This looks like #1813:
 * If the `urlTemplate` does not change, tiles are evicted but not loaded until the map is moved (and possibly the tiles are not evicted correctly because they appear to reload quickly, so maybe from cache, but `reset` should empty cache)
 * If the `urlTemplate` does change, tiles are evicted but then instantly reloaded (see above), and new tiles load in gradually replacing/overlaying the old ones - the old tiles should not become visible in-between

Therefore, this will need a follow up PR.


lib/src/layer/tile_layer/tile_layer.dart
```diff
@@ -345,10 +344,14 @@ class _TileLayerState extends State<TileLayer> with TickerProviderStateMixin {
   StreamSubscription<TileUpdateEvent>? _tileUpdateSubscription;
   Timer? _pruneLater;
 
-  late final _resetSub = widget.reset?.listen((_) {
-    _tileImageManager.removeAll(widget.evictErrorTileStrategy);
-    if (mounted) _loadAndPruneInVisibleBounds(MapCamera.of(context));
-  });
+  StreamSubscription<void>? _resetSub;
```
josxha, line 347: Stream needs to get canceled in dispose().

lib/src/layer/tile_layer/tile_layer.dart
```diff
@@ -706,6 +714,11 @@ class _TileLayerState extends State<TileLayer> with TickerProviderStateMixin {
 
   bool _outsideZoomLimits(num zoom) =>
       zoom < widget.minZoom || zoom > widget.maxZoom;
+
+  void _resetStreamHandler(void _) {
+    _tileImageManager.removeAll(widget.evictErrorTileStrategy);
+    if (mounted) _loadAndPruneInVisibleBounds(MapCamera.of(context));
```
josxha, line 720: Does this handle the tile buffer around the visible tiles, too?

josxha (commented):
Nice pull request @JaffaKetchup!👍Added two small comments.
Could be nice to have a test for this, but no pressure. 😄

lib/src/layer/tile_layer/tile_layer.dart
JaffaKetchup: Ah, forgot to do this, but did it in update :D

lib/src/layer/tile_layer/tile_layer.dart
JaffaKetchup: This doesn't change the reset logic, it only ensures the handler is fired. Another PR is needed to look into this and fix #1813.

lib/src/layer/tile_layer/tile_layer.dart
JaffaKetchup: Stream is already cancelled in `dispose`:  https://github.com/fleaflet/flutter_map/blob/8136ef0fc0b75b759ea5e02eeaa2e1eac3dca245/lib/src/layer/tile_layer/tile_layer.dart#L469

JaffaKetchup:
We can add a more detailed test for the pruning etc. logic when #1813 is fixed.

lib/src/layer/tile_layer/tile_layer.dart
josxha: My bad sorry, haven't seen it while reviewing.

josxha (approved):
👍

PR #1967, opened 2024-09-18
fix: remove `hitValue` from `Polyline.renderHashCode`

JaffaKetchup:
Should result in a performance improvement when using interactive polylines.


mootw (approved):
lgtm

PR #1969, opened 2024-09-22
fix: 1338 - longitude +-180 with correct polylines and polygons

monsieurtanuki:
### What
* This PR deals with polylines and polygons (with or without holes) across longitudes +-180.
* The main issue was that longitudes -179 and 179 were projected in opposite sides of the world, although they are geographically nearby.
* The solution is to build incrementally the list of projected points: if two consecutive projected points are obviously too far away (more than half a world away), we add/remove one world width to the latter point.
* Doing so, we avoid jumping around -180 and +180, and we ensure the continuity among the points.
* The PR doesn't deal with the "teleportation" effect. This will be done in another PR, once we're confident with that "add/remove a world width" strategy.

### Screenshots
| Polyline: before / after |
| -- |
| Screenshot_1727023511 |
| Screenshot_1727023556 |

| Polygon with hole: before / after |
| -- |
| Screenshot_1727023664 |
| Screenshot_1727023641 |

### Impacted files
* `crs.dart`: new methods `getHalfWorldWidth` and `projectList`
* `painter.dart`: refactored using pre-computed `List<Double>`
* `polygon.dart`: added an example around longitude 180
* `polyline.dart`: added an example around longitude 180
* `polyline_layer.dart`: we don't cull polylines that go beyond longitude 180
* `projected_polygon.dart`: using new method `Projection.projectList`
* `projected_polyline.dart`: using new method `Projection.projectList`

### DIFF
--- a/lib/src/geo/crs.dart
+++ b/lib/src/geo/crs.dart
@@ -2,6 +2,7 @@ import 'dart:math' as math hide Point;
 import 'dart:math' show Point;
 
 import 'package:flutter_map/src/misc/bounds.dart';
+import 'package:flutter_map/src/misc/simplify.dart';
 import 'package:latlong2/latlong.dart';
 import 'package:meta/meta.dart';
 import 'package:proj4dart/proj4dart.dart' as proj4;
@@ -388,6 +389,45 @@ abstract class Projection {
 
   /// unproject cartesian x,y coordinates to [LatLng].
   LatLng unprojectXY(double x, double y);
+
+  /// Returns half the width of the world in geometry coordinates.
+  double getHalfWorldWidth() {
+    final (x0, _) = projectXY(const LatLng(0, 0));
+    final (x180, _) = projectXY(const LatLng(0, 180));
+    return x0 > x180 ? x0 - x180 : x180 - x0;
+  }
+
+  /// Projects a list of [LatLng]s into geometry coordinates.
+  ///
+  /// All resulting points gather somehow around the first point, or the
+  /// optional [referencePoint] if provided.
+  /// The typical use-case is when you display the whole world: you don't want
+  /// longitudes -179 and 179 to be projected each on one side.
+  /// [referencePoint] is used for polygon holes: we want the holes to be
+  /// displayed close to the polygon, not on the other side of the world.
+  List<DoublePoint> projectList(List<LatLng> points, {LatLng? referencePoint}) {
+    late double previousX;
+    final halfWorldWith = getHalfWorldWidth();
+    return List<DoublePoint>.generate(
+      points.length,
+      (j) {
+        if (j == 0 && referencePoint != null) {
+          (previousX, _) = projectXY(referencePoint);
+        }
+        var (x, y) = projectXY(points[j]);
+        if (j > 0 || referencePoint != null) {
+          if (x - previousX > halfWorldWith) {
+            x -= 2 * halfWorldWith;
+          } else if (x - previousX < -halfWorldWith) {
+            x += 2 * halfWorldWith;
+          }
+        }
+        previousX = x;
+        return DoublePoint(x, y);
+      },
+      growable: false,
+    );
+  }
 }
 
 class _LonLat extends Projection {
--- a/lib/src/layer/polyline_layer/projected_polyline.dart
+++ b/lib/src/layer/polyline_layer/projected_polyline.dart
@@ -16,13 +16,16 @@ class _ProjectedPolyline<R extends Object> with HitDetectableElement<R> {
   _ProjectedPolyline._fromPolyline(Projection projection, Polyline<R> polyline)
       : this._(
           polyline: polyline,
-          points: List<DoublePoint>.generate(
-            polyline.points.length,
-            (j) {
-              final (x, y) = projection.projectXY(polyline.points[j]);
-              return DoublePoint(x, y);
-            },
-            growable: false,
-          ),
+          points: projection.projectList(polyline.points),
         );
+
+  /// Returns true if the points stretch on different versions of the world.
+  bool goesBeyondTheUniverse(double halfWorldWidth) {
+    for (final point in points) {
+      if (point.x > halfWorldWidth || point.x < -halfWorldWidth) {
+        return true;
+      }
+    }
+    return false;
+  }
 }
--- a/example/lib/pages/polygon.dart
+++ b/example/lib/pages/polygon.dart
@@ -330,6 +330,35 @@ class _PolygonPageState extends State<PolygonPage> {
                 polygons: [
                   Polygon(
                     points: const [
+                      LatLng(40, 150),
+                      LatLng(45, 160),
+                      LatLng(50, 170),
+                      LatLng(55, 180),
+                      LatLng(50, -170),
+                      LatLng(45, -160),
+                      LatLng(40, -150),
+                      LatLng(35, -160),
+                      LatLng(30, -170),
+                      LatLng(25, -180),
+                      LatLng(30, 170),
+                      LatLng(35, 160),
+                    ],
+                    holePointsList: const [
+                      [
+                        LatLng(45, 175),
+                        LatLng(45, -175),
+                        LatLng(35, -175),
+                        LatLng(35, 175),
+                      ],
+                    ],
+                    color: const Color(0xFFFF0000),
+                    hitValue: (
+                      title: 'Red Line',
+                      subtitle: 'Across the universe...',
+                    ),
+                  ),
+                  Polygon(
+                    points: const [
                       LatLng(50, -18),
                       LatLng(50, -14),
                       LatLng(51.5, -12.5),
--- a/example/lib/pages/polyline.dart
+++ b/example/lib/pages/polyline.dart
@@ -23,6 +23,23 @@ class _PolylinePageState extends State<PolylinePage> {
 
   final _polylinesRaw = <Polyline<HitValue>>[
     Polyline(
+      points: const [
+        LatLng(40, 150),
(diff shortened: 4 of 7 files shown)


lib/src/layer/polyline_layer/projected_polyline.dart
```diff
@@ -16,13 +16,16 @@ class _ProjectedPolyline<R extends Object> with HitDetectableElement<R> {
...
-              final (x, y) = projection.projectXY(polyline.points[j]);
-              return DoublePoint(x, y);
-            },
-            growable: false,
-          ),
+          points: projection.projectList(polyline.points),
         );
+
+  /// Returns true if the points stretch on different versions of the world.
+  bool goesBeyondTheUniverse(double halfWorldWidth) {
```
josxha, line 23: As much as I like this name, a declarative name would be more understandable in the future. 😄

josxha (commented):
Thanks for the pull request @monsieurtanuki! I did some quick testing and your added examples work like intended (thanks for adding them).

Some additional thoughts:
The main use case of line wrapping are paths that cross the -180/180 longitude for example when we want to display flights. We need to figure out a solution when the user follows a polyline or polygon accross the world border. Currently the line suddenly disapears. This, of course, is not in scope for this this pull request. (:

https://github.com/user-attachments/assets/1546321e-fae2-4666-8fc5-3b5deaa976e7

monsieurtanuki:
> Some additional thoughts: The main use case of line wrapping are paths that cross the -180/180 longitude for example when we want to display flights. We need to figure out a solution when the user follows a polyline or polygon accross the world border. Currently the line suddenly disapears. This, of course, is not in scope for this this pull request. (:

Oh it's a bug then! Probably related to the "teleportation" issue.
I'll have a look at it, and I may fix that in the current PR.
I keep you posted.

josxha:
> Oh it's a bug then! Probably related to the "teleportation" issue.
I'll have a look at it, and I may fix that in the current PR.
I actually don't think it's an bug. The line just moves to the new world that comes into center. So the line is still active but basically on the other side of the world. If you plan to look into this as well how it can get improved that's awesome of couse. (:

monsieurtanuki:
@josxha Your description is correct: it's the "teleportation" issue (I can see a unique polyline jumping from world to world on the same zoom 0 map) but when the zoom is high - the "other" world is not visible.

I still think it's a bug, probably not a new bug, but a fair user expectation is to see at least one polyline.

Working on it. The current PR is relatively small (7 files changed), so I'll see how review-able it would become when including that bug fix.

monsieurtanuki:
@josxha Fixed! One visible copy of the polyline, with a small or a large zoom.

lib/src/layer/polyline_layer/projected_polyline.dart
JaffaKetchup: Agreed with this point, or at least a little more documentation (what is `halfWorldWidth`, why does it matter, where does it come from, etc.).

JaffaKetchup (approved):
LGTM in general! Some documentation about what the whole 'half world width' stuff is about would be nice. I'll leave to @josxha to give final review :) Thanks!

JaffaKetchup:
Just a thought, what happens with `CircleMarker`s?

lib/src/layer/polyline_layer/projected_polyline.dart
mootw: agree

monsieurtanuki:
@JaffaKetchup @josxha I've just pushed some changes that make the code (somehow) more readable.
In particular, I've switched to `getWorldWidth` (no more "half") and I've added comments:
```dart
  /// Returns the width of the world in geometry coordinates.
  ///
  /// Is used at least in 2 cases:
  /// * my polyline crosses longitude 180, and I somehow need to "add a world"
  /// to the coordinates in order to display a continuous polyline
  /// * when my map scrolls around longitude 180 and I have a marker in this
  /// area, the marker may be projected a world away, depending on the map being
  /// centered either in the 179 or the -179 part - again, we can "add a world"
  double getWorldWidth() {
    final (x0, _) = projectXY(const LatLng(0, 0));
    final (x180, _) = projectXY(const LatLng(0, 180));
    return 2 * (x0 > x180 ? x0 - x180 : x180 - x0);
  }
```

josxha:
> I'll leave to @josxha to give final review :)

Sorry, I won't find time to review it this week. 😶 I'll try to review it next week if no one gets ahead of me. (:

monsieurtanuki:
The review should probably be put on hold until the fix for #1976 is coded (by me) (like, today), reviewed and merged.

JaffaKetchup:
I have found another issue, trying to use polygons with the points at the corners of the world (+-180, +-90). Do I need to use different coordinates?
Without a hole, it does not appear. With a hole, the hole is actually filled and the rest of the polygon is not.

monsieurtanuki:
> No worries! I have found another issue, trying to use polygons with the points at the corners of the world (+-180, +-90). Do I need to use different coordinates? Without a hole, it does not appear. With a hole, the hole is actually filled and the rest of the polygon is not. This appears on master.

You mean, +-90 _latitude_? Hard-code stuff, isn't it? Please provide an example.

JaffaKetchup:
Can't give a demo right now, but these points:
```
                points: [
                  const LatLng(-90, 180),
                  const LatLng(90, 180),
                  const LatLng(90, -180),
                  const LatLng(-90, -180),
                ],
```

monsieurtanuki:
I did understand the concept of +-90 latitude, but for a webmercator projection going beyond the 85.0511287798 limit seems rather bold.
And a polygon that covers the whole world was not part of the PR: I only imagined normal areas that happened to be around +-180.
Did it work before the recent changes? Wouldn't it be fair to create a new issue / feature request like "display a polygon inside out"?

JaffaKetchup:
There's nothing wrong! I was depending on my local cloned copy which was broken instead of the git branch. Sorry!

monsieurtanuki:
> The review should probably be put on hold until the fix for #1976 is coded (by me) (like, today), reviewed and merged.

Now back to polygons and polylines in multi-worlds!

monsieurtanuki:
Ping

JaffaKetchup:
Oops, apologies, thought you needed more time to work on this. I'll review now.

example/pubspec.yaml
```diff
@@ -18,6 +18,7 @@ dependencies:
   proj4dart: ^2.1.0
   shared_preferences: ^2.3.2
   url_launcher: ^6.3.0
+  url_launcher_android: 6.3.2
```
JaffaKetchup, line 21: Do we need to depend on this directly @monsieurtanuki?

JaffaKetchup (approved):
LGTM, and some nice removal of duplications as well :) One question, then ready to merge.

---

This also appears on master, but is caused by the infinite scroll feature. Not sure if its a bug exactly? Using a `cameraConstraint` with the entire world will cause the map to stick against the edges during a fling animation, but will jump once the fling is complete. A drag does not work at all (causes sticking).

```dart
cameraConstraint: CameraConstraint.contain(
	bounds: LatLngBounds(
		const LatLng(-90, -180),
        const LatLng(90, 180),
    ),
),
```

https://github.com/user-attachments/assets/d951703d-b69c-4e59-9581-b6d3614679f6

lib/src/geo/crs.dart
```diff
@@ -2,6 +2,7 @@ import 'dart:math' as math hide Point;
 import 'dart:math' show Point;
 
 import 'package:flutter_map/src/misc/bounds.dart';
+import 'package:flutter_map/src/misc/simplify.dart';
```
JaffaKetchup, line 5: (for the future: if we're starting to migrate to `DoublePoint` throughout (which I agree with), we should probably separate it into its own file, and make the internal changes all at once as much as possible. Currently, I think we're split between `Offset`, `Point<double>`, `DoublePoint` and records - they are all used in different places obviously, but we should try to get just one or two (eg. `Offset` cannot be avoided in the painter).)

mootw: I will handle this in #1996

mootw (approved):
looks good pending the one dependency comment!

example/pubspec.yaml
monsieurtanuki: I guess not, indeed. I think I was trapped in a WTF situation that was somehow solved later by #1985.

monsieurtanuki:
@JaffaKetchup I'm not an expert on `cameraConstraint` and fling effect, but I think the possible bug you're mentioning goes beyond this PR and would deserve its own issue.
@mootw Removed the useless new explicit package dependency.

monsieurtanuki:
Thank you @JaffaKetchup @mootw @josxha for your reviews!

JaffaKetchup:
Hey @monsieurtanuki, just plotting our next release schedule. Do you think you'll be able to work on the teleporting issue any time soon? Of course, no problem at all if not! Thanks for all your hard work so far :)

monsieurtanuki:
@JaffaKetchup I guess I should be able to fix the teleporting effect within a month - including reviews and feedbacks.

Keep in mind though that:
* the teleporting is to be coded specifically for each widget (marker, circlemarker, polygon and polyline) - several potential sources of bugs
* #1976 (reported by @arneke) about the multiworld feature having non webmercator maps not working anymore
* the teleporting goes a bit beyond the "multiworld" and "pacific polygons" issues, adding a "low zoom" flavor

Imagining an arbitrary "let's publish a release before 2025" goal, I believe it would be safer not to jeopardize the next release, possibly to test again, and to keep the teleporting for afterwards.

Issue #1976, opened 2024-10-10
[BUG] unbounded horizontal scroll breaks most tile grids


arneke:
### What is the bug?

My EPSG:3006 WMTS layer works in 7.0.2, but stopped working after I switched to the master branch. On closer inspection, the reason was that flutter_map is requesting the wrong x,y,z tiles for a given lat,long position.

I've not completely isolated #1948 as the cause, but I think it's the most likely thing in the commit log, and there is an assumption about the modulus that is not generally true.

### How can we reproduce it?

Most non-WebMercator grids should trigger this, but here is one.

```
// OSM based, uses somewhat strange bounds. No access restrictions or fees in GetCapabilities as of 2024-10-15
FlutterMap(
    options: MapOptions(
        initialCameraFit: CameraFit.bounds(
          bounds: LatLngBounds(
              LatLng(61.285991891313344, 17.006922572652666),
              LatLng(61.279648385340494, 17.018309853620366)),
        ),
        crs: Proj4Crs.fromFactory(
          code: 'EPSG:3006',
          proj4Projection: proj4.Projection.add(
            'EPSG:3006',
            '+proj=utm +zone=33 +ellps=GRS80 +towgs84=0,0,0,0,0,0,0 +units=m +no_defs +type=crs',
          ),
          origins: [
            Point(-58122915.354077406, 19995929.885879364),
          ],
          resolutions: const <double>[
            4096,
            2048,
            1024,
            512,
            256,
            128,
            64,
            32,
            16,
            8,
            4,
            2,
            1,
            0.5,
            0.25,
            0.125
          ],
        )),
    children: [
      TileLayer(
          tileBuilder: coordinateDebugTileBuilder,
          urlTemplate:
              'https://maps.trafikinfo.trafikverket.se/MapService/wmts.axd/BakgrundskartaNorden.gpkg?layer=Background&style=default&tilematrixset=Default.256.3006&Service=WMTS&Request=GetTile&Version=1.0.0&Format=image%2Fpng&TileMatrix={z}&TileCol={x}&TileRow={y}')
    ],
  )
```

### Do you have a potential solution?

I don't uderstand all the code in #1948 , but I believe the problem originates in

``` final modulo = 1 << coordinates.z; ```

https://github.com/fleaflet/flutter_map/blob/d73d2f1a230957b79a12f82d96de420b68715006/lib/src/layer/tile_layer/tile_coordinates.dart#L28

I don't think the code currently handles cases where
- the grid is not square
[...]

PR #1978, opened 2024-10-16
fix: make horizontal repetition CRS dependent

monsieurtanuki:
### What
- cf. #1976
- The recent changes #1948 enhanced world replications.
- Unfortunately, in less straightforward examples than webmercator the world replications don't have much sense. Which provoked display bugs.
- Now, we can enjoy both webmercator with world replications, and other projections without world replications.
- cc @arneke

### Files
New file:
* `epsg3006_crs.dart`: new example around EPSG3006, needed for tests

Impacted files:
* `camera.dart`: special case around `replicatesWorldLongitude`
* `crs.dart`: new `bool` getter `replicatesWorldLongitude`
* `main.dart`: added EPSG3006 example
* `map_interactive_viewer.dart`: special case around `replicatesWorldLongitude`
* `menu_drawer.dart`: added EPSG3006 example
* `tile_coordinates.dart`: moved code to new class `TileCoordinatesSimplifier`
* `tile_image_manager.dart`: now using a `TileCoordinatesSimplifier` for more flexibility
* `tile_image_view.dart`: now using a `TileCoordinatesSimplifier` for more flexibility
* `tile_layer.dart`: special case around `replicatesWorldLongitude`
* `tile_range.dart`: special case around `replicatesWorldLongitude`


arneke:
This branch works well in my application. 

Still think the `1 << z` is a bit problematic.  Does https://demo.fleaflet.dev/crs_epsg4326 work for you locally ? (that is a grid with two tiles at z = 0)

monsieurtanuki:
> This branch works well in my application.

Cool!

> Still think the `1 << z` is a bit problematic. Does https://demo.fleaflet.dev/crs_epsg4326 work for you locally ? (that is a grid with two tiles at z = 0)

@arneke You're right: I've just fixed that. Thank you for your meticulous review!

lib/src/layer/tile_layer/tile_coordinates.dart
```diff
@@ -70,3 +47,45 @@ class TileCoordinates extends Point<int> {
...
+/// On maps with world replications, different tile coordinates may actually
+/// refer to the same "simplified" tile coordinate - the coordinate that starts
+/// from 0.
+/// For instance, on zoom level 0, all tile coordinates can be simplified to
+/// (0,0), which is the only tile.
+/// On zoom level 1, (0, 1) and (2, 1) can be simplified to (0, 1), as they both
+/// mean the bottom left tile.
+/// And when we're not in the context of world replications, we don't have to
+/// simplify the tile coordinates: we just return the same value.
+class TileCoordinatesSimplifier {
```
JaffaKetchup, line 62: I'm not 100% sure this is necessary. It adds a little more overhead, and ideally we'd skip it all together if `replicatesWorldLongitude` is `false`. This means the method can be made static, and at that point, is the complexity of another class in this space worth it? Would it be better to just pass a conditional around and have a static function somewhere else (like it was before), or maybe pass an anonymous function around to avoid passing the conditional everywhere?
If we're keeping it, maybe a better name would be "Resolver" or "Replicator". Not sure though.

JaffaKetchup (changes requested):
Sorry for the long wait, had a busy week!

lib/src/layer/tile_layer/tile_coordinates.dart
monsieurtanuki: @JaffaKetchup I'd have to double-check that, but that's a matter of OOP (rather my way of coding) against performances (nothing wrong with that) (but in that case not sure the performances would be that much impacted).

I remember that basically I had to create this `TileCoordinatesSimplifier` because:
* the `immutable` clause for `TileImageManager` made it impossible to just set a `bool replicatesWorldLongitude` field in `TileImageManager`
* I wanted the `crs` to set the `bool` value instead of explicitly setting (even "forcing") the same `bool` in the `TileImageManager`, especially given that we need the `camera` for that - and it's available through a `MapCamera.of(context)` that somehow needs a call to `didChangeDependencies` and cannot be set at `initState` time

JaffaKetchup (approved):
(Sorry for the wait, I've had a hugely busy few weeks!)

I think I'm OK with this, but would appreciate a review from someone else!

monsieurtanuki:
I'm open to new reviews and new suggestions.

example/lib/pages/epsg3006_crs.dart
```diff
@@ -0,0 +1,76 @@
...
+                    0.5,
+                    0.25,
+                    0.125
+                  ],
+                )),
+            children: [
+              TileLayer(
+                  tileBuilder: coordinateDebugTileBuilder,
+                  urlTemplate:
+                      'https://maps.trafikinfo.trafikverket.se/MapService/wmts.axd/BakgrundskartaNorden.gpkg?layer=Background&style=default&tilematrixset=Default.256.3006&Service=WMTS&Request=GetTile&Version=1.0.0&Format=image%2Fpng&TileMatrix={z}&TileCol={x}&TileRow={y}')
```
josxha, line 69: What are the usage policy for this tile source? Source attribution would be a good think, too.

The tile source has long loading times for me. When I run the app on windows the tiles don't load but I got an exception. Not sure if this only happens on my machine though.
```console
======== Exception caught by image resource service ================================================
The following HandshakeException was thrown resolving an image codec:
Handshake error in client (OS Error: 
	CERTIFICATE_VERIFY_FAILED: unable to get local issuer certificate(../../../flutter/third_party/boringssl/src/ssl/handshake.cc:393))

When the exception was thrown, this was the stack: 
#0      _SecureFilterImpl._handshake (dart:io-patch/secure_socket_patch.dart:99:46)
#1      _SecureFilterImpl.handshake (dart:io-patch/secure_socket_patch.dart:143:25)
#2      _RawSecureSocket._secureHandshake (dart:io/secure_socket.dart:920:54)
#3      _RawSecureSocket._tryFilter (dart:io/secure_socket.dart:1049:19)
<asynchronous suspension>
#4      _RawSecureSocket._secureHandshake (dart:io/secure_socket.dart:928:9)
<asynchronous suspension>
#5      _RawSecureSocket._tryFilter (dart:io/secure_socket.dart:1049:13)
<asynchronous suspension>
URL: https://maps.trafikinfo.trafikverket.se/MapService/wmts.axd/BakgrundskartaNorden.gpkg?layer=Background&style=default&tilematrixset=Default.256.3006&Service=WMTS&Request=GetTile&Version=1.0.0&Format=image%2Fpng&TileMatrix=13&TileCol=466563&TileRow=111610
Fallback URL: null
Current provider: MapNetworkImageProvider()
====================================================================================================
flutter: HandshakeException: Handshake error in client (OS Error: 
flutter: 	CERTIFICATE_VERIFY_FAILED: unable to get local issuer certificate(../../../flutter/third_party/boringssl/src/ssl/handshake.cc:393))
flutter: HandshakeException: Handshake error in client (OS Error: 
flutter: 	CERTIFICATE_VERIFY_FAILED: unable to get local issuer certificate(../../../flutter/third_party/boringssl/src/ssl/handshake.cc:393))
flutter: HandshakeException: Handshake error in client (OS Error: 
flutter: 	CERTIFICATE_VERIFY_FAILED: unable to get local issuer certificate(../../../flutter/third_party/boringssl/src/ssl/handshake.cc:393))
flutter: HandshakeException: Handshake error in client (OS Error: 
flutter: 	CERTIFICATE_VERIFY_FAILED: unable to get local issuer certificate(../../../flutter/third_party/boringssl/src/ssl/handshake.cc:393))
flutter: HandshakeException: Handshake error in client (OS Error: 
flutter: 	CERTIFICATE_VERIFY_FAILED: unable to get local issuer certificate(../../../flutter/third_party/boringssl/src/ssl/handshake.cc:393))
flutter: HandshakeException: Handshake error in client (OS Error: 
flutter: 	CERTIFICATE_VERIFY_FAILED: unable to get local issuer certificate(../../../flutter/third_party/boringssl/src/ssl/handshake.cc:393))
```

lib/src/layer/tile_layer/tile_coordinates.dart
josxha: Your arguments sound reasonable but what is there reason to keep `TileImageManager` immutable since it's not really a data class?

josxha (commented):
I have the feeling that the fling animation has stutters when using a map with longitude repetition. Most notifable when crossing the -180/180 longitude. Could anyone check if it happens to them as well?

lib/src/layer/tile_layer/tile_coordinates.dart
monsieurtanuki: @josxha I'm not familiar with `@immutable`, but from what I've just read we use only one occurrence of it and there doesn't seem to be added value in that metatag here.
I'll recode without `@immutable`: probably cleaner to read and to maintain.

monsieurtanuki:
@JaffaKetchup @josxha Actually `TileCoordinatesSimplifier` isn't only used in `TileImageManager`, but also in `TileImageView`.
That said, removing `@immutable` enables me to code as I would have in a first approach, cf. latest push.

example/lib/pages/epsg3006_crs.dart
monsieurtanuki: > What are the usage policy for this tile source? Source attribution would be a good think, too.

cc @arneke

Regarding CERTIFICATE errors, I've never experienced one, working on an Android emulator working on Windows.

monsieurtanuki:
> I have the feeling that the fling animation has stutters when using a map with longitude repetition. Most notifable when crossing the -180/180 longitude. Could anyone check if it happens to them as well?

I suspect that my latest push made things slightly faster, with `const` classes and fields. Not a big user of flings, but haven't noticed anything problematic.

lib/src/layer/tile_layer/tile_coordinates.dart
josxha: Looking good, thanks!

josxha:
Do you know why the map is offset between zoom level 1 and 2? 
I've seen some thin black lines shimmering through from the custom tile builder. Do you know if it is related to this pr?

https://github.com/user-attachments/assets/86be795a-a0ae-456c-be8a-e9ae565018b7

example/lib/pages/epsg3006_crs.dart
arneke: There's nothing published (I mention this at the top of the example in #1976). I included it to make it easy to reproduce the bug, was not really expecting this to end up on the Flutter Map website.

I can see if I can make a public service with dummy data available at some point, but not immediately and no guarantees about uptime or longevity.

monsieurtanuki: @arneke My initial point was to be able to test the fixes, for instance with the "3006" projection.
Any "non web-mercator" would have done the trick.
I've just removed "your 3006" from the PR.

josxha (approved):
Removing the example page, solves my open points. 😄
Looks good! Thanks for the contribution.

monsieurtanuki:
@JaffaKetchup @josxha Thank you for your reviews and comments!

Still, there seems to be an issue with the "Build Android Example App" CI/CD. A bit cryptic to me: maybe the gradle version, maybe the android ndk version.

JaffaKetchup:
#1985 fixes the build for Android (would appreciate a review @josxha!), then I can fix this.

PR #1985, opened 2024-11-17
chore: fixed Android example app build for CI/CD

JaffaKetchup:


JaffaKetchup:
Yep, copied it out of one of my other projects that I recently regenerated and fixed this same issue on (and ran AGP on), then just changed the necessary strings.

PR #1987, opened 2024-11-19
feat: added keyboard controls for map gestures

JaffaKetchup:
This PR adds keyboard controls for the map, each of which may be toggled individually. By default, only the new arrow key gestures work - but maybe it would be better to enable them all?

Supports arrow & WASD keys for panning, QE keys for rotation, and RF keys for zoom (assuming QWERTY). All keys are physical not logical, so whatever key is in those positions on a QWERTY keyboard will work the same on others).

A potential extension could be to allow a key to be defined which resets the map rotation. And/or, to allow more customizability with keys (eg. to better support rotation and zooming for left handed users using the arrow keys).


JaffaKetchup:
Here's a testing link: https://fleaflet-firebase--arrow-keys-pscghofy.web.app/interactive_flags_page

One thing to note is that when switching from an 'axis'(?)-movement to diagonal-movement (even just by very briefly pressing on a key in the perpendicular axis, but not briefly enough to trigger a leap), it appears as though there is a very fast change in movement velocity. I'm not 100% sure why this occurs, but IMO it shouldn't block this merge.

josxha (approved):
[link to discord discussion](https://discord.com/channels/951867686378409984/951879268227485707/1335662780031963299)

PR #1990, opened 2024-11-24
feat: expose `BaseOverlayImage` for external implementation

JaffaKetchup:
* Switch to `abstract` instead of `sealed`
* Renamed `_render` to `layout` (added protected metadata)
* Removed non-virtual metadata from `build`
* Replaced `Image.color` & `.colorBlendMode` for `opacity` support with normal `.opacity` field using `AlwaysStoppedAnimation`
* Removed ambient `MapCamera` from `layout` arguments

I feel like we're overcomplicating things, and maybe doing too much for the user? It's much more flexible if we allow the user to pass their own `Image`, and not much more difficult if we allow them to do so. I'm not sure if it would work, but then wouldn't any widget work (like https://github.com/TimBaumgart/flutter_map_polywidget), not just images?

### DIFF
--- a/lib/src/layer/overlay_image_layer/overlay_image.dart
+++ b/lib/src/layer/overlay_image_layer/overlay_image.dart
@@ -1,9 +1,12 @@
 part of 'overlay_image_layer.dart';
 
-/// Base class for all overlay images.
+/// Display an [Image] on the map at a specific coordinate location, within an
+/// [OverlayImageLayer]
+///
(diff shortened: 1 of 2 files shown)


mootw (approved):
looks good. i like this. i think it should be possible to expand to any widget in the future. though, the library you mentioned has extra features that honestly might be out of scope for us to implement and maintain

josxha (approved):

PR #1991, opened 2024-11-24
fix(perf): fixed massive performance issue in simplification steps for `Polyline/gonLayer`s

JaffaKetchup:
The caching of the simplification was not working correctly, and creating massive overheads.
Essentially, we were storing a lazily-generated `Iterable` in the simplification cache. This means that it was fully re-generated (re-simplified) on every cache retrieval: simplification had to run every frame. We now store a non-growable `List`.

Before:
image

After:
image


josxha (approved):

corepuncher:
Sounds exciting, can't wait to test.  As you know, I am into "massive performance increases" myself :-)

PR #1993, opened 2024-11-24
feat(perf): perform bbox culling prior to aggressive culling for polylines

JaffaKetchup:
Both types of culling are still performed after simplification.

It is expected that this will have the most effect on longer polylines that are not always within the viewport - for example, the user is zoomed into see part of one specific long polyline in a map with many. Whilst this does add more work, it avoids many iterations of a loop that would be useless. To minimize the expense of constructing the bounding box for the polylines, they are cached on the `Polyline` (the same as we do for `Polygon`s).
Note that the bbox culling still works in geographic space - this is what we do for polygons, and to avoid this, a new class would be needed to maintain efficiency, mixing `Bounds` and `DoublePoint`.

In the example stress test, with simplification set to 0.02 to increase stress on the culling, with the entire polyline off screen, it seems to reduce the load on the UI thread by about ~1-2ms.

Before:
Screenshot 2024-11-24 164159

After:
Screenshot 2024-11-24 164354


lib/src/layer/polyline_layer/polyline_layer.dart
```diff
@@ -143,6 +143,10 @@ class _PolylineLayerState<R extends Object> extends State<PolylineLayer<R>>
     for (final projectedPolyline in polylines) {
       final polyline = projectedPolyline.polyline;
 
+      // Test bounding boxes to avoid potentially expensive aggressive culling
+      // where none of the line is visibles
```
mootw, line 147:
```suggestion
      // when none of the lines are visible
```

small grammar fix

mootw (changes requested):
is the bbox re-used for post-simplification culling or is that based on the simplified polygon?

JaffaKetchup:
@fleaflet/maintainers

JaffaKetchup:
@mootw That's a good point. It is using the original shape not the simplified shape for the bounding box. However, it looks as though the polygon culling is also doing this.

mootw (approved):

PR #1994, opened 2024-11-24
refactor: bump min flutter version to 3.27.0, fix deprecations

JaffaKetchup:
Colors are changing in 3.27: https://docs.flutter.dev/release/breaking-changes/wide-gamut-framework.

Not released yet, so no need to worry about merging just yet.


monsieurtanuki:
@JaffaKetchup Maybe it's time now, cf. https://github.com/fleaflet/flutter_map/actions/runs/12331534552/job/34418263958

image

josxha:
Two additional changes:
- I removed the used workflow action to check the package score with cli commands as it didn't used the latest flutter version.
- The new methods that deprecated old flutter api isn't available in previous versions. That's why I bumped the minimum flutter version to 3.27 and with it the minimum dart version to 3.6.0. (new dart functionalities have been unlocked😮)

josxha (approved):

corepuncher:
Regarding requiring 3.27 for next flutter_map version, I am currently experiencing GPU OOM crashes on Android. iOS seem ok. 

I downgraded back to 3.24.5, and the crashes went away. At first I thought it was Impeller-induced on Android, but even after adding the impeller-disabling block to my AndroidManifest.xml, it would still crash, so there must be something GPU/memory related OTHER than "straight impeller" that is causing this.  Just something to be aware of. 

I am currently working on better memory management to mitigate these crashes. Your OverlayImage PR seems promising because PART of my issues are those.

monsieurtanuki:
Given that new flutter versions may be problematic, a solution would be to (virtually) downgrade flutter_map to 3.24 and handle the "deprecated" messages with lousy "ignore" comments.
This way developers would be able to use both 3.24 and 3.27.

PR #1996, opened 2024-12-02
feat!: migrate `Point<double>` to `Offset` internally

### LINKED ISSUE #1769: [FEATURE] Use classes instead of `Record`s for efficiency
The research in https://github.com/fleaflet/flutter_map/pull/1750#issuecomment-1846490749 and https://github.com/fleaflet/flutter_map/pull/1750#issuecomment-1847661696 points out the reasons to avoid records. 
Records have no advantage on memory consumtion while they require about double the time to be initialized.

As a consequence we should remove records from the package and use immutable classes in favour of it.

mootw:
This will yield significant performance benefits due to a reduction in casting. This likely breaks the external API but it is worth it

Migrations:
 `Point<double>` -> `Offset` or `Size`
`Bounds<double>` -> `Rect`
`Bounds<int>` -> (rename and marked internal) `IntegerBounds`

maybe fixes #1769 (it is not a full migration, as there are still record types (they are okay in certain conditions honestly, especially as an interface between public apis), but it does fix a TON of issues with math.Point, which is much slower than the difference between Objects and Records)

there is a modest 7% performance boost in the polygon and polyline tests in both the UI and Raster threads in my debug mode profiling. i expect other layers to benefit and more performance boost in ideal situations


lib/src/map/camera/camera.dart
```diff
@@ -208,37 +207,38 @@ class MapCamera {
...
 
   /// Calculates the [LatLng] for the given [point] using this camera's
   /// [crs] and [zoom] (or the provided [zoom]).
-  LatLng unproject(Point point, [double? zoom]) =>
+  LatLng unproject(Offset point, [double? zoom]) =>
       crs.pointToLatLng(point, zoom ?? this.zoom);
 
   /// Same as the [unproject] function.
-  LatLng layerPointToLatLng(Point point) => unproject(point);
+  LatLng layerPointToLatLng(Offset point) => unproject(point);
```
JaffaKetchup, line 241: We might want to deprecate this. It's just duplication.

lib/src/map/camera/camera.dart
```diff
@@ -208,37 +207,38 @@ class MapCamera {
...
   /// [crs] and [zoom] (or the provided [zoom]).
-  Point<double> project(LatLng latlng, [double? zoom]) =>
+  Offset project(LatLng latlng, [double? zoom]) =>
       crs.latLngToPoint(latlng, zoom ?? this.zoom);
 
   /// Calculates the [LatLng] for the given [point] using this camera's
   /// [crs] and [zoom] (or the provided [zoom]).
-  LatLng unproject(Point point, [double? zoom]) =>
+  LatLng unproject(Offset point, [double? zoom]) =>
       crs.pointToLatLng(point, zoom ?? this.zoom);
```
JaffaKetchup, line 238: It might make sense to deprecate these for simplicity? The user can just use the CRS/projection directly without a wrapper.

lib/src/map/camera/camera.dart
```diff
@@ -274,38 +272,39 @@ class MapCamera {
...
 
     return point - nonRotatedPixelOrigin;
   }
 
   /// Calculate the [LatLng] coordinates for a [localPoint].
-  LatLng pointToLatLng(Point localPoint) {
-    final localPointCenterDistance = Point(
-      (nonRotatedSize.x / 2) - localPoint.x,
-      (nonRotatedSize.y / 2) - localPoint.y,
+  LatLng pointToLatLng(Offset localPoint) {
```
JaffaKetchup, line 297: Consistency, rename?

Now or in the future, we may also want to better document these or rename them to differentiate them from the CRS/projection level, which can be a confusing difference.

lib/src/map/camera/camera.dart
```diff
@@ -18,7 +17,7 @@ class MapCamera {
   /// in a subsequent build to the actual constraints. We set the size to this
   /// impossible (negative) value initially and only change it once Flutter
   /// provides real constraints.
-  static const kImpossibleSize = Point<double>(-1, -1);
+  static const kImpossibleSize = Offset(-1, -1);
```
JaffaKetchup, line 20: Should we consider using const `Offset.infinite`?

lib/src/map/controller/map_controller_impl.dart
```diff
@@ -224,7 +224,8 @@ class MapControllerImpl extends ValueNotifier<_MapControllerState>
   /// the map.
   MoveAndRotateResult rotateAroundPointRaw(
     double degree, {
-    required Point<double>? point,
+    //TODO this looks insanely jank
+    required Offset? point,
     required Offset? offset,
```
JaffaKetchup, line 229: Could this be better documented? It is indeed quite strange.

lib/src/misc/bounds.dart
```diff
@@ -41,17 +43,17 @@ class Bounds<T extends num> {
   const Bounds.unsafe(this.min, this.max);
 
   /// Create a [Bounds] as bounding box of a list of points.
-  static Bounds<double> containing(Iterable<Point<double>> points) {
+  static Bounds<double> containing(Iterable<Offset> points) {
```
JaffaKetchup, line 46: This feels wrong. I think we should move `Bounds` to `Offset` completely, but otherwise we should not be mixing `Offset` and `Point`. I understand this is better for performance, but it is not great for consistency.

lib/src/misc/extensions.dart
```diff
@@ -3,62 +3,34 @@ import 'dart:ui';
 
 /// Extension methods for the math.[Point] class
 extension PointExtension<T extends num> on Point<T> {
```
JaffaKetchup, line 5: We should deprecate this entire extension method and all methods/operators within, or mark it as internal at least.

lib/src/misc/extensions.dart
```diff
@@ -3,62 +3,34 @@ import 'dart:ui';
...
-  /// Maps the [Point] to an [Offset].
+  /// Converts to offset
   Offset toOffset() => Offset(x.toDouble(), y.toDouble());
 }
 
 /// Extension methods for [Offset]
 extension OffsetToPointExtension on Offset {
-  /// Creates a [Point] representation of this offset.
+  /// Creates a [Point] representation of this offset. This is ONLY used for backwards compatibility
   Point<double> toPoint() => Point(dx, dy);
```
JaffaKetchup, line 19: Deprecate this method.

JaffaKetchup (changes requested):
Some things I would be even more heavy handed/breaking on. I would be in favour of implemented deprecated methods just to ease the transition for external users, but I think we need to be more thorough. 

Is there any reason for keeping `Bounds` using `Point`?

JaffaKetchup (changes requested):
Some things I would be even more heavy handed/breaking on. I would be in favour of implemented deprecated methods just to ease the transition for external users, but I think we need to be more thorough. 

Is there any reason for keeping `Bounds` using `Point`?

mootw:
> Some things I would be even more heavy handed/breaking on. I would be in favour of implemented deprecated methods just to ease the transition for external users, but I think we need to be more thorough.
> 
> Is there any reason for keeping `Bounds` using `Point`?

Yes, I have only migrated the Point<double> interface methods in `Bounds` because the `DiscreteTileRange` uses `Bounds<int>` which would mean that i need to implement a new `IntegerBounds` class on top of migrating Bounds or change `DiscreteTileRange` to no longer require integers

lib/src/geo/crs.dart
```diff
@@ -51,13 +52,13 @@ abstract class Crs {
...
-  Point<double> latLngToPoint(LatLng latlng, double zoom) {
+  Offset latLngToPoint(LatLng latlng, double zoom) {
     final (x, y) = latLngToXY(latlng, scale(zoom));
-    return Point<double>(x, y);
+    return Offset(x, y);
   }
 
   /// Converts a map point to the sphere coordinate (at a certain zoom).
-  LatLng pointToLatLng(Point point, double zoom);
+  LatLng pointToLatLng(Offset point, double zoom);
```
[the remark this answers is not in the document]
mootw: agreed, where appropriate. i think in some situations it makes more sense to use the term "Point", and only to use "Offset" where it is directly related to the UI rendering

lib/src/map/camera/camera.dart
mootw: yeah it isn't even used

lib/src/misc/bounds.dart
mootw: I only migrated the interfaces that are `Bounds<double>`, as the intention is to convert entirely to using Offset instead of Points in the future. Reverting this back would actually require explicit conversion from Offset -> Point in multiple other parts of the code.

lib/src/layer/shared/layer_interactivity/internal_hit_detectable.dart
```diff
@@ -98,7 +96,7 @@ base mixin HitTestRequiresCameraOrigin<R extends Object,
   /// Calculated [MapCamera] origin, using the following formula:
   ///
   /// ```dart
-  /// camera.project(camera.center).toOffset() - camera.size.toOffset() / 2
+  /// camera.project(camera.center) - camera.size. / 2
```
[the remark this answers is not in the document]
mootw: I am not seeing this extra `.` locally at all which is super weird

lib/src/map/camera/camera.dart
mootw: I think this wrapper is potentially useful since it includes the camera's zoom by default

lib/src/layer/shared/layer_interactivity/internal_hit_detectable.dart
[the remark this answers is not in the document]
mootw: nvm it is in the comment lol

lib/src/geo/crs.dart
```diff
@@ -51,13 +52,13 @@ abstract class Crs {
   (double, double) latLngToXY(LatLng latlng, double scale);
 
   /// Similar to [latLngToXY] but converts the XY coordinates to a [Point].
-  Point<double> latLngToPoint(LatLng latlng, double zoom) {
+  Offset latLngToPoint(LatLng latlng, double zoom) {
     final (x, y) = latLngToXY(latlng, scale(zoom));
-    return Point<double>(x, y);
+    return Offset(x, y);
   }
```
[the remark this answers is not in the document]
mootw: I think that is out of scope of this PR. and generally, while i dislike having "wrappers", the MapCamera ones add zoom.

lib/src/layer/polygon_layer/projected_polygon.dart
```diff
@@ -18,11 +18,11 @@ class _ProjectedPolygon<R extends Object> with HitDetectableElement<R> {
   _ProjectedPolygon._fromPolygon(Projection projection, Polygon<R> polygon)
       : this._(
           polygon: polygon,
-          points: List<DoublePoint>.generate(
+          points: List<Offset>.generate(
             polygon.points.length,
             (j) {
               final (x, y) = projection.projectXY(polygon.points[j]);
-              return DoublePoint(x, y);
+              return Offset(x, y);
```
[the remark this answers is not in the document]
mootw: This just wraps to an offset, I think it is okay for this wrapper to exist as it is frequently used

mootw:
Tried to migrate Bounds entirely, however it requires TileCoordinates which extends Point<int>. I also tried to convert Bounds to just use doubles (this is fine arguably, because integers are only representable to 2^53 on web due to javascript representing ints as doubles. I decided against pushing that for this PR since it got a little messy with tilecoordinates and converting all of the existing ints with doubles would yield confusion, and as-is the PR allows for a 90% conversion to using Offset internally and externally, already saving a lot of complexity which other PRs could expand on in the future. The only thing is there would likely be other breaking changes in the future PRs which isn't ideal for end-users, so it will be something that is done over time. I just do not have the time to invest in doing a 100% proper migration, which would end up touching so many things right now.

mootw:
I am also going to migrate "size" values to use the `Size` class which implements `OffsetBase`. It makes a big difference in readability of the code and shows where redundancies are

lib/src/misc/bounds.dart
mootw: Update, I migrated everything

lib/src/geo/crs.dart
```diff
@@ -52,13 +53,13 @@ abstract class Crs {
   (double, double) latLngToXY(LatLng latlng, double scale);
 
   /// Similar to [latLngToXY] but converts the XY coordinates to a [Point].
```
JaffaKetchup, line 55:
```suggestion
  /// Similar to [latLngToXY] but converts the XY coordinates to an [Offset].
```

lib/src/geo/crs.dart
```diff
@@ -371,18 +371,18 @@ class Proj4Crs extends Crs {
 /// Inherit from this class if you want to create or implement your own CRS.
 @immutable
 abstract class Projection {
-  /// The [Bounds] for the coordinates of this [Projection].
-  final Bounds<double>? bounds;
+  /// The [IntegerBounds] for the coordinates of this [Projection].
```
JaffaKetchup, line 374:
```suggestion
  /// The bounds for the coordinates of this [Projection].
```

lib/src/geo/crs.dart
```diff
@@ -371,18 +371,18 @@ class Proj4Crs extends Crs {
...
+  final Rect? bounds;
 
   /// Base constructor for the abstract [Projection] class that sets the
   /// required fields.
   const Projection(this.bounds);
 
   /// Converts a [LatLng] to a coordinates and returns them as [Point] object.
```
JaffaKetchup, line 381:
```suggestion
  /// Converts a [LatLng] to a coordinates and returns them as an [Offset].
```

lib/src/layer/shared/layer_interactivity/internal_hit_detectable.dart
```diff
@@ -98,7 +96,7 @@ base mixin HitTestRequiresCameraOrigin<R extends Object,
   /// Calculated [MapCamera] origin, using the following formula:
   ///
   /// ```dart
-  /// camera.project(camera.center).toOffset() - camera.size.toOffset() / 2
+  /// camera.project(camera.center) - camera.size / 2
```
JaffaKetchup, line 99:
```suggestion
  /// camera.project(camera.center) - camera.size.center(Offset.zero)
```

lib/src/map/camera/camera.dart
```diff
@@ -249,87 +248,89 @@ class MapCamera {
...
 
     return point - nonRotatedPixelOrigin;
   }
 
   /// Calculate the [LatLng] coordinates for a [localPoint].
-  LatLng pointToLatLng(Point localPoint) {
-    final localPointCenterDistance = Point(
-      (nonRotatedSize.x / 2) - localPoint.x,
-      (nonRotatedSize.y / 2) - localPoint.y,
+  LatLng offsetToLatLng(Offset localPoint) {
```
JaffaKetchup, line 300: Should this be called `screenOffsetToLatLng` for consitency, or are they different?

lib/src/map/camera/camera.dart
```diff
@@ -249,87 +248,89 @@ class MapCamera {
...
   }
 
   /// Sometimes we need to make allowances that a rotation already exists, so
   /// it needs to be reversed (pointToLatLng), and sometimes we want to use
   /// the same rotation to create a new position (latLngToScreenpoint).
   /// counterRotation just makes allowances this for this.
-  Point<double> rotatePoint(
-    Point<double> mapCenter,
-    Point<double> point, {
+  Offset rotateOffset(
```
JaffaKetchup, line 320: Whilst we use 'point' for variable and parameter names, I would suggest keeping this as `rotatePoint`, but I'm not sure (it would be inconsistent with other external methods). What do you think?

JaffaKetchup (changes requested):
Just a couple things left, mostly nits.

lib/src/map/camera/camera.dart
mootw: agreed!

JaffaKetchup (approved):
LGTM 🎉. Performance seems to have improved (about 0.2ms on home page example app, 0.3-0.4ms on basic polygon page, untested others), enough to actually make things feel slightly smoother (but maybe that's just a placebo). Don't notice any bugs, and code looks good.

josxha (approved):
I like these changes. 😄 
Maybe a feature release would be good before merging breaking changes?

Edit: I wouldn't be sad if we'd not use records here, but pr is fine as is too. (:
```dart
final (minx, miny) =
        transformation.transform(b.min.dx, b.min.dy, zoomScale);
    final (maxx, maxy) =
        transformation.transform(b.max.dx, b.max.dy, zoomScale);
```

mootw:
i like that for future changes. I want to avoid creeping scope on this PR too much for now. we could wait to release for more changes

lib/src/map/controller/map_controller_impl.dart
```diff
@@ -254,17 +244,14 @@ class MapControllerImpl extends ValueNotifier<_MapControllerState>
     }
 
     final rotationDiff = degree - camera.rotation;
-    final rotationCenter = camera.project(camera.center) +
-        (point != null
-                ? (point - (camera.nonRotatedSize / 2.0))
-                : Point(offset!.dx, offset.dy))
-            .rotate(camera.rotationRad);
+    final rotationCenter = (camera.projectAtZoom(camera.center) + offset!)
```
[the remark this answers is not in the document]
JaffaKetchup: Hi @hket-tommyluk,
Sorry if we've made a mistake! Can you please file a bug report if this is causing an issue and we'll look into it. I'm not clear what this would cause.

tlserver: This is wrong. It should be without parentheses.
```
    final rotationCenter = camera.projectAtZoom(camera.center) + 
        offset!.rotate(camera.rotationRad);
```
Sorry, commented from wrong account before.

PR #2007, opened 2025-01-06
fix: resolve most tile pruning/state issues

### LINKED ISSUE #1837: [BUG] Old tiles remain visible beneath new tiles after zoom until new gesture
### What is the bug?

After a fast change in zoom, my simple state/country border webp tiles do not redraw correctly. My GUESS is that the "old" tiles are not being REMOVED until a map interaction. All other tiles redraw fine.

### How can we reproduce it?

I am using transparent webp tiles for my maps.  They have black or white state/country lines, that's it. They are actually VERY small, the largest being 1-2kb.

Make a set of transparent tiles with only state boundaries (or I can send you mine),  then zoom in or out fast to the point of needing to download a new zoom level of tiles.

After the zoom gesture, you will notice the lines appear "brighter".  That is because temporarily, there are two layers of tiles loaded.  Since the old (now stretched) tile image from the lower zoom level has "thicker" lines, I can actually see the new lines plot on top of the previous lines. Then when I move the map, the old/stretched tiles disappear.  

- Can someone please contact me privately so I can send you a video of the behavior? I would rather not showcase the app yet, thank you.

- I can upload 2 or 3 dirs of my tiles if you would like to test them.

Just a simple tile layer:

                    TileLayer(
                        tileProvider: CancellableNetworkTileProvider(),
                        urlTemplate: 'https://site.com/{z}/{x}/{y}.webp',
                        tms: false,
                        minNativeZoom: 4,
                        maxNativeZoom: 10,
                      ),

Since I have never seen this issue before with all my other tiles, it makes me wonder:

Has this been going on before, but I only noticed now because these are transparent tiles? In other words, the previous set of tiles remain on the screen, but most of the time, with opaque tiles, you don't see it?  Anyway, it would be nice if once the new tiles are downloaded, the old would immediately disappear, without a map gesture.

Also, I have ruled out: Cancellable vs NetworkTileProvider...no difference.
Also commended out keep/pan buffers and tileDisplay property.  No difference.

### Platforms

Galaxy S23 

### Severity

mootw:
fixes #1837

there are still some instances I can get the tiles to not prune when zooming, but i am not sure the exact conditions. it might be related to multi-level zooms, or prune timers? though it seems like they all appropriately callback.

As part of this PR i also looked to add optimistic tile pruning of the parent when all 4 children are loaded, however quickly ran into state consistency issues. This is a simple quick fix to resolve most of the problem, but the tile layer really needs a rewrite as some of the state is held inside of a `Manager` which cannot call setState when making changes to the tiles, leading to visual bugs, like this one. I tried to add a setState callback, but ran into flutter issues (as it is generally bad practice) to do this.

So as a mostly-fix, I am adding a setState callback, which adds a few more calls to setState, but the performance impact is minimal (meaning only causing a few more widget draws, which isn't a big deal at all, however we should be able to just call setState under the manager and only update when we update the tiles), and not making the widget drawing slower. I did not notice a meaningful difference in the numbe of rendered frames after moving the map/zooming/animating when profiling.

This setState call also conveniently fixes another bug where changing the tileUrl didn't always update the drawing when it finishes loading.

### DIFF
--- a/lib/src/layer/tile_layer/tile_layer.dart
+++ b/lib/src/layer/tile_layer/tile_layer.dart
@@ -749,6 +749,7 @@ class _TileLayerState extends State<TileLayer> with TickerProviderStateMixin {
       pruneBuffer: max(widget.panBuffer, widget.keepBuffer),
       evictStrategy: widget.evictErrorTileStrategy,
     );
+    setState(() {});
   }
 
   bool _outsideZoomLimits(num zoom) =>


JaffaKetchup:
> This setState call also conveniently fixes another bug where changing the tileUrl didn't always update the drawing when it finishes loading.

This is the purpose of `reset` I believe, and there was previously an option on the `TileLayer` which changed whether it would be reloaded when a URL was changed. I wonder if the bug was introduced when that was removed. We might also think about removing `reset`?

JaffaKetchup (approved):

PR #2011, opened 2025-01-14
fix: use `Client` instead of `BaseClient` in network tile/image provider

### LINKED ISSUE #2010: [BUG] Incorrect http class used BaseClient is mixin, should be interface Client
### What is the bug?

NetworkTileProvider and MapNetworkImageProvider cannot be used with overridden Clients.

### How can we reproduce it?

Supply a overridden Client and classes require incorrect BaseClient implementation and not interface.

I have a PR being submitted

### Platforms

All

### Severity

slightfoot:
Fixes #2010

## Issue
`NetworkTileProvider` and `MapNetworkImageProvider` cannot be used with implementations of the http `Client` library.

## Resolution
Switch out incorrect `BaseClient` usage with `Client`.

## References

- https://pub.dev/documentation/http/latest/http/BaseClient-class.html
- https://pub.dev/documentation/http/latest/http/Client-class.html

## Notes

This fix uses the interface class and not the mixin implementation class that is used at the moment.

This change is backwards compatible and will not affect current usage of these classes as BaseClient implements Client interface.

@JaffaKetchup I am working on a project that would need this change. I'd be most greatful if we can get this out in a release. Thanks!

### DIFF
--- a/lib/src/layer/tile_layer/tile_provider/network_image_provider.dart
+++ b/lib/src/layer/tile_layer/tile_provider/network_image_provider.dart
@@ -32,7 +32,7 @@ class MapNetworkImageProvider extends ImageProvider<MapNetworkImageProvider> {
   /// The HTTP client to use to make network requests
   ///
   /// Not included in [operator==].
-  final BaseClient httpClient;
+  final Client httpClient;
 
   /// Whether to ignore exceptions and errors that occur whilst fetching tiles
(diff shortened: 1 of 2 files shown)


TesteurManiak (approved):
LGTM! 👍

JaffaKetchup:
Hi @slightfoot, we plan to release the next major version soon :)

PR #2012, opened 2025-01-14
fix: avoid closing externally created `http.Client` in `NetworkTileProvider`

### LINKED ISSUE #2009: [BUG] `NetworkTileProvider` and `CancellableNetworkTileProvider` close externally created clients
### What is the bug?

`NetworkTileProvider` and `CancellableNetworkTileProvider` close `httpClient`/`dioClient` even when the client was created externally and may be used elsewhere for other purposes.

The tile provider should only dispose the client when the client was also created by the tile provider.

### How can we reproduce it?

The issue can be seen in the `NetworkTileProvider` source code:

```
  @override
  Future<void> dispose() async {
    if (_tilesInProgress.isNotEmpty) {
      await Future.wait(_tilesInProgress.values.map((c) => c.future));
    }
    // Both internally and externally created clients are closed 
    // when the tile layer/tile provider is disposed.
    _httpClient.close();
    super.dispose();
  }
```

A possible fix is to keep track of whether or not the client was created internally:

```
class NetworkTileProvider extends TileProvider {
  NetworkTileProvider({
    super.headers,
    BaseClient? httpClient,
    this.silenceExceptions = false,
  })  : _closeClientOnDispose = (httpClient == null),
        _httpClient = httpClient ?? RetryClient(Client());

  final bool silenceExceptions;
  final BaseClient _httpClient;
  final bool _closeClientOnDispose;

  // ...

  @override
  Future<void> dispose() async {
    if (_tilesInProgress.isNotEmpty) {
      await Future.wait(_tilesInProgress.values.map((c) => c.future));
    }

    if (_closeClientOnDispose) {
      _httpClient.close();
    }

    super.dispose();
  }
}
```

### Severity

JaffaKetchup:
Fixes #2009

### DIFF
--- a/lib/src/layer/tile_layer/tile_provider/network_tile_provider.dart
+++ b/lib/src/layer/tile_layer/tile_provider/network_tile_provider.dart
@@ -36,7 +36,8 @@ class NetworkTileProvider extends TileProvider {
     super.headers,
     Client? httpClient,
     this.silenceExceptions = false,
-  }) : _httpClient = httpClient ?? RetryClient(Client());
+  })  : _isInternallyCreatedClient = httpClient == null,
+        _httpClient = httpClient ?? RetryClient(Client());
 
   /// Whether to ignore exceptions and errors that occur whilst fetching tiles
   /// over the network, and just return a transparent tile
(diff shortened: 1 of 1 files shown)


mootw (approved):
lgtm

Issue #1970, opened 2024-09-23
[BUG] `CircleMarker.useRadiusInMeter` also applied to border stroke width


JaffaKetchup:
### What is the bug?

When `CircleMarker.useRadiusInMeter` is `true`, it's effect is also applied to the border width. Ideally, it should be a seperate property.

### How can we reproduce it?

Set `CircleMarker.useRadiusInMeter` `true` and a border radius of non-zero thickness.

### Do you have a potential solution?

Add a seperate property.

### Platforms

All

### Severity

Obtrusive: Prevents normal functioning but causes no errors in the console

PR #2018, opened 2025-01-26
feat: add multi-world support to `CircleMarker`s

monsieurtanuki:
Now `CircleMarker`s are replicated on all visible worlds.
That includes both display and hit detection.
Some refactoring involved.

Impacted files:
* `multi_worlds.dart`: added a couple of `CircleMarker`s
* `painter.dart`: added replication of circle display and hit detection on all visible worlds

Screenshot_1737904593

Screenshot_1737904630

### DIFF
--- a/example/lib/pages/circle.dart
+++ b/example/lib/pages/circle.dart
@@ -21,12 +21,15 @@ class _CirclePageState extends State<CirclePage> {
   List<HitValue>? _prevHitValues;
   List<CircleMarker<HitValue>>? _hoverCircles;
 
+  static const double _initialBorderStrokeWidth = 2;
+  static const double _hoverBorderStrokeWidth = 15;
+
   final _circlesRaw = <CircleMarker<HitValue>>[
     CircleMarker(
       point: const LatLng(51.5, -0.09),
       color: Colors.white.withAlpha(178),
       borderColor: Colors.black,
-      borderStrokeWidth: 2,
+      borderStrokeWidth: _initialBorderStrokeWidth,
       useRadiusInMeter: false,
       radius: 100,
       hitValue: (title: 'White', subtitle: 'Radius in logical pixels'),
@@ -35,7 +38,7 @@ class _CirclePageState extends State<CirclePage> {
       point: const LatLng(51.5, -0.09),
       color: Colors.black.withAlpha(178),
       borderColor: Colors.black,
-      borderStrokeWidth: 2,
+      borderStrokeWidth: _initialBorderStrokeWidth,
       useRadiusInMeter: false,
       radius: 50,
       hitValue: (
@@ -48,9 +51,10 @@ class _CirclePageState extends State<CirclePage> {
       // Dorney Lake is ~2km long
       color: Colors.green.withAlpha(229),
       borderColor: Colors.black,
-      borderStrokeWidth: 2,
+      borderStrokeWidth: _initialBorderStrokeWidth,
       useRadiusInMeter: true,
-      radius: 1000, // 1000 meters
+      radius: 1000,
+      // 1000 meters
       hitValue: (
         title: 'Green',
         subtitle: 'Radius in meters, calibrated over ~2km rowing lake'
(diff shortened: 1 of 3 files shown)


JaffaKetchup (approved):
LGTM, thanks :)

JaffaKetchup (changes requested):
Hey @monsieurtanuki, thanks for everything, it looks like it's working correctly. However, it looks like the circle test has broken and is taking too long? I'm not sure why that would be, do you have any idea?

JaffaKetchup (changes requested):
This seems to have broken something, sorry :|. It looks like the radius 'ring' is no longer painted correctly, and appears detached from the circle itself. See master vs PR below.

image

image

monsieurtanuki:
Hi @JaffaKetchup!
I did change something about the border (something like adding half borderWidth to the circle radius), but that's because the result wasn't correct (without that, the border overlapped the partially transparent disk).
Is your screenshot taken from an existing example, or would you share the related code?

JaffaKetchup:
It's taken from the circles example. I have experimented with the painter recently to resolve that exact issue and the entire painter seems to be quite fragile IYSWIM. I did find a way to do it, but I'm not sure how resilient it is (for example to varying opacities etc.).

1. Move the "`paintBorder`" block to the end
2. Re-enable anti-aliasing on the first "`paintPoint`" block
3. Do this in the "`paintBorder`" block:
	```diff
	- _paintCircle(canvas, offset, radius, radiusPaint);
	+ _paintCircle(canvas, offset, radius + borderWidth / 2, radiusPaint);
	```

monsieurtanuki:
Hi @JaffaKetchup!

First remark: in the "old" version there's a bug regarding the border. For optimization reasons, we have a slightly different way of displaying borders when the disk is opaque or not. An undesired side-effect is that in one case the border width is considered (potentially) in meters, and in the other case it's considered in pixels.
### "your" version, opaque, with an invisible 2-meter border width
Screenshot_1738137797
### "your" version, with alpha and a visible 2-pixel border width:
Screenshot_1738137779
First question: obviously the border width behavior should be the same in both opaque/withAlpha cases, but which one is the correct one? I guess it's the border width always as pixel, right? cf. #1970.
For the record in the current PR I considered the border width as potentially in meters, and I'm about to fix that.

For the "hovered" example with additional green border, I think there's a couple of bugs too.
This is the code of the additional "hover" green border:
```dart
CircleMarker<HitValue>(
  point: original.point,
  radius: original.radius + 6.5,
  useRadiusInMeter: original.useRadiusInMeter,
  color: Colors.transparent,
  borderStrokeWidth: 15,
  borderColor: Colors.green,
);
```
For the record the initial border is `2` in both disks.
In your version, the additional green border overlaps the initial black border - we don't see the black border anymore.
My assumption is that we should display a green circle around the existing disk, with no overlapping.
That means that we need the initial radius + half the initial border + half the green circle border, therefore 2/2+15/2 = 8.5.

Besides, that doesn't make sense to add pixels to meters, so `original.radius + 6.5` (or `original.radius + 8.5`, whatever) cannot really be applied with `useRadiusInMeter: true`.

With my latest push:
Screenshot_1738166020
Screenshot_1738166028

JaffaKetchup:
> An undesired side-effect is that in one case the border width is considered (potentially) in meters, and in the other case it's considered in pixels.

Ah, so this would explain https://github.com/fleaflet/flutter_map/issues/1970 and why I struggled to reproduce it consistenly the other day, because I had the root cause wrong.

Yes, it should always be in pixels - or a separate parameter to use meters, but that wouldn't be consistent with polylines, so that's out of scope for now.

> In your version, the additional green border overlaps the initial black border - we don't see the black border anymore.

Actually, that was what I was aiming for, but it doesn't really matter and maybe this makes more sense. I was also aware when writing it that it wouldn't work for meters but I think it just ignored it maybe?

JaffaKetchup (changes requested):
Hmm, there's yet another issue now (sorry!) :D. It only seems to happen with the multi-worlds example page.

https://github.com/user-attachments/assets/212642fa-ee4f-499b-8f36-b829f742a6a9

monsieurtanuki:
> Hmm, there's yet another issue now (sorry!) :D. It only seems to happen with the multi-worlds example page.

Of course, I fixed the border width to be always in pixels. But kept the value in meter. 100k...
Sorry about that.
About to push.

JaffaKetchup (approved):
LGTM, thanks :)

PR #2021, opened 2025-01-27
chore: prepare for v8 release

JaffaKetchup:
The usual + regenerated example app configs once again + fork the geojson package to fix conflict.

Once FM v8 is released, I can make a proper version to submit a PR to the geojson package to support v8. Then we can update the example app again.

Includes #1987 & #2018 in the CHANGELOG.


analysis_options.yaml
```diff
@@ -17,7 +17,6 @@ linter:
     cancel_subscriptions: true
     close_sinks: true
     directives_ordering: true
-    package_api_docs: true
```
josxha, line 20: What is the reason to remove this lint rule?

example/pubspec.yaml
```diff
@@ -25,6 +25,10 @@ dependencies:
 dependency_overrides:
   flutter_map:
     path: ../
+  flutter_map_geojson:
+    git:
+      url: https://github.com/JaffaKetchup/flutter_map_geojson.git
+      ref: fm-v8
```
josxha, line 31: Maybe it's worth to think about a first party solution if we rely that havily on the package?

analysis_options.yaml
JaffaKetchup: Removed in the latest versions of Flutter. Maybe only on beta for now.

josxha: > NOTE: This lint is deprecated because it is has not been fully functional since at least Dart 2.0. Remove all inclusions of this lint from your analysis options.

Oh that's interesting.

example/pubspec.yaml
JaffaKetchup: It would be nice, but I don't have time to do that right now. I guess the ideal thing to do would be to preprocess the geojson data into some form that's  very easy to decode (given we don't use very much of its functionality, maybe just a list of numbers or something).
If the PR I make isn't accepted I can look into this.

analysis_options.yaml
JaffaKetchup: Yeah, bit strange. Always worked well for me.

example/pubspec.yaml
josxha: This comment is no merge blocker. Just a note 😄

josxha (approved):

example/pubspec.yaml
josxha: > preprocess the geojson data into some form that's very easy to decode

Yes that would be ideal.

PR #2025, opened 2025-01-30
fix: invoke map interaction callbacks with coordinates in primary world

### LINKED ISSUE #2008: [BUG] Map callbacks with coordinate arguments do not work correctly when invoked on 'non-primary' world
### What is the bug?

The `onTap` (and probably the rest of the handlers in `MapOptions`) callback in the Markers page in the example does not add the marker to the location tapped, if that location is not in the main world. It adds it to the closest point in the main world.

See https://github.com/fleaflet/flutter_map/pull/2000#pullrequestreview-2516201365.

_Note that this bug does not yet exist in a released version._

### How can we reproduce it?

Zoom out in the Markers example, so that a world is not the main 'focus' (centered) in the view. Tap to add a marker to this world. It is added along the anti-meridian at the same latitude, in the main world.

_No response_

### Platforms

All (unreleased)

### Severity

monsieurtanuki:
### What
- Tapping on the map will be redirected to the main world, instead of being truncated on a -180 or 180 longitude.
- Fixes #2008.

### Impacted files
* `camera.dart`: now looking for the best world before unprojecting a pixel
* `multi_worlds.dart`: added an `onTap` event to create on the fly markers

### Adding markers around New Zealand and longitude 180
Screenshot_1738239927

### Adding markers in a multi-world
Screenshot_1738236612

### DIFF
--- a/example/lib/pages/multi_worlds.dart
+++ b/example/lib/pages/multi_worlds.dart
@@ -17,6 +17,24 @@ class MultiWorldsPage extends StatefulWidget {
 class _MultiWorldsPageState extends State<MultiWorldsPage> {
   final LayerHitNotifier<String> _hitNotifier = ValueNotifier(null);
 
+  final _customMarkers = <Marker>[];
+
+  Marker _buildPin(LatLng point) => Marker(
+        point: point,
+        width: 60,
+        height: 60,
+        child: GestureDetector(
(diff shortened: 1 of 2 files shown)


JaffaKetchup (approved):
LGTM, thanks!

PR #2035, opened 2025-02-16
fix: ensure movement gestures emit events when starting

### LINKED ISSUE #1939: Gesture start events not received
### What is the bug?

Hello everyone! I tried to find the same issue, but I didn't find it. If you forward mapController to FlutterMap widget, and will be listen mapEventStream from mapController, you will receive onMultiFinger and onDrag events, while trigger onDrag and onMultiFinger gestures accordingly. You also will be receive dragEnd and multiFingerEnd events, but dragStart and multiFingerStart doesn't. Is it expected behaviour, or bug?

### How can we reproduce it?

Take FlutterMap widget, forward mapController to it, and start listen mapEventStream from mapController in main, with printing events to console. 

I didn't learn code, responsible for sending mapEvents, but I think, this bug doesn't trigger big code changes

### Platforms

Mac Os Sonoma

### Severity

JaffaKetchup:
Fixes #1939.

### DIFF
--- a/lib/src/gestures/map_interactive_viewer.dart
+++ b/lib/src/gestures/map_interactive_viewer.dart
@@ -528,6 +528,7 @@ class MapInteractiveViewerState extends State<MapInteractiveViewer>
   void _handleScaleStart(ScaleStartDetails details) {
     _dragMode = _pointerCounter == 1;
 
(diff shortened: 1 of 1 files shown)


mootw (approved):
lgtm!

PR #2046, opened 2025-02-26
feat: add inverted fill option to `PolygonLayer`

### LINKED ISSUE #2034: [💰] Add option for `PolygonLayer` to have fill color around `Polygon`s
### What is the bug?

When attempting to display a `Polygon` that is large, it fails to display incorrectly with the new support for unbounded horizontal scrolling. When the polygon has a hole defined, the hole may be shown inverted in some cases, or not at all (culled incorrectly?) in others.

The issue lies in the new `projectList` method in the CRS. 

https://github.com/fleaflet/flutter_map/blob/e9272e28f910d4433aba662fd58753c4a345c424/lib/src/geo/crs.dart#L420-L443

(Originally reported in the Discord server: https://discord.com/channels/951867686378409984/1339879842610675794)

### How can we reproduce it?

Use a polygon with a defined set of points creating a large area that spans nearly to the edge of the world, with a fill color.

If the `Polygon`'s points are as such, then no polygon appears at all.

```dart
points: const [
  LatLng(-90, -180),
  LatLng(90, -180),
  LatLng(90, 180),
  LatLng(-90, 180),
],
```

If the points are as such, then the issue can be more clearly seen. The fill is applied to the areas between the worlds, outside of the polygon itself.

```dart
points: const [
  LatLng(-90, -180),
  LatLng(90, -180),
  LatLng(90, 180),
  LatLng(-90, 180),
],
```

On adding a hole, for example as so, then it may be shown correctly in some cases, or completely removed in others.

```dart
holePointsList: [
  [
    LatLng(51.5, -0.09),
    LatLng(51.5, -1.09),
    LatLng(52.5, -1.09),
    LatLng(52.5, -0.09),
  ]
],
```

Image

_No response_

---

## Bounty

This issue is eligible for a bounty!

This continues from #1582, and also includes a whole load of other work done since then to improve unbounded horizontal scrolling support.

### Bounty Criteria

To be eligible to receive the bounty (as outlined in the meta information below), the solution must:
* implement a parameter on the `PolygonLayer` to fill the space outside of any polygon with any color
* provide the expected result (to be discussed in further detail during implementation) in a wide variety of cases, tested on the example application

The bounty is also awarded via this bounty for a number of other issue resolutions related to unbounded horizontal scrolling support.

Due to circumstances outside of the control of any contributor (bugs within Flutter), the solution does not need to work fully on the web platform.

The decision to award the bounty is entirely at the discretion of the maintainer team.

### Bounty Meta Information

flutter_map uses [Polar.sh](https://polar.sh/fleaflet) to provide a bounty on this issue! Contributors who resolve this issue using the criteria listed above will receive the full amount pledged below, after all fees have been paid.

The flutter_map team funds issues we want to see fixed but are relatively low priority for us, and we do this using your donations! When you [donate](https://docs.fleaflet.dev/supporters#support-us), 15% goes to the OpenStreetMap Foundation, and the rest is held as donations, some of which we fund these issues with!

Of course, we appriciate any size donation you can make: it supports us directly, and allows us to fund these issues, but if this issue is particularly important to you, you can pledge money directly to this issue using the button below.
<br><br>

<a href="https://polar.sh/fleaflet/flutter_map/issues/2034">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/fleaflet/flutter_map/issues/2034/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

monsieurtanuki:
### What
- Introduced a `bool justHoles` field for polygon:
  - the "full" points won't be considered and will be replaced by the full map
  - the "holes" points will be considered as the only "not colored" parts of the polygon
- New example named "Advanced polygons".
- The typical use-case would be using a single polygon with several holes.
- In that case only the holes are clickable.
- Could be that eventually we may consider that feature to be applied to the map instead, and have all polygons considered as holes. The thing is that currently we cannot say: "just hit hole number 2", because the label is attached to the "polygon", and all holes share the same polygon.
- cc @enricostrijks and #2034

Fixes #2034. Potentially eligible for bounty.

### Screenshots
| new menu item | example |
| -- | -- |
| Screenshot_1740564922 | Screenshot_1740564737 |

| multi-world example | hit example |
| -- | -- |
| Screenshot_1740564526 | Screenshot_1740564628 |

### Files
New file:
* `advanced_polygons.dart`: Example dedicated to polygons with advanced features.

Impacted files:
* `main.dart`: added the new `AdvancedPolygonsPage` page
* `menu_drawer.dart`: added the new `AdvancedPolygonsPage` page
* `painter.dart`: managed the `justHoles` case; refactored a bit `borderPaint`
* `polygon.dart`: added the `bool justHoles` field
* `polygon_layer.dart`: special case for `justHoles`
* `projected_polygon.dart`: special case for `justHoles


enricostrijks:
LGTM

JaffaKetchup:
Thanks :). I've made a couple changes, namely creating a new constructor for the polygons instead of a property, which allows us to enforce some usage patterns, and renamed it to "inverted".

One question, do we always want the hitting to be inverted? Surely sometimes they want to detect not hitting a hole still?

monsieurtanuki:
> Thanks :). I've made a couple changes, namely creating a new constructor for the polygons instead of a property, which allows us to enforce some usage patterns, and renamed it to "inverted".

I'm OK with that.

> Two questions, do we always want the hitting to be inverted? Surely sometimes they want to detect not hitting a hole still? Also, it's now impossible to have multiple holes with different `hitValues` - is there anything we can do about this (I'm aware that's quite a tricky problem in itself, and that you've now done quite a lot of work extra for all of this!)

As already mentioned, it's a bit strange to set this property to a polygon, as it impacts the whole map (e.g. the alpha background).
Perhaps a map property called `bool invertedPolygons` instead? In that case we would have more control over `hitValues`, as we would be displaying several distinct polygons, but as holes from the map.
About the hitting being inverted or not, it could indeed be deactivated, so that the hit ignores the polygons altogether.

All of that depends on typical use-cases:
* would it be expected to display inverted polygons and "normal" polygons on the same map?
* would it be expected to display inverted polygons with holes?

JaffaKetchup:
I would say we should support both those use-cases - as we did before.

That becomes a really difficult solution I guess. A solution is to allow an entire `PolygonLayer` to be inverted instead (`PolygonLayer.inverted`)? At that point, I'm not sure how much that would complicate the painter. And we'd somehow have to factor in a color to the layer. Maybe `Color? invertedFill`?

I'm also not 100% sure still why we couldn't make the current usage pattern work, also for the `PolylineLayer`. Would it be possible to have a property that stops the element from passing over the anti-meridian? I haven't tested it, but surely this current solution would still have the exact same issue as before. As unlikely the real world use case, what if someone wanted something like below (ofc they could just use multiple polygons, but as far as I understand, they can at the moment as well):

image

monsieurtanuki:
> I would say we should support both those use-cases - as we did before.

Ok.

> That becomes a really difficult solution I guess [...]

`PolygonLayer.inverted` with `Color invertedFill` sounds reasonable.
Or even just `Color? invertedFill` without any new constructor.

> I'm also not 100% sure still why we couldn't make the current usage pattern work, also for the `PolylineLayer`. Would it be possible to have a property that stops the element from passing over the anti-meridian?

Just adding intermediary longitude points would do the trick, as the next point of a poly is the closest (the problem is when you try to jump _directly_ from -180 to 180).

> I haven't tested it, but surely this current solution would still have the exact same issue as before. As unlikely the real world use case, what if someone wanted something like below (ofc they could just use multiple polygons, but as far as I understand, they can at the moment as well):
> 
> image

No problem with intermediary longitude points.

That wouldn't support the "hit an inverted polygon/hole" feature, though.

monsieurtanuki:
Fun fact: with my 1980's background the "inverted" polygon feature looks a bit like what was then called "video reverse". But it wasn't easy to find that term on the web now because I rather found tutorials on how to play a video backwards - which wasn't a topic back then.

https://en.m.wikipedia.org/wiki/Reverse_video

JaffaKetchup:
Yeah, so I guess we introduce a new property `Color? invertedFill`, and we can also introduce `bool invertHitting`. We don't particularly need a new constructor for it, and both of those properties can be used seperately. `Polygon.fillColor` then applies to holes in each polygon. How much do you think that would complicate the painter?

(Yep, I can imagine "reverse video" being difficult to search for :D)

JaffaKetchup:
I have realised that `bool invertHitting` doesn't really make sense. Instead, we could introduce a `R? hitValue` on the layer level, returning that as hit if no elements are hit. Again, this could be used independently.

JaffaKetchup:
> The more I think about it, the more I do believe we should put the "inverted" info at the PolygonLayer level, with a new Color? invertedFill field.

Agreed.

> We could even have a fully alpha map (except for the polygons inside this layer) PLUS alpha polygons with a different color.
Basically, that would say that we would somehow keep this PR's display of polygons as holes, and then display normally each polygon.

I'm not quite following what you mean here? If you mean moving the inverted option all the way to `MapOptions`, then that might not give enough flexibility.

In regards to allowing the hit testing to be inverted, I've opened #2050. It also gives some other options, and applies to all the feature layers.

monsieurtanuki:
> I'm not quite following what you mean here?

I'll show you a screenshot when I'm ready.

> In regards to allowing the hit testing to be inverted, I've opened #2050.

I'm not sure we need the hit testing to be inverted, or at least I cannot imagine a typical use-case.

monsieurtanuki:
@JaffaKetchup This is what I'm coding right now:
* `Color? invertedFill` at the layer level
* "normal" polygons, with points
* clickable polygons
* polygons with different border colors, potentially with no fill color

Something like that:
Screenshot_1741339449

monsieurtanuki:
@JaffaKetchup I think I've nailed it (latest push).
Screenshot_1741346997
Screenshot_1741347008
Screenshot_1741347033

JaffaKetchup (changes requested):
This has really complicated things huh :D. Thanks for spending even more time on this again.

Unfortunately I think this is the sort of thing really difficult to test and iron out all of the edge cases.

Here's one (inverting the existing example page) - the intersection of two overlapping polygons should be transparent (as both the polygons are), but the inverted color is shown instead:

image

image

When hovering, another transparent polygon is added. Again, it should be transparent through to the map. Instead, the aforementioned intersection becomes transparent, and the other area becomes the inverted fill. I guess the blending or winding is wrong again - there's no issue with transparent over filled, just transparent over transparent or intersecting.

image

monsieurtanuki:
@JaffaKetchup I guess the code needs some fixing.
I wouldn't say that things are "more complicated" now, though. I believe the "hole polygons" idea was a dead-end (e.g. for border styles and hit values).
The code is easier to read now, especially if you read the diffs from before this PR - and not from before the latest push.

We're now dealing with features not simple to apprehend: intersecting polygons on an inverse layer. Not an "of course we should display the map that way" situation.

I'm not sure I understand correctly all the issues, but I think what you'd want would be:
* painting all the map in black
* painting all the polygons (points and holes) as white
* use this "map" as a mask on top of the real map
* and then, display "normally" the polygons

Or something like that.

Another solution would be to say that the inverted polygon layer isn't compatible (temporarily?) with all features, e.g. intersecting polygons or holes.

JaffaKetchup:
Yeah, I think what you've said there is what I would be aiming for, but we want holes in inverted polygons to appear black (as they currently are in this PR). It should be the same behaviour as could be created before (by creating the big polygon with several holes).

One thing to consider is overlaps and outlines, and overlaps and holes. Take the following picture:

Untitled

What do we do with the hole area here, which overlaps the polygon underneath? What do we do with the outlines?

monsieurtanuki:
@JaffaKetchup Again, that's going too far, I cannot picture what we want in the end.

I suggest we switch from theoretical mode ("what alpha polygons (with partially outside holes) intersecting each other are supposed to look like on an inverted alpha polygon layer?") (no idea) to engineer mode ("alpha inverted polygon layers would be a nice feature - displaying full polygons makes sense, displaying polygons with realistic holes makes sense too, but beyond that there are probably no use-cases and anyway the output wouldn't be implicitly understandable) (which is probably in contradiction with the use of a map)

JaffaKetchup:
Ok, sorry, not 100% sure what you mean! I fully agree that the new approach is much better than doing it per-`Polygon`.

> displaying full polygons makes sense, displaying polygons with realistic holes makes sense too, but beyond that there are probably no use-cases and anyway the output wouldn't be implicitly understandable) (which is probably in contradiction with the use of a map)

I'm not suggesting necessarily doing anything beyond. In my mind, inverted means everything outside of any polygon should be the specified color. Alternatively, it could mean literally painting the opposite of what we're currently doing.
In that case, an intersection of two polygons in the normal view, if they were both transparent, left a transparency. Therefore, this should not leave a transparency - it should fill with the specified color. That is what this PR is currently doing. Maybe that's a good approach, but I feel as though it could be unexpected? Wdyt?
However, in the first case (everything outside of any polygon should be the specified color), the overlapping transparencies should remain transparent, since that area is still within a polygon. It also leads that holes should appear in the specified color since they are out of a polygon. But it is not clear to me what would be expected if the hole overlapped another polygon. This isn't true-inverse though, so I can see why maybe that's not clear.

monsieurtanuki:
@JaffaKetchup I'm about to push a new version, at least to make a clean diff with the master code.
As a work version, I added `invertedFill: Colors.orangeAccent.withAlpha(64),` to the first polygon layer of the "Polygons" example; I'll remove it later.

Now the code's much easier to understand.
Before the standard loop on polygons, we add an optional step: if `invertedFill` isn't `null`, we fill the map except all "full" polygons (no triangles, no holes).
Not sure if it matches all your expectations.

There's still one thing that doesn't quite work anyway.
In the "hover" case on the "Polygons" example, we temporarily add a copy of the selected polygon, with a thick green border.
The problem is that if you add twice the same polygon as a hole on the same `Path`, there's no hole anymore, they somehow neutralize each other.

A solution could be to add a subtile distinction between a `null` fill color and a transparent fill color. If we decided that the "green border" polygon had a `null` fill and that we don't add `null` filled polygons as holes, we could make the hover effect work. But there would probably be side-effects with legacy code.

A similar solution would be to add something like a new polygon field called `bool ignoreFill`: we could make the hover effect work with impact on the existing apps.

Tell me what you think.

image

JaffaKetchup:
As far as I can see, it still has the same 'issue'(?) as before:
image

Additionally, in this latest commit, it appears some polygons' holes are rendered transparently, not with the `invertedFill` - maybe only for non-performant-rendered polygons:
image

Here, the strange invalid polygons results in a darker color. In non-inverse mode, the areas appear in the same color as the rest of the polygon.
image

I think we're possibly still coming at this from different directions. My 'inverse' is everything that isn't in any polygon is painted with `inverseFill`. I'm not sure if that's what you're aiming for as well.

---

I've had a poke around, and I think some of the issues come from the way we're mixing fills. We're using `oddOrEven` - there's a complicated reason for this, but the short of it is that we don't need to worry about it anymore, and it looks as though using `Path.combine` has helped. I'm not 100% sure of the performance implications, and it needs thorough testing, but it's pretty much as I envisaged it. It resolves the issues above and the overlap issues.

Let me know what you think of 55f3a98.

JaffaKetchup:
Just looking at debugging to solve that second issue, if you change line 408 in the painter to `PathOperation.reverseDifference`, the polygon comes out like this:

image

Why do both holes not appear in orange?

monsieurtanuki:
Hi @JaffaKetchup!

I'm glad you had a positive experience with `Path.combine`. I used it in my tests but there was something wrong (maybe on my side).
I gave some reflection about it: I think the idea would be to "combine/union" all the full polygons, AND THEN subtract that combination from the "full map" polygon.
Anyway, your current solution doesn't look good on my Chrome browser, with obvious side-effects (I have a small screen).
| map | slightly moved |
| -- | -- |
| image | image |

Still trying to find a solution. My latest thought is that "combine/union" does really work (random results with `xor` and `difference`), therefore using a "mask" canvas may do the trick.

JaffaKetchup:
Yeah, there's quite a few bugs with `Path.combine` as it turns out, also with text rendering. On web WASM or non-WASM, the text appears in the wrong color?

image

Might have to file a few bug reports to Flutter, because it's making things difficult.

See https://github.com/flutter/flutter/issues/124675, which reported the issue (the OP was working on FM when he filed it I think). I forgot about this, and thought the only blocker was the lack of HTML support (which doesn't matter anymore). See also https://github.com/flutter/flutter/issues/149743. Not sure whether to open another report or not. Annoying that they've been forgotten about despite being P2, but maybe it's not a simple fix?

monsieurtanuki:
@JaffaKetchup The bugs around `Path.combine` won't probably be fixed in a near future, and may not be fixed similarly on all platforms.

That's why I'd go back to my previous idea of supporting invertedFill with limited polygon features only. Without using `Path.combine` but rather `evenOdd` - that we already use and seems reliable.

JaffaKetchup:
> That's why I'd go back to my previous idea of supporting invertedFill with limited polygon features only. Without using Path.combine but rather evenOdd - that we already use and seems reliable.

I don't think supporting limited polygon features is a good idea - what would you say would not make the cut? From what I can see, it's not necessarily the features of each polygon that are problematic, it's the interaction between multiple polygons, which we can't check for. It could lead to confusion. `evenOdd` doesn't seem to fully support what we want to do - but it's at least better.

I would rather get as close as possible to the correct behaviour on as many platforms as possible - given that it works almost perfectly everywhere but web (bar the second point of holes being filled wrong, which does need fixing) and maybe just disable or give a console logged warning (or some other obvious indication) on the web - maybe if it's easy to fallback to `evenOdd` we can do that, but that's more complexity.

I've added a comment to the old Flutter issue (https://github.com/flutter/flutter/issues/124675). With any luck it will get a bit of attention in the meantime. I might also open an issue to report the text color issues (inconsistency with default color on web vs native) - I'll see if I can manage to get some attention on the aforementioned issue via that as well. Although I have the same concerns as you about a fix anytime in the foreseeable future: given that the Wonderous app experienced this (a major 'customer') and it was not investigated further.

What do you think? Sorry again that this is turning out to be a larger job than expected - hopefully it's easy to copy across the other layers once we've got a solution, and I'll look into adding a bounty to this PR (given that the 50 USD ended up a lot more watered-down than expected last time)! Wdyt @fleaflet/maintainers?

monsieurtanuki:
@JaffaKetchup My point was to say to the users something like : "hey, we've got a new feature called "inverted `PolygonLayer`", but be careful that in some exotic cases (e.g. intersecting polygon holes) the rendering might be inaccurate, especially on web".

Regarding `Path.combine` versus `evenOdd`, the code is almost identical as we can start in both case with a loop on all polygons. Once it's done, we draw the polygons as we used to.

I'm not sure I would use `Path.combine` for normal polygon drawing (I mean, not the new inverted polygon layer preparatory step). At least not in that PR. The idea would be to say: "business as usual, we just had an optional inverted color feature on top". As opposed to: "oh and we changed the way we draw polygons too".
Or at least I would add a code comment about "we used to use evenOdd that worked OK too, switch back to it if needed".
Or, I would make it an optim parameter about "how we remove polygons from polygons", so far either `Path.combine/difference` or `evenOdd`. With the default to `evenOdd` for legacy reasons. That would help running perf tests too.

JaffaKetchup:
> but be careful that in some exotic cases (e.g. intersecting polygon holes) the rendering might be inaccurate, especially on web".

I think the bigger problem is that ANY polygon overlap of two transparent polygons (and given the feature, I'd say transparent polygons are likely) results in an issue. But yeah, overall, I agree - it might be a good approach to just say "problems could arise on web except for the simplest of use-cases". See below.

---

I wonder if perhaps some of the issues with `Path.combine` on the web is us not being precise enough - I wonder if native just has more tolerance with winding directions. 

Running with `Path.combine`. Make the layer non-inverse again. Take the orange non-malformed polygon and run on web. Notice how, on web, one of the holes is filled (incorrectly), but the others aren't. On native it's fine. Add `.reversed.toList()` to the the middle hole list on orange to quickly reverse it - this affects all the holes, and suddenly they've changed again on the web. Is it possible we need to be more careful with winding directions?

Then again, I still can't understand why the thing completely breaks when moving or zooming.

I feel like there must be a way of getting this working - but I have no good idea how, other than the ways suggested in this PR.

---

> Or, I would make it an optim parameter about "how we remove polygons from polygons", so far either Path.combine/difference or evenOdd. With the default to evenOdd for legacy reasons. That would help running perf tests too.

This sounds like the best idea. I would suggest the making the default platform dependent: on web, use `oddEven` for everything; on native, use `Path.combine` for everything.

On native, I can't see any downsides to `Path.combine` for non-inverted - we do need to test performance though, that could be a factor. When inverted, only `Path.combine` gives the correct result anyway.

On web, `Path.combine` seems just too broken. If the user enables inverted mode, we log a console warning (in debug mode).

---

Still a big blocker regardless is the transparency of holes - they should not be transparent. I guess in the first step, we need to cut them in(?) and fill with `invertedFill`?

---

It looks like we'll add a bounty to this issue, maybe a bit larger than last time. I'll write a spec - not sure yet if it'll include the same feature for polylines and circle markers.

monsieurtanuki:
I believe the "inverted holes" used to work: it's just that in the latest PRs we somehow ignored them and considered just the main polygon.
To be tested (I'll do that tomorrow): adding the holes when computing the inverted area.

monsieurtanuki:
@JaffaKetchup I've just tested on Android emulator the 4 cases: `evenOdd` or `Path.combine`, excluding the holes from the inversion or including them.
I guess your favorite combination is `Path.combine` + "include the holes".
For the record I also tested on Chrome browser, but `Path.combine` gives random results.
| x | excluding holes | including holes |
| -- | -- | -- |
| combine | Screenshot_1741855397 | Screenshot_1741855614 |
| evenOdd | Screenshot_1741855495 | Screenshot_1741855671 |

monsieurtanuki:
@JaffaKetchup I may have found a solution: filling the holes with inverted fill.
Quickly coded, probably some fine-tuning needed.
| filling holes | not filling holes |
| -- | -- |
| Screenshot_1741895399 | Screenshot_1741895331 |

JaffaKetchup:
That's looking good for sure! I'll see if I can give it a test tomorrow. Thanks :)

JaffaKetchup:
Just given it a test, and made `useEvenOdd` dependent on `kIsWeb`. I think it's all good, but I'll give it a more thorough test tomorrow!

lib/src/layer/polygon_layer/painter.dart
```diff
@@ -49,9 +55,40 @@ class _PolygonPainter<R extends Object> extends CustomPainter
...
+
+  // Do we also fill the holes with inverted fill?
+  // Should be `true`.
+  static const _fillInvertedHoles = true;
+
+  // Whether to draw the batch of polygons when a polygon with translucency is
+  // encountered.
+  // Should be `true`.
+  // TODO: Verify if still necessary.
+  static const _flushBatchOnTranslucency = true;
```
josxha, line 90:
```suggestion
  /// Corner coordinates of the polygon painted onto the entire world when using
  /// inverted fill.
  static const _minMaxLatitude = [LatLng(90, 0), LatLng(-90, 0)];

  /// Whether to use `PathFillType.evenOdd` (true) or `Path.combine` (false).
  ///
  ///  * `Path.combine` doesn't work & isn't stable/consistent on web
  ///  * `evenOdd` gives broken results when polygons intersect when inverted
  ///
  /// The best option is to use `evenOdd` on web, as it at least works sometimes,
  /// and `Path.combine` otherwise, as it gives correct results on native
  /// platforms.
  ///
  /// See https://github.com/fleaflet/flutter_map/pull/2046.
  static const _useEvenOdd = kIsWeb;

  /// Do we also remove the holes from the inverted map?
  /// Should be `true`.
  static const _invertedHoles = true;

  /// Do we also fill the holes with inverted fill?
  /// Should be `true`.
  static const _fillInvertedHoles = true;

  /// Whether to draw the batch of polygons when a polygon with translucency is
  /// encountered.
  /// Should be `true`.
  ///
  /// TODO: Verify if still necessary.
  static const _flushBatchOnTranslucency = true;
```

lib/src/layer/polygon_layer/painter.dart
```diff
@@ -112,101 +149,102 @@ class _PolygonPainter<R extends Object> extends CustomPainter
...
 
-    final filledPath = Path();
+    Path filledPath = Path();
     final borderPath = Path();
-    Polygon? lastPolygon;
+    Color? lastColor;
     int? lastHash;
+    Paint? borderPaint;
+
+    // Draw polygon outline
```
josxha, line 164:
```suggestion
    /// Draw polygon outline
```

lib/src/layer/polygon_layer/painter.dart
```diff
@@ -238,11 +276,107 @@ class _PolygonPainter<R extends Object> extends CustomPainter
       return WorldWorkControl.visible;
     }
 
-    // Main loop constructing batched fill and border paths from given polygons.
+    Path holePaths = Path();
+
+    void addPolygon(List<Offset> offsets) {
+      // ignore: dead_code
+      if (!_fillInvertedHoles) return;
```
josxha, line 283: Is ignoring this lint really required? If so, please add a comment why.

lib/src/layer/polygon_layer/polygon_layer.dart
```diff
@@ -66,6 +70,16 @@ base class PolygonLayer<R extends Object>
...
   final bool drawLabelsLast;
 
+  /// Color to apply to the map where not covered by a polygon
+  ///
+  /// > [!WARNING]
+  /// > On the web, inverted filling may not work as expected in some cases.
+  /// > It will not match the behaviour seen on native platforms. Avoid allowing
+  /// > polygons to intersect, and avoid using holes within polygons.
+  /// > This is due to multiple limitations/bugs within Flutter. See online
+  /// > documentation for more info.
```
josxha, line 77: If possible, link the online resources like https://github.com/flutter/flutter/issues/124675

lib/src/layer/polygon_layer/polygon_layer.dart
```diff
@@ -90,6 +105,21 @@ class _PolygonLayerState<R extends Object> extends State<PolygonLayer<R>>
...
+  @override
+  void initState() {
+    if (kDebugMode && kIsWeb && widget.invertedFill != null) {
+      Logger(printer: PrettyPrinter(methodCount: 0)).w(
+        '\x1B[1m\x1B[3mflutter_map\x1B[0m\nOn the web, inverted filling may '
+        'not work as expected in some cases. It will not match the behaviour\n'
+        'seen on native platforms.\nAvoid allowing polygons to intersect, and '
+        'avoid using holes within polygons.\nThis is due to multiple '
+        'limitations/bugs within Flutter. See online documentation for more '
+        'info.',
```
josxha, line 114: If possible, link the online resources like https://github.com/flutter/flutter/issues/124675

JaffaKetchup: This is linked from our online docs: https://docs.fleaflet.dev/layers/polygon-layer#inverted-filling

lib/src/layer/polygon_layer/polygon_layer.dart
JaffaKetchup: This is linked from our online docs: [docs.fleaflet.dev/layers/polygon-layer#inverted-filling](https://docs.fleaflet.dev/layers/polygon-layer#inverted-filling)

lib/src/layer/polygon_layer/painter.dart
JaffaKetchup: This is retained for debugging purposes. Added comment.

JaffaKetchup:
I've completed the testing, it seems to work fine. There does unfortunately appear to be a performance penalty with `Path.combine`, I guess due to the repetitive variable re-assignment. Is there any way around this?

lib/src/layer/polygon_layer/polygon_layer.dart
JaffaKetchup: Done. Also added elsewhere.

monsieurtanuki:
> I've completed the testing, it seems to work fine. There does unfortunately appear to be a performance penalty with `Path.combine`, I guess due to the repetitive variable re-assignment. Is there any way around this?

@JaffaKetchup I'm not sure what you mean with "repetitive variable re-assignment"
```dart
      holePaths = Path.combine(
        PathOperation.union,
        holePaths,
        Path()..addPolygon(offsets, true),
      );
```
1. You mean `holePaths` being used as a return value AND as a parameter? Don't know how much it costs. If needed we can use a dedicated buffer variable.
2. As a former Java boy, I would say repeatedly using a new `Path()` isn't such a good idea. In that case, the optim would be to reset the same `Path tmpPath` initialized as `Path()`.
3. Perhaps, for the `PathOperation.union` we would be better off just adding the polygon, like `holePaths.addPolygon(offsets, true)`, but TBH I've lost the track of when to use what.
4. Another solution would be to use `Path.combine` only in the inverted fill mode, assuming that in the "standard" mode the `evenOdd` mode was working with the same results and faster.

josxha (approved):
I'm currently not able to test the behavior but the code changes look alright to me. @monsieurtanuki thanks for your pull request!

monsieurtanuki:
Thank you @JaffaKetchup @josxha for your reviews and comments!

JaffaKetchup:
> You mean holePaths being used as a return value AND as a parameter? Don't know how much it costs. If needed we can use a dedicated buffer variable.

Yeah, this is kinda what I meant. It creates a new `_NativePath` behind the scenes everytime.

> As a former Java boy, I would say repeatedly using a new Path() isn't such a good idea. In that case, the optim would be to reset the same Path tmpPath initialized as Path().

Just experimented with this, it makes no significant difference whatsoever - maybe 1ms in the stress test on the raster thread, so almost nothing in the real world.

Also tried minimizing the number of combine operations by adding multiple polys to one path first, but of course, that gives the wrong result (since each poly cuts into each other).

> Perhaps, for the PathOperation.union we would be better off just adding the polygon, like holePaths.addPolygon(offsets, true), but TBH I've lost the track of when to use what.

This is what we do for `evenOdd`, and it doesn't seem to work properly.

> Another solution would be to use Path.combine only in the inverted fill mode, assuming that in the "standard" mode the evenOdd mode was working with the same results and faster.

I've done this. `evenOdd` does work on native for non-inverted, so I've enabled that.

JaffaKetchup (approved):
LGTM at last! Before we merge, I'll write a bounty spec and assign this.
I won't include the feature for `PolylineLayer` and `CircleLayer` - I'll leave this open for the future, since it wasn't possible before. If I have time, I'll try to implement it based on this.

monsieurtanuki:
@JaffaKetchup We may not need the "Inverted Polygons" / "Advanced polygons" example page anymore, now that you've updated the standard "Polygon Layer" page.
Anyway "Inverted Polygons" is a deprecated term now ;)

JaffaKetchup:
Ah, good point, forgot to remove it. I'll do it at some point.

I've setup the bounty on Polar, but it turns out they're sunsetting bounties soon (which is a bit annoying), so they need to mark this one as complete manually due to a bug or something. Hopefully it won't take too long.

JaffaKetchup:
We've paid the bounty through GitHub Sponsors instead :). You should get the entire $170 USD (we paid $10.20 in fees).

monsieurtanuki:
> We've paid the bounty through GitHub Sponsors instead :). You should get the entire $170 USD (we paid $10.20 in fees).

Thank you so much!

JaffaKetchup:
Just a note:

We might need to change the default of `_useEvenOdd` again. I think the performance impacts (~3ms on the stress test) of combine are not bad enough to justify preferring even/odd. And the artifacts from even/odd on intersecting holes are annoying, especially as they are not consistent with the behaviour in inverted mode (which is correct, since that always uses combine) which is confusing.

monsieurtanuki:
> We might need to change the default of _useEvenOdd again.

Whatever works. As long as it works, the performances are good and it's easy to maintain ;)

PR #2050, opened 2025-03-04
feat: added `LayerHitTestStrategy` to feature layers

JaffaKetchup:
This is designed to go well in conjunction with #2046, which will allow for the painting of inverted polygons.

It adds more options to control hitting.

### DIFF
--- a/example/lib/pages/polygon.dart
+++ b/example/lib/pages/polygon.dart
@@ -1,4 +1,3 @@
-import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_map/flutter_map.dart';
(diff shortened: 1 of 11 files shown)


JaffaKetchup:
I honestly can't decide myself whether this feature is necessary. I thought I could think of a use-case for it, but I can't particularly (that MapOptions.onTap wouldn't already). Happy to close if you think it's not useful, lmk.

monsieurtanuki:
@JaffaKetchup I'm not sure it still makes sense to keep that PR as we moved the "inverted" feature from polygon holes to polygon layers.

JaffaKetchup:
I think I'm going to close this then. Maybe the ability to not consider non-interactive polygons as hittable could be useful, but that can be more easily/better implemented at some other point.

Issue #2052, opened 2025-03-06
Labelled polygons cause fatal crash when map rotated


redbaty:
### What is the bug?

If there's a polygon layer present, and you try to zoom in, the app will hang or crash on android, this happens both on a real phone and on emulator

Not sure if it happens on other platforms

### How can we reproduce it?

The example app provided has the exact same problem, just open the "Polygons" page and try to zoom in:

![Image](https://github.com/user-attachments/assets/65fc4605-2639-4de2-aef9-8962ad30830f)

The "zooming" level required to hang depends, but keep on zooming in until it happens

JaffaKetchup:
Hi @redbaty,
Can you add the Flutter version & device information please (preferably not an emulator), and check whether running without impeller changes the result?

redbaty:
Hello @JaffaKetchup 

Sorry, forgot to add

This my the output of `flutter doctor`

Doctor summary (to see all details, run flutter doctor -v):
[✓] Flutter (Channel stable, 3.29.0, on Microsoft Windows [Version 10.0.26100.3323], locale en-GB)
[✓] Windows Version (11 Pro 64-bit, 24H2, 2009)
[✓] Android toolchain - develop for Android devices (Android SDK version 35.0.0)
[✓] Chrome - develop for the web
[✓] Visual Studio - develop Windows apps (Visual Studio Community 2022 17.13.2)
[✓] Android Studio (version 2024.3)
[✓] VS Code (version 1.98.0)
[✓] Connected device (4 available)
[✓] Network resources

And I've tested on two devices:

* Motorola G7 Power running Android 9
* Samsung S24 Ultra running Android 14

Using the flag `--no-enable-impeller` it still hangs

JaffaKetchup:
@redbaty Did this occur in a released version for you, or can you only reproduce on the example app?

redbaty:
@JaffaKetchup Occurred both in the example app as well as in my application with version 8.1.0:

```
  flutter_map:
    dependency: "direct main"
    description:
      name: flutter_map
      sha256: bbf145e8220531f2f727608c431871c7457f3b134e513543913afd00fdc1cd47
      url: "https://pub.dev"
    source: hosted
    version: "8.1.0"
```

JaffaKetchup:
What's your `Polygon` configuration look like? I can reproduce.

JaffaKetchup:
[...]

PR #2054, opened 2025-03-08
fix: prevent infinite looping when polygon labels displayed on rotated map

monsieurtanuki:
This #2052 fix is made of 2 separate parts.

One part is saying "No!" to forever loops: we assume that after trying to display on multiple worlds, if we keep trying it's because we'll never get an exit permit. And in that case we send an exception, which is slightly more refined than an ANR.

The second part deals more specifically with labels.
Only in the "no rotation" case we computed if the label was visible or not.
That means that in all non-0 rotation cases, we presumed that the label was visible, hence the forever loops.
Now we roughly compute cases when the label is invisible in all rotations. No more forever loops for labels.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,17 @@ Please consider [donating](https://docs.fleaflet.dev/supporters#support-us) or [
 
 This CHANGELOG does not include every commit and/or PR - it is a hand picked selection of the ones that have an effect on you. For a full list of changes, please check the GitHub repository releases/tags.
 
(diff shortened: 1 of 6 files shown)


JaffaKetchup (approved):
LGTM, thanks :). I agree having both is the best solution here as well. I'm releasing this as a hotfix immediately due to the impacts of the bug.

monsieurtanuki:
Thank you @JaffaKetchup for your quick review and release!

PR #2058, opened 2025-03-20
fix: raise attribution widgets above screen insets

JaffaKetchup:
Noticed on Android 15 with the new edge-to-edge mode that the attribution was now hidden below the navigation/gesture bar. This just wraps them with `SafeArea` to push them up.

I've also removed the asymmetric border rounding from the `RichAttributionWidget` because it doesn't look to good on most devices and was making things unnecessarily complex. I wouldn't consider this a breaking fix.


monsieurtanuki (approved):
@JaffaKetchup LGTM

josxha (approved):
Probably relevant for some iOS devices as well.

Issue #2042, opened 2025-02-20
Emulated retina mode displays incorrect tiles


stefangaller:
### What is the bug?

After upgrading from version `7.0.2` to `8.0.0`, `flutter_map` is trying to fetch the wrong coordinates.

Flutter doctor: 
```
Doctor summary (to see all details, run flutter doctor -v):
[✓] Flutter (Channel stable, 3.29.0, on macOS 15.3.1 24D70 darwin-arm64, locale en-AT)
[✓] Android toolchain - develop for Android devices (Android SDK version 35.0.0)
[✓] Xcode - develop for iOS and macOS (Xcode 16.2)
[✓] Chrome - develop for the web
[✓] Android Studio (version 2024.2)
[✓] IntelliJ IDEA Ultimate Edition (version 2024.3.3)
[✓] VS Code (version 1.91.0)
[✓] Connected device (5 available)
[✓] Network resources
```

### How can we reproduce it?

``` dart
FlutterMap(
  children: [
      TileLayer(
        retinaMode: true,
        urlTemplate: 'https://a.tile.opentopomap.org/{z}/{x}/{y}.png',
      )
  ],
),
```

**Output Version 7.0.2**
<img width="700" alt="Image" src="https://github.com/user-attachments/assets/4ac66216-1447-4455-9583-4e074bd6a654" />

**Output Version 8.0.0**
<img width="703" alt="Image" src="https://github.com/user-attachments/assets/9f12d654-8dff-4dbe-8982-d2fcd498529e" />

Zoomed out it now looks like this:

<img width="692" alt="Image" src="https://github.com/user-attachments/assets/ce96cb24-81d6-4d84-8a5b-e265112adcaa" />

JaffaKetchup:
It should be done automatically, but can you please try setting `zoomOffset` to -1? If that doesn't work, try -2?

stefangaller:
I did try with zoomOffset, which didn't seem to make any difference. I tried with -1, -2, +1, +2,... 
Shouldn't it still work the same way with both version 7.0.2 and 8.0.0 ? 
I now downgraded the package, and everything works as expected.

JaffaKetchup:
Ok, thanks for trying! We'll look into it.

JaffaKetchup:
This was introduced in 182c4fc0c73411e2be10774dd227d0da9cf4276a, alongside unbounded horizontal scrolling.

JaffaKetchup:
[...]

PR #2060, opened 2025-03-22
fix: consider `zoomOffset` & emulated retina mode when generating tile coordinates

monsieurtanuki:
The fix was to add the concept of `zoomOffset` to `TileCoordinatesResolver`, in order to compute correct modulus.
Fixes: #2042

The world as normal:
Capture d'écran 2025-03-22 091437

The world as simulated retina:
Capture d'écran 2025-03-22 091457


JaffaKetchup (approved):
LGTM! Thanks again :)

monsieurtanuki:
Thank you @JaffaKetchup for your review and your hunch about modulus that helped me fix the bug.

PR #2062, opened 2025-03-23
chore: prepare for v8.2.0 release

JaffaKetchup:
Updates CHANGELOG, and adds a commit hash indication to the example app.

This also replaces the funding link with the new GitHub Sponsors account. The docs site will follow after.


josxha (approved):
Nice changes!

JaffaKetchup:
Hey @enricostrijks,

We're currently waiting for some other maintainers (@fleaflet/maintainers) hopefully to do some reviews. We need to merge:

* #2082 
* #2093 
* #2102 (and need to verify that it's non-breaking, I think it is, but reviews need to check that since it's a big change - technically I can review and merge this, but I've made a very large contribution, so that wouldn't be ideal)
* #2114

This PR is mostly ready I think - just need to verify that the CHANGELOG is up to date, then everything's ready.

If you wanted to help, then I guess some reviews would be useful just to add another level of certainty and maybe aid the other maintainers. Which feature/commit are you waiting for exactly?

JaffaKetchup:
Unfortunately no estimated date for now. I would also like to get it done ASAP, but I'm not sure when anyone else will be available next.

JaffaKetchup:
@josxha I might revert the changes to the tile layer that I made here. Wdyt? I think they might be a problem.

JaffaKetchup:
I've just pushed the 2 current ones into `initState`. Now they don't annoy users.

PR #2066, opened 2025-03-27
fix: stop ongoing animations when scroll zooming

### LINKED ISSUE #1767: Scroll zoom does not stop ongoing fling and zoom animations
### What is the bug?

On web, when using mousedown-drag-mouseup to move the map, and then using mousewheel to zoom in/out, the movement continues and acts unexpectedly.

### How can we reproduce it?

-

See Google Maps, Bing Maps, or Apple Maps for industry standard implementation.

### Platforms

web

### Severity

JaffaKetchup:
Fixes #1767.


josxha (approved):

josxha:
tested on web

PR #2068, opened 2025-04-04
feat: add double tap zoom specificity

Quirille:
Right now there is no option to configure the zoom speed for double tap zooming. This PR adds it to the MapOptions as a double.

We use this in our project as a result of a request from the client.


JaffaKetchup:
Thanks for contributing! Unfortunately I can't proceed forward with this PR, since there's a few issues with it and I don't seem to be able to edit the PR. I'll open a new one momentarily with the same feature but slightly modified (and some other nice features).

PureTryOut:
Just a note, this response is not really nice. The new PR you made isn't _that_ different from this PR and the author here could have made these changes just fine themselves if they were told about it. This kind of approach deters people from contributing as credits for the change are taken away and the response can feel kind of insulting.

Issue #2067, opened 2025-04-01
Polylines draw beyond primary world bounds when CRS has replicatesWorldLongitude as false


mattmyne:
### What is the bug?

PolylinesLayer (and others?) draw beyond primary world longitude bounds when CRS has replicatesWorldLongitude as false. It should only draw within the primary/original map bounds, not draw over to the next repeated map space if it's closer (e.g. drawing from 150 to -170 longitude draws a short line to the east repeated world, rather than a long line back west to the same world's -170).

### How can we reproduce it?

Disable repeated horizontal scroll/"infinite worlds" for the map, which currently seems to require overriding the default EPSG3857 CRS and setting the flag false:

```dart
@immutable
class Epsg3857NoRepeat extends Epsg3857 {
  @override
  bool get replicatesWorldLongitude => false;
}
```

and then including this in the `MapOptions`:

```dart
crs: Epsg3857NoRepeat(),
```

Though this stops the polylines repeating, it does not stop the lines "hopping over" to adjacent worlds space rather than taking a longer route within the single world space.

If world longitude it not to replicate, these additional spaces should not be drawn on.

Example with older flutter_map (I'm not sure what version), with polylineLayer drawing a custom world grid:

![Image](https://github.com/user-attachments/assets/5303bdf0-12c5-4573-81cd-7648d7c9b5d8)

and now, with some polylines hopping over the antimeridian into repeated worlds:

![Image](https://github.com/user-attachments/assets/85024dd5-af3a-4ac7-b2b1-b19fa64d7180)

There does not seem to be a way in code to prevent this happening outside of the library.

mattmyne:
Another example, where flutter_map is "choosing" to draw some of the polylineLayer content on the repeated world to the east, rather than the primary world where it should be.

![Image](https://github.com/user-attachments/assets/1f65025f-1ab1-45b0-8ad6-4abe59a26e2a)

monsieurtanuki:
Related to https://github.com/fleaflet/flutter_map/discussions/2061 and https://github.com/fleaflet/flutter_map/issues/2022#issuecomment-2761236172

My point of view was:
> Generally speaking, a list of LatLng should include a policy about what to do with -180/180 crossing and half a world distance apart.
That would make sense also for a couple of LatLng, e.g. regarding comparison.

Would a "one world only" `bool` be enough? I'll have a go.

monsieurtanuki:
@mattmyne Just to be sure, in the "one world" case you would prefer
[...]

PR #2069, opened 2025-04-06
feat: added `Poly*Layer.drawInSingleWorld`

monsieurtanuki:
### What
- We have a new `bool oneWorld` parameter for `PolygonLayer`s and `PolylineLayer`s.
- Default value is `false`.
- If that parameter is `true`, each list of points will be projected in the same (central) world - for instance, -179 and 179 longitudes will be considered almost a world away for 2 consecutive points.
- If that parameter is `false`, each "next" point will be as close as possible to the "previous" point of the list - for instance, -179 and 179 longitudes will be considered very close for 2 consecutive points.
- As it's a parameter at the level of `PolygonLayer`s and `PolylineLayer`s, we can display both "oneWorld" behaviors on the same map (if needed).
- Closes: #2067

### Screenshot
Taken from the new "One World - or not - Layers" demo:
image

### Files
New file:
* `one_world_or_not.dart`: Demo of how the new `oneWorld` parameter works on Poly*Layer's.

Impacted files:
* `crs.dart`: added a `bool oneWorld` parameter to `projectList` method, and that's where the list of continuous segments are computed.
* `main.dart`: added `OneWorldOrNotPage` demo.
* `menu_drawer.dart`: added `OneWorldOrNotPage` demo.
* `polygon_layer/painter.dart`: added a `bool oneWorld` parameter.
* `polygon_layer.dart`: added a `bool oneWorld` parameter.
* `polyline_layer.dart`: added a `bool oneWorld` parameter.
* `projected_polygon.dart`: added a `bool oneWorld` parameter.
* `projected_polyline.dart`: added a `bool oneWorld` parameter.

### DIFF
--- a/example/lib/pages/single_world_polys.dart
+++ b/example/lib/pages/single_world_polys.dart
@@ -6,26 +6,26 @@ import 'package:latlong2/latlong.dart';
 
 @immutable
 class Epsg3857NoRepeat extends Epsg3857 {
+  const Epsg3857NoRepeat();
+
   @override
   bool get replicatesWorldLongitude => false;
 }
 
-typedef HitValue = ({String title, String subtitle});
-
-/// Demo of how the new `oneWorld` parameter works on Poly*Layer's.
+/// Demo of how the new `drawInSingleWorld` parameter works on Poly*Layer's.
 ///
 /// cf. https://github.com/fleaflet/flutter_map/issues/2067
-class OneWorldOrNotPage extends StatefulWidget {
-  static const String route = '/one_world';
+class SingleWorldPolysPage extends StatefulWidget {
+  static const String route = '/single_world_polys';
 
-  const OneWorldOrNotPage({super.key});
+  const SingleWorldPolysPage({super.key});
 
   @override
-  State<OneWorldOrNotPage> createState() => _OneWorldOrNotPageState();
+  State<SingleWorldPolysPage> createState() => _SingleWorldPolysPageState();
 }
 
-class _OneWorldOrNotPageState extends State<OneWorldOrNotPage> {
-  final LayerHitNotifier<HitValue> _hitNotifier = ValueNotifier(null);
+class _SingleWorldPolysPageState extends State<SingleWorldPolysPage> {
+  bool _repeatLongitudes = false;
 
   static const _polylinePoints = [
     LatLng(40, 150),
@@ -62,23 +62,24 @@ class _OneWorldOrNotPageState extends State<OneWorldOrNotPage> {
   @override
   Widget build(BuildContext context) {
     return Scaffold(
-      appBar: AppBar(title: const Text('One World - or not')),
-      drawer: const MenuDrawer(OneWorldOrNotPage.route),
+      appBar: AppBar(title: const Text('Single World Polys')),
+      drawer: const MenuDrawer(SingleWorldPolysPage.route),
       body: Stack(
         children: [
           FlutterMap(
             options: MapOptions(
               initialCenter: const LatLng(0, 0),
               initialZoom: 0,
-              crs: Epsg3857NoRepeat(),
+              crs: _repeatLongitudes
+                  ? const Epsg3857()
+                  : const Epsg3857NoRepeat(),
             ),
             children: [
               openStreetMapTileLayer,
               PolygonLayer(
-                hitNotifier: _hitNotifier,
                 simplificationTolerance: 0,
-                oneWorld: false,
-                polygons: <Polygon<HitValue>>[
+                drawInSingleWorld: false,
+                polygons: [
                   Polygon(
                     points: _polygonPoints
                         .map((latLng) =>
@@ -91,10 +92,9 @@ class _OneWorldOrNotPageState extends State<OneWorldOrNotPage> {
                           .toList(),
                     ],
                     color: const Color(0xFF0000FF),
-                    hitValue: (
-                      title: 'Blue Line',
-                      subtitle: 'Across the universe...',
-                    ),
+                    borderColor: Colors.purple,
+                    borderStrokeWidth: 6,
+                    pattern: const StrokePattern.dotted(),
                   ),
                   Polygon(
                     points: _polygonPoints
@@ -102,18 +102,13 @@ class _OneWorldOrNotPageState extends State<OneWorldOrNotPage> {
                             LatLng(latLng.latitude - 80, latLng.longitude))
                         .toList(),
                     color: const Color(0xFF000000),
-                    hitValue: (
-                      title: 'Black Line',
-                      subtitle: 'Across the universe...',
-                    ),
                   ),
                 ],
               ),
               PolygonLayer(
-                hitNotifier: _hitNotifier,
                 simplificationTolerance: 0,
-                oneWorld: true,
-                polygons: <Polygon<HitValue>>[
+                drawInSingleWorld: true,
+                polygons: [
                   Polygon(
                     points: _polygonPoints
                         .map((latLng) =>
@@ -125,30 +120,24 @@ class _OneWorldOrNotPageState extends State<OneWorldOrNotPage> {
                               LatLng(latLng.latitude - 55, latLng.longitude))
                           .toList(),
                     ],
-                    color: const Color(0xFF00FFFF),
-                    hitValue: (
-                      title: 'Teal Line',
-                      subtitle: 'Across the universe...',
-                    ),
+                    color: const Color(0xFF0000FF),
+                    borderColor: Colors.purple,
+                    borderStrokeWidth: 6,
+                    pattern: const StrokePattern.dotted(),
                   ),
                   Polygon(
                     points: _polygonPoints
                         .map((latLng) =>
                             LatLng(latLng.latitude - 110, latLng.longitude))
                         .toList(),
-                    color: const Color(0xFFFFFF00),
-                    hitValue: (
-                      title: 'Yellow Line',
-                      subtitle: 'Across the universe...',
-                    ),
+                    color: const Color(0xFF000000),
                   ),
                 ],
               ),
               PolylineLayer(
-                hitNotifier: _hitNotifier,
                 simplificationTolerance: 0,
-                oneWorld: false,
-                polylines: <Polyline<HitValue>>[
+                drawInSingleWorld: false,
+                polylines: [
                   Polyline(
                     points: _polylinePoints
                         .map((latLng) =>
@@ -156,10 +145,7 @@ class _OneWorldOrNotPageState extends State<OneWorldOrNotPage> {
                         .toList(),
                     strokeWidth: 8,
                     color: const Color(0xFFFF0000),
-                    hitValue: (
-                      title: 'Red Line',
-                      subtitle: 'Across the universe...',
-                    ),
+                    pattern: const StrokePattern.dotted(),
                   ),
                   Polyline(
                     points: const [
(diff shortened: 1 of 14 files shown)


mattmyne:
> cc. @JaffaKetchup @mattmyne

Many thanks for adding this

JaffaKetchup:
https://github.com/user-attachments/assets/65377ee0-d749-42f3-8bc4-ff2ed3a4c2e8

Could you reproduce this? I guess it occurs as sides of the holes go across the boundary?

And how do we 'decide' on which world the polys appear on? It doesn't always appear to be the one closest to the middle? I'm assuming its some kind of result of `worldWidth` and `project`?

mattmyne:
> Recording.2025-04-06.151145.mp4 
> Could you reproduce this? I guess it occurs as sides of the holes go across the boundary?
> 
> And how do we 'decide' on which world the polys appear on? It doesn't always appear to be the one closest to the middle? I'm assuming its some kind of result of `worldWidth` and `project`?
> 
> And thanks for looking into this so quickly!

Surely it shouldn't jump to other worlds at all, even when scrolling - isn't that the aim of setting the oneWorld flag?

monsieurtanuki:
It's a side-effect of the side-effect. I'll fix it tomorrow.
To explain what happened, the main polygon was projected in one world and the hole was projected in another. The fix is about making the first point of the hole be as close as possible to the first point of the polygon. Which happens automatically for the default `oneWorld: false` mode.

monsieurtanuki:
@mattmyne @JaffaKetchup I've just pushed the fixes.

### Impacted files
* `offsets.dart`: added parameters to `getOffsetsXY` in order to "force" which world to use in some cases (polygon holes and "oneWorld")
* `polygon_layer/painter.dart`: now forcing which world to use when relevant
* `polyline_layer/painter.dart`: added a `bool oneWorld` parameter; now forcing which world to use when relevant
* `polyline_layer.dart`: added a `bool oneWorld` parameter.

### Screenshots
The green, teal and yellow objects (all "oneWorld") stay in the same world.
image
Scrolling...
image




There are still problems with the code:
* an obvious bad naming for `MyLousyDouble` class - I think we should move `getOffsetsXY` in a helper class instead
* just noticed that Circles and Markers are not projected on the best world. Shouldn't they?
* confusion between the concepts of "oneWorld" at the layers' level and "replicatesWorldLongitude" at the Crs level

JaffaKetchup:
Ok all this projection/offsets stuff is getting quite complicated now :D 

It might be a good idea to write a little readme on how the projection stuff works especially in these more complicated cases. Not sure tho.

> an obvious bad naming for MyLousyDouble class - I think we should move getOffsetsXY in a helper class instead

I think we should either:

- Rename to `_PassByReference`
- Use something like `ValueNotifier` (which is a bit overkill maybe)
- I'm not sure how moving `getOffsetsXY` into its own class will look, but maybe it could help? Could also join with the effort to improve some documentation and include all the offsets stuff. 

> confusion between the concepts of "oneWorld" at the layers' level and "replicatesWorldLongitude" at the Crs level

Perhaps we should introduce a map option. If the CRS doesn't support it, it's forced false. Otherwise we default true (but it can be changed to false). Layers can then check the option. This might also lead to being able to reimplement some sort of constraint so the user doesn't see the other worlds at all - but messing with the constraint system is probably out of scope for this PR (and maybe goes against the point of some of it?).

> just noticed that Circles and Markers are not projected on the best world. Shouldn't they?

Not sure what you mean by this, but I can't test right now. I'll have a look.

monsieurtanuki:
> * Rename to `_PassByReference`

I don't think we can make it private.

> * Use something like `ValueNotifier` (which is a bit overkill maybe)

Let's keep it simple: we just want to return an extra `double`.
A solution for `getOffsetsXY` would also be to return `(List<Offset>, double)` instead of `List<Offset>`. But I didn't want to change too much the method in a first approach.

> * I'm not sure how moving `getOffsetsXY` into its own class will look, but maybe it could help? Could also join with the effort to improve some documentation and include all the offsets stuff.

I'll have a go with both "class" and "(List<Offset>, double)" options.

> Perhaps we should introduce a map option...

Anything that can make things easier to understand for users and developers would be welcome indeed.

> > just noticed that Circles and Markers are not projected on the best world. Shouldn't they?
> 
> Not sure what you mean by this, but I can't test right now. I'll have a look.

When we're displaying polylines, we display them in the "most centered world" (e.g. the red polyline in the screenshot).
When we're displaying circles and markers, we just display them in the first suggested world, which will be basically on the left side. Would be consistent to use the same pattern as polylines and polygons.

monsieurtanuki:
### Screenshots
The "one world" example:
image

Twisted "one world" example, with "replicate longitude = true"
image

### What
I think I've just found the logic behind the concepts:
* "oneWorld" (potentially to be renamed)
  * means (if `true`) that we want _all_ the coordinates to be projected in the standard -180/180 world
  * for instance, -179 to 179 means a VERY long distance
  * it's attached to a poly*layer (polygon or polyline), because it makes sense only if we have consecutive points and wonder how to link them (typically west or east).
* "replicate longitude"
  * means that we display everything on all the worlds

### Impacted files
* `offsets.dart`: transformed into a helper class
* `one_world_or_not.dart`: new examples
* `polygon_layer/painter.dart`: refactored using new class `OffsetHelper`; removed `oneWorld` parameter
* `polyline_layer/painter.dart`: refactored using new class `OffsetHelper`; removed `oneWorld` parameter
* `polygon_layer.dart`: minor refactoring
* `polyline_layer.dart`: minor refactoring

JaffaKetchup:
This occurs when `oneWorld` and `replicateWorldLongitude` are both `true`:

https://github.com/user-attachments/assets/d756540c-8753-4016-8710-fdcf4356c1aa

image

I'm not sure whether this is a significant issue, especially as this configuration seems unlikely for most users ? There is a point where the complications of fixing it and trying to maintain that become too great for these minor bugs.

KovynevA:
> Я не уверен, что это серьёзная проблема,

And yet it is a mistake. It is impossible to correctly display any polyline with a difference in longitudes greater than 180 degrees. For example, the flight path of an airplane from longitude -15 to longitude 166, the airplane flies in the other direction)
I really look forward to fixing this.

JaffaKetchup:
In that use-case, why would you be using `oneWorld` and `replicateWorldLongitude`?

KovynevA:
> In that use-case, why would you be using `oneWorld` and `replicateWorldLongitude`?

Sorry, did I understand that the oneWorld flag cancels the search for the shortest path for a polyline or polygon? Or did I misunderstand and wait in vain :)

JaffaKetchup:
You can indeed use them both as true, and in this case, [the issue above](https://github.com/fleaflet/flutter_map/pull/2069#issuecomment-2786997220) appears. However, it only appears in these very narrow circumstances where the camera is split across the boundary and the polyline is also.

> I'm not sure whether this is a significant issue

When I say this, I'm referring only to [that comment](https://github.com/fleaflet/flutter_map/pull/2069#issuecomment-2786997220). I would agree that it should be fixed, if the fix is very complicated, we need to consider whether there is an easier way to do this all, otherwise we'll keep running into these edge cases (as we are now with camera constraints and positioning for example).

monsieurtanuki:
> I really need the ability to choose whether to draw a line along the shortest path or project into the same world

That would be `oneWorld` (`false`: "shortest path", `true`: same world) at the `PolylineLayer` and `PolygonLayer` levels.

> and copy it into other worlds.

`replicateWorldLongitude`: `true` at the `Projection` level.

monsieurtanuki:
> This occurs when `oneWorld` and `replicateWorldLongitude` are both `true`:

I'll have a look at it.

monsieurtanuki:
> > This occurs when `oneWorld` and `replicateWorldLongitude` are both `true`:
> 
> I'll have a look at it.

Quickly said, it looks like a side-effect issue with the "polyline hiker". Investigating...

monsieurtanuki:
> Quickly said, it looks like a side-effect issue with the "polyline hiker". Investigating...

Turns out it's due to "aggressive culling". Just putting `cullingMargin: null` in `PolylineLayer` makes the bug disappear.
That was somehow hinted in the code.
```dart
      // TODO: think about how to cull polylines that go beyond -180/180.
      if (stretchesBeyondTheLimits()) {
        yield projectedPolyline;
        continue;
      }
```
Here we're in a new specific case - we want to _force_ the polyline to be in -180/180.

monsieurtanuki:
Seems to be fixed now.

### Screenshots
replicate world as true, oneWorld as true, before the fix:
image
replicate world as true, oneWorld as true, after the fix:
image
replicate world as false, oneWorld as true, part 1:
image
replicate world as false, oneWorld as true, part 2:
image

### Impacted files
* `offsets.dart`: tidied
* `polyline_layer.dart`: removed aggressive culling for "oneWorld" polylines

monsieurtanuki:
I've just pushed a fix regarding @WorldpixelSoftware's https://github.com/fleaflet/flutter_map/issues/1338#issuecomment-2799927382
Again, it was a "culling" problem. Basically if we truncate the camera bounds into -180/180, we cannot correctly compare polyline points with the bounds.

JaffaKetchup:
I believe this is a breaking change, which complicates things a bit. Is there any way we can make it backwards compatible - eg. the new option in `Projection.projectList` is required?

monsieurtanuki:
> I believe this is a breaking change, which complicates things a bit. Is there any way we can make it backwards compatible - eg. the new option in `Projection.projectList` is required?

@JaffaKetchup I've just PR'ed: was it what you were calling a breaking change?

lib/src/geo/crs.dart
```diff
@@ -417,7 +417,11 @@ abstract class Projection {
   /// longitudes -179 and 179 to be projected each on one side.
   /// [referencePoint] is used for polygon holes: we want the holes to be
   /// displayed close to the polygon, not on the other side of the world.
-  List<Offset> projectList(List<LatLng> points, {LatLng? referencePoint}) {
+  List<Offset> projectList(
+    List<LatLng> points, {
+    LatLng? referencePoint,
+    required bool oneWorld,
```
JaffaKetchup, line 423: @monsieurtanuki I was referring to this line. Can we default it to `false`?

I think the changes to offsets.dart were OK because it wasn't exported. Maybe it's easier to revert that last commit?

JaffaKetchup (changes requested):

lib/src/geo/crs.dart
monsieurtanuki: @JaffaKetchup Done.

example/lib/pages/single_world_polys.dart
```diff
@@ -0,0 +1,250 @@
...
+                        LatLng(46, -9),
+                        LatLng(46, -8),
+                        LatLng(45.5, -7.5),
+                        LatLng(45, -8),
+                        LatLng(45, -9),
+                      ]
+                    ],
+                  ),
+                ],
+              ),*/
```
JaffaKetchup, line 212: This demonstrates a question relating to the recently added `invertedFill`: is it still necessary. Is it now possible to reliably achieve everything we did before the multi-world change, and/or with the feature set introduced by `invertedFill`, or is there still some advantage to keeping `invertedFill`?

Wdyt?

lib/src/layer/polyline_layer/polyline_layer.dart
```diff
@@ -146,17 +165,59 @@ class _PolylineLayerState<R extends Object> extends State<PolylineLayer<R>>
...
         continue;
       }
 
+      // TODO: think about how to cull polylines in a forced "drawInSingleWorld".
+      // As what may be culled in a world may not be culled in a next world.
+      if (widget.drawInSingleWorld) {
+        yield projectedPolyline;
+        continue;
+      }
+
```
JaffaKetchup, line 220: Is this culling something that could be more easily implemented if we have a method - or are we limited by the tools we have again. For example, I realise `LatLngBounds` can't really support what we're doing, so could something like that be an issue?

JaffaKetchup (commented):
LGTM, thanks :) And sorry it took me so long to get to this.

I did have two questions which are not directly related to the code, but are still important before we introduce this, I can approve once we've sorted those out.

JaffaKetchup:
Also see #2080.

example/lib/pages/single_world_polys.dart
monsieurtanuki: @JaffaKetchup `invertedFill` is still necessary if you want to display polylines as transparent holes.
image

lib/src/layer/polyline_layer/polyline_layer.dart
JaffaKetchup: Ok, I've removed this check then. I think we're ready to go with this hopefully :)

JaffaKetchup (approved):
LGTM, huge thanks again :)

monsieurtanuki:
@JaffaKetchup You're welcome!

PR #2070, opened 2025-04-06
feat: added options to customize double-tap (drag) zoom gestures

JaffaKetchup:
Continuation of #2068. The original PR was not suitable for merging without significant work and was not open for editing by maintainers. I also took the opportunity to introduce some other new options.

Now, it's possible to change the amount by which the double-tap-drag zoom gesture works, and the duration and curve of the normal double tap zoom gesture. I've also changed the default curve for this animation from linear to `fastOutSlowIn`, which is more similar to what can be expected from other mobile map apps.

Thanks for the initial idea @Quirille.


tobiasleijs (approved):

freekvandeven (approved):

tobiasleijs:
@josxha could you review this pull request?

lib/src/gestures/map_interactive_viewer.dart
```diff
@@ -919,7 +915,9 @@ class MapInteractiveViewerState extends State<MapInteractiveViewer>
     final flags = _interactionOptions.flags;
     if (InteractiveFlag.hasDoubleTapDragZoom(flags)) {
       final verticalOffset = (_focalStartLocal - details.localFocalPoint).dy;
-      final newZoom = _mapZoomStart - verticalOffset / 360 * _camera.zoom;
+      final zoomFactor = 1 / _interactionOptions.doubleTapDragZoomFactor;
+      final newZoom =
+          _mapZoomStart - verticalOffset * zoomFactor * _camera.zoom;
```
josxha, line 920: I think it's a bit counter intuitive that a lower value increases the zoom speed. What are the benefits of using a factor here instead of an absolute value?

JaffaKetchup: Do you mean to remove the `1 / _` from `zoomFactor`? We could do that, but then any normal values would be very small (default would be 1/360). But yeah, maybe makes more sense that way around? Wdyt?

JaffaKetchup: @josxha In b69d4b4 I've flipped it, so the user now specifies the `1/x` themselves.

In terms of including `_camera.zoom` as a term, if that's what you mean, I'm not quite sure why it's like that - it's possibly been that way since it was introduced. It's kind of a nice feature IMO, although it can get a big aggressive with the default factor at very high zoom levels. But it means a user doesn't have to repeat the gesture so many times to make large zoom differences.

If we wanted the most flexibility, we could just make it a callback and leave the calculation to the user, as we already do elsewhere. Wdyt?

JaffaKetchup: I've changed it to a function.

josxha (approved):
sorry for the long wait for a code review, looks good to me.

JaffaKetchup:
Thanks again @Quirille & @freekvandeven :)

PR #2073, opened 2025-04-09
feat: add `ContainCameraLatitude`

### LINKED ISSUE #2071: Constraining only latitude with unconstrained horizontal scroll
**This issue is actually a feature request : it is labelled as a bug because it originated from a bug and I originally planned on only creating this issue, however after trying to understand what causes this behaviour, I created issue #2072 which is the underlying bug. They are two separate issues and solving #2072 will not "solve" this feature request**

### What is the bug?

If `cameraConstraint` is set from -90;-180 to 90;180, panning to the horizontal edges of the map will be inconsistent and there is no way to have unconstrained horizontal scroll with latitude boundaries of -90 to 90. If you don't pan with enough speed it will stop at the border and if you pan with enough speed it will jump to the other side with some lag and it will be quite jerky.

Seems to be the same issue as #2005 don't know why he marked it as completed though ? Is there already a way to fix it I missed ?

### How can we reproduce it?

Create a `FlutterMap` with the following property set in `MapOptions` : `cameraConstraint: CameraConstraint.contain(LatLngBounds(const LatLng(-90, -180), const LatLng(90, 180))),`

I see two ways to fix the problem :
- Ignore the longitude constraints when they're 180 or -180, but it would make it impossible for someone to actually constrain a map to these longitudes if they want to
- Create a camera constraint that only constrains latitude

To work around it I opted for the second solution and made the following camera constraint class inspired from the other classes, it only constrains the camera on the latitude. 

```dart

@immutable
class ContainCameraVertically extends CameraConstraint {
  const ContainCameraVertically(this.a, this.b);

  final double a;
  final double b;

  @override
  MapCamera? constrain(MapCamera camera) {
    final double testZoom = camera.zoom;
    final LatLng testCenter = camera.center;

    final Offset aPixel = camera.projectAtZoom(LatLng(a, 0), testZoom);
    final Offset bPixel = camera.projectAtZoom(LatLng(b, 0), testZoom);

    final Size halfSize = camera.size / 2;

    // Find the limits for the map center which would keep the camera within the
    // [a] and [b] bounds.
    final double topOkCenter = math.min(aPixel.dy, bPixel.dy) + halfSize.height;
    final double botOkCenter = math.max(aPixel.dy, bPixel.dy) - halfSize.height;

    // Stop if we are zoomed out so far that the camera cannot be translated to
    // stay within the [a] and [b] bounds.
    if (topOkCenter > botOkCenter) {
      return null;
    }

    final Offset centerPix = camera.projectAtZoom(testCenter, testZoom);
    final newCenterPix = Offset(centerPix.dx, centerPix.dy.clamp(topOkCenter, botOkCenter));

    if (newCenterPix == centerPix) {
      return camera;
    }

    return camera.withPosition(center: camera.unprojectAtZoom(newCenterPix, testZoom));
  }

  @override
  bool operator ==(Object other) {
    return other is ContainCameraVertically && other.a == a && other.b == b;
  }

  @override
  int get hashCode => a.hashCode ^ b.hashCode;
}
```

mbenoukaiss:
Allows constraining the camera vertically and not horizontally #2071 

- `ContainCameraVertically` or `ContainCameraLatitudinally` ?
- I'm not quite happy with the `a` and `b` parameters naming, but I don't like anything that would mention a notion of order (like `min` and `max`) because it does not quite make sense, let my know if this is alright or if you have a suggestion
- You mentioned "this could be made the new default (maybe only if Crs.replicateWorldLongitude is true)", even though I suppose not many people, if any, use the default `CameraConstraint.unconstrained()`, it would potentially be a breaking change. Also not 100% familiar with dart but I could not find a way to keep the constructor constant with a condition on `crs.replicatesWorldLongitude`.


JaffaKetchup (changes requested):
Thanks @mbenoukaiss! Sorry I haven't got to this earlier.

We do need to find some way to describe `a` & `b`. I can live with the names (although they're not ideal), but we do need to put a docstring on each of them. I know you said you'd prefer not to imply order, but I'm not sure why? Would `positive` & `negative` or `north` and `south` work?

I'm assuming you set `a` & `b` to the latitudes, such as 180 and -180? Maybe it's a good idea to add this to one of the examples, but I can find a space for that.

I think `ContainCameraLatitude` might be a better name, as it doesn't matter about rotation.

On that note, I need to test this to make sure it works correctly when the map is rotated as well. Would it be possible to add some unit tests for this as well?

> Also not 100% familiar with dart but I could not find a way to keep the constructor constant with a condition on crs.replicatesWorldLongitude.

That's a good point. I'll think about it, but I think this is probably OK for now.

JaffaKetchup (approved):
LGTM, thanks :). Seems to work with rotation.

PR #2093, opened 2025-05-16
fix: calculate `LayerHitResult.coordinate` correctly

### LINKED ISSUE #2092: Hit notifier returns incorrect `LayerHitResult.coordinate` when rotated
### What is the bug?

The `LayerHitResult.coordinate` seems to be in the wrong location (compared to the accurate `MapOptions.onTap`) when the map rotation is not 0.

Initially reported and initial MRE created at https://discord.com/channels/951867686378409984/1372837133546684437.

### How can we reproduce it?

```dart

void main() {
  runApp(const MyApp());
}

class MyApp extends StatelessWidget {
  const MyApp({super.key});

  @override
    return const MaterialApp(title: 'Flutter Demo', home: MyHomePage());
  }
}

class MyHomePage extends StatefulWidget {
  const MyHomePage({super.key});

  @override
  State<MyHomePage> createState() => _MyHomePageState();
}

class _MyHomePageState extends State<MyHomePage> {
  final LayerHitNotifier _polygonHitNotifier = ValueNotifier(null);

  LatLng? onTap;
  LatLng? notifier;

  @override
    return Scaffold(
      body: FlutterMap(
        options: MapOptions(
          initialCenter: const LatLng(30.940609, 40.236738),
          initialZoom: 8,
          onTap: (tapPosition, point) {
            setState(() {
              onTap = point;
              notifier = _polygonHitNotifier.value?.coordinate;
            });
          },
        ),
        children: [
          TileLayer(
          ),
          PolygonLayer(
            hitNotifier: _polygonHitNotifier,
            polygons: [
              Polygon(
                points: const [LatLng(30, 40), LatLng(31, 40), LatLng(32, 41)],
                color: Colors.blue,
              ),
            ],
          ),
          MarkerLayer(
            markers: [
              if (onTap != null)
                Marker(
                  alignment: Alignment.topCenter,
                  height: 32,
                  width: 32,
                  point: onTap!,
                  child: const Icon(
                    Icons.location_on,
                    color: Colors.purple,
                    size: 32,
                  ),
                ),
              if (notifier != null)
                Marker(
                  alignment: Alignment.topCenter,
                  height: 24,
                  width: 24,
                  point: notifier!,
                  child: const Icon(
                    Icons.location_on,
                    color: Colors.red,
                  ),
                ),
            ],
          ),
        ],
      ),
    );
  }
}
```

_No response_

JaffaKetchup:
Fixes #2092.

Also added `.coordinate` demo to Polygons example.

I also removed `FeatureLayerUtils.origin` in favour of `MapCamera.pixelOrigin`. I'm not 100% if this is correct, but it seems to make sense.

### DIFF
--- a/example/lib/pages/polygon.dart
+++ b/example/lib/pages/polygon.dart
@@ -438,6 +438,8 @@ class _PolygonPageState extends State<PolygonPage> {
                 ],
               ),
               IgnorePointer(
+                // Ignore the marker so it doesn't capture the pointer
+                // (as the marker is over the pointer)
                 child: ValueListenableBuilder(
                   valueListenable: _hitNotifier,
                   builder: (context, value, _) => MarkerLayer(
--- a/example/lib/pages/polygon.dart
+++ b/example/lib/pages/polygon.dart
@@ -426,6 +426,40 @@ class _PolygonPageState extends State<PolygonPage> {
                   ),
                 ],
               ),
+              IgnorePointer(
+                child: ValueListenableBuilder(
+                  valueListenable: _hitNotifier,
+                  builder: (context, value, _) => MarkerLayer(
+                    rotate: true,
+                    markers: [
+                      if (value != null)
+                        Marker(
+                          height: 22,
+                          width: 24,
+                          alignment: Alignment.topCenter,
+                          point: value.coordinate,
+                          child: DecoratedBox(
+                            decoration: BoxDecoration(
+                              shape: BoxShape.circle,
+                              boxShadow: [
+                                BoxShadow(
+                                  blurRadius: 15,
+                                  spreadRadius: 5,
+                                  color: Colors.black.withAlpha(255 ~/ 3),
+                                  offset: const Offset(0, 11),
+                                ),
+                              ],
+                            ),
+                            child: const Icon(
+                              Icons.location_on,
+                              color: Colors.green,
+                            ),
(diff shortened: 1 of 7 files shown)


JaffaKetchup:
@monsieurtanuki I believe it was while we were messing about with the feature layer painting that `origin` ended up in `FeatureLayerUtils`. In this PR, I've removed it because I couldn't see a difference to `MapCamera.origin`. They seem to work the same, but maybe you had something in mind to put `origin` in `FeatureLayerUtils` as well (caching messing something up?)? Does this change seem OK to you?

monsieurtanuki:
@JaffaKetchup I'll try to have a look at it this week. Sorry about the delay.

example/lib/pages/polygon.dart
```diff
@@ -437,6 +437,40 @@ class _PolygonPageState extends State<PolygonPage> {
                   ),
                 ],
               ),
+              IgnorePointer(
+                child: ValueListenableBuilder(
+                  valueListenable: _hitNotifier,
+                  builder: (context, value, _) => MarkerLayer(
```
josxha, line 443: A comment would be good why an IgnorePointer is used here.

JaffaKetchup: Done!

josxha (approved):

Issue #2076, opened 2025-04-15
Improve efficiency of painting many `CircleMarker`s with meter radii


bombicod:
### What do you want implemented?

The techniques can increase the performance of flutter_map several times. Please tell me if it makes sense to implement these techniques in your wonderful package? [article](https://plugfox.dev/high-performance-canvas-rendering/)

JaffaKetchup:
Hi @bombicod,

I've had a read through the article. A lot of it isn't directly applicable to flutter_map, but I'll break it down.

> Moving to a custom RenderObject

We'd do this for the existing canvas layers, ie. the feature layers. However, I'm not sure of the performance impacts - we do have to redraw very often (every frame of map movement), but we already do extensive caching and batching to try to compensate.

> Optimizing Repaints with Repaint Boundaries

I haven't looked into this, but again, I'm not sure it would help, as everything is being rendered anyway. We can experiment just by wrapping each layer, I'd be interested in the result.

> Avoids the performance overhead associated with setState() and unnecessary widget tree recalculations.

This is potentially a good idea. But it requires a very large rethink of the internal state system. 

The remainder seems like more general advice, some of which we already do, some of which doesn't apply without a major rethink.

I'm all up for improving the performance, but I think we need to do a more detailed performance review and tracing before changing things too much. The recent changes to support multi-world have also complicated things massively, and may have reduced performance, but that's unavoidable as we have to do more work.

Are you having any specific performance issues? Please be aware that, especially for FM, performance in debug mode is significantly worse than performance in release mode, even for simple maps.

bombicod:
Thank you for your reply!
In release mode the speed is much better, but here's what was discovered. on the video
30k points with the use of a circle layer (important in meters) + 30k polylines at any zoom freezes very much.
- by experiment we determined that the whole thing is in the circle layers in meters.
Please tell me how we can implement such a case without freezing?

https://github.com/user-attachments/assets/6ef9af85-bd94-417b-9d46-46ddc0151168

monsieurtanuki:
Regarding CircleMarkers with radius in meters, we systematically recompute the radius for each CircleMarker:
```dart
  /// Converts a distance in meters to the equivalent distance in screen pixels,
  /// at the geographic coordinates specified.
  double metersToScreenPixels(LatLng point, double meters) =>
      (camera.getOffsetFromOrigin(point) -
              camera.getOffsetFromOrigin(_distance.offset(point, meters, 180)))
          .distance;
```
[...]

PR #2101, opened 2025-05-29
feat: add `optimizeRadiusInMeters` option to `CircleLayer`

monsieurtanuki:
### What
- With the `useRadiusInMeter` parameter, we compute again and again the radius of each `CircleMarker`. Which depends on its latitude and requires heavy computation.
- Now, with the new `optimizeRadiusInMeters` parameter for `CircleLayer`, we are able to cache the number of pixels per meter, and to compute the value once for all markers.
- With this new feature, displaying 20k `CircleMarker`s doesn't freeze anymore.
- cc. @bombicod
- Closes: #2076

### Impacted files
* `circle_layer.dart`: new `bool optimizeRadiusInMeters` field
* `many_circles.dart`: now displaying radius of 100km (instead of 5 pixels), with `optimizeRadiusInMeters: true`
* `painter.dart`: caching the meter in pixels value

### DIFF
--- a/example/lib/pages/many_circles.dart
+++ b/example/lib/pages/many_circles.dart
@@ -31,6 +31,7 @@ class _ManyCirclesPageState extends State<ManyCirclesPage> {
   int numOfCircles = _maxCirclesCount ~/ 10;
   List<CircleMarker> allCircles = [];
 
+  bool useBorders = false;
   bool useRadiusInMeters = false;
   bool optimizeRadiusInMeters = true;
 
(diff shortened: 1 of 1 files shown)


monsieurtanuki:
> Ok, that seems like a fair tradeoff :) Definitely needs some docs though - I think describing those scenarios is a good idea. If you wanted to write that, feel free, or I can add some docs.
> 
> I'll also try to go through and add a more dynamic demo.

@JaffaKetchup If you find inspiration for a demo, I guess you'll also find inspiration for docs/comments: I let you deal with both.

JaffaKetchup (approved):
LGTM, thanks again :)

monsieurtanuki:

PR #2110, opened 2025-06-10
fix: added Not-null assertion operators

ReinisSprogis:
Produces compile time error without them.


ReinisSprogis:
Hi. Yes. My bad. It just like that went away. I don't know why I was getting error there and then suddenly I did not - after submitting pull request.

Issue #2111, opened 2025-06-11
Stack Overflow Error in feature_layer_utils


Finni123:
### What is the bug?

A crash was reported on one of my user's devices.
Unfortunately, no logs were sent with it so I only have the stacktrace, but I don't know what happened...

I am very sorry that I can't provide a minimal reproducible example or anything else, but I thought it's probably good to report this here anyways.

This occurred on a non-rooted Android 15 phone in portrait mode. That's all I know :(

This is what my map looks like:
```dart
    
  final MapController _mapController = MapController();
  late final StreamController<double?> _alignPositionStreamController;
  /// ...

    return FlutterMap(
      mapController: _mapController,
      options: MapOptions(
        keepAlive: true,
        initialCenter: provider.center,
        initialZoom: provider.center.zoom,
        initialRotation: provider.center.rotation,
      ),
      children: [
        TileLayer(
          urlTemplate: 'https://tile.openstreetmap.org/{z}/{x}/{y}.png',
          userAgentPackageName: _packageName,
          tileProvider: NetworkTileProvider(silenceExceptions: true),
        ),
        _floorLayer(provider),
        if (locationReady) _currentLocationLayer(),
        if (_showMarker) _markerLayer(provider),
      ],
    );

/// here are the layers:
  Widget _floorLayer(AppDataProvider provider) {
    return Consumer<AppDataProvider>(
      builder: (context, provider, _) {
        String mapAsset = provider.mapAsset;
        return IgnorePointer(
        child: OverlayImageLayer(
          overlayImages: [
            RotatedOverlayImage(
              bottomLeftCorner: provider.mapBotLeft,
              bottomRightCorner: provider.mapBotRight,
              topLeftCorner: provider.mapTopLeft,
              imageProvider:
                  AssetImage(mapAsset),
            ),
          ],
        ),
      );
      },
    );
  }

  Widget _currentLocationLayer() {
    return IgnorePointer(
      child: CurrentLocationLayer(
        alignPositionStream: _alignPositionStreamController.stream,
        style: LocationMarkerStyle(
          accuracyCircleColor: AppConfig.instance.color.withAlpha(32),
          headingSectorColor: AppConfig.instance.color.withAlpha(32),
          headingSectorRadius: 10,
          marker: DefaultLocationMarker(color: color),
        ),
      ),
    );
  }

  Widget _markerLayer(AppDataProvider provider) {
    return IgnorePointer(
      child: MarkerLayer(markers: [
        Marker(
[...]

PR #2113, opened 2025-06-24
fix: adjust infinite loop protection in `workAcrossWorlds`

monsieurtanuki:
As discussed in https://github.com/fleaflet/flutter_map/issues/2111#issuecomment-2996083734, some fine-tuning about infinite loop error.


JaffaKetchup (approved):
LGTM thanks :)

PR #2122, opened 2025-07-08
fix(example): updated marker stress test to use more stressful markers

ReinisSprogis:
This better reflects real life usage and looks better than rectangles.  Updated to match single marker demo.


example/lib/pages/many_markers.dart
```diff
@@ -39,20 +38,45 @@ class ManyMarkersPageState extends State<ManyMarkersPage> {
...
+              ),
+              child: Icon(Icons.location_pin, size: 30, color: getRandomColor()),
+            )));
       }
       setState(() {});
     });
   }
 
+  Color getRandomColor() {
+    final source = Random();
```
josxha, line 71: Try to reuse the `Random` instance from line 40 for color generation. Having a seed might be a good thing as well to have a random but consistent result.

example/lib/pages/many_markers.dart
```diff
@@ -62,18 +86,8 @@ class ManyMarkersPageState extends State<ManyMarkersPage> {
...
-                ),
-                padding: const EdgeInsets.only(
-                  left: 16,
-                  right: 16,
-                  top: 88,
-                  bottom: 192,
-                ),
-              ),
+              initialCenter: london,
+              initialZoom: 10,
```
josxha, line 90: initialZoom might have problems on different screen sizes. I'd personally stick with initial camera bounds.

example/lib/pages/many_markers.dart
ReinisSprogis: Hi. Not sure how to implement seed.

josxha (approved):

ReinisSprogis:
@JaffaKetchupI don't think it's related to the demo itself. I tested the original demo by simply swapping ColoredBox with Icon and observed the same result. ColoredBox might have been too primitive to reveal the issue. Replacing it with any slightly more complex widget causes noticeable lag.

I'm just speculating here, but maybe when all markers use the same widget, same color, size, and structure. Flutter applies some optimization under the hood, recognizing they're identical. As you can see, there's a clear difference between using colored markers and markers with a single color.

Still, it's unusual that removing markers would degrade performance, unless there's inefficient filtering happening under the hood.

```dart
 Marker(
     point: LatLng(doubleInRange(r, 37, 55), doubleInRange(r, -9, 30)),
     height: 12,
     width: 12,
     child: Icon(
         Icons.location_on,
         size: 12,
         color: Colors.primaries[r.nextInt(Colors.primaries.length)]
     ),
 ),
 ```

https://github.com/user-attachments/assets/02dabe28-0149-40ba-ab33-78300ce65a38

https://github.com/user-attachments/assets/fcb2ba36-d8d5-4b90-beab-aed214e8b6c4

PR #2125, opened 2025-07-11
fix: ensure tiles still load when failing to generate cache metadata

### LINKED ISSUE #2124: Tiles fail to load when built-in caching fails to generate a tile's metadata due to a HTTP spec non-conformity in the tile server
### What is the bug?

When using the built-in caching feature of FlutterMap 8.2.0, an issue arises if the tile provider’s HTTP headers include a date that does not conform to RFC-1123 or RFC-850 (e.g., Stadia Maps). 

In such cases, `HttpDate.parse(nonConformDate)` fails to parse the date, resulting in an HttpException: Invalid HTTP date.

As a result, no tiles are displayed at all.

While this behavior is not a bug in the library itself, it would be helpful if such cases were handled more gracefully—for example, by falling back to the fallbackFreshnessAge setting when an invalid date is encountered.

### How can we reproduce it?

Using Stadia Maps as the tile provider, the received HTTP headers (attached as an image) contain a date in the last-modified field that is incorrectly formatted.

For example:
Last-Modified: Fri, 11 Jul 2025 09:54:56 +0000

According to the standard (RFC-1123), the timezone should be specified as GMT, not +0000. So the correct format would be:
Last-Modified: Fri, 11 Jul 2025 09:54:56 GMT

Image

- Gracefully handle these cases by falling back to fallbackFreshnessAge when a date header cannot be parsed.

- Alternatively, support non-standard date formats like +0000 (though this is less ideal and potentially error-prone).

Personally, I would recommend the first approach, as it’s safer and more predictable. I can open a PR it if you want.

JaffaKetchup:
Also warn when failing to generate cache metadata or using fallback freshness age

Fixes #2124.

### DIFF
--- a/lib/src/layer/tile_layer/tile_provider/network/caching/tile_metadata.dart
+++ b/lib/src/layer/tile_layer/tile_provider/network/caching/tile_metadata.dart
@@ -1,8 +1,9 @@
 import 'dart:io' show HttpHeaders, HttpDate; // web safe!
 import 'dart:math';
 
+import 'package:flutter/foundation.dart';
 import 'package:flutter_map/flutter_map.dart';
-import 'package:meta/meta.dart';
+import 'package:logger/logger.dart';
 
 /// Metadata about a tile cached with a [MapCachingProvider]
 ///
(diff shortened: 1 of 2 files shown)


scapin01:
@JaffaKetchup Many thanks for building a fix so quickly. I tested it directly in my App (RiverApp) and it works like expected (a proper warning, and the error is catched).

mootw (approved):
not tested but looks good

PR #2128, opened 2025-07-13
fix: Preserving tile state on rebuild 

ReinisSprogis:
Previously each tile widget was discarded and rebuilt on every map gesture. Now it just gets repositioned. I still need to check if this doesn't cause any implications. I think it should improve performance, but "I think" is not a metric, so still want to check on that.


ReinisSprogis:
Yeah. I had the same feeling. And with new TileLayer on horizon, might not be worth it.

PR #2131, opened 2025-07-21
fix: Added value key and RepaintBoundary to preserve Marker state.

ReinisSprogis:
Added ValueKey and RepaintBoundary to preserve Marker state. m.key was always null. However, might be that implementation can produce duplicate Keys if coordinates of the two markers are the same. Need a better sollution for key, but this is start.

### DIFF
--- a/lib/src/layer/marker_layer/marker_layer.dart
+++ b/lib/src/layer/marker_layer/marker_layer.dart
@@ -72,23 +72,22 @@ class MarkerLayer extends StatelessWidget {
 
               // Shift original coordinate along worlds, then move into relative
               // to origin space
-              final shiftedLocalPoint =
-                  Offset(shiftedX, pxPoint.dy) - map.pixelOrigin;
-
+              final shiftedLocalPoint = Offset(shiftedX, pxPoint.dy) - map.pixelOrigin;
+             
+              final markerKey = m.key ?? '${m.point.latitude}:${m.point.longitude}';
               return Positioned(
-                key: m.key,
-                width: m.width,
-                height: m.height,
-                left: shiftedLocalPoint.dx - right,
-                top: shiftedLocalPoint.dy - bottom,
-                child: (m.rotate ?? rotate)
-                    ? Transform.rotate(
-                        angle: -map.rotationRad,
-                        alignment: (m.alignment ?? alignment) * -1,
-                        child: m.child,
-                      )
-                    : m.child,
-              );
+                  key: ValueKey('$markerKey@$shiftedX'),
+                  width: m.width,
+                  height: m.height,
+                  left: shiftedLocalPoint.dx - right,
+                  top: shiftedLocalPoint.dy - bottom,
+                  child: RepaintBoundary(child: (m.rotate ?? rotate)
+                      ? Transform.rotate(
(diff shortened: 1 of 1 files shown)


ReinisSprogis:
I found out that m.key is null for all markers, So as soon as children of Stack changed they all rebuild. 
You can see that in this video with highlite repaints how they all repaint.

https://github.com/user-attachments/assets/9061be11-a2d5-4ed7-919b-67d793aab605

So now with non null key and RepaintBoundary we preserve the state. 

https://github.com/user-attachments/assets/05523848-04b6-4f49-9345-4756311ff518

ReinisSprogis:
Ahh. I missclicked and closed PR. So reopened. 
It looks to improve with this fix as there is no more of that jank on ui thread. 

https://github.com/user-attachments/assets/14dd12a9-27b6-4b63-bebf-da6c8eef4a1a

Unlike before it was lagging as soon as any marker left the screen. 

https://github.com/user-attachments/assets/1cd451c7-2666-4988-82db-3fcd6625374e

But need to figure out how to give it a stable and uniqe Key. If you have any ideas, welcome to contribute.

JaffaKetchup:
This is a nice fix, thanks. But I would first like to find out when this issue was introduced, since I don't remember it always being there. Is this ready for review?

ReinisSprogis:
Hi. Not yet. Still need to figure out what key to use. Right now, the key is based on the marker’s location, but if two markers share the same location, it throws a “Duplicate keys found” exception.

As I mentioned, I’m not sure what the best key would be in that case. One option might be to use a UniqueKey when creating each marker, which would avoid the duplicate issue, but then I wouldn’t be able to declare the Marker as const. That could reduce some of the performance optimizations Flutter provides.

ReinisSprogis:
Ah yes ObjectKey can be used. #2134 looks good, Not sure I understand it all. I don't have much experiance with custom RenderObjects.
Regarding culling, need to test and see whats better. Maybe checking if outside the screen and toggle visibility in Offstage widget https://api.flutter.dev/flutter/widgets/Offstage-class.html
 I think It would be better to rename Marker to MapWidget and WidgetLayer (or similar) as it has nothing to do with Marker as such.  If user want's to put marker as child, thats ok.

JaffaKetchup:
For the moment, #2134 is nice but low priority for me right now (I've got limited time for the next while). I think that's the "correct" way to do things - moreso the "right" way to support multi-worlds (as it just renders multiple times instead of building multiple times) - but it's a lot to figure out and I'm struggling with the rotation stuff & coordinate maths (as I mentioned there). It's close to working, but not there just yet. + research is needed into culling need.

Therefore, for the time being, this might be worth perusing if you're still interested. 

Renaming stuff is not possible for the time being. If a user wants to overlay a widget more precisely, there's a plugin for that which we could consider incorporating in the future, but the marker API & usage is pretty standardised across mapping libraries and so doesn't need changing.

But if using the `Marker` as the `ObjectKey(...)` works, then I think this could be a very good improvement.

ReinisSprogis:
I don't know why, but seems like performance is overal reduced with this fix. But removes that lag when markers are culled.

PR #2133, opened 2025-07-22
fix: prevent negative left-shift when using negative `zoomOffset`

monsieurtanuki:
Fixed the case where zoomOffset is negative.
Couldn't test it properly, e.g. with Retina Tile Layer example.

{Edit] I did manage to reproduce the bug though, by forcing zoomOffset to -1.

### DIFF
--- a/example/lib/pages/tile_builder.dart
+++ b/example/lib/pages/tile_builder.dart
@@ -32,23 +32,27 @@ class TileBuilderPageState extends State<TileBuilderPage> {
         children: [
           tileWidget,
           if (showLoadingTime || showCoordinates)
-            Column(
-              mainAxisAlignment: MainAxisAlignment.center,
-              children: [
-                if (showCoordinates)
-                  Text(
-                    '${coords.x} : ${coords.y} : ${coords.z}',
-                    style: Theme.of(context).textTheme.headlineSmall,
-                  ),
-                if (showLoadingTime)
-                  Text(
-                    tile.loadFinishedAt == null
-                        ? 'Loading'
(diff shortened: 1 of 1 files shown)


JaffaKetchup (changes requested):
This doesn't look like it works quite right :(

image

JaffaKetchup:
To be fair, I don't know necessarily what I would expect to see with a negative zoom offset - maybe that screenshot is what it's supposed to be? I'll have a look at what it looked like on an older version.

JaffaKetchup:
Maybe this is working correctly. Using this code to test:

```dart
TileLayer(
      zoomOffset: -1,
      minNativeZoom: 1,
      maxNativeZoom: 18,
      tileProvider: NetworkTileProvider(httpClient: httpClient),
    );
```

I think yesterday I tested without a `minNativeZoom`, which would obviously cause problems. Nevermind!

> [Ah, everything's got to be perfect with you!](https://www.youtube.com/watch?v=UucXz3ZGmF4&t=90s) 😉

Sorry, you're right 😅

JaffaKetchup (approved):
I think this works correctly. Will try to ask original reporter to test, but seems to be OK. Thanks and apologies :)

JaffaKetchup:
Hey @mash-g, I'd be interested to know if this actually fixes an issue you have - I guess a negative offset is more common than I expected 😂

But in answer, we have no fixed release schedule. Likely this one will be waiting for a little bit unless there's another more critical fix soon - I would like to release them faster, but it requires coordination from the rest of the maintainer team, so it's usually not too quick.

I'll open a PR for the release, but have to wait for a review before it can be merged and released.

PR #2137, opened 2025-07-26
feat: add pointer move callback to MapOptions and handle it in MapInteractiveViewer

ReinisSprogis:
I needed a onPointerMove functionality but there was only onPointerHover. For example when I want to drag graphics on map, I disable map pan onTapDown and can use onPointerMove to reposition graphics. This cannot be done with onPointerHover as it does not respond when tap is also down. There already was an _onPointerMove method just not exposed to user. Any reason not to do this?

### DIFF
--- a/lib/src/gestures/map_interactive_viewer.dart
+++ b/lib/src/gestures/map_interactive_viewer.dart
@@ -420,6 +420,10 @@ class MapInteractiveViewerState extends State<MapInteractiveViewer>
   }
 
   void _onPointerMove(PointerMoveEvent event) {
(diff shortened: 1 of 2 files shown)


JaffaKetchup:
Hey @ReinisSprogis, just wondering why one of the existing movement listeners wouldn't work?

The `MapOptions.onPositionChanged` callback has an argument which is set if the movement was gesture initiated (instead of controller initiated), if that's what you're looking for: https://pub.dev/documentation/flutter_map/latest/flutter_map/PositionCallback.html.

PR #2145, opened 2025-08-25
feat: add nativeZooms property for custom zoom level support

awoni:
- Add nativeZooms property to specify available discrete zoom levels
- Add NativeZoomStrategy enum with multiple selection strategies
- Implement intelligent zoom clamping based on distance and strategy
- Maintain backward compatibility with existing zoom behavior

### DIFF
--- a/lib/src/layer/tile_layer/tile_layer.dart
+++ b/lib/src/layer/tile_layer/tile_layer.dart
@@ -29,22 +29,21 @@ part 'wms_tile_layer_options.dart';
 /// https://docs.fleaflet.dev/usage/layers/tile-layer. Some are important to
 /// avoid issues.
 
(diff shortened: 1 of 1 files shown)


JaffaKetchup:
Hey, sorry for leaving this for so long. I like the concept behind it, and I can see how it would be useful, but it would be good to get an idea from another @fleaflet/maintainers on what they think about the API, as this does complicate the already extensive list of options on the `TileLayer` a little more.

JaffaKetchup:
Hey @awoni, I'm just looking to start tweaking this so we can merge it. Please could you tick the box to allow us maintainers to edit the branch attached to this PR, or if this isn't OK, please let me know? Thanks!

PR #2158, opened 2025-10-18
fix: improve fling behaviour when pointer leaves window

ben-milanko:
This PR makes some minor changes to the drag behaviour aimed at improving the experience on web and desktop.

First, if the cursor leaves the screen while dragging, the fling drag behaviour will send the map camera moving in the opposite direction.

Before:

https://github.com/user-attachments/assets/3307aacd-f0a9-4039-ae61-c221b69e3bbe

After:

https://github.com/user-attachments/assets/34cbb458-b9ac-4249-9996-94c9f38fbe29

Secondly, I would like to modify the damping behaviour of the fling for desktop users. This has been made part of the public API via `InteractionOptions` the default has been left per the current.

I've found a value of 2 is closer to the behaviour of Google Maps on web.


JaffaKetchup:
Ah sorry, misread! Ok, I'll have a look :)

JaffaKetchup (dismissed):
Hey @ben-milanko, thanks, and sorry it's taken so long to get to this!

The feature addition of `flingAnimationDampingRatio` looks good to me (bar the quick documentation fix).

However, it looks like the fix part of the PR has introduced a regression.

I will say that I found it difficult to reproduce the original issue on my computer, but I could in a few cases. This appears to have been fixed from what I can see.

However, when the user now has a curved fling before release, the direction of the fling is incorrect. The direction should be the last direction of the pointer, not the overall direction of the preceding gesture.


Correct behaviour on master:

https://github.com/user-attachments/assets/1ccdaefd-adcb-49fc-b357-1a4789ea0222

Incorrect behaviour on touchscreen:

https://github.com/user-attachments/assets/8abc3640-c920-4093-8409-8647ebbea4c1

Incorrect behaviour on desktop:

https://github.com/user-attachments/assets/2bd77978-b6fa-4820-a09d-18baa6e55c21

Do you think it makes sense to split out the fix part from this PR so we can merge the feature, and merge the fix later (if you're still willing to work on it?).

ben-milanko:
Hey thanks for looking at the PR @JaffaKetchup. I can see the issue from your videos.

The original issue with the rebound behaviour seems to still be there as well on the base branch, I don't think this was fixed.
It's easier to replicate with the window moved to the edge of the screen.

I've added an additional tracked point so the fling can correctly use the last swiped direction for the fling rather than the direction from the start to end point. Are you okay with this change?

JaffaKetchup:
Whoops, sorry, I'll try and review this soon.

JaffaKetchup (commented):
https://github.com/fleaflet/flutter_map/pull/2158#issuecomment-3678886504

mootw (commented):
even broken, the new behavior is an overall net positive. it would be nice to see the last-second angle change fixed before merging though. code lgtm!

JaffaKetchup (approved):
I'm going to approve this for now. I think it would be good to fix in the future the last remaining glitch, but as @mootw said, I think this is a move in the right direction. Thanks for all your work :)

PR #2165, opened 2025-11-18
chore: upgrade example app build configuration & dependencies

JaffaKetchup:
Also fixed a minor display bug in the example app

(Replaces #2164 with fixed line endings)

### DIFF
--- a/.metadata
+++ b/.metadata
@@ -1,29 +1,10 @@
-# This file should be version controlled and should not be manually edited
+# This file tracks properties of this Flutter project.
+# Used by Flutter tool to assess capabilities and perform upgrades etc.
+#
+# This file should be version controlled and should not be manually edited.
 
 version:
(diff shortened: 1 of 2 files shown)


.metadata
```diff
@@ -1,10 +1,29 @@
...
+# This file should be version controlled and should not be manually edited
 
 version:
-  revision: 8962f6dc68ec8e2206ac2fa874da4a453856c7d3
-  channel: stable
+  revision: "cbfa8f452d1837e8d8b3fd0f25e22fff0f0f5bba"
+  channel: "main"
 
-project_type: package
+project_type: app
```
josxha: Not as important but this file is kinda messed up. It should have project_type "package" and be created from the stable channel.

JaffaKetchup: Yeah gone back to stable and regenerated them 👍

PR #2168, opened 2025-11-28
fix: 2022 - no fling beyond the antimeridian if constrained

### LINKED ISSUE #2022: [BUG] Map warps to other side of the planet when map center hits 180 W
### What is the bug?

When I pan my map west to -180, instead of stopping like a wall, it flips around to the other side of the globe.  

Is there a way to put the "wall" behavior back? I enjoyed that natural blocking behavior actually. 

### How can we reproduce it?

Unless it's something in my code specifically, just pan/fling westward beyond -180 and see what happens.  

If you cannot reproduce, I'll give more info on my setup for troubleshooting.

_No response_

### Platforms

All

### Severity

monsieurtanuki:
### What
Fix in 2 parts:
* during a fling, if at one step we cannot move, we stop immediately the rest of the fling animation
* when we apply the LatLngBounds constraint, we now take into consideration the specific -180,180 "constraint". Of course the longitude will always be between -180,180, so we need an extra test to check if we got beyond the limits.
* Fixes: #2022

### Files
New file:
* `antimeridian.dart`: Testing if we can fling beyond the world. We can't, due to the constraint.

Impacted files:
* `camera_constraint.dart`: special case when the constraint is the world
* `main.dart`: added new `AntimeridianPage` page.
* `map_interactive_viewer.dart`: stop fling when the `moveRaw` fails
* `menu_drawer.dart`: added new `AntimeridianPage` page.


JaffaKetchup:
Hey @monsieurtanuki, thanks for giving it a shot! If you could give me any hints as to what you were thinking to do next if anything (if you can remember), I'll see if I can take a look at it. For the time being, I'm going to close this PR (feel free to ping if you want it reopened :)).

PR #2184, opened 2026-02-23
fix: consider `strokeWidth` when calculating polyline visibility

braindamagedman:
For thick polylines, setting `cullingMargin: null` is not always sufficient: if all polyline points are outside the viewport, the polyline is not rendered. This can be undesirable in many cases.

ezgif-27183e0c0d11c842-ezgif com-loop-count


JaffaKetchup (approved):
LGTM, thanks! Although would be good to get another @fleaflet/maintainers' review.

This does result in an additional call to `metersToScreenPixels` (which involves projections) for each polyline+world when it may have been culled before, but if this gives the correct result, it's necessary so its fine.

(Just a note to readers, this doesn't fix the case where polylines disappear to early when the stroke width is 'large' and `cullingMargin` isn't `null` and hasn't been increased to account for this.)

PR #2195, opened 2026-04-10
perf: replace unstable `Tile` widget key with dedicated stable `TileKey`

braindamagedman:
Previously tiles were keyed with `ObjectKey(tileRenderer)`, but `TileRenderer` instances are recreated on every build. Because of that, the key was unstable even when the logical tile did not change. The root cause of this issue was introduced in commit
[182c4fc](https://github.com/fleaflet/flutter_map/commit/182c4fc0c73411e2be10774dd227d0da9cf4276a) when unbounded horizontal scroll was introduced.

As a result, Flutter could not match the new widget with the existing element and treated the tile as a new widget. This led to repeated disposal and recreation of `_TileState` instead of updating the existing state.

The fix replaces that key with a stable `TileKey` based on `TileImage` identity and `positionCoordinates`. This preserves `_TileState` for the same rendered tile and avoids unnecessary remounting.

Before:
before

After fix:
after

You can check simple code to reproduce before/after examples here:
https://github.com/braindamagedman/flutter_map/tree/test/tile-key

### DIFF
--- a/lib/src/layer/tile_layer/tile_key.dart
+++ b/lib/src/layer/tile_layer/tile_key.dart
@@ -1,0 +1,37 @@
+import 'package:flutter/foundation.dart';
+import 'package:flutter_map/src/layer/tile_layer/tile_coordinates.dart';
+import 'package:flutter_map/src/layer/tile_layer/tile_image.dart';
+import 'package:flutter_map/src/layer/tile_layer/tile_renderer.dart';
+
+/// A key that identifies rendering of a [TileImage] at given [TileCoordinates].
+///
+/// Two [TileKey]s are equal when they reference the same [TileImage]
+/// instance and have equal [positionCoordinates].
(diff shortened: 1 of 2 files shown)


JaffaKetchup:
Oo, thanks for this, we should've caught that. That's a useful debug text, we might add that to the example app. 

We need to make sure that this isn't preventing the tiles from changing when we actually do want the state to change, since we've had far too many issues there in the past, but I think your fix should be good.

JaffaKetchup (approved):
Ok, LGTM, thanks! I am ofc hesitant in these parts but I think the change itself looks sensible and I can't see any issues with testing. Indeed, it looks like it has improved the performance a bit!

(I don't know why 100% we couldn't have just changed the equality of the `tileRenderer`, but I think this makes sense anyway.)

PR #2198, opened 2026-04-13
feat: add smooth scroll zooming

vinlet:
https://github.com/user-attachments/assets/d5def771-e272-47f8-bcff-9bdd9b91e4bc

With some help from Claude Code and Opus 4.6 (I'm sorry) I ported MapLibre GL JS's implementation of animated smooth zooming for mouse wheel scrolls. I feel like this was sorely missing from `flutter_map`, though I'm aware web/desktop users are probably not in the majority.

MapLibre's implementation tries to differentiate mouse wheels from trackpads through some magic (whic is apparently [damn](https://github.com/mapbox/mapbox-gl-js/issues/7572) [hard](https://github.com/facebookarchive/fixed-data-table/blame/master/src/vendor_upstream/dom/normalizeWheel.js)). I ported that logic verbatim but not sure how well it applies to Flutter. Apparently pretty well.

On top of MapLibre's implementation I made the scroll rate and animation duration configurable, but non-defaults probably dont't play well with some of the other hardcoded delays from MapLibre's implementation.

The old behavior still exists but I made smooth scrolling the default which is technically a breaking change, so let me know if that's not ok...

Manually tested it in:

- Windows with mouse wheel and trackpad 
- Chrome with mouse wheel and trackpad
- Firefox with mouse wheel and trackpad
- Android with mouse wheel

Random notes:

- On Windows and Chrome, my mouse seems to produce scroll deltas in 100 increments. Firefox does 102 🤷‍♂️. Chrome with trackpad does only integer deltas, Firefox with trackpad does deltas with fractional parts as well. Android with mouse wheel did 64.
- Added tests. From what I've seen there weren't really tests for gestures, which is understandable because it's a pain. Main pain point was the usage of `DateTime.now()`, but I found somewhat of a clever/ugly hack to fake it out with `package:clock` without adding it as a dependency.
- Should I somehow dedupe the tests? (most code is duplicated between scroll up/down tests)
- Should I somehow test the exact zoom values that smooth zooming should arrive at?
- Initially I duplicated `TestApp` in `scroll_wheel_zoom_test.dart` to add in custom `interactionOptions` but realized it's a bad idea long term. So instead I added an `interactionOptions` param to `TestApp`. But this still feels like pointless busywork. In my projects I typically have something like `wrapInTestApp(Widget child)` where you can pass in any `FlutterMap` you like. I didn't do this here because the PR would blow up.
- Removed the OSM URL from `TestApp`, tests still work and don't pointlessly print messages when running tests.
- We could deprecate `InteractionOptions.scrollWheelVelocity` and move it into `ScrollZoomOptions`, maybe even merge it with `ScrollZoomOptions.wheelZoomRate` somehow?
- Is this better as a plugin? Currently I don't see how to spin it off easily into a plugin.
- Please nitpick the PR :) I'm not really familiar with the codebase entirely, so if I missed something or my approach could be improved, let me know.

Credit where credit is due:
- https://github.com/maplibre/maplibre-gl-js/blob/main/src/ui/handler/scroll_zoom.ts
- https://github.com/mapbox/unitbezier/blob/master/index.js


JaffaKetchup:
Hey @vinlet,

Wow! This is hugely appreciated - and something I've wanted to implement and has been top of my personal priority list for a while, but I've always been put off by needing to handle the whole trackpad/mouse thing (it's a well known issue in Flutter as well, see https://github.com/flutter/flutter/issues/32120).

Although it's true that most users use flutter_map on mobile platforms, one of flutter_map's biggest advantages is that it does actually run on web and desktop unlike the alternatives, so any improvements that bring that support even further ahead are great.

Given this is such a big thing to test and get right, and because the gesture handling has historically been a messy part of flutter_map, this might not be merged quickly (especially as I'm about to be unavailable for coding/testing for a while). But we'll look into it. Thanks again!

vinlet:
@JaffaKetchup that's ok! I'll try to be available and respond if there's news.
Note to self: I missed that MapLibre also has tests for smooth scrolling, should have looked into those as well

JaffaKetchup (approved):
Hey, thanks for this again! I think this looks good to me. I'm tempted to release this into a prerelease first to see whether we can drum up some testers, since I'm aware that this part of the code can be very very fiddly and behave differently with different devices and on different platforms - but I'm not sure yet.

I'm also considering making a couple other changes to the example app and API surface as you mentioned.

JaffaKetchup:
I've updated the example app and the API. Also removed a couple tests that I believe are redundant (the ones testing defaults of the constructors of `ScrollZoomOptions`).

vinlet:
hey, sorry for the late reply, haven't logged on in a while. Yeah we should do quite a bit more testing especially if we make this the default.

If anyone stumbles across this PR and wants this feature, please test it locally with all the platforms and mice and laptop trackpads you have and report the results :) Ideally it would be great to see the scroll deltas and timings if you find an issue, but you'll have to add logging yourself...

JaffaKetchup:
I've managed to test this on a mac mini and my Ubuntu dualboot, seems to be working OK both on platform and on web on each \:)

Unfortunately I don't have a special mouse or anything (like an Apple mouse), but I have been testing with a trackpad on Windows and Ubuntu (platform+web) as well, and that seems to be working.

I've also changed the API a little again, and improved the behaviour of the deprecation and transition. The new smooth behaviour will be the default for all users in v8.4+, unless the current/old scroll wheel velocity argument is not at its default (and the new arguments have not been changed from their defaults).

Looking good so far, definitely looking to get this merged in now!

PR #1964, opened 2024-09-12
fix: added polygon validity check before hit testing

JaffaKetchup:
Fixes #1933 using method suggested by @mootw (and which I think also makes sense).

Issue #1933, opened 2024-07-17
[BUG] Hit testing `Polygon` with less than 3 points fails assertion


longch:
### What is the bug?

When using PolygonLayer with simplificationTolerance, the program reports an error when sliding or zooming the PolygonLayer.:

```
I/flutter (26644): zone _handleError 'package:flutter_map/src/misc/point_in_polygon.dart': Failed assertion: line 11 pos 10: 'len >= 3': not a polygon stack #0      _AssertionError._doThrowNew (dart:core-patch/errors_patch.dart:51:61)
I/flutter (26644): #1      _AssertionError._throwNew (dart:core-patch/errors_patch.dart:40:5)
I/flutter (26644): #2      isPointInPolygon (package:flutter_map/src/misc/point_in_polygon.dart:11:10)
I/flutter (26644): #3      _PolygonPainter.elementHitTest.<anonymous closure> (package:flutter_map/src/layer/polygon_layer/painter.dart:86:17)
I/flutter (26644): #4      _PolygonPainter.elementHitTest (package:flutter_map/src/layer/polygon_layer/painter.dart:91:10)
I/flutter (26644): #5      HitDetectablePainter.hitTest (package:flutter_map/src/layer/shared/layer_interactivity/internal_hit_detectable.dart:75:11)
I/flutter (26644): #6      HitTestRequiresCameraOrigin.hitTest (package:flutter_map/src/layer/shared/layer_interactivity/internal_hit_detectable.dart:112:18)
I/flutter (26644): #7      RenderCustomPaint.hitTestSelf (package:flutter/src/rendering/custom_paint.dart:564:43)
```

Disabling simplificationTolerance will not report an error.

flutter_map: ^7.0.2
flutter_map_geojson: ^1.0.8
maps_toolkit2: ^1.0.1+1

### How can we reproduce it?

```import 'package:flutter/material.dart';
import 'package:flutter_map/flutter_map.dart';
import 'package:maps_toolkit2/maps_toolkit2.dart';
import 'package:flutter_map_geojson/flutter_map_geojson.dart';

class MapPage extends StatefulWidget {
  const MapPage({super.key});

  @override
  State<MapPage> createState() => MapPageState();
}

class MapPageState extends State<MapPage> {
  final testGeoJson = {
    "type": "FeatureCollection",
    "features": [
      {
        "type": "Feature",
        "geometry": {
          "type": "Polygon",
          "coordinates": [
            [
              [107.07002, 26.32992],
[...]

PR #2204, opened 2026-04-26
fix: remove polygons with fewer than 3 points after simplification

rekire:
When you have a (multi) polygon with lots of tiny polygons the simplification can shrink a polygon to e.g. two points, this will cause a this assertion error:

```
The following assertion was thrown while handling a pointer data packet: not a polygon 'package:flutter_map/src/misc/point_in_polygon.dart': Failed assertion: line 10 pos 10: 'len >= 3'
When the exception was thrown, this was the stack:
#2 isPointInPolygon (package:flutter_map/src/misc/point_in_polygon.dart:10:10)
#3 _PolygonPainter.elementHitTest.<anonymous> (package:flutter_map/src/layer/polygon_layer/painter.dart:85:17)
#4 _PolygonPainter.elementHitTest (package:flutter_map/src/layer/polygon_layer/painter.dart:90:10)
#5 HitDetectablePainter.hitTest (package:flutter_map/src/layer/shared/layer_interactivity/internal_hit_detectable.dart:75:11)
#6 HitTestRequiresCameraOrigin.hitTest (package:flutter_map/src/layer/shared/layer_interactivity/internal_hit_detectable.dart:112:18)
#7 RenderCustomPaint.hitTestSelf (package:flutter/src/rendering/custom_paint.dart:569:43)
#8 RenderBox.hitTest (package:flutter/src/rendering/box.dart:2953:58)</anonymous>
```

My fix simply removes those polygons entirely. That happens when you hover a `PolygonLayer` as workaround you can use `simplificationTolerance: 0` but that is based on the geometry quite bad for the performance.

### DIFF
--- a/lib/src/layer/polygon_layer/polygon_layer.dart
+++ b/lib/src/layer/polygon_layer/polygon_layer.dart
@@ -201,7 +201,7 @@ class _PolygonLayerState<R extends Object> extends State<PolygonLayer<R>>
             highQuality: true,
           ),
           growable: false,
-        ),
+        ).where((h) => h.length >= 3).toList(),
       );
 
   @override


JaffaKetchup:
Hi @rekire,
This was already reported a while ago by #1933 and fixed by #1964, released in v8. Could you please confirm that you are seeing this issue in the latest version, and if so, please open a new bug report.
I think it was decided that it's better just to exclude these from hit-testing rather than remove them visually altogether (as if they have a border stroke, for example, they may still be visible?).

rekire:
I'm using my fork right now to avoid this issue. In my `pubspec.yaml` is `flutter_map: ^7.0.2`... I am some major versions behind and I cannot remember right now why. Maybe some incompatible dependencies. I'll try to find it out. I will inform you with the results with a new comment.

rekire:
Okay I found it out: I am using vector_map_tiles which requires that old version and the latest beta does not compile at the moment. So the real issue is that I am using a very old version.

JaffaKetchup:
Ah no worries :) Thanks for contributing anyway, and please feel free to again in future!

PR #2207, opened 2026-05-15
chore: allow 'pkg:latlong2' v0.10.x dependency

ThexXTURBOXx:
As already mentioned, 0.10.x was not really a breaking release. Hence, it can be supported side-by-side with 0.9.x.
1.x will be breaking - but that will take a few more rounds of alpha tests before being published...


JaffaKetchup (approved):

JaffaKetchup:
LGTM, thanks! Looks like we've got flaky checks again.

PR #2211, opened 2026-06-10
perf: avoid quadratic hole-point iteration in `PolygonLayer` offset generation & triangulation

ben-milanko:
While profiling flutter_map in a production app on low-powered hardware, I found that `OffsetHelper.getOffsetsXY` and the `PolygonLayer` Earcut coordinate flattening index a lazily concatenated iterable (`points.followedBy(holePoints.expand(...))`) with `elementAt(i)`. On a lazy iterable that walks from the start on every call, so both paths are **O(n²) in the total point count — every frame** for any polygon with holes.

This PR flattens the points + hole points into a fixed-length list before the loops (and writes the Earcut input into a `Float64List` directly).

## Results

Measured with the benchmark harness added in the first commit (`flutter test benchmark/feature_layer_benchmark_test.dart`, JIT, best-of-reps):

| | before | after |
|---|---|---|
| `getOffsetsXY`, 500-point polygon + 10 holes × 200 points | 26,761 µs/call | 33 µs/call (~800×) |

All existing tests pass. No behavioural change intended — output offsets are identical.

The harness commit is shared with #2212 and #2213 (identical file content, so whichever PR merges first, it collapses out of the others' diffs). It lives in `benchmark/` rather than `test/` so it doesn't extend CI runtime; numbers are only meaningful relative to each other on the same machine.

### DIFF
--- a/benchmark/feature_layer_benchmark_test.dart
+++ b/benchmark/feature_layer_benchmark_test.dart
@@ -1,0 +1,277 @@
+// Layer-level CPU benchmarks, used to validate performance work.
+//
+// Run with:
+//   flutter test benchmark/feature_layer_benchmark_test.dart --plain-name=benchmark -r expanded
+//
+// Numbers are JIT/debug-mode and only meaningful *relative* to each other
+// (before/after a change on the same machine). Each scenario warms up, then
+// times repeated pumps while the camera pans, and reports the best repetition
+// (min) to reduce GC/scheduling noise.
+import 'dart:math' as math;
+
+import 'package:flutter/material.dart';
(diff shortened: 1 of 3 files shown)


ben-milanko:
@JaffaKetchup here's the demonstration for this change (I know the vid doesn't look great, I think the frame rate of the recording couldn't keep up with the demo)

https://github.com/user-attachments/assets/ecd5a0af-28cf-4e13-8667-ed96df1003fd


https://github.com/ben-milanko/flutter_map_perf_demo

JaffaKetchup (approved):
LGTM, thanks again!

PR #2212, opened 2026-06-10
perf: cull polylines via cached projected bounding boxes & chore: prepare for v8.3.1 release

ben-milanko:
While profiling flutter_map in a production app on low-powered hardware, I found two per-frame O(points) scans in `PolylineLayer`'s aggressive culling:

- `stretchesBeyondTheLimits()` iterates every point of every polyline that overlaps the viewport latitudes, every frame
- fully-visible polylines still go through the per-segment `aabbContainsLine` scan (and `sublist` allocations), even though nothing needs to be culled

This PR caches the projected-space bounding box on `_ProjectedPolyline` (computed lazily — culled fragments never compute one; cached instances are created once per zoom level alongside simplification). The world-stretch check becomes an O(1) bbox comparison, and polylines whose bbox is contained in the viewport bounds are yielded whole, skipping the segment scan entirely.

## Results

Measured with `flutter test benchmark/feature_layer_benchmark_test.dart` (JIT, best-of-reps):

| | before | after |
|---|---|---|
| 600 polylines × 60 pts, all visible, panning | 2,631 µs/frame | 2,484 µs/frame (~6%) |
| Mostly-culled scenario | unchanged | unchanged |

Modest in this paint-dominated scenario, but the cull step itself no longer scales with point count, which matters as polyline density grows.

All existing tests pass. The benchmark harness commit is shared with #2211 (identical file content — it collapses out of this diff once either merges).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,29 @@ Please consider [donating](https://github.com/sponsors/fleaflet) or [contributin
 
 This CHANGELOG does not include every commit and/or PR - it is a hand picked selection of the ones that have an effect on most users. For a full list of changes, please check the GitHub repository releases/tags. We also release highlights for some releases on the docs site.
 
+## [8.3.1] - 2026/07/xx
+
+Contains the following user-affecting bug fixes:
+
+- Revert [#2182](https://github.com/fleaflet/flutter_map/pull/2182) to fix multiple internal errors - [#2218](https://github.com/fleaflet/flutter_map/pull/2218) for [#2199](https://github.com/fleaflet/flutter_map/issues/2199)
+- Prevent memory leak and crash when `Marker`s with NaN coordinates provided - [#2213](https://github.com/fleaflet/flutter_map/pull/2213) for [#2178](https://github.com/fleaflet/flutter_map/issues/2178)
+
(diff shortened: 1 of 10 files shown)


JaffaKetchup:
Hey, thanks for these performance improvements! (Slightly annoyed that I left gaps in when putting all the caching stuff together lol 😂). 

I'm still travelling right now, but I'll try to give these a review when I'm back in a couple weeks :)

ben-milanko:
Thanks @JaffaKetchup . Full disclosure these came up when I was doing some AI debugging and optimisations.

To provide some verification of the claims, I put together a repo with samples

There's pretty minimal improvements on this optimisation, as seen below.

https://github.com/user-attachments/assets/2358bcbc-1a97-4df1-8b36-6338ba3c76b7

Full repo is https://github.com/ben-milanko/flutter_map_perf_demo

JaffaKetchup (approved):
LGTM I think, thanks again!

(I'm not super happy with the `_ProjectedPolyline` no longer having a `const` constructor and departing from symmetry with the `_ProjectedPolygon`, but it's still immutable which is the important thing.)

I've also hijacked this PR to prepare for a patch release (v8.3.1).

PR #2216, opened 2026-06-27
fix latlong2: ^0.10.1

ivanzhadnov:


JaffaKetchup:
Hey, #2207 already added support for this version - it hasn't been released yet though. Please use a dependency override to depend on 0.10.1 if you need to for now.

PR #2217, opened 2026-06-28
fix: load tiles during double-tap-drag zoom gesture

ProstoSanja:
# fix: load tiles during double-tap-and-drag (one finger) zoom gesture

## Problem

The double-tap-and-drag zoom gesture (double-tap, hold, then drag to zoom with
one finger — `InteractiveFlag.doubleTapDragZoom`) doesn't load new tiles while
the gesture is ongoing. The map zooms visually, but `TileLayer` keeps showing
the stale/upscaled/missing tiles and only loads the correct ones once the finger is
released. Pinch-to-zoom is unaffected by this behaviour.

## Root cause

`TileLayer` only loads tiles in response to events on
`MapController.mapEventStream`. Every gesture-driven camera change flows through
`MapControllerImpl.moveRaw`, which builds its event via
`MapEventWithMove.fromSource(...)` (`lib/src/gestures/map_events.dart`):

```dart
switch (source) {
  ...
  MapEventSource.onDrag ||
  MapEventSource.onMultiFinger ||   // pinch zoom
  MapEventSource.mapController ||
  MapEventSource.keyboard ||
  MapEventSource.custom            => MapEventMove(...),
  _                                => null,   // ← doubleTapHold lands here
}
```

`MapInteractiveViewer._handleDoubleTapHold` drives the gesture with
`moveRaw(..., source: MapEventSource.doubleTapHold)`, but `doubleTapHold` isn't
in the switch, so `fromSource` returns `null` and **no `MapEvent` is emitted**
for the whole gesture.

The existing `//! PREVENTS TILE LOADING IF SOURCE NOT CONFIGURED IN CONSTRUCTOR`
comment in `moveRaw` points at exactly this gap.

## Testing

- Added `test/map/map_events_test.dart`, asserting `fromSource(... source:
  doubleTapHold)` returns a `MapEventMove` (with `onMultiFinger` as a baseline).

## AI usage disclosure

- I have identified regressive behaviour myself in the application I am developing, but I used Claude Code to find the cause of the bug and analyse the code for potential fixes. 
- Claude Code has been used to generate a test file
- I have manually verified the solution myself in the application I am developing.

### DIFF
--- a/lib/src/gestures/map_events.dart
+++ b/lib/src/gestures/map_events.dart
@@ -133,6 +133,7 @@ abstract class MapEventWithMove extends MapEvent {
           ),
         MapEventSource.onDrag ||
         MapEventSource.onMultiFinger ||
+        MapEventSource.doubleTapHold ||
         MapEventSource.mapController ||
         MapEventSource.keyboard ||
         MapEventSource.custom =>
--- a/test/map/map_events_test.dart
+++ b/test/map/map_events_test.dart
@@ -1,0 +1,25 @@
+import 'package:flutter_map/flutter_map.dart';
+import 'package:flutter_test/flutter_test.dart';
+
+import '../test_utils/test_app.dart';
+
+void main() {
+  testWidgets(
+    'MapEventWithMove.fromSource produces a move event for gesture zooms',
+    (tester) async {
+      final controller = MapController();
+      await tester.pumpWidget(TestApp(controller: controller));
+      final camera = controller.camera;
+
+      MapEventWithMove? eventFor(MapEventSource source) =>
+          MapEventWithMove.fromSource(
+            oldCamera: camera,
+            camera: camera,
+            hasGesture: true,
+            source: source,
+          );
+
+      expect(eventFor(MapEventSource.doubleTapHold), isA<MapEventMove>());
+    },
+  );
+}


JaffaKetchup:
Hey, thanks for the contribution!

Just to fully document, it's worth noting that this was considered a feature by at least one user (https://github.com/fleaflet/flutter_map/issues/1602#issuecomment-1827251873) and was therefore left in intentionally somewhat (or at least never hunted down or reported).

Therefore, this needs some discussion amongst the team (and community input would be massively appreciated) before merging.

I think it's also worth opening a bug report for this for better tracking - if we decide it's not a bug, we've got a more trackable/referable source of decision making in future. Would you be able to do this, then link this PR to close it?

JaffaKetchup:
I'm going to accept this, as it really was a bug even if we kept it around a while. 

If users prefer the old behaviour, they should be able to set a custom `TileUpdateTransformer` and ignore ongoing zoom events with the `MapEventSource.doubleTapHold`, or debounce them.

At some point we should bring in a feature to prevent tiles loading during a gesture if they are destroyed quickly (although the abortable loading has taken care of some of that problem).

JaffaKetchup (approved):
LGTM! Thanks :)

ProstoSanja:
Thanks @JaffaKetchup , sorry I didn't have time to catch up on the thread during the week. I assume all is clear now? Or is there any followup steps you want me to do?

JaffaKetchup:
It's all good, thanks!

Issue #2199, opened 2026-04-14
LatLng is not finite: LatLng(latitude:NaN, longitude:NaN) from map_interactive_viewer.dart


matisiekpl:
### What is the bug?

Map is stuck and stacktrace is thrown:
```
_Exception: Exception: LatLng is not finite: LatLng(latitude:NaN, longitude:NaN)
#0      Crs.checkLatLng (package:flutter_map/src/geo/crs.dart:82)
#1      Epsg3857.latLngToOffset (package:flutter_map/src/geo/crs.dart:189)
#2      MapCamera.projectAtZoom (package:flutter_map/src/map/camera/camera.dart:278)
#3      MapCamera.offsetToCrs (package:flutter_map/src/map/camera/camera.dart:400)
#4      MapInteractiveViewerState._handleScaleStart (package:flutter_map/src/gestures/map_interactive_viewer.dart:549)
#5      ScaleGestureRecognizer._dispatchOnStartCallbackIfNeeded.<fn> (package:flutter/src/gestures/scale.dart:787)
#6      GestureRecognizer.invokeCallback (package:flutter/src/gestures/recognizer.dart:345)
#7      ScaleGestureRecognizer._dispatchOnStartCallbackIfNeeded (package:flutter/src/gestures/scale.dart:786)
#8      ScaleGestureRecognizer.acceptGesture (package:flutter/src/gestures/scale.dart:809)
#9      _CombiningGestureArenaMember.acceptGesture (package:flutter/src/gestures/team.dart:51)
#10      GestureArenaManager._resolveByDefault (package:flutter/src/gestures/arena.dart:275)
#11      GestureArenaManager._tryToResolveArena.<fn> (package:flutter/src/gestures/arena.dart:255)
#12      _microtaskLoop (dart:async/schedule_microtask.dart:40)
#13      _startMicrotaskLoop (dart:async/schedule_microtask.dart:49)
```

### How can we reproduce it?

```bash
git clone https://github.com/matisiekpl/flutter_map_supercluster
cd flutter_map_supercluster/example
flutter build ios
flutter run ios

# After launching app, go to any map from drawer and try to move in mulitple directions. After few moves, you should get grey screen.
```
Doctor output:
```
[✓] Flutter (Channel stable, 3.41.6, on macOS 26.4.1 25E253 darwin-arm64, locale pl-PL)
```

nisenbeck:
[...]

PR #2220, opened 2026-07-01
fix: prevent NaN fling direction when tracked pointer offsets are zero

### LINKED ISSUE #2221: Gestures with very close start/end points causes division by zero
## What happened

A fling (pan-release) gesture can corrupt the map camera's position with
`NaN`/`Infinity`, which then crashes with:

```
Unhandled error: Unsupported operation: Infinity or NaN toInt
#0      double.toInt (dart:core-patch/double.dart)
#1      double.floor (dart:core-patch/double.dart:201:34)
#2      _floor (package:flutter_map/src/layer/tile_layer/tile_range.dart:36:25)
#3      new DiscreteTileRange.fromPixelBounds (package:flutter_map/src/layer/tile_layer/tile_range.dart:61:9)
#4      TileRangeCalculator.calculate (package:flutter_map/src/layer/tile_layer/tile_range_calculator.dart:31:30)
#5      _TileLayerState._onTileUpdateEvent (package:flutter_map/src/layer/tile_layer/tile_layer.dart:632:51)
```

Once the camera's center is `NaN`, other layers that read `MapCamera.visibleBounds`
also fail, e.g. `flutter_map_marker_cluster` throws:

```
The north latitude can't be bigger than 90.0: NaN
'package:flutter_map/src/geo/latlng_bounds.dart': line 78: 'north <= maxLatitude'
#2      new LatLngBounds.worldSafe (package:flutter_map/src/geo/latlng_bounds.dart:78:16)
#3      MapCamera._computeVisibleBounds (package:flutter_map/src/map/camera/camera.dart:71:25)
```

## Expected vs actual behaviour

- **Expected:** flinging the map, however fast or however the gesture curves,
  never produces a non-finite camera position.
- **Actual:** under the above conditions, the camera's center/zoom become
  `NaN`, and the app crashes on the very next tile-range or visible-bounds
  computation.

## Root cause

[#2158](https://github.com/fleaflet/flutter_map/pull/2158)
(`c5e4909`, "improve fling behaviour when pointer leaves window", released
in 8.3.0) changed `MapInteractiveViewerState._handleScaleEnd` to derive the
fling direction from tracked pointer-position deltas instead of from
`details.velocity` directly:

```dart
final flingOffset = _focalStartLocal - _lastFocalLocal;
final finalSegment = _prevFocalLocal - _lastFocalLocal;
final finalSegmentDistance = finalSegment.distance;

final Offset direction;
if (finalSegmentDistance > 0) {
  direction = finalSegment / finalSegmentDistance;
} else {
  final flingDistance = flingOffset.distance;
  direction = flingOffset / flingDistance;
}
```

The guard above this only checks that `details.velocity.pixelsPerSecond.distance`
(the recognizer's fitted velocity) is above the fling threshold - it does
**not** guarantee that the tracked position deltas (`finalSegment`,
`flingOffset`) are non-zero. Velocity and sampled position deltas are
independently computed by the gesture recognizer, so it's entirely possible
for a fast flick to report a high velocity while the last two tracked pointer
samples coincide *and* the overall drag ends back near its start point. In
that case both `finalSegment` and `flingOffset` are `Offset.zero`, so
`flingOffset / flingDistance` evaluates to `0.0 / 0.0 = NaN`. This `NaN`
direction is fed straight into the fling `Tween`, and every subsequent
animation tick then moves the camera to a `NaN`/`Infinity` position.

I confirmed Flutter's own `VelocityTracker` / `LeastSquaresSolver` cannot be
the source of a `NaN` velocity here - they fall back to `Offset.zero` /
`null` on ill-conditioned samples rather than producing `NaN` - so the bug
is isolated to the division above.

## Affected versions

- Reproduces on **8.3.0** and **8.3.1**.
- Does **not** reproduce on **8.2.2** (confirmed by manual testing on the
  same device) - 8.2.2 derives the fling direction directly from
  `details.velocity.pixelsPerSecond / magnitude`, which is safe because
  `magnitude` is already guarded to be non-zero at that point.

### How can we reproduce it?

A minimal reproduction is available at
https://github.com/nisenbeck/flutter_map_mre, branch `repro/fling-nan-crash`:

```bash
git clone https://github.com/nisenbeck/flutter_map_mre
cd flutter_map_mre
git checkout repro/fling-nan-crash
flutter pub get
flutter run
```

Or in any `FlutterMap` (any `TileLayer`, default `MapOptions`) on a physical
touchscreen device (debug or release):

1. Perform a fast pan/flick gesture where the finger ends close to where it
   started (e.g. a quick back-and-forth "shake" or a fast flick that curves
   back near the start point) - the pattern that reliably triggers it is a
   fling with high recorded *velocity* but where the last couple of tracked
   pointer samples happen to coincide.
2. The app crashes (or, if the exception is caught by a higher-level error
   handler, the map silently jumps to an invalid/blank state).

This only reproduces reliably on a real touchscreen - the exact micro-timing
of pointer samples is very hard to force through synthetic `WidgetTester`
gestures, which is presumably why this hasn't shown up in the existing test
suite.

Screen recording of the reproduction is attached below.

https://github.com/user-attachments/assets/740db2e9-c3f9-4cbe-aada-93c4c7bbaffd

Fixed in [#2220](https://github.com/fleaflet/flutter_map/pull/2220).

### LINKED ISSUE #2227: Unsupported operation: Infinity or NaN toInt
### What is the bug?

Zooming-out throwing `Infinity or NaN toInt`.
Per the stacktrace, the root seems to stem from the `TileLayer` class.

```
E/flutter ( 3750): [ERROR:flutter/runtime/dart_vm_initializer.cc(40)] Unhandled Exception: Unsupported operation: Infinity or NaN toInt
E/flutter ( 3750): #0      double.toInt (dart:core-patch/double.dart)
E/flutter ( 3750): #1      double.floor (dart:core-patch/double.dart:201:34)
E/flutter ( 3750): #2      _floor (package:flutter_map/src/layer/tile_layer/tile_range.dart:36:25)
E/flutter ( 3750): #3      new DiscreteTileRange.fromPixelBounds (package:flutter_map/src/layer/tile_layer/tile_range.dart:61:9)
E/flutter ( 3750): #4      TileRangeCalculator.calculate (package:flutter_map/src/layer/tile_layer/tile_range_calculator.dart:31:30)
E/flutter ( 3750): #5      _TileLayerState._onTileUpdateEvent (package:flutter_map/src/layer/tile_layer/tile_layer.dart:632:51)
E/flutter ( 3750): #6      _RootZone.runUnaryGuarded (dart:async/zone.dart:891:10)
E/flutter ( 3750): #7      _BufferingStreamSubscription._sendData (dart:async/stream_impl.dart:381:11)
E/flutter ( 3750): #8      _BufferingStreamSubscription._add (dart:async/stream_impl.dart:312:7)
E/flutter ( 3750): #9      _SinkTransformerStreamSubscription._add (dart:async/stream_transformers.dart:67:11)
E/flutter ( 3750): #10     _EventSinkWrapper.add (dart:async/stream_transformers.dart:13:11)
E/flutter ( 3750): #11     TileUpdateTransformers.ignoreTapEvents.<anonymous closure> (package:flutter_map/src/layer/tile_layer/tile_update_transformer.dart:54:42)
E/flutter ( 3750): #12     _HandlerEventSink.add (dart:async/stream_transformers.dart:230:17)
E/flutter ( 3750): #13     _SinkTransformerStreamSubscription._handleData (dart:async/stream_transformers.dart:115:24)
E/flutter ( 3750): #14     _RootZone.runUnaryGuarded (dart:async/zone.dart:891:10)
E/flutter ( 3750): #15     _BufferingStreamSubscription._sendData (dart:async/stream_impl.dart:381:11)
E/flutter ( 3750): #16     _BufferingStreamSubscription._add (dart:async/stream_impl.dart:312:7)
E/flutter ( 3750): #17     _ForwardingStreamSubscription._add (dart:async/stream_pipe.dart:154:11)
E/flutter ( 3750): #18     _MapStream._handleData (dart:async/stream_pipe.dart:252:10)
E/flutter ( 3750): #19     _ForwardingStreamSubscription._handleData (dart:async/stream_pipe.dart:184:13)
E/flutter ( 3750): #20     _RootZone.runUnaryGuarded (dart:async/zone.dart:891:10)
E/flutter ( 3750): #21     _BufferingStreamSubscription._sendData (dart:async/stream_impl.dart:381:11)
E/flutter ( 3750): #22     _DelayedData.perform (dart:async/stream_impl.dart:573:14)
E/flutter ( 3750): #23     _PendingEvents.handleNext (dart:async/stream_impl.dart:678:11)
E/flutter ( 3750): #24     _PendingEvents.schedule.<anonymous closure> (dart:async/stream_impl.dart:649:7)
E/flutter ( 3750): #25     _microtaskLoop (dart:async/schedule_microtask.dart:40:35)
E/flutter ( 3750): #26     _startMicrotaskLoop (dart:async/schedule_microtask.dart:49:5)
```

https://github.com/user-attachments/assets/3c20c4dc-363f-425f-bba3-9fd770f22954

### How can we reproduce it?

`flutter_map: 8.3.1`

flutter doctor:
```
[✓] Flutter (Channel stable, 3.44.1, on macOS 26.5.1 25F80 darwin-arm64, locale en-IL) [659ms]
    • Flutter version 3.44.1 on channel stable at /Users/XXX/Developer/flutter
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision 924134a44c (6 weeks ago), 2026-05-29 12:13:22 -0400
    • Engine revision c416acfeb8
    • Dart version 3.12.1
    • DevTools version 2.57.0
    • Feature flags: enable-web, enable-linux-desktop, enable-macos-desktop, enable-windows-desktop, enable-android, enable-ios, cli-animations, enable-native-assets,
      enable-swift-package-manager, omit-legacy-version-file, enable-lldb-debugging, enable-uiscene-migration

[✓] Android toolchain - develop for Android devices (Android SDK version 36.0.0) [3.9s]
    • Android SDK at /Users/XXX/Library/Android/sdk
    • Emulator version 36.6.11.0 (build_id 15507667) (CL:N/A)
    • Platform android-36, build-tools 36.0.0
    • ANDROID_HOME = /Users/XXX/Library/Android/sdk
    • Java binary at: /Applications/Android Studio.app/Contents/jbr/Contents/Home/bin/java
      This JDK is specified in your Flutter configuration.
      To change the current JDK, run: `flutter config --jdk-dir="path/to/jdk"`.
    • Java version OpenJDK Runtime Environment (build 21.0.10+-117844308-b1163.108)
    • All Android licenses accepted.

[✓] Xcode - develop for iOS and macOS (Xcode 26.5) [1,351ms]
    • Xcode at /Applications/Xcode.app/Contents/Developer
    • Build 17F42
    • CocoaPods version 1.16.2

[✓] Chrome - develop for the web [5ms]
    • Chrome at /Applications/Google Chrome.app/Contents/MacOS/Google Chrome

[✓] Connected device (4 available) [6.6s]
    • SM A556E (mobile)                   • XXXXXXX               • android-arm64  • Android 16 (API 36)

[✓] Network resources [1,713ms]
    • All expected network resources are available.

• No issues found!
```

MRE:
```

void main() {
  runApp(const MyApp());
}

class MyApp extends StatelessWidget {
  const MyApp({super.key});

  @override
    return MaterialApp(
      title: 'Flutter Demo',
      theme: ThemeData(colorScheme: .fromSeed(seedColor: Colors.deepPurple)),
      home: const MyHomePage(title: 'Flutter Demo Home Page'),
    );
  }
}

class MyHomePage extends StatefulWidget {
  const MyHomePage({super.key, required this.title});
  final String title;

  @override
  State<MyHomePage> createState() => _MyHomePageState();
}

class _MyHomePageState extends State<MyHomePage> {
  final MapController _flutterMapController = MapController();

  @override
    return Scaffold(
      appBar: AppBar(backgroundColor: Theme.of(context).colorScheme.inversePrimary, title: Text(widget.title)),
      body: FlutterMap(
        mapController: _flutterMapController,
        options: MapOptions(
          interactionOptions: InteractionOptions(flags: InteractiveFlag.all & ~InteractiveFlag.rotate),
        ),
        children: [
          TileLayer(
            urlTemplate:
                'https://services.arcgisonline.com/ArcGIS/rest/services/World_Imagery/MapServer/tile/{z}/{y}/{x}',
            maxNativeZoom: 18,
            tileProvider: NetworkTileProvider(silenceExceptions: true),
          ),
        ],
      ),
    );
  }
}

```

_No response_

### LINKED ISSUE #2230: Zero-length fling generates NaN camera causing white frozen map on Android
### Description

During rapid repeated pinch-zoom gestures on Android, flutter_map can start a fling with a zero-length vector.

In MapInteractiveViewerState._handleScaleEnd():

```dart
direction = flingOffset / flingDistance;
```

If:

```dart
flingDistance == 0.0
```

then:

```
direction = Offset(NaN, NaN)
```

This propagates through the camera:

```
Offset(NaN, NaN)
→ LatLng(NaN, NaN)
→ moveRaw()
→ TileRangeCalculator
→ Unsupported operation: Infinity or NaN toInt
```

The visible result is that the map becomes completely white and stops responding to gestures while the rest of the Flutter application continues working.

### Environment

- flutter_map 8.3.1
- Flutter 3.44.4
- Android 16
- Xiaomi HyperOS
- Impeller (Vulkan)

### Investigation

The issue reproduces even with:

- no PMTiles
- no custom layers
- no overlays
- no Geolocator
- only FlutterMap + TileLayer(OpenStreetMap)

Instrumentation showed the first invalid value is produced during the fling calculation before TileRangeCalculator is reached.

### Proposed fix

Before normalizing the fling vector, verify that the distance is finite and greater than zero.

Example:

```dart
if (!flingDistance.isFinite || flingDistance <= 0) {
  widget.controller.flingNotStarted(eventSource);
  return;
}
```

Also guard against non-finite offsets and directions before calling moveRaw().

This preserves normal fling behaviour while preventing NaN camera states.
if (!flingDistance.isFinite || flingDistance <= 0) {
  widget.controller.flingNotStarted(eventSource);
  return;
}

### How can we reproduce it?

Reproduction steps
Create a new Flutter application.
Add flutter_map 8.3.1.
Display a FlutterMap containing only a standard TileLayer (OpenStreetMap).
Run the app on an Android device (the issue was reproduced on a Xiaomi HyperOS / Android 16 device).
Open the map.
Repeatedly perform very fast pinch-zoom in/out gestures, alternating between zooming in and out without pausing.
Occasionally pan the map between pinch gestures to keep the gesture sequence continuous.
Continue for approximately 20–60 seconds until the gesture ends with an almost zero-length fling vector.
Expected result

The map should continue responding normally.

Actual result

Eventually the map becomes completely white and stops responding to pan and pinch gestures, while the rest of the Flutter application remains responsive.

The exception is:

Unsupported operation: Infinity or NaN toInt

The investigation showed the first invalid value is produced in MapInteractiveViewerState._handleScaleEnd() when:

flingDistance == 0.0

leading to:

direction = flingOffset / flingDistance;

which produces:

Offset(NaN, NaN)
→ LatLng(NaN, NaN)
→ moveRaw()
→ TileRangeCalculator
→ Unsupported operation: Infinity or NaN toInt

_No response_

nisenbeck:
Fixes #2221 - see that issue for the full bug report,
reproduction steps and root-cause analysis.

## Fix

`MapInteractiveViewerState._handleScaleEnd` could divide `flingOffset` by a
zero `flingDistance` (`0.0 / 0.0 = NaN`) when both the final tracked pointer
segment and the overall drag offset happened to have zero length, even
though the gesture's reported velocity was above the fling threshold.

Extracted the direction calculation into a small, pure, unit-testable
`flingDirection` helper, and added a final fallback to the (already
guaranteed non-zero) velocity direction instead of dividing by zero:

```dart
if (finalSegmentDistance > 0) return finalSegment / finalSegmentDistance;
if (flingDistance > 0) return flingOffset / flingDistance;
return velocityDirection;
```

This preserves the existing behaviour (prefer the final tracked segment,
then the overall drag direction) and only changes behaviour for the
previously-undefined zero/zero edge case, where it now falls back to the
same velocity-based direction that flutter_map used prior to #2158 (8.2.2
and earlier).

`velocityDirection` is guaranteed finite: `magnitude` is already checked to
be `>= _kMinFlingVelocity` before this is called, and Flutter's own
`VelocityTracker` / `LeastSquaresSolver` never return `NaN` - they fall back
to `Offset.zero` / `null` on ill-conditioned samples - so this can never
itself be a source of `NaN`.

## Testing

- Added `test/gestures/map_interactive_viewer_test.dart` covering all three
  branches of `flingDirection`, including a regression test for the
  zero/zero case.
- Full existing test suite passes locally with no regressions.
- Manually verified on a physical Android device: confirmed the crash on
  8.3.0/8.3.1, confirmed it does *not* occur on 8.2.2, and confirmed this
  fix resolves it while pointing a reproduction app at this branch via a
  `dependency_overrides: path:` entry.

## AI disclosure

Per `CONTRIBUTING.md`: I used AI assistance (Claude Code) to help diagnose
the root cause and draft this fix and its tests. I've reviewed the change,
verified it against the project's lint rules and test suite, and manually
confirmed the fix resolves the crash on a physical device.

### DIFF
--- a/lib/src/gestures/map_interactive_viewer.dart
+++ b/lib/src/gestures/map_interactive_viewer.dart
@@ -802,17 +802,14 @@ class MapInteractiveViewerState extends State<MapInteractiveViewer>
     // gestures where the user changes direction during the drag.
     final flingOffset = _focalStartLocal - _lastFocalLocal;
     final finalSegment = _prevFocalLocal - _lastFocalLocal;
-    final finalSegmentDistance = finalSegment.distance;
 
-    // Use final segment direction if available, otherwise fall back to overall
-    // direction for edge cases where the final segment has no movement.
-    final Offset direction;
-    if (finalSegmentDistance > 0) {
-      direction = finalSegment / finalSegmentDistance;
-    } else {
-      final flingDistance = flingOffset.distance;
-      direction = flingOffset / flingDistance;
-    }
+    final direction = flingDirection(
+      finalSegment: finalSegment,
+      flingOffset: flingOffset,
+      // `magnitude` is checked to be non-zero above, so this is always a
+      // finite, non-zero-length direction and is a safe final fallback.
+      velocityDirection: details.velocity.pixelsPerSecond / magnitude,
+    );
     final distance = (Offset.zero & _camera.nonRotatedSize).shortestSide;
 
     _flingAnimation = Tween<Offset>(
@@ -829,6 +826,31 @@ class MapInteractiveViewerState extends State<MapInteractiveViewer>
             stiffness: 1000,
             ratio: _interactionOptions.flingAnimationDampingRatio,
           ));
+  }
+
+  /// Calculates the direction a fling gesture should continue in.
+  ///
+  /// Prefers the direction of the final tracked pointer segment, falling
+  /// back to the overall drag direction. If both [finalSegment] and
+  /// [flingOffset] have zero length - which can happen even when the
+  /// gesture recognizer reports a velocity above the fling threshold, since
+  /// the velocity is fitted over multiple recent samples and is not
+  /// necessarily proportional to the last tracked position deltas - falls
+  /// back to [velocityDirection] to avoid a division by zero (which would
+  /// otherwise produce a `NaN` direction and corrupt the camera position).
+  @visibleForTesting
+  static Offset flingDirection({
+    required Offset finalSegment,
+    required Offset flingOffset,
+    required Offset velocityDirection,
+  }) {
+    final finalSegmentDistance = finalSegment.distance;
+    if (finalSegmentDistance > 0) return finalSegment / finalSegmentDistance;
+
+    final flingDistance = flingOffset.distance;
+    if (flingDistance > 0) return flingOffset / flingDistance;
+
+    return velocityDirection;
   }
 
   void _handleTap(TapPosition position) {
--- a/test/gestures/map_interactive_viewer_test.dart
+++ b/test/gestures/map_interactive_viewer_test.dart
@@ -1,0 +1,49 @@
+import 'package:flutter_map/src/gestures/map_interactive_viewer.dart';
+import 'package:flutter_test/flutter_test.dart';
+
+void main() {
+  group('MapInteractiveViewerState.flingDirection', () {
+    test('uses the final segment direction when it has non-zero length', () {
+      final direction = MapInteractiveViewerState.flingDirection(
+        finalSegment: const Offset(10, 0),
+        flingOffset: const Offset(100, 0),
+        velocityDirection: const Offset(0, 1),
+      );
+
+      expect(direction, const Offset(1, 0));
+    });
+
+    test(
+      'falls back to the overall drag direction when the final segment has '
+      'zero length',
+      () {
+        final direction = MapInteractiveViewerState.flingDirection(
+          finalSegment: Offset.zero,
+          flingOffset: const Offset(0, -50),
+          velocityDirection: const Offset(1, 0),
+        );
+
+        expect(direction, const Offset(0, -1));
+      },
+    );
+
+    test(
+      'falls back to the velocity direction instead of dividing by zero '
+      'when both the final segment and the overall drag offset have zero '
+      'length (regression test: this previously produced a NaN direction, '
+      'which corrupted the camera position - '
+      'https://github.com/fleaflet/flutter_map/issues/2199)',
+      () {
+        final direction = MapInteractiveViewerState.flingDirection(
+          finalSegment: Offset.zero,
+          flingOffset: Offset.zero,
+          velocityDirection: const Offset(0.6, 0.8),
+        );
+
+        expect(direction, const Offset(0.6, 0.8));
+        expect(direction.dx.isFinite, isTrue);
+        expect(direction.dy.isFinite, isTrue);
+      },
+    );
+  });
+}


JaffaKetchup:
Hey, thanks for this. Please open a bug report issue, and move your details of the issue there - this helps with tracking. The PR should only detail the fix. It would also be good to get a screen recording of you reproducing this issue.

nisenbeck:
Thanks for the quick feedback! I've split this up as requested:

- Bug report with full reproduction steps, root-cause analysis and a screen recording: #2221
- This PR description now only covers the fix itself

Let me know if anything else is needed.

JaffaKetchup (approved):
LGTM!

Really not sure how you ran into this bug in practise since it was quite difficult to reproduce, but I guess it is the sort of thing that seems to just randomly happen. I also would expect the offset never to be zero since I would've expected that to require more precision than is available, but I guess it happens.

From what I can tell, this fixes the bug. Thanks :)

nisenbeck:
Thanks for merging! Found it while checking whether #2199 ("LatLng is not finite") was actually resolved on 8.3.1 - was just doing normal pan/fling testing on the map and hit this and a separate issue (#2222), unrelated to #2199 itself.

PR #2223, opened 2026-07-01
fix: don't apply `initialCameraFit` on a zero-sized camera

### LINKED ISSUE #2222: initialCameraFit can silently apply against a zero-sized parent and never retry with the correct size
### What is the bug?

`initialCameraFit` is meant to fit the camera to a given `LatLngBounds` once, when the map first appears. If the map is nested under a parent that doesn't have its final size on the very first layout pass (e.g. an `Expanded`/`Flexible` inside a not-yet-measured parent, an `AnimatedSize`, a lazily-revealed tab, ...), the fit can be computed against a `Size(0, 0)` camera instead of the real, final size. The result is a degenerate (fully zoomed-out, whole-world) camera position - and because the "initial fit applied" flag is set regardless, it is **never retried** once the real size becomes available a moment later. The map keeps showing the wrong view for as long as that `FlutterMap` widget stays mounted, until the user manually pans or zooms (the camera itself isn't frozen - only the automatic one-time fit never re-runs).

This doesn't crash the app, but it silently defeats the one thing `initialCameraFit` is meant to do, with no error or log to point at the cause - I'd classify this as a moderate-severity, silent UX bug rather than critical.

## Root cause

`_FlutterMapStateContainer._parentConstraintsAreSet` (`lib/src/map/widget.dart`):

```dart
/// During Flutter startup the native platform resolution is not immediately
/// available which can cause constraints to be zero before they are updated
/// in a subsequent build to the actual constraints. This check allows us to
/// differentiate zero constraints caused by missing platform resolution vs
/// zero constraints which were actually provided by the parent widget.
bool _parentConstraintsAreSet(
        BuildContext context, BoxConstraints constraints) =>
    constraints.maxWidth != 0 || MediaQuery.sizeOf(context) != Size.zero;
```

This is meant to distinguish "constraints are zero because the platform hasn't resolved yet" from "constraints are zero because the parent widget actually gave zero space". In practice, `MediaQuery.sizeOf(context)` is almost always already non-zero (the device's screen size is known essentially immediately), even while `constraints.maxWidth` is still `0` for one frame because *this specific widget's* layout hasn't settled yet - the two are independent signals, and the `||` makes the check pass in exactly the case it's supposed to catch.

`_applyInitialCameraFit` then runs with `constraints` still zero-sized:

```dart
void _applyInitialCameraFit(BoxConstraints constraints) {
  if (!_initialCameraFitApplied &&
      widget.options.initialCameraFit != null &&
      _parentConstraintsAreSet(context, constraints)) {
    _initialCameraFitApplied = true;

    _mapController.fitCamera(widget.options.initialCameraFit!);
  }
}
```

`_initialCameraFitApplied` is set to `true` unconditionally once this runs, so even though `_updateAndEmitSizeIfConstraintsChanged` will be called again with the real, non-zero size a moment later, the fit is never recomputed.

### How can we reproduce it?

A minimal reproduction is available at https://github.com/nisenbeck/flutter_map_mre, branch `repro/zero-size-initial-fit`:

```bash
git clone https://github.com/nisenbeck/flutter_map_mre
cd flutter_map_mre
git checkout repro/zero-size-initial-fit
flutter pub get
flutter run
```

It wraps `FlutterMap` in a small widget that reports `Size.zero` for exactly one frame before resizing to fill the screen - mimicking the real-world layout patterns mentioned above.

- **Expected:** the map opens framed on the configured bounds (roughly Germany/Switzerland/Austria).
- **Actual (when the bug reproduces):** the map opens zoomed all the way out to the whole world, and stays that way until manual interaction.

This is timing-sensitive - if it doesn't reproduce on the very first launch, a few hot restarts (`R` in the `flutter run` terminal) reliably trigger it.

I also have a deterministic widget test (no device/timing dependency) demonstrating the same thing directly against `FlutterMap`, which I'll include with the fix PR.

Fixed in [#2223](https://github.com/fleaflet/flutter_map/pull/2223).

nisenbeck:
Fixes #2222 - see that issue for the full bug report, reproduction steps and root-cause analysis.

## Fix

`_applyInitialCameraFit` could apply `initialCameraFit` against a zero-sized camera (`constraints.maxWidth`/`maxHeight == 0`) and mark it as permanently applied, because `_parentConstraintsAreSet` can return `true` in exactly that situation (its `MediaQuery`-based fallback is satisfied almost immediately, independently of whether *this* widget's own layout has actually settled).

Added an explicit, additional guard so the fit is only applied - and only latched as applied - once the constraints are genuinely non-zero in both dimensions:

```dart
if (!_initialCameraFitApplied &&
    widget.options.initialCameraFit != null &&
    _parentConstraintsAreSet(context, constraints) &&
    constraints.maxWidth > 0 &&
    constraints.maxHeight > 0) {
  _initialCameraFitApplied = true;
  _mapController.fitCamera(widget.options.initialCameraFit!);
}
```

I deliberately left `_parentConstraintsAreSet` itself untouched, rather than rewriting its heuristic - there's no history explaining why the `MediaQuery` fallback was added, so I didn't want to risk changing behaviour I can't fully verify. This fix instead directly prevents the harmful outcome (latching against a degenerate size) regardless of what `_parentConstraintsAreSet` decides. If the real constraints never arrive (the widget is genuinely given zero space forever), the fit simply never applies, which is a strictly better outcome than silently locking in a degenerate zoom.

## Testing

Added `test/map/zero_size_camera_fit_test.dart` with two cases (there was no existing test coverage for `initialCameraFit` at all):

- A real `FlutterMap` mounted inside a widget that reports `Size.zero` for one frame before revealing its real size: asserts the resulting `camera.zoom` matches a direct fit against the real size (previously it settled on `0.0`, the degenerate zoomed-out result).
- The same fit with the real size available from the very first frame (no race), to confirm the already-working case is unaffected by this change.

Full existing test suite passes locally with no regressions.

Manually verified on a physical Android device using the reproduction linked in the issue, pointed at this branch via a `dependency_overrides: path:` entry.

## AI disclosure

Per `CONTRIBUTING.md`: I used AI assistance (Claude Code) to help diagnose the root cause and draft this fix and its tests. I've reviewed the change, verified it against the project's lint rules and test suite, and manually confirmed the fix resolves the issue on a physical device.

### DIFF
--- a/lib/src/map/widget.dart
+++ b/lib/src/map/widget.dart
@@ -119,7 +119,15 @@ class _FlutterMapStateContainer extends State<FlutterMap>
 
     if (!_initialCameraFitApplied &&
         widget.options.initialCameraFit != null &&
-        _parentConstraintsAreSet(context, constraints)) {
+        _parentConstraintsAreSet(context, constraints) &&
+        // `_parentConstraintsAreSet` can return true while `constraints` is
+        // still zero-sized (e.g. on the first build of a not-yet-measured
+        // parent, while `MediaQuery` already reports a non-zero screen
+        // size). Fitting against a zero size produces a degenerate camera,
+        // so wait for a real, non-zero size instead of latching
+        // `_initialCameraFitApplied` against it.
+        constraints.maxWidth > 0 &&
+        constraints.maxHeight > 0) {
       _initialCameraFitApplied = true;
 
       _mapController.fitCamera(widget.options.initialCameraFit!);
--- a/test/map/zero_size_camera_fit_test.dart
+++ b/test/map/zero_size_camera_fit_test.dart
@@ -1,0 +1,167 @@
+import 'package:flutter/material.dart';
+import 'package:flutter_map/flutter_map.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:latlong2/latlong.dart';
+
+/// Wraps [child] in a box that reports [Size.zero] to the [LayoutBuilder]
+/// constraints for the first build, then switches to [realSize] once
+/// `reveal()` is called - while [MediaQuery] (from the ambient test binding)
+/// stays at its real, non-zero size throughout.
+///
+/// This reproduces the situation described in
+/// `_FlutterMapStateContainer._parentConstraintsAreSet`'s doc comment:
+/// zero `LayoutBuilder` constraints while the platform/`MediaQuery` size is
+/// already known - which can happen any time a `FlutterMap` is nested under
+/// a widget that doesn't have its final size on the first layout pass (e.g.
+/// `Expanded`/`Flexible` inside a not-yet-measured parent, `AnimatedSize`,
+/// a lazily-revealed tab, etc.), not just during platform startup.
+class _DelayedSizeBox extends StatefulWidget {
+  const _DelayedSizeBox({
+    super.key,
+    required this.realSize,
+    required this.child,
+  });
+
+  final Size realSize;
+  final Widget child;
+
+  @override
+  State<_DelayedSizeBox> createState() => _DelayedSizeBoxState();
+}
+
+class _DelayedSizeBoxState extends State<_DelayedSizeBox> {
+  bool _revealed = false;
+
+  void reveal() => setState(() => _revealed = true);
+
+  @override
+  Widget build(BuildContext context) {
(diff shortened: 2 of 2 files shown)


JaffaKetchup (approved):
LGTM! Thanks again :)

PR #2226, opened 2026-07-03
fix: use smoothed velocity direction for touch flings to prevent off-axis glides

### LINKED ISSUE #2225: Fling direction is derived from the noisy last pointer segment on touch: fast swipes glide off-axis / zigzag since 8.3.0
### What is the bug?

Since v8.3.0, fast swipes on **touch devices** launch the fling/momentum glide visibly off-axis: a fast, near-vertical flick coasts diagonally, and repeatedly flicking in the same direction makes the map zigzag left and right. Slow drags are unaffected (the finger-down pan itself tracks correctly) — only the direction of the post-release glide is wrong.

Reproduced on flutter_map 8.3.0 and 8.3.1 (and current master), Android and iOS physical devices. 8.2.2 behaves correctly.

**Cause:** #2158 (released in 8.3.0) changed the fling direction source in `_handleScaleEnd` from the gesture recognizer's velocity to the **single last tracked pointer segment**:

```dart
// 8.2.2
final direction = details.velocity.pixelsPerSecond / magnitude;
...
end: flingOffset - direction * distance,

// 8.3.0+ (master: flingDirection(), which prefers finalSegment)
final finalSegment = _prevFocalLocal - _lastFocalLocal;
direction = finalSegment / finalSegmentDistance;
...
end: flingOffset + direction * distance,
```

`details.velocity` comes from Flutter's `VelocityTracker` — a least-squares fit over the last ~100 ms of pointer samples — so its direction is stable. `_prevFocalLocal - _lastFocalLocal` is one single frame-to-frame delta: at fast swipe speeds this segment is dominated by touch-sampling noise and the small thumb-roll at lift-off, so its direction can deviate wildly from the actual swipe axis (routinely by 30–90° in our testing).

#2158 solved a real problem — on web/desktop the pointer can leave the window mid-drag, corrupting the velocity estimate — but that failure mode cannot occur with touch input, while the cost (noisy single-segment direction) hits touch hardest, since that's where fast flicks are a primary navigation gesture.

**A related detail — the `velocityDirection` fallback from #2220 points backwards:** the three candidate vectors in `flingDirection()` don't share a sign convention. `finalSegment` and `flingOffset` run *opposite* to the pointer motion (they're used as `end: flingOffset + direction * distance`), but the fallback passes `details.velocity.pixelsPerSecond / magnitude`, which points *along* the pointer motion (8.2.2 used it with `- direction * distance`). So in the degenerate zero/zero case the glide coasts against the swipe. Rarely visible (both tracked offsets must be zero-length), but worth fixing in the same place.

### How can we reproduce it?

1. Run any `FlutterMap` with `InteractiveFlag.flingAnimation` enabled on a physical phone (or `flutter create` + the flutter_map README example, unchanged).
2. Flick the map quickly upwards several times in a row, as you would to travel a long distance.
3. Watch the glide direction: instead of continuing along the swipe axis, each glide veers to a different side depending on sub-frame noise in the last pointer segment.

```dart
FlutterMap(
  options: MapOptions(
    initialCenter: const LatLng(52.0, 9.0),
    initialZoom: 13,
    // default flags include flingAnimation
  ),
  children: [
    TileLayer(
    ),
  ],
)
```

Yes — keep #2158's tracked-segment approach where it's needed (mouse/trackpad, where the pointer can leave the window), and use the smoothed velocity direction for `PointerDeviceKind.touch`; also negate the velocity fallback so all three candidates share the same sign convention. PR incoming: it extends `flingDirection()` with a `preferVelocityDirection` flag set from the last pointer-down's device kind, plus tests.

(We currently ship with `flingAnimation` disabled as a stopgap, which loses momentum panning entirely — hence the report.)

JonasGrunau:
Fixes #2225.

## What

On touch input, derive the fling direction from the gesture recognizer's velocity (Flutter's `VelocityTracker`, a least-squares fit over the last ~100 ms of samples) instead of the single last tracked pointer segment. Mouse/trackpad input keeps the tracked-segment behaviour introduced in #2158, which handles the pointer leaving the window mid-drag — a case that cannot occur with touch.

Also negates the `velocityDirection` passed into `flingDirection()` as the division-by-zero fallback (#2220), so all three direction candidates share the tracked-offset sign convention (`end: flingOffset + direction * distance`). Previously the degenerate zero/zero case glided *against* the swipe, since `details.velocity.pixelsPerSecond` points along the pointer motion while `finalSegment`/`flingOffset` run opposite to it (8.2.2 used the velocity with `- direction * distance`).

## Why

The last tracked segment is one frame-to-frame delta. At fast swipe speeds it is dominated by touch-sampling noise and lift-off thumb-roll, so its direction routinely deviates 30–90° from the swipe axis: fast flicks glide off-axis and repeated flicks visibly zigzag (see #2225 for the full analysis). The fitted velocity direction is stable — it's what the pre-8.3.0 fling used and what platform map apps use.

## How

- `MapInteractiveViewerState` records the device kind of the last pointer-down.
- `flingDirection()` gains a `preferVelocityDirection` parameter; when set (touch), the already-normalized `velocityDirection` is returned directly, otherwise the existing preference order (final segment → overall drag offset → velocity fallback) is unchanged.
- Existing `flingDirection` tests updated for the new parameter; new test covers the touch preference.

All existing tests pass (`flutter test`), `dart analyze` clean, `dart format` clean.

### DIFF
--- a/lib/src/gestures/map_interactive_viewer.dart
+++ b/lib/src/gestures/map_interactive_viewer.dart
@@ -64,6 +64,7 @@ class MapInteractiveViewerState extends State<MapInteractiveViewer>
   bool _dragMode = false;
   int _gestureWinner = MultiFingerGesture.none;
   int _pointerCounter = 0;
+  PointerDeviceKind? _lastPointerKind;
   bool _isListeningForInterruptions = false;
 
   var _rotationStarted = false;
@@ -372,6 +373,7 @@ class MapInteractiveViewerState extends State<MapInteractiveViewer>
 
   void _onPointerDown(PointerDownEvent event) {
     ++_pointerCounter;
+    _lastPointerKind = event.kind;
 
     if (_ckrTriggered.value) {
       _ckrInitialDegrees = _camera.rotation;
@@ -808,7 +810,16 @@ class MapInteractiveViewerState extends State<MapInteractiveViewer>
       flingOffset: flingOffset,
       // `magnitude` is checked to be non-zero above, so this is always a
       // finite, non-zero-length direction and is a safe final fallback.
-      velocityDirection: details.velocity.pixelsPerSecond / magnitude,
+      // Negated so that it shares the convention of the tracked offsets,
+      // which run opposite to the pointer motion.
+      velocityDirection: -details.velocity.pixelsPerSecond / magnitude,
+      // On touch devices the pointer cannot leave the window mid-drag (the
+      // case the tracked offsets handle better than the velocity, see
+      // #2158), and the velocity - fitted over multiple recent samples - is
+      // far more stable than the final tracked segment: deriving the
+      // direction from the single last, noisy pointer segment sends fast
+      // swipes off-axis, so repeated flicks visibly zigzag.
+      preferVelocityDirection: _lastPointerKind == PointerDeviceKind.touch,
     );
     final distance = (Offset.zero & _camera.nonRotatedSize).shortestSide;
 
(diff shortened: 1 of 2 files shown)


JaffaKetchup:
> On touch devices the pointer cannot leave the window mid-drag

Is this true? What about things like touchscreen laptops? Unfortunately I don't have one to test with.

Issue #1759, opened 2023-12-05
[BUG] Unsupported operation: Infinity or NaN toInt


RicardoRB:
### What is the bug?

When I'm launching my app in a --release mode, this exception is happening

```dart
I/flutter (25342): Unsupported operation: Infinity or NaN toInt
I/flutter (25342): #0      double.floor (dart:core-patch/double.dart)
I/flutter (25342): #1      PointExtension.floor (package:flutter_map/src/misc/point_extensions.dart:38)
I/flutter (25342): #2      MapCamera.pixelBoundsAtZoom (package:flutter_map/src/map/camera/camera.dart:262)
I/flutter (25342): #3      MapCamera.pixelBounds (package:flutter_map/src/map/camera/camera.dart:253)
I/flutter (25342): #4      MapCamera.visibleBounds (package:flutter_map/src/map/camera/camera.dart:67)
I/flutter (25342): #5      FlutterMapInternalController.move (package:flutter_map/src/map/controller/internal.dart:104)
I/flutter (25342): #6      FlutterMapInternalController.fitCamera (package:flutter_map/src/map/controller/internal.dart:245)
I/flutter (25342): #7      _FlutterMapStateContainer._applyInitialCameraFit (package:flutter_map/src/map/widget.dart:215)
I/flutter (25342): #8      _FlutterMapStateContainer._updateAndEmitSizeIfConstraintsChanged.<anonymous closure> (package:flutter_map/src/map/widget.dart:242)
I/flutter (25342): #9      SchedulerBinding._invokeFrameCallback (package:flutter/src/scheduler/binding.dart:1325)
I/flutter (25342): #10     SchedulerBinding.handleDrawFrame (package:flutter/src/scheduler/binding.dart:1264)
I/flutter (25342): #11     SchedulerBinding._handleDrawFrame (package:flutter/src/scheduler/binding.dart:1113)
I/flutter (25342): #12     _invoke (dart:ui/hooks.dart:312)
I/flutter (25342): #13     PlatformDispatcher._drawFrame (dart:ui/platform_dispatcher.dart:383)
I/flutter (25342): #14     _drawFrame (dart:ui/hooks.dart:283)
E/flutter (25342): [ERROR:flutter/runtime/dart_vm_initializer.cc(41)] Unhandled Exception: Unsupported operation: Infinity or NaN toInt
[...]

Issue #1370, opened 2022-09-20
[BUG] fitting bounds throws Infinity or NaN toInt or does not fit map correctly


typexy:
### What is the bug?

when fitbounds is called, we have to distinguish 2 cases:

1. multiple points are given -> zoom level is not calculated correctly, it shows whole map
2. one point is given -> Unsupported operation: Infinity or NaN toInt -> is thrown at following function:

/// Create new [CustomPoint] whose [x] and [y] values are rounded down
  /// to int
  CustomPoint<T> floor() {
    return CustomPoint<T>(x.floor(), y.floor());
  }

### What is the expected behaviour?

Expected behavior is to show a map centered at the middle of the given LatLng points and to not throw an error.

### How can we reproduce this issue?

```dart
class Maps extends StatefulWidget {
  final LatLng? initMarker;
  final List<LatLng> markers;

  Maps({Key? key, this.initMarker, required this.markers}) : super(key: key) {
    var multipleTestMarkrs = [
      LatLng(52.527194793354084, 13.371265245125935),
      LatLng(48.14384446905119, 11.57992496759276)
    ];
    var singleTestMarker = [LatLng(52.527194793354084, 13.371265245125935)];
    this.markers = singleTestMarker;
  }

  @override
  _MapsState createState() => _MapsState();
}

class _MapsState extends State<Maps> {
  bool hasMarkers() {
    return widget.markers.isNotEmpty;
  }

  @override
  Widget build(BuildContext context) {
    final options = MapOptions(
        interactiveFlags: InteractiveFlag.none,
        bounds: LatLngBounds.fromPoints(widget.markers),
        center: computeCentroid(widget.markers),
        maxBounds: LatLngBounds.fromPoints(widget.markers),
        screenSize: MediaQuery.of(context).size,
        boundsOptions: FitBoundsOptions(padding: EdgeInsets.all(20)));

    return hasMarkers()
        ? Container(
            height: 300,
            child: IgnorePointer(
              child: FlutterMap(
                options: options,
                children: [
                  TileLayer(
                    retinaMode:
                        true && MediaQuery.of(context).devicePixelRatio > 1.0,
                    urlTemplate:
                        "https://cartodb-basemaps-{s}.global.ssl.fastly.net/light_all/{z}/{x}/{y}.png",
                    subdomains: ['a', 'b', 'c'],
                  ),
                  MarkerLayer(
                    markers: [
                      for (var marker in widget.markers)
                        Marker(
[...]

Issue #524, opened 2020-02-20
unsupported operation infinity or nan toint


riccardopirani:
Hi I'm trying to implement the flutter_maps in my flutter application when I run the app but I have this error. what is it due to and how can i solve it?

**Flutter Code:**
```
import 'package:flutter_map/flutter_map.dart';
import 'package:Myapp/View/Cantieri/mediterranesn_diet_view.dart'
    show MediterranesnDietView;
import 'package:Myapp/View/Cantieri/title_view.dart' show TitleView;
import 'package:Myapp/View/Home_theme.dart';
import 'package:Myapp/View/MarcaTempo/meals_list_view.dart' show MealsListView;
import 'package:flutter/material.dart';
import 'package:latlong/latlong.dart';

class MyDiaryScreen extends StatefulWidget {
  const MyDiaryScreen({Key key, this.animationController}) : super(key: key);

  final AnimationController animationController;
  @override
  _MyDiaryScreenState createState() => _MyDiaryScreenState();
}

class _MyDiaryScreenState extends State<MyDiaryScreen>
    with TickerProviderStateMixin {
  Animation<double> topBarAnimation;

  List<Widget> listViews = <Widget>[];
  final ScrollController scrollController = ScrollController();
  double topBarOpacity = 0.0;

  @override
  void initState() {
    topBarAnimation = Tween<double>(begin: 0.0, end: 1.0).animate(
        CurvedAnimation(
            parent: widget.animationController,
            curve: Interval(0, 0.5, curve: Curves.fastOutSlowIn)));
    addAllListData();

    scrollController.addListener(() {
      if (scrollController.offset >= 24) {
        if (topBarOpacity != 1.0) {
          setState(() {
            topBarOpacity = 1.0;
          });
        }
      } else if (scrollController.offset <= 24 &&
          scrollController.offset >= 0) {
        if (topBarOpacity != scrollController.offset / 24) {
          setState(() {
            topBarOpacity = scrollController.offset / 24;
          });
        }
      } else if (scrollController.offset <= 0) {
        if (topBarOpacity != 0.0) {
          setState(() {
            topBarOpacity = 0.0;
          });
        }
      }
    });
    super.initState();
  }

  void addAllListData() {
    const int count = 9;

    //Prima view che comprende il titolo della Posizione
    listViews.add(
      TitleView(
        titleTxt: 'Posizione',
        subTxt: '',
[...]

Issue #409, opened 2019-09-02
Unsupported operation: Infinity or NaN toInt


viplike:
**widget code:**
```dart
return Scaffold(
        appBar: AppBar(
          title: Text('Test'),
        ),
        drawer: GuestDrawer(),
        body: Column(children: <Widget>[
              Container(
                alignment: Alignment.centerLeft,
                child: FlutterMap(
                  options: MapOptions(
                    center: LatLng(51.5, -0.09),
                    zoom: 13.0,
                  ),
                  layers: [
                    TileLayerOptions(
                      urlTemplate: "https://api.tiles.mapbox.com/v4/"
                          "{id}/{z}/{x}/{y}@2x.png?access_token={accessToken}",
                      additionalOptions: {
                        'accessToken':
                            '<YOUR_KEY>',
                        'id': 'mapbox.streets',
                      },
                    ),
                    MarkerLayerOptions(
                      markers: [
                        Marker(
                          width: 80.0,
                          height: 80.0,
                          point: LatLng(51.5, -0.09),
                          builder: (ctx) => Container(
                            child: Image.asset(
                                'assets/img/leaflet/marker_blue.png'),
                          ),
                        ),
                      ],
                    ),
                  ],
                ),
                padding: EdgeInsets.all(10),
              )
            ]));
  }
```

**error stack:** 
```
I/flutter (16858): ══╡ EXCEPTION CAUGHT BY WIDGETS LIBRARY ╞═══════════════════════════════════════════════════════════
I/flutter (16858): The following UnsupportedError was thrown building LayoutBuilder:
I/flutter (16858): Unsupported operation: Infinity or NaN toInt
I/flutter (16858): When the exception was thrown, this was the stack:
I/flutter (16858): #0      double.toInt (dart:core-patch/double.dart:183:36)
I/flutter (16858): #1      double.round (dart:core-patch/double.dart:151:34)
D/WebView (16858): WebView<init>
I/flutter (16858): #2      CustomPoint.round (package:flutter_map/src/core/point.dart:43:39)
I/flutter (16858): #3      MapState.getNewPixelOrigin (package:flutter_map/src/map/map.dart:269:47)
I/flutter (16858): #4      MapState.move (package:flutter_map/src/map/map.dart:124:20)
I/flutter (16858): #5      MapState._init (package:flutter_map/src/map/map.dart:104:5)
I/flutter (16858): #6      MapState.size= (package:flutter_map/src/map/map.dart:90:7)
[...]

PR #2228, opened 2026-07-09
fix: sanitise non-finite zoom in MapCamera.clampZoom

Pistacchione:
## What & why

During a rapid pinch zoom-out, the gesture handler calls `_getZoomForScale`,
which computes `startZoom + math.log(scale) / math.ln2`. When
`details.scale + _scaleCorrector` momentarily becomes non-positive (the scale
corrector can be negative), `math.log` returns `NaN` or `-Infinity`.

`clampZoom` relied on `num.clamp`, which returns `NaN` unchanged for a `NaN`
input and therefore cannot sanitise it. The invalid zoom was committed to the
camera and propagated into:

- **`TileRangeCalculator`** — `halfSize = size / (scale * 2)` becomes non-finite
  and `DiscreteTileRange.fromPixelBounds` throws
  `Unsupported operation: Infinity or NaN toInt` from `.floor()`.
- **`MarkerClusterLayer`** — the visible bounds become `NaN`, tripping the
  `north <= maxLatitude` assertion.

The crash is transient/self-healing (the next finite scale update recomputes a
valid zoom), but each bad frame surfaces as an uncaught framework error.

## Fix

`clampZoom` now treats any non-finite input as an invalid update and returns the
current (finite) zoom, keeping the camera in a valid state. Being the single
chokepoint every zoom mutation passes through (gestures and `moveRaw`), this
covers the whole class of NaN-zoom bugs. It also fixes the `minZoom == null`
case, where a `-Infinity` zoom crashed as well.

## Tests

Added `test/map/camera/camera_test.dart` covering the non-finite guard
(`NaN`/`±Infinity`), normal min/max clamping, and the no-min/max case. The
existing `camera/` tests still pass.

## Notes

- Related issues: #1759, #1370, #409, #524.

## AI disclosure

This change was written with the assistance of AI (Claude).

### DIFF
--- a/lib/src/map/camera/camera.dart
+++ b/lib/src/map/camera/camera.dart
@@ -388,10 +388,22 @@ class MapCamera {
 
   /// Clamps the provided [zoom] to the range specified by [minZoom] and
   /// [maxZoom], if set.
-  double clampZoom(double zoom) => zoom.clamp(
-        minZoom ?? double.negativeInfinity,
-        maxZoom ?? double.infinity,
-      );
+  ///
+  /// A non-finite [zoom] (`NaN`/`Infinity`) is treated as an invalid update and
+  /// the current [zoom] is returned unchanged. Without this guard such a value
+  /// would flow into the camera and crash the tile range and marker cluster
+  /// layers — `num.clamp` returns `NaN` for a `NaN` input, so it cannot sanitise
+  /// it. This happens in practice during a rapid pinch zoom-out, where
+  /// `math.log` is fed a non-positive scale and produces a non-finite zoom.
+  double clampZoom(double zoom) {
+    if (!zoom.isFinite) {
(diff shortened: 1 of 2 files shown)


JaffaKetchup:
Please open a bug report with an MRE. Ensure you test to see whether your original issue is resolved on the latest master.

As is, this PR does not fix anything and instead just buries the problem, if there is one.

Pistacchione:
Hi @JaffaKetchup

You were right, thanks for the pointer to test against master.
That's already fixed on master by #2220 (issue #2199) via the velocityDirection fallback in flingDirection, and I've confirmed the crash no longer occurs when pinning to master, so this was never a clampZoom problem.
My PR was masking an already-solved issue, as you said.

Closing the loop here; the PR can stay closed. Thanks again.

PR #2236, opened 2026-07-30
fix: exclude the map's gesture detector from semantics

### LINKED ISSUE #2176: ensureSemantics() breaks onTap callback on Web platform
### What is the bug?

For enabling advanced accessibility for Web Screen Readers Flutter offers to set:

`SemanticsBinding.instance.ensureSemantics();`

See: [https://blog.flutter.dev/accessibility-in-flutter-on-the-web-51bfc558b7d3](https://blog.flutter.dev/accessibility-in-flutter-on-the-web-51bfc558b7d3)

When enabled however, the onTap callback of FlutterMap is not handling the correct coordinates of the tap. Rather the displayed map center.

However LongTap is not affected.

I created a small demo project for the bug: [https://gitlab.com/kybernetik/flutter-web-map-tap-test](https://gitlab.com/kybernetik/flutter-web-map-tap-test)

You can clone and build it for web (see README).  

`onTap` creates a marker at the wrong position. `onLongTap` at the right position.

This bug only affects the web version. onTap works fine in the Android Emulator, even with activated Semantics.

Since FlutterMap is the only Widget I encountered a problem with Semantics so far, I believe the bug report belongs here.

### How can we reproduce it?

Activate `SemanticsBinding.instance.ensureSemantics();` in main.dart main method and watch LatLng coordinates of FlutterMaps onTap callback.

Try demo project.

No clue.

wcjord:
Fixes #2176.

### The mechanism

`MapInteractiveViewer` builds `RawGestureDetector(gestures: _gestures, ...)` and leaves `excludeFromSemantics` at its default `false`. `_gestures` contains a `TapGestureRecognizer` and a `LongPressGestureRecognizer`, so Flutter wraps the child in `_GestureSemantics` and publishes `SemanticsAction.tap` / `longPress` on a semantics node covering the detector's whole hit area — i.e. the entire map.

Two consequences once the semantics tree is enabled (`SemanticsBinding.instance.ensureSemantics()`, or automatically when assistive technology is detected):

**1. `onTap` reports the map centre (#2176).** A semantics-driven activation doesn't carry a pointer position, so the framework synthesises one at the node's centre and the recognizer fires with `kind: unknown`. Since the node is the size of the map, that centre *is* the map centre — which is exactly the reported symptom, and why `kind` changes from `mouse` to `unknown`. Long-press is affected in principle for the same reason; drag recognizers publish no semantics action, which is why panning/zooming were unaffected and only `onTap` looked broken.

**2. On web it blankets any `HtmlElementView` over the map.** Flutter web gives a tappable semantics node `pointer-events: all`. Platform views are *not* part of the semantics tree, so a map-sized tappable node sits above any `<iframe>` / `<video>` composited over the map and swallows its mouse events — the embed renders and plays but its own controls never respond (a YouTube embed never even shows its control bar, because it never receives `mousemove`).

I confirmed (2) in a Flutter web app by measuring `document.elementFromPoint` over a playing embed: the top element was the map's `flt-semantics` node, viewport-sized, `pointer-events: all`, every ancestor correctly `none`. Setting `pointer-events: none` on that one node — changing nothing else — immediately restored the embed's full controls.

### The change

`excludeFromSemantics: true` on that `RawGestureDetector`. The map's gestures are pointer gestures, not semantic actions: a map-sized "tap" target isn't a meaningful screen-reader affordance, and its only current effect is to mis-report the tap position. Markers and other children keep their own semantics, so nothing meaningfully activatable loses a node.

### Testing

- New regression test in `test/gestures/map_interactive_viewer_test.dart`: with semantics enabled, a bare `FlutterMap` publishes no `tap`/`longPress` semantics action. It fails on `master` and passes with this change.
- Full suite green locally (118/118).

Happy to adjust the approach — if you'd rather keep a semantic affordance on the map and fix the position synthesis instead, say so and I'll rework it.

### AI disclosure

Per CONTRIBUTING: this was written with AI assistance (Claude). I've reviewed the diagnosis and the diff, ran the tests locally, and take responsibility for the change.

### DIFF
--- a/lib/src/gestures/map_interactive_viewer.dart
+++ b/lib/src/gestures/map_interactive_viewer.dart
@@ -359,6 +359,24 @@ class MapInteractiveViewerState extends State<MapInteractiveViewer>
                   : Duration.zero,
           child: RawGestureDetector(
             gestures: _gestures,
+            // The map's gestures are pointer gestures, not semantic actions.
+            // Left included, `_gestures`' tap/long-press recognizers publish
+            // `SemanticsAction.tap`/`longPress` on a semantics node covering the
+            // map's whole hit area, which causes two problems once the semantics
+            // tree is enabled:
+            //
+            //  * A semantics-driven tap synthesises its position at the node's
+            //    centre, so `onTap` reports the map centre instead of where the
+            //    user tapped, with `kind: unknown` (#2176).
+            //  * On web, Flutter gives a tappable semantics node
+            //    `pointer-events: all`. A map-sized node therefore covers any
+            //    HtmlElementView layered above the map (`<iframe>`, `<video>`)
+            //    and swallows the mouse events those embeds need, leaving their
+            //    own controls unusable.
+            //
+            // Markers and other children keep their own semantics, so nothing
+            // that is meaningfully activatable loses its node here.
+            excludeFromSemantics: true,
             child: widget.builder(
               context,
               widget.controller.options,
--- a/test/gestures/map_interactive_viewer_test.dart
+++ b/test/gestures/map_interactive_viewer_test.dart
@@ -1,3 +1,6 @@
+import 'package:flutter/material.dart';
+import 'package:flutter/semantics.dart';
+import 'package:flutter_map/flutter_map.dart';
 import 'package:flutter_map/src/gestures/map_interactive_viewer.dart';
 import 'package:flutter_test/flutter_test.dart';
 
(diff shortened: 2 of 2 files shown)


JaffaKetchup:
Thanks for opening this PR!

Please see https://github.com/fleaflet/flutter_map/issues/2176#issuecomment-5158206651.

Therefore, this PR is a workaround. Unfortunately I don't think the best path forwards is a workaround. As you have done in your own PR on another project, if any reader wants a workaround, wrapping the map with `ExcludeSemantics` seems to work, with some drawbacks of course.

A tap on a map could be semantically important if that tap, for example, performs an action.

Also, your point 2 seems to talk about platform views and embeds which are not relevant to flutter_map.

PR #2239, opened 2026-08-05
fix: dispose the image handed to a tile pruned during dispatch

dinin92-del:
`TileImage._onImageLoadSuccess` stores the `ImageInfo` even when the tile is already disposed. `ImageStreamCompleter.setImage` gives every listener its own handle and the listener owns it: a displayed tile passes ownership to `RenderImage`, which disposes it — but a disposed tile never builds one, so the decoded image stays alive for the lifetime of the process.

## Why `dispose()` removing the listener isn't enough

This is **not** reachable by simply disposing a tile and then completing its image — `dispose()` removes the listener and `setImage` early-returns on an empty listener list.

It is reachable because `setImage` dispatches over a **copy** of the listener list:

```dart
// Make a copy to allow for concurrent modification.
final localListeners = List<ImageStreamListener>.of(_listeners);
for (final listener in localListeners) {
  listener.onImage(image.clone(), false);
}
```

A listener removed from inside that loop is still called. Two tiles resolving equal keys share one completer, and `onLoadComplete` is where tiles get pruned — as `TileImageManager.reloadImages` already notes:

> If a TileImage's imageInfo is already available when load() is called it will call its onLoadComplete callback synchronously which can trigger pruning.

So the first tile's completion can dispose the second tile mid-dispatch, and the second tile is handed an image anyway.

## Impact

Unnoticeable with 256×256 tiles (256 KB). With 768×768 RGBA tiles (2.25 MB each) it killed an app on iOS while browsing the map — the process was terminated at its memory limit after a few minutes of panning.

## Test

Adds `test/layer/tile_layer/tile_image_test.dart`, which reproduces the dispatch race with two tiles sharing one completer and asserts both that the disposed tile keeps no `ImageInfo` and that the handle it was given is released (`debugGetOpenHandleStackTraces`).

On the current code it fails with:

```
Expected: null
  Actual: ImageInfo:<[8x8] @ 1.0x>
```

Full suite passes with the change (118/118).

---

**AI usage disclosure** (per CONTRIBUTING): this patch and its test were written with AI assistance (Claude). I reviewed the change and the reasoning, ran the test suite, and verified that the added test fails on unpatched `master` and passes with the fix. I take responsibility for the code.

### DIFF
--- a/lib/src/layer/tile_layer/tile.dart
+++ b/lib/src/layer/tile_layer/tile.dart
@@ -1,3 +1,5 @@
+import 'dart:ui' as ui show Image;
+
 import 'package:flutter/widgets.dart';
 import 'package:flutter_map/flutter_map.dart';
 
@@ -89,7 +91,7 @@ class _TileState extends State<Tile> {
       );
     } else if (widget.tileImage.animation == null) {
       return RawImage(
-        image: widget.tileImage.imageInfo?.image,
+        image: _handOff(widget.tileImage),
         fit: BoxFit.fill,
         opacity: widget.tileImage.opacity == 1
             ? null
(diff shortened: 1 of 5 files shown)


JaffaKetchup:
Thanks for the contribution! I'll try to take a look at this soon.

JaffaKetchup:
Hey, could you please open a bug report and move the bug report from the PR description into it, and create an MRE so I can manually confirm that this fix fixes the memory leak? Thanks!

bandy87:
Opened #2265 with a full reproduction and before/after measurements on a physical device (Samsung SM S938B, Android 16, release build), as requested.

I can confirm this PR's fix works: with these changes applied on top of v8.3.2, memory stays flat through 130+ seconds of continuous rapid pinch zooming (Private Other 34-77 MB), where the unpatched build was killed by the OS after ~50 seconds.

The issue also lists five hypotheses I ruled out by measurement before arriving here (smaller tiles, disabling the built-in disk cache, forcing a layer reset, upgrading to 8.3.2, and reducing tile creation), in case that saves anyone repeating them.

One note on CI: the only red check is "Analyse Code", which fails at `dart format --set-exit-if-changed` — the test jobs pass on both Flutter channels.

PR #2241, opened 2026-08-13
fix: don't pass a non-positive pinch scale to math.log

### LINKED ISSUE #2237: Pinch zoom can pass a non-positive scale to math.log, poisoning the camera with NaN
### What is the bug?

`_getZoomForScale` can be handed a non-positive scale, making `math.log()` return `NaN`/`-Infinity`, which then propagates into the camera via `moveRaw` unvalidated.

In [`map_interactive_viewer.dart`](https://github.com/fleaflet/flutter_map/blob/master/lib/src/gestures/map_interactive_viewer.dart), the pinch-zoom branch guards `details.scale` but passes `details.scale + _scaleCorrector`:

```dart
// L650-654
if (hasPinchZoom && details.scale > 0.0) {
  newZoom = _getZoomForScale(
    _mapZoomStart,
    details.scale + _scaleCorrector,   // <- the sum is never checked
  );
```

`_scaleCorrector` is set when the finger count changes mid-gesture (L617):

```dart
_scaleCorrector = 1.0 - _lastScale;
```

so it is negative whenever `_lastScale > 1` (i.e. the user was zooming in when a finger was added/removed). The sum is then `details.scale + 1.0 - _lastScale`, which is `<= 0` as soon as `details.scale <= _lastScale - 1`. Concretely: a finger lifts at `_lastScale == 2.5` (`_scaleCorrector == -1.5`), the recognizer restarts the new gesture near `details.scale == 1.0`, and the sum is `-0.5`.

`_getZoomForScale` has no guard, and `clampZoom` cannot recover it because `NaN.clamp()` returns `NaN` in Dart:

```dart
// L1404-1408
double _getZoomForScale(double startZoom, double scale) {
  final resultZoom =
      scale == 1.0 ? startZoom : startZoom + math.log(scale) / math.ln2;
  return _camera.clampZoom(resultZoom);
}
```

`log(negative)` is `NaN` and `log(0)` is `-Infinity`, so `newZoom` becomes non-finite and is passed straight to `controller.moveRaw(newCenter, newZoom, ...)`. `_calculatePinchZoomAndMove` also consumes that same `zoomAfterPinchZoom`, so the resulting `center` goes `NaN` too.

Once the camera holds a non-finite center/zoom, every layer that reads it breaks. The `MapCamera` constructor's `assert(zoom.isFinite)` only fires in debug builds, and it covers `zoom` but not `center`, so release builds store the poisoned camera silently.

I want to flag how this interacts with the recent NaN work, because I think it makes the bug easier to miss now rather than harder:

- #2182 (in 8.3.0) added the `Crs.checkLatLng` throw. That surfaced this as a hard crash, which I believe is why #2214 reported "only happens starting with version 8.3, no problems with 8.2" — the NaN was likely already being produced, and 8.3.0 just made it fatal.
- #2218 reverted #2182 for 8.3.1, and `checkLatLng` is now gone from master.

The revert removes the exception, but the arithmetic above still produces the NaN. So on 8.3.1+ the same gesture should yield a silently non-finite camera — a stuck/blank map rather than a stack trace. That matches the "Map is stuck" wording in #2199. I mention it only because a silent version of this is harder to diagnose than the crashing one, not to argue the revert was wrong.

**Honest caveat on evidence:** this is code-path analysis plus production telemetry, not a captured repro. I have Crashlytics reports from a released Android app on 8.3.0 showing `LatLng(latitude:NaN, longitude:NaN)` reaching `Crs.checkLatLng` from four different layers (`MarkerClusterLayer.didUpdateWidget`, `MarkerLayer.build`, `TileLayer._onTileUpdateEvent`, `_buildMobileTransformStack`) — the signature of an already-poisoned camera rather than one bad input. I ruled out my own call sites: every `move`/`fitCamera` in the app is validated and instrumented, and the rejection counter never fired once. The pinch path is the one route to the camera that bypasses application code, and the unguarded sum above is a real defect on master regardless of whether it is what my users hit.

### How can we reproduce it?

I have not been able to reliably reproduce the gesture by hand — the window where `details.scale + _scaleCorrector` goes non-positive requires a finger count change while zoomed in, then a rapid continuation. #2214's "rapidly zooming in and out" with varying finger counts on Android is the closest description I have.

The arithmetic itself is directly demonstrable without gestures:

```dart
// _scaleCorrector == 1.0 - _lastScale, with _lastScale == 2.5
const scaleCorrector = 1.0 - 2.5;   // -1.5
const detailsScale = 1.0;           // recognizer restarts near 1.0
final scale = detailsScale + scaleCorrector;  // -0.5

// mirrors _getZoomForScale
final resultZoom = 13.0 + math.log(scale) / math.ln2;
print(resultZoom);              // NaN
print(resultZoom.clamp(0, 20)); // NaN — clampZoom cannot recover it
```

I am happy to try to build a proper `WidgetTester` gesture repro (driving `ScaleUpdateDetails` with a changing `pointerCount`) if that would be useful — I did not want to guess at the harness conventions before checking whether you would want the test at all.

Guard the value actually passed to `math.log`, rather than `details.scale` alone. Either check the sum at the call site:

```dart
final scale = details.scale + _scaleCorrector;
if (hasPinchZoom && scale > 0.0) {
  newZoom = _getZoomForScale(_mapZoomStart, scale);
```

or make `_getZoomForScale` total, which also covers the other two call sites (L745 and L908):

```dart
double _getZoomForScale(double startZoom, double scale) {
  if (scale <= 0.0 || !scale.isFinite) return _camera.clampZoom(startZoom);
  final resultZoom =
      scale == 1.0 ? startZoom : startZoom + math.log(scale) / math.ln2;
  return _camera.clampZoom(resultZoom);
}
```

Separately, it may be worth rejecting non-finite values in `moveRaw` itself. With `checkLatLng` reverted there is now no runtime backstop between a non-finite gesture computation and a permanently poisoned camera, and a single bad frame persists until the app is restarted. Happy to open a PR for either or both if you have a preference on the approach.

AzazelSensei:
Fixes #2237

The pinch-zoom path only checked `details.scale > 0` before calling `math.log`. The value that actually gets logged is `details.scale + _scaleCorrector`. After a multi-finger gesture race win, that corrector is `1 - lastScale`, so the sum can be 0 or negative. `log` of that is NaN, and it goes straight into the camera.

I now check the corrected scale at the call site. `zoomForScale` also keeps the start zoom if the scale isn't a positive finite number, so the other two callers are covered too.

There's a unit test with the numbers from the issue (`lastScale = 2.5`, new gesture at `details.scale = 1.0`).

### DIFF
--- a/lib/src/gestures/map_interactive_viewer.dart
+++ b/lib/src/gestures/map_interactive_viewer.dart
@@ -647,10 +647,13 @@ class MapInteractiveViewerState extends State<MapInteractiveViewer>
     var newZoom = _camera.zoom;
 
     // Handle pinch zoom.
-    if (hasPinchZoom && details.scale > 0.0) {
+    // After a gesture-race win, _scaleCorrector is 1 - lastScale, so this
+    // sum can be <= 0 even when details.scale itself is positive.
+    final scale = details.scale + _scaleCorrector;
+    if (hasPinchZoom && scale > 0.0) {
       newZoom = _getZoomForScale(
         _mapZoomStart,
-        details.scale + _scaleCorrector,
+        scale,
       );
 
       // Handle starting of pinch zoom.
@@ -1401,10 +1404,19 @@ class MapInteractiveViewerState extends State<MapInteractiveViewer>
 
   // Utilities
 
+  /// Converts a pinch [scale] into a zoom level starting from [startZoom].
+  ///
+  /// Returns [startZoom] when [scale] is not a positive finite number.
+  /// `math.log` is undefined for those inputs and would produce `NaN` or
+  /// `-Infinity`, which then poison the camera.
+  @visibleForTesting
+  static double zoomForScale(double startZoom, double scale) {
+    if (scale <= 0.0 || !scale.isFinite) return startZoom;
+    return scale == 1.0 ? startZoom : startZoom + math.log(scale) / math.ln2;
+  }
+
   double _getZoomForScale(double startZoom, double scale) {
-    final resultZoom =
-        scale == 1.0 ? startZoom : startZoom + math.log(scale) / math.ln2;
-    return _camera.clampZoom(resultZoom);
+    return _camera.clampZoom(zoomForScale(startZoom, scale));
   }
 
   Offset _rotateOffset(Offset offset) {
--- a/test/gestures/map_interactive_viewer_test.dart
+++ b/test/gestures/map_interactive_viewer_test.dart
@@ -46,4 +46,42 @@ void main() {
       },
     );
   });
+
+  group('MapInteractiveViewerState.zoomForScale', () {
+    test('adds log2(scale) to the start zoom', () {
+      expect(MapInteractiveViewerState.zoomForScale(13, 2), 14);
+      expect(MapInteractiveViewerState.zoomForScale(13, 1), 13);
+      expect(MapInteractiveViewerState.zoomForScale(13, 0.5), 12);
+    });
+
+    test(
+      'returns the start zoom when the corrected pinch scale is not '
+      'positive (regression: details.scale + _scaleCorrector can be '
+      '<= 0 after a gesture-race win, and math.log of that is NaN - '
+      'https://github.com/fleaflet/flutter_map/issues/2237)',
+      () {
+        // _scaleCorrector = 1.0 - _lastScale, with _lastScale == 2.5
+        const scaleCorrector = 1.0 - 2.5;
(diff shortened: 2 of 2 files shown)


JaffaKetchup:
Thanks for the contribution! This may need to be reopened in case the issue is not resolved, but please test v8.3.2 to see if you can still reproduce your issue. If you can, please ping me. In the meantime, I'm going to close this as a potential duplicate.

Issue #2178, opened 2025-12-31
`LatLng` with `double.nan` in `Marker`/`MarkerLayer` triggers memory leak


xoofee:
### What is the bug?

 NaN LatLng point in Marker/MakerLayer cause Out of memory error
` dart
MarkerLayer(markers: [Marker(point: latlong2.LatLng(0/0, 0/0), child: Icon(Icons.keyboard_arrow_up, color: Colors.white, size: 20))]),
`

### How can we reproduce it?

```dart
import 'package:flutter/material.dart';
import 'package:flutter_map/flutter_map.dart';
import 'package:latlong2/latlong.dart' as latlong2;

void main() {
  runApp(const MyApp());
}

class MyApp extends StatelessWidget {
  const MyApp({super.key});

  @override
  Widget build(BuildContext context) {
    return MaterialApp(
      title: 'NaN Marker Test',
      theme: ThemeData(
        colorScheme: ColorScheme.fromSeed(seedColor: Colors.deepPurple),
        useMaterial3: true,
      ),
      home: const MapPage(),
    );
  }
}

class MapPage extends StatelessWidget {
  const MapPage({super.key});

  @override
  Widget build(BuildContext context) {
    return Scaffold(
      appBar: AppBar(
        backgroundColor: Theme.of(context).colorScheme.inversePrimary,
        title: const Text('NaN Marker OOM Test'),
      ),
      body: FlutterMap(
        options: MapOptions(
          initialCenter: latlong2.LatLng(0.0, 0.0),
          initialZoom: 2.0,
        ),
        children: [
          TileLayer(
            urlTemplate: 'https://tile.openstreetmap.org/{z}/{x}/{y}.png',
            userAgentPackageName: 'com.example.findeasy',
          ),
          MarkerLayer(
            markers: [
              Marker(
                // point: latlong2.LatLng(0.0, 0.0), // this is ok
                point: latlong2.LatLng(0/0, 0/0), // NaN coordinates cause Out of Memory
                child: const Icon(
                  Icons.keyboard_arrow_up,
                  color: Colors.white,
                  size: 20,
                ),
              ),
            ],
          ),
        ],
      ),
    );
  }
}
```

pubspec.lock
```
  flutter_map:
    dependency: "direct main"
    description:
      name: flutter_map
      sha256: "391e7dc95cc3f5190748210a69d4cfeb5d8f84dcdfa9c3235d0a9d7742ccb3f8"
      url: "https://pub.dev"
[...]

PR #2242, opened 2026-08-13
fix: stop MarkerLayer wrap loops on a non-finite camera

### LINKED ISSUE #2240: A non-finite camera makes MarkerLayer's world-wrap loops unbounded (OOM/ANR) — _projectPoints guards the points but nothing guards the camera
### What is the bug?

`#2178` fixed the case where a `Marker.point` is `NaN`, by validating points in
`_MarkerLayerState._projectPoints`. The same out-of-memory failure is still
reachable from the other side: when a **camera** value is non-finite, every
marker point is perfectly valid, so that guard never fires.

In `_MarkerLayerState.build`, culling decides when to stop wrapping:

```dart
Positioned? getPositioned(double worldShift) {
  if (!pixelBounds.overlaps(Rect.fromPoints(...))) return null;   // cull
  ...
}

for (double shift = -worldWidth;; shift -= worldWidth) {
  final additional = getPositioned(shift);
  if (additional == null) break;      // <- only exit
  yield additional;
}
```

If the camera is non-finite, `pixelBounds` is `Rect.fromLTRB(NaN, NaN, NaN, NaN)`.
`Rect.overlaps` early-returns `false` on ordered comparisons, and **every
comparison against NaN is false**, so it falls through and returns `true` —
culling never culls, `getPositioned` never returns `null`, and the `sync*`
generator yields `Positioned` widgets until the process dies.

`feature_layer_utils.workAcrossWorlds` (polyline/polygon/circle) already has a
`maxShiftsCount = 30` backstop from #2052/#2111. `MarkerLayer` has no equivalent.

### How can we reproduce it?

Twelve lines, no gestures — `flutter test` this against `flutter_map: 8.3.1`:

```dart
testWidgets('non-finite camera rotation hangs MarkerLayer', (tester) async {
  final controller = MapController();
  await tester.pumpWidget(MaterialApp(
    home: FlutterMap(
      mapController: controller,
      options: const MapOptions(
        initialCenter: LatLng(16.6, 120.9),
        initialZoom: 12,
      ),
      children: const [
        MarkerLayer(markers: [
          Marker(
            point: LatLng(16.6, 120.9),   // finite: #2178's guard cannot fire
            width: 40,
            height: 40,
            child: SizedBox.shrink(),
          ),
        ]),
      ],
    ),
  ));

  controller.rotate(double.nan);   // any non-finite camera value will do
  await tester.pump();             // never returns
});
```

`flutter_tester` reached **16 GB** resident before we killed it. A test
`Timeout` does not fire, because the loop is synchronous.

`MapController.rotate(double.nan)` is just the cheapest injection point. We hit
this in production from an uncalibrated magnetometer: the sensor emitted a `NaN`
heading which our own code passed to `rotate()`. Because NaN comparisons are all
false, it slipped through several range checks on the way — which is the same
property that defeats the culling here. That was our bug to fix, and we fixed it;
we're reporting the library-side amplification, where one bad value costs the
whole process instead of one bad frame.

Note `camera.dart:148` already asserts `zoom.isFinite`, so a non-finite **zoom**
is caught in debug. There is no equivalent assert for `rotation`, and asserts are
stripped in release builds, so both fields reach `MarkerLayer` in production.

### Suggested fix

Two independent options; we apply both locally.

1. Bail out of the wrap loops on non-finite input (mirrors the `_projectPoints`
   guard, on the camera side):

```dart
if (worldWidth == 0 ||
    !worldWidth.isFinite ||
    !pxPoint.dx.isFinite || !pxPoint.dy.isFinite ||
    !pixelBounds.left.isFinite || !pixelBounds.right.isFinite) {
  continue;
}
```

2. Give `MarkerLayer` the same `maxShiftsCount` backstop that
   `feature_layer_utils.workAcrossWorlds` already has, so any future
   never-culls bug costs an assertion instead of the process.

Asserting `rotation.isFinite` in `MapCamera` alongside the existing
`zoom.isFinite` assert would also surface the root cause much earlier in debug.

### Environment

- flutter_map 8.3.1
- Reproduced in `flutter test` (host) and on a Galaxy S24 Ultra, Android 16
  (10s+ input-dispatch ANR, `am_anr` in logcat)

### Related

- #2178 — same OOM, marker-point side, fixed; camera side still open
- #2052 / #2111 — why `feature_layer_utils` gained `maxShiftsCount`
- #2237 — a *source* of NaN cameras (unrelated fix; this report is about the
  amplification, and would still matter once #2237 lands)

AzazelSensei:
Fixes #2240

#2178 only rejects a non-finite `Marker.point`. A NaN camera still makes `Rect.overlaps` return true for every world copy, so the wrap loops never stop.

I skip `MarkerLayer` when the camera isn't finite, and I skip wrapping a marker if its projected point isn't finite either. There's also a 30-shift cap like `workAcrossWorlds`, so a future never-culls bug throws instead of taking the process down.

Widget test: `rotate(double.nan)`, then `pump()` has to return.

### DIFF
--- a/lib/src/layer/marker_layer/marker_layer.dart
+++ b/lib/src/layer/marker_layer/marker_layer.dart
@@ -98,6 +98,16 @@ class _MarkerLayerState extends State<MarkerLayer> {
     final pixelOrigin = map.pixelOrigin;
     final markers = widget.markers;
 
+    // A non-finite camera makes Rect.overlaps return true for every world
+    // copy, so the wrap loops never terminate (OOM/ANR). #2178 only guards
+    // Marker.point. See https://github.com/fleaflet/flutter_map/issues/2240.
+    if (!worldWidth.isFinite ||
+        !pixelBounds.left.isFinite ||
+        !pixelBounds.right.isFinite ||
+        !map.rotation.isFinite) {
+      return const SizedBox.shrink();
+    }
+
     return MobileLayerTransformer(
       child: Stack(
         children: () sync* {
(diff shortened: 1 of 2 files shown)


JaffaKetchup:
Thanks for your contribution!

The root issue should now be fixed, please ping me if it isn't. And as previously discussed, this is a partial patch which doesn't solve the root issue. That's not to say that this isn't a good idea, as it is defensive against similar bugs in future, but I am not completely happy with the implementation either.

So for now, I'm going to close this PR. But please feel free to contribute again in future!

PR #2252, opened 2026-09-03
chore(example): prepare for upgrade to built-in Kotlin

JaffaKetchup:
Flutter 3.47 supports built-in Kotlin. To best prepare for this, we should remove shared_preferences:

shared_preferences_android needs to be upgraded to support built-in Kotlin. However, this requires version 3.12 of the Dart SDK, while we currently support 3.6. To ensure that we can build on Flutter 3.27 to prove that flutter_map works on that version, shared_preferences must therefore be removed. This comes with other benefits as well (makes the app smaller, faster, less complicated, 100% cookie free). Therefore, a small redesign is required to remove the first-time popup previously presented to Live Web Demo users which asked them to use the demo responsibly, and used shared_preferences to store whether it was a user's first visit. Therefore, this message has been moved into the drawer header (as shown in the image below). To reduce the intrusiveness of the attribution popup on first load, the sentence there has also been removed, and instead the attribution is now available on every page (although it still only opens by default on the home page). To save space, and because the attribution widget now shows the flutter_map logo, the logo has been removed from next to the menu button.

new_drawer

image

Once upgraded to built-in Kotlin on v3.47, it will no longer be buildable for any version less than 3.47 (or maybe 3.44, I'm not sure), unless we maintain two separate build configs.


josxha (approved):
good changes 👍

PR #2259, opened 2026-09-06
feat: show a background grid while map tiles load

ben-milanko:
Empty map areas currently show a flat background while tiles load. Add a subtle grid beneath all map layers, enabled by default with 64 logical pixel cells at integer zoom levels. It follows panning, zooming, and rotation, and opaque tiles cover it as they fade in. Existing fallback tiles continue to cover the grid while replacement tiles load.

Loading grid behind partially rendered map tiles

Expose `MapOptions.backgroundGridColor` and `MapOptions.backgroundGridSpacing` for customization. Setting `backgroundGridColor: null` restores the plain background. Fully transparent map backgrounds remain transparent. When disabled, the background uses the existing plain `ColoredBox` without a grid painter or camera dependency.

This intentionally changes the default map appearance. The grid is a background decoration, so it also remains visible through transparent tiles and in areas without tiles. The Tile Builder example includes a toggle, and the API documentation describes these behaviors.

Validation:

- All 148 tests pass on the clean upstream-based branch, including 8 new tests for rendered grid pixels, partial loading, fade-in, fallback tiles, camera movement, customization, transparency, and gestures.
- Analysis passes for the library, tests, and example page; formatting checks pass for all changed Dart files.
- Tested with Flutter 3.41.9 / Dart 3.11.5. Minimum-supported-SDK validation remains for CI.

Developed and checked with OpenAI Codex; the validation listed above ran locally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,11 @@
 
 Flutter's №1 non-commercially aimed map client: it's easy-to-use, versatile, vendor-free, fully cross-platform, and 100% pure-Flutter
 
+Maps show a subtle background grid while tiles load. It moves with the map and
+is covered as tiles render. Customize it with `MapOptions.backgroundGridColor`
+and `MapOptions.backgroundGridSpacing`, or set `backgroundGridColor: null` to
+use a plain background. Fully transparent map backgrounds remain transparent.
+
 * ### [📖 **Read the docs**](https://docs.fleaflet.dev/)
 
 * ### [🚀 **Launch the web demo**](https://demo.fleaflet.dev/)
--- a/example/lib/pages/tile_builder.dart
+++ b/example/lib/pages/tile_builder.dart
@@ -14,6 +14,7 @@ class TileBuilderPage extends StatefulWidget {
 
 class TileBuilderPageState extends State<TileBuilderPage> {
   bool enableGrid = true;
+  bool showBackgroundGrid = true;
   bool showCoordinates = true;
   bool showLoadingTime = true;
   bool darkMode = true;
@@ -82,6 +83,15 @@ class TileBuilderPageState extends State<TileBuilderPage> {
                   ),
                   const SizedBox.square(dimension: 12),
                   const Tooltip(
+                    message: 'Show Background Grid While Tiles Load',
+                    child: Icon(Icons.grid_on),
+                  ),
+                  Switch.adaptive(
+                    value: showBackgroundGrid,
+                    onChanged: (v) => setState(() => showBackgroundGrid = v),
+                  ),
+                  const SizedBox.square(dimension: 12),
+                  const Tooltip(
                     message: 'Show Coordinates',
                     child: Icon(Icons.location_on),
                   ),
@@ -113,9 +123,11 @@ class TileBuilderPageState extends State<TileBuilderPage> {
           ),
           Expanded(
             child: FlutterMap(
-              options: const MapOptions(
-                initialCenter: LatLng(51.5, -0.09),
+              options: MapOptions(
+                initialCenter: const LatLng(51.5, -0.09),
                 initialZoom: 5,
+                backgroundGridColor:
+                    showBackgroundGrid ? const Color(0x14000000) : null,
               ),
               children: [
                 _darkModeContainerIfEnabled(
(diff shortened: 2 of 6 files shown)


example/lib/pages/tile_builder.dart
```diff
@@ -14,6 +14,7 @@ class TileBuilderPage extends StatefulWidget {
 
 class TileBuilderPageState extends State<TileBuilderPage> {
   bool enableGrid = true;
+  bool showBackgroundGrid = true;
```
JaffaKetchup, line 17: This is the wrong place to demo this functionality. 

Since it will be enabled by default, and it is minimally different, I don't think any example is necessary.

lib/src/map/options/options.dart
```diff
@@ -84,6 +84,22 @@ class MapOptions {
...
   final Color backgroundColor;
 
+  /// Color of the grid drawn over [backgroundColor], beneath all map layers.
+  ///
+  /// The grid is visible while tiles load and is covered as opaque tiles fade
+  /// in. It also remains visible through transparent tiles and in empty areas.
+  /// Set to `null` to disable it. A fully transparent [backgroundColor] also
+  /// disables the grid.
+  ///
+  /// Enabled by default with a subtle black line color (`0x14000000`).
```
JaffaKetchup, line 94: I like the idea that a transparent `backgroundColor` automatically disables the grid.

However, this we should also automatically adjust the colour of the grid based on the background colour (to improve DX). Maybe by converting the background color into HSL, then ± the lightness value by maybe 10% (or something similar) based on whether the lightness is less than or more than 50%.

This means that `null` now means auto-calculate, not disabled. Therefore, another switch is required to enable/disable the grid. Therefore, there's two options:

* Use a sub-options class/object called `BackgroundGridOptions` - then if the parameter `backgroundGrid` is `null`, it's disabled, but the default is a `const BackgroundGridOptions()` which is enabled and automatic colour and default grid spacing
* Convert the whole grid into a layer - the obvious drawback of this is it requires adding manually even though the grid behaviour is a nice improvement that I would imagine would be welcomed (if even noticed) by most users, and it breaks the semantics by having the background configured in two different places

I think it's worth keeping it in the options, even though it technically is more like another layer. So the first option is probably the best, but willing to hear otherwise from anyone \:) (@fleaflet/maintainers ?)

lib/src/map/background.dart
```diff
@@ -0,0 +1,98 @@
...
+  /// The grid cell size at integer zoom levels.
+  final double gridSpacing;
+
+  @override
+  Widget build(BuildContext context) {
+    final gridColor = this.gridColor;
+    if (gridColor == null || gridColor.a == 0 || color.a == 0) {
+      return ColoredBox(color: color);
+    }
+    final camera = MapCamera.of(context);
```
JaffaKetchup, line 32: Nit: this should be the first line in `build`.

lib/src/map/background.dart
```diff
@@ -0,0 +1,98 @@
...
+
+    return RepaintBoundary(
+      child: ColoredBox(
+        color: color,
+        child: IgnorePointer(
```
JaffaKetchup, line 37: Nit: what is the point in this `IgnorePointer`? There might be one, I just can't find any difference with or without?

README.md
```diff
@@ -2,6 +2,11 @@
 
 Flutter's №1 non-commercially aimed map client: it's easy-to-use, versatile, vendor-free, fully cross-platform, and 100% pure-Flutter
 
+Maps show a subtle background grid while tiles load. It moves with the map and
+is covered as tiles render. Customize it with `MapOptions.backgroundGridColor`
+and `MapOptions.backgroundGridSpacing`, or set `backgroundGridColor: null` to
+use a plain background. Fully transparent map backgrounds remain transparent.
+
```
JaffaKetchup, line 9: We don't add documentation into the README. I'll add it into the docs later if merged.

JaffaKetchup (changes requested):
Hey, thanks for contributing! There are a few things to follow up on, but overall I think this is a nice addition.

(For completeness, it might be worth mentioning the old [plugin which added a latlng grid layer](https://github.com/matthiasdittmer/lat_lon_grid_plugin), but that's a different purpose and also isn't updated to the latest versions, so it's mostly irrelevant here.)

JaffaKetchup:
Hey @ben-milanko, I've made some fixes to the above requests, and also some other improvements. However, I don't seem to be able to push these to the PR/branch. Could you please check that you've checked the box allowing maintainers to make changes? If you're not happy to do this, let me know and I'll open a new PR and cherry pick your commit across to that.

PR #2260, opened 2026-09-07
fix(tile_layer): Don't mark a failed tile ready to display without an error image

nebkat:
`readyToDisplay` documents itself as `true` when "loading errored but an error image is configured", but `_display()` set it on any error regardless. A tile that failed and has nothing to paint therefore counted as covering its area, so `renderTiles` discarded the coarser ancestor standing in for it and the area went blank instead of staying upscaled. Services whose coverage depth varies by location show this plainly: past the deepest level available at a given place the tiles 404, and zooming in lost detail that was already on screen.

Set it from `errorImage`, as documented. Tiles that have an error image are unaffected, and must stay ready - opacity is gated on this, so the error image would otherwise never be painted.


nebkat:
Closing for now — I want to verify the downstream effect end to end before asking for review time. Will reopen once I have that.

